PR #217, opened 2025-10-29
[Feature]: Auto detect kata mode and self configure monitoring

### LINKED ISSUE #142: [Feature]: Auto detect kata mode and self configure monitoring
### Prerequisites

### Feature Summary

Currently to enable the solution on kata environments, users have to set a flag in the values file, this becomes complex in heterogenous environments. NVSentinel should be able to detect if kata is enabled on the node and it should update the configuration for monitoring accordingly.

### Problem/Use Case

Make enabling of the solution in kata environments easier

### Proposed Solution

We could looks for the type of GPUs advertised and use that for configuration, but that may not be the ideal way to go about it. This will need a lot more though and design.

### Component

Health Monitor

mchmarny:
## Summary

closes #142

Implements detection of Kata-enabled nodes using k8s API, with caching and concurrency support.

How It Works:
1) Labeler Module sets the kata label on nodes: 1nvsentinel.dgxc.nvidia.com/kata.enabled=true ` (or custom label from `kataLabelOverride`)
2) Two Separate DaemonSets are deployed:
* `daemonset-kata.yaml` - runs only on kata-enabled nodes
* `daemonset-regular.yaml` - runs only on non-kata nodes
3) NodeAffinity ensures correct placement:
```yaml
nodeAffinity:
  requiredDuringSchedulingIgnoredDuringExecution:
    nodeSelectorTerms:
      - matchExpressions:
        - key: nvsentinel.dgxc.nvidia.com/kata.enabled
          operator: In/NotIn
          values: ["true"]
```

4) The `--kata-enabled` flag is simply passed as a configuration parameter:

* Line 120 in `_helpers.tpl`: - `{{ $kataLabel | quote }}`
* This tells the monitor "you're running on a kata node" (determined by which DaemonSet deployed you)
5) No Runtime Detection Needed - Kubernetes already placed the pod on the right node type, so the syslog-health-monitor just needs to know which variant it is to adjust its behavior

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Health Monitors
- [ ] Core Services
- [ ] Fault Management
- [ ] Documentation/CI
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Release Notes

* **New Features**
  * Added Kata Containers detection and automatic node labeling capability
  * Introduced annotation prefix-matching utility functions
  * Added string truthy value helper for configuration parsing
  * Extended test infrastructure supporting Kata-enabled test nodes

* **Tests**
  * Topology validation test for Kata labeling verification
  * Comprehensive labeler test coverage including Kata scenarios

* **Documentation**
  * Development guidelines and architecture reference documentation

* **Chores**
  * Updated .gitignore entries for build artifacts
  * Dependency optimization cleanup


distros/kubernetes/nvsentinel/charts/syslog-health-monitor/templates/daemonset-regular.yaml
```diff
@@ -0,0 +1,166 @@
...
+  template:
+    metadata:
+      {{- with (.Values.global.podAnnotations | default .Values.podAnnotations) }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "syslog-health-monitor.selectorLabels" . | nindent 8 }}
+        nvsentinel.dgxc.nvidia.com/kata: "false"
+    spec:
```
lalitadithya: Since all of this same between the two daemonsets, can we make a helm variable and use the same in both places to avoid drift?

labeler-module/pkg/labeler/labeler.go
```diff
@@ -229,13 +235,41 @@ func (l *Labeler) getDriverLabelForNode(nodeName string) (string, error) {
...
+			return LabelValueTrue, nil
 		}
 	}
 
 	return "", nil
 }
 
+// getKataLabelForNode detects if Kata is enabled on the specified node
+// Returns "true" if Kata is enabled, "false" if not, or error if detection fails
+func (l *Labeler) getKataLabelForNode(nodeName string) (string, error) {
```
lalitadithya: I'm not sure I follow the changes. Since we have 2 daemonsets, can't we pass an additional arg to avoid the API call and cache here?

distros/kubernetes/nvsentinel/charts/syslog-health-monitor/templates/daemonset-kata.yaml
```diff
@@ -27,6 +27,7 @@ spec:
   selector:
     matchLabels:
       {{- include "syslog-health-monitor.selectorLabels" . | nindent 6 }}
+      nvsentinel.dgxc.nvidia.com/kata: "true"
```
lalitadithya: how is label set?

labeler-module/pkg/labeler/labeler.go
mchmarny: OK, so 
Labeler detects Kata → Sets node labels (API calls happen here - already cached)
DS use node labels for scheduling → Pods get deployed with --kata-enabled flag
Syslog pods receive the flag → No detection needed, just informational

mchmarny:
We have copilot and coderabbit reviewing this PR now, let me finish applying their feedback first before any humans start review. Will remove draft when done to indicate it's ready

distros/kubernetes/nvsentinel/charts/syslog-health-monitor/templates/_helpers.tpl
```diff
@@ -60,3 +60,198 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
```
lalitadithya, line 64: is this no longer needed -- https://github.com/NVIDIA/NVSentinel/blob/main/distros/kubernetes/nvsentinel/charts/syslog-health-monitor/templates/configmap.yaml#L40  ? Can we drop it?

labeler-module/go.mod
```diff
@@ -85,3 +85,5 @@ replace github.com/nvidia/nvsentinel/store-client-sdk => ../store-client-sdk
 replace github.com/nvidia/nvsentinel/data-models => ../data-models
 
 replace github.com/nvidia/nvsentinel/commons => ../commons
+
+replace github.com/nvidia/nvsentinel/health-monitors/syslog-health-monitor => ../health-monitors/syslog-health-monitor
```
lalitadithya: I'd prefer to not interlink modules; we recently went through an exercise to remove all the code level dependencies between modules so that they interact with each other via well-defined APIs and this makes testing much more easier. Would it be possible to move the detection logic into the labeler itself?

labeler-module/pkg/labeler/labeler.go
```diff
@@ -229,13 +239,69 @@ func (l *Labeler) getDriverLabelForNode(nodeName string) (string, error) {
...
+	// Get or create cached detector for this node
+	cachedDetector, err := l.getOrCreateKataDetector(nodeName)
```
lalitadithya: We don't need an additional cache, the informers are already backed by a cache and we should be able to query the cache either using an indexer or a lister from the informer

labeler-module/pkg/labeler/labeler.go
```diff
@@ -229,13 +239,69 @@ func (l *Labeler) getDriverLabelForNode(nodeName string) (string, error) {
...
+		return LabelValueTrue, nil
+	}
+
+	return LabelValueFalse, nil
+}
+
+// getOrCreateKataDetector returns the cached detector for a node, creating it if needed
+func (l *Labeler) getOrCreateKataDetector(nodeName string) (*kata.CachedDetector, error) {
+	// Fast path: check if detector exists with read lock
+	l.kataDetectorMu.RLock()
```
lalitadithya: I'm not sure if we need a lock, the labeler should be single threaded, so we should be able to get away with using a simple lookup.

health-monitors/syslog-health-monitor/pkg/kata/detector.go
```diff
@@ -0,0 +1,545 @@
...
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+		return nil
+	}
+}
+
+// validateClientset ensures the Kubernetes clientset is available
+func (d *Detector) validateClientset() error {
+	if d.clientset == nil {
```
lalitadithya: Can we instead ensure the clientset is always passed? Maybe in main we exit with status code 1 if we can't get the clientset so we don't have to keep doing the nil check everywhere

health-monitors/syslog-health-monitor/pkg/kata/detector.go
```diff
@@ -0,0 +1,545 @@
...
+
+	if retryErr != nil {
+		return nil, fmt.Errorf("failed to get node after retries: %w", retryErr)
+	}
+
+	return node, nil
+}
+
+// checkNodeMetadata examines node metadata for Kata indicators.
+func (d *Detector) checkNodeMetadata(node *corev1.Node) bool {
```
lalitadithya: Maybe I'm missing something, do we need all the concurrency? Wouldn't it be simpler for us to just do a if check?

health-monitors/syslog-health-monitor/pkg/kata/detector.go
```diff
@@ -0,0 +1,545 @@
...
+}
+
+// checkNodeLabels looks for Kata-related node labels.
+func (d *Detector) checkNodeLabels(node *corev1.Node) bool {
+	kataLabels := []string{
+		"katacontainers.io/kata-runtime",
+		"kata-containers.io/runtime",
+		"node.kubernetes.io/kata-enabled",
+		"kata.io/enabled",
+		"runtime.kata",
```
lalitadithya: Damn, this is quite a list. Is this because of the different components that apply labels -- GPU operator, kata etc? 

I'm also wondering if we can just check for `katacontainers.io/kata-runtime` or the GPU operator's cc enabled I think it is called. 

Also more broadly, if we have a label, do we need to relabel, can we just use the labels that other components have already applied?

health-monitors/syslog-health-monitor/pkg/kata/detector.go
```diff
@@ -0,0 +1,545 @@
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// detectViaRuntimeClass checks RuntimeClass resources for Kata handlers.
+// Uses TTL-based caching to minimize API load in large clusters (2000+ nodes).
+func (d *Detector) detectViaRuntimeClass(ctx context.Context) (bool, error) {
```
lalitadithya: Aren't runtime classes across nodes? How do we map a runtime class to do a node, I guess using the selectors?

tilt/KATA_TESTING.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Overview
+
+The Tilt setup creates two types of KWOK (fake) nodes:
+- **Regular nodes** (75% of total): Standard container runtime, kata detection returns `false`
+- **Kata nodes** (25% of total): Kata-enabled runtime, kata detection returns `true`
+
+## Node Distribution
+
+With the default `NUM_GPU_NODES=50`:
```
lalitadithya: We have another PR that tests all 50 nodes, would it be possible to add a different set of nodes outside of this with kata enabled? Or even better would be to add them for this test and then delete them at the end

distros/kubernetes/nvsentinel/charts/syslog-health-monitor/templates/_helpers.tpl
mchmarny: yep, no longer needed

labeler-module/go.mod
mchmarny: ok, let me refactor

labeler-module/pkg/labeler/labeler.go
mchmarny: What's the downside here? It's a basic read lock so if for some reason we had concurrency here in the future

health-monitors/syslog-health-monitor/pkg/kata/detector.go
mchmarny: agree, fail fast

labeler-module/pkg/labeler/labeler.go
mchmarny: k, will remove

health-monitors/syslog-health-monitor/pkg/kata/detector.go
mchmarny: Are we sure `katacontainers.io/kata-runtime` is always there regardless how kata was installed? Some nodes might have kata installed but not labeled I suppose too. 
Runtime detection (checking `containerRuntimeVersion`) gives us nice/single places to manage how we detect so rest of the code is isloated from the choices we made there

mchmarny: Based on GH search:

| Label | Example Value | Source / Context |
|-------|----------------|-----------------|
| **`katacontainers.io/kata-runtime=true`** | `"true"` | Most common, used by Kata project itself |
| **`runtime.kata.containers=enabled`** | `"enabled"` | Kada (v1.x) or older custom installer playbooks |
| **`io.katacontainers.runtime=true`** | `"true"` | Used in some Red Hat-based OpenShift environments |
| **`kata-runtime.handler=<handler-name>`** | `"kata-qemu"` or `"kata-clh"` | Sometimes in containerd/CRI-O runtime |
| **`containerd.kata.enabled=true`** | `"true"` | Used in containerd-based detection scripts |
| **`kubernetes.io/runtimeclass=kata-containers`** | `"kata-containers"` | Some automation tools seem to use that one |
| **`nvidia.com/gpu.workload.config=vm-passthrough`** | `"vm-passthrough"` | Kata + GPU Operator |
| **`node.kubernetes.io/virtualization=kata`** | `"kata"` | OpenShift 4.x |
| **`kata.installed=true`** | `"true"` | Seems to be common manual label for Kata-enabled node pools |

lalitadithya: I see... 

> Are we sure katacontainers.io/kata-runtime is always there regardless how kata was installed? Some nodes might have kata installed but not labeled I suppose too.

I think so. From what I can tell that it the official one and it used in all the examples. Can we start with `katacontainers.io/kata-runtime` since that is the most common and then add more labels as we start to see them. 

> Runtime detection (checking containerRuntimeVersion) gives us nice/single places to manage how we detect so rest of the code is isloated from the choices we made there

Yeah, I like this option too, but in the kata cluster that I have, I don't see the kata string in the version. I only see `containerRuntimeVersion: containerd://1.7.27` and looking at the containerd releases it doesn't look like they have kata specific releases, maybe you are looking at a different container runtime?

mchmarny: Removed runtime, not a good signal. Trim labels to the documented one and provided overide on label to use as an escape hatch

labeler-module/pkg/labeler/labeler.go
```diff
@@ -229,13 +245,73 @@ func (l *Labeler) getDriverLabelForNode(nodeName string) (string, error) {
...
 }
 
+// getKataLabelForNode detects if Kata is enabled on the specified node by checking node metadata.
+// This uses the Kubernetes API which is backed by client-go's cache, so it's efficient.
+// Returns "true" if Kata is enabled, "false" if not, or error if detection fails.
+func (l *Labeler) getKataLabelForNode(nodeName string) (string, error) {
+	ctx, cancel := context.WithTimeout(l.ctx, 5*time.Second)
+	defer cancel()
+
+	// Get node from Kubernetes API (backed by client-go cache)
```
lalitadithya: client-go isn't cached, informers are. Since this is a node-to-node label change and not a pod to node label change, you will likely have to set up a node informer similar to the pod one we have in `NewLabeler`

labeler-module/pkg/labeler/labeler.go
```diff
@@ -229,13 +245,73 @@ func (l *Labeler) getDriverLabelForNode(nodeName string) (string, error) {
...
+	}
+
+	// Fallback to annotations
+	kataAnnotations := []string{
+		"kata-runtime.io/enabled",
+		"io.katacontainers.config",
+	}
+
+	for _, annotation := range kataAnnotations {
+		if value, exists := node.Annotations[annotation]; exists && value != "" && stringutil.IsTruthyValue(value) {
```
lalitadithya: This needs to be a prefix match right -- https://github.com/confidential-containers/confidentialcontainers.org/blob/44bef545077033c03dc3d04781e383612ecd1f24/content/en/docs/features/initdata.md?plain=1#L144

So, the annotation is `io.katacontainers.config.foo.bar.baz`, so we will likely need to loop over the annotation list and find ones that match the prefix.

labeler-module/pkg/labeler/labeler_test.go
```diff
@@ -553,7 +553,7 @@ func TestLabeler_handlePodEvent(t *testing.T) {
 				require.NoError(t, err, "failed to update pod status")
 			}
 
-			labeler, err := NewLabeler(cli, time.Minute, "nvidia-dcgm", "nvidia-driver-daemonset")
+			labeler, err := NewLabeler(cli, time.Minute, "nvidia-dcgm", "nvidia-driver-daemonset", "")
```
lalitadithya, line 556: Can we add a test that tests the overrides?

health-monitors/syslog-health-monitor/pkg/kata/cache.go
lalitadithya, line 1: Can we move this into the labeler package so that we don't have dependencies between modules?

labeler-module/main.go
```diff
@@ -41,9 +41,12 @@ var (
 	date    = "unknown"
 
 	kubeconfig     = flag.String("kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
-	metricsPort    = flag.String("metrics-port", "2112", "port to expose Prometheus metrics on")
+	metricsPort    = flag.String("metrics-port", "2112", "Port to expose Prometheus metrics on")
 	dcgmAppLabel   = flag.String("dcgm-app-label", "nvidia-dcgm", "App label value for DCGM pods")
 	driverAppLabel = flag.String("driver-app-label", "nvidia-driver-daemonset", "App label value for driver pods")
+	kataLabel      = flag.String("kata-label", "",
+		"Custom node label to check for Kata Containers support. "+
+			"If empty, uses default 'nvsentinel.dgxc.nvidia.com/kata.enabled'")
```
lalitadithya: if empty it uses katacontainers.io/kata-runtime right?

labeler-module/pkg/labeler/labeler.go
```diff
@@ -383,7 +469,16 @@ func (l *Labeler) handlePodDeleteEvent(obj any) error {
 		return fmt.Errorf("failed to get driver label for node %s excluding deleted pod: %w", pod.Spec.NodeName, err)
 	}
 
-	return l.updateNodeLabels(pod.Spec.NodeName, expectedDCGMVersion, expectedDriverLabel)
+	// Kata detection is independent of pod state, so we detect it normally
+	expectedKataLabel, err := l.getKataLabelForNode(pod.Spec.NodeName)
```
lalitadithya: Do we still need this in the `handlePodDeleteEvent`?

labeler-module/pkg/labeler/labeler.go
```diff
@@ -408,5 +503,14 @@ func (l *Labeler) handlePodEvent(obj any) error {
 		return fmt.Errorf("failed to get driver label for node %s: %w", pod.Spec.NodeName, err)
 	}
 
-	return l.updateNodeLabels(pod.Spec.NodeName, expectedDCGMVersion, expectedDriverLabel)
+	// Detect Kata runtime on the node
+	expectedKataLabel, err := l.getKataLabelForNode(pod.Spec.NodeName)
```
lalitadithya: this should be in the node informer handler right? Do we need this in the `handlePodEvent`?

labeler-module/pkg/labeler/labeler.go
```diff
@@ -48,32 +54,49 @@ var (
...
 			options.LabelSelector = labelSelector.String()
 		}),
 	)
 
-	err = informerFactory.Core().V1().Pods().Informer().GetIndexer().AddIndexers(
+	// Create node informer factory (no filtering needed)
+	nodeInformerFactory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
+
+	podInformer := podInformerFactory.Core().V1().Pods().Informer()
+	nodeInformer := nodeInformerFactory.Core().V1().Nodes().Informer()
```
lalitadithya, line 96: We also need event handlers for the nodeInformer right?

labeler-module/pkg/labeler/labeler_test.go
```diff
@@ -637,3 +638,100 @@ func TestLabeler_handlePodEvent(t *testing.T) {
...
+		{
+			name:       "with custom override",
+			override:   "custom.io/kata-enabled",
+			wantLabels: []string{"katacontainers.io/kata-runtime", "custom.io/kata-enabled"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			clientset := fake.NewSimpleClientset()
```
lalitadithya: Please don't use the fake client, instead use the envtest client we have in the other test. The fake client has a lot of gaps, and it leads to very subtle bugs that we find in prod. envtest is much more robust option.

labeler-module/pkg/labeler/labeler_test.go
lalitadithya: can we add some tests for the kata labels in here please?

tilt/kata-runtimeclass.yaml
lalitadithya, line 1: We don't use the runtime detector anymore, right? Do we need this?

labeler-module/main.go
mchmarny: Correct

mchmarny: It's defaulted in the labeler module

lalitadithya: yep, the documentation says `nvsentinel.dgxc.nvidia.com/kata.enabled`

labeler-module/pkg/labeler/labeler_test.go
mchmarny: Other than the `TestKataLabelOverride` and `TestKataLabelOverrideIsolation` that are already there?

lalitadithya: yeah, maybe something that does the full flow, i.e., create a node, run the informer and then check the labels. Similar to what we have for the other labels.

labeler-module/pkg/labeler/labeler.go
mchmarny: Hm, I think so: 
* Every time we update node labels (add/update/delete events), we should set all three labels atomically
* On initial setup, if the first event for a node happens to be a pod deletion, kata label still needs to be set
* kata label value won't change, but including it ensures the label is always there

mchmarny: I could be missing something in lifecycle here though

lalitadithya: Hmm, so what I'm thinking is that we hook into the node lifecycle rather than the pod lifecycle. In the node informer we can have event handlers for nodeAdded and nodeUpdated. We don't care about the nodeDeleted. 

So, we will have the following scenarios:
1. New node joins the cluster with the label: the nodeAdded handler should be able to detect the kata label and apply our label
2. Node joins the cluster without the label, but is then updated to include the kata label: the nodeUpdated handler should be able to detect the label change and relabel the node
3. Node joins the cluster with the label but is then updated to remove the kata label: same as above. 

> Every time we update node labels (add/update/delete events), we should set all three labels atomically

I don't think we need to make this atomic since the labels stem from different lifecycles, for example, it is entirely possible for a node that has kata enable to never trigger the pod handlers because we look for very specific pods such as DCGM and the driver. In this scenario we will never label the node so we will end up losing monitoring on the node. Also, for the pod informer it is possible that the pod informer updates the driver and dcgm specific labels, but the state of the node doesn't change, for example, the driver starts to crash on the node, so the label is removed and, in such cases, we don't have to update/recompute the kata enablement state.

labeler-module/pkg/labeler/labeler.go
```diff
@@ -358,6 +464,66 @@ func (l *Labeler) updateNodeLabels(nodeName, expectedDCGMVersion, expectedDriver
...
 }
 
+// handleNodeEvent processes node events to update kata detection label
+func (l *Labeler) handleNodeEvent(obj any) error {
+	node, ok := obj.(*v1.Node)
+	if !ok {
+		return fmt.Errorf("node event: expected Node object, got %T", obj)
+	}
+
+	expectedKataLabel, err := l.getKataLabelForNode(node.Name)
```
lalitadithya: Since we already have the node, can just use the object, instead of passing the name and then doing a re-fetch from cache?

labeler-module/pkg/labeler/labeler.go
```diff
@@ -358,6 +464,66 @@ func (l *Labeler) updateNodeLabels(nodeName, expectedDCGMVersion, expectedDriver
...
+			slog.Info("Removing Kata enabled label from node", "node", nodeName)
+		} else {
+			node.Labels[KataEnabledLabel] = expectedKataLabel
+			slog.Info("Setting Kata enabled label on node", "node", nodeName, "kata", expectedKataLabel)
+		}
+
+		_, err = l.clientset.CoreV1().Nodes().Update(l.ctx, node, metav1.UpdateOptions{})
+
+		if err != nil {
+			return fmt.Errorf("failed to update node %s: %w", nodeName, err)
```
lalitadithya: We have return the error as is, if we wrap it, then the retry doesn't work i believe

labeler-module/pkg/labeler/labeler_test.go
```diff
@@ -637,3 +637,274 @@ func TestLabeler_handlePodEvent(t *testing.T) {
...
+					t.Logf("Node %s missing kata.enabled label", tt.node.Name)
+					return false
+				}
+
+				if kataLabel != tt.expectedKataVal {
+					t.Logf("Node %s has wrong kata label: got %s, want %s", tt.node.Name, kataLabel, tt.expectedKataVal)
+					return false
+				}
+
+				t.Logf("✓ Node %s correctly has kata.enabled=%s", tt.node.Name, kataLabel)
```
lalitadithya: nit: we can drop this log? The test log will be become large.

tilt/KATA_TESTING.md
lalitadithya, line 1: I think this is evident from the test case, do we want this md? What do you think?

labeler-module/pkg/labeler/labeler.go
mchmarny: good call

health-monitors/syslog-health-monitor/main.go
```diff
@@ -129,6 +131,16 @@ func run() error {
 		return fmt.Errorf("no checks defined in the config file")
 	}
 
+	// Add kata-specific journal tags if running in Kata mode
+	if *kataEnabled == "true" {
+		slog.Info("Kata mode enabled, adding containerd service filter to journal checks")
+
+		for i := range config.Checks {
+			// Add "-u containerd.service" tag to filter for containerd logs in systemd journal
+			config.Checks[i].Tags = append(config.Checks[i].Tags, "-u", "containerd.service")
```
lalitadithya: Can `.Tags` be nil? should we create a new array?

lalitadithya (approved):

PR #223, opened 2025-10-31
feat(platform-connectors): add node metadata enrichment for health ev…

### LINKED ISSUE #119: [Feature]: Enhance the health event with additional metadata about the node
### Prerequisites

### Feature Summary

Currently the health events contain the node name which works nicely for a limited number of small to medium clusters. As the fleet grows, it becomes harder to guarantee uniqueness of node names across the fleet, especially if the nodes names are simply IP addresses. This will make it harder to co-relate health event data across several clusters. Furthermore, in environments that have multiple interconnected racks, comparing health events across racks becomes cumbersome as manual lookup is needed to determine the which rack the health event originated from.

Adding the following information will be very helpful:
1. Provider ID ~~(bonus points if the provider ID is decoded)~~
2. A set of allow-listed node labels to include

### Problem/Use Case

The ability to more easily slice/dice and co-relate data stored in mongodb across clusters

### Proposed Solution

The health event proto already has a metadata field which can hold arbitrary key/value pairs and can be a good start for storing this information. The platform connectors can add additional metadata about the node before the health event is fanned out. The node metadata doesn't change frequently, so an adequate cache should be used to avoid too many requests. 

### Component

Core Service

Gyan172004:
## Summary

**Implements Feature Request #119**

This PR enriches health events with node metadata (provider IDs and labels) before sending them to connectors. This makes it easier to track and correlate GPU health issues across multiple clusters, especially when node names aren't unique.

**Key additions:**
- Fetches node metadata from Kubernetes API (`CoreV1().Nodes().Get()`) with configurable timeout and rate limiting
- Decodes cloud provider IDs into structured fields (extracts region, zone, instance ID from provider-specific formats)
- Implements LRU cache with TTL using `hashicorp/golang-lru` to minimize API load (default: 50 entries, 1-hour expiration)
- Adds double-check locking pattern with mutex to prevent duplicate API calls when multiple events arrive for the same node simultaneously
- Integrates at the platform connector level - enriches events before fanout to MongoDB, Kubernetes connector, and other backends
- Configurable allowlist of 17 default node labels covering topology, instance types, GPU metadata, and physical infrastructure
- Graceful error handling - metadata fetch failures don't block health event processing (logged as warnings)

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Health Monitors
- [X] Core Services
- [ ] Fault Management
- [ ] Documentation/CI
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Release Notes

* **New Features**
  * Added node metadata enrichment to health events, automatically capturing cloud provider IDs, topology labels, instance types, CUDA driver/runtime versions, and custom Kubernetes labels across AWS, GCP, Azure, and OCI environments.

* **Documentation**
  * Updated configuration documentation with node metadata enrichment settings, including cache configuration and allowed labels.


platform-connectors/pkg/nodemetadata/README.md
lalitadithya, line 1: Let's not maintain package level readmes, instead let's try and make the code as self documenting as possible

platform-connectors/pkg/nodemetadata/cache.go
```diff
@@ -0,0 +1,97 @@
...
+
+// GetOrFetch retrieves metadata from cache or fetches it if not present.
+// Uses double-check pattern to prevent duplicate fetches.
+func (c *Cache) GetOrFetch(ctx context.Context, nodeName string) (*NodeMetadata, error) {
+	// Fast path: check cache without lock
+	if metadata, found := c.lru.Get(nodeName); found {
+		return metadata, nil
+	}
+
+	// Slow path: acquire lock and check again
```
lalitadithya: Do we need these obvious comments? Isn't the code self-explanatory in most cases?

platform-connectors/pkg/nodemetadata/cache.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/golang-lru/v2/expirable"
+)
+
+// Cache wraps hashicorp/golang-lru for node metadata storage with coordinated fetching.
+type Cache struct {
```
lalitadithya: I'm not sure if we need this additional abstraction layer, the cache in this module is only going to be used for one thing and we what it is, so let's not create unnecessary abstractions. If we need the cache for something else, we can create the abstraction then.

platform-connectors/pkg/nodemetadata/cache_test.go
```diff
@@ -0,0 +1,490 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cache := NewCache(tt.cacheSize, tt.ttl, mockFetchFunc(t))
+			tt.setup(t, cache)
+			tt.verify(t, cache)
+		})
+	}
+}
+
+// ============================================================================
```
lalitadithya: These tests confuse me; these seem to be testing the functionality of the OSS package which is already tests. Can you please explain a bit further on why we need these tests?

platform-connectors/pkg/nodemetadata/config.go
```diff
@@ -0,0 +1,168 @@
...
+	DefaultQPS              = 5.0
+	DefaultBurst            = 10
+	DefaultMaxRetries       = 3
+	DefaultDecodeProviderID = true
+)
+
+// DefaultAllowedLabels defines the recommended set of node labels to include in health events.
+// These labels are carefully selected to provide maximum operational value while minimizing
+// metadata bloat and cache churn.
+var DefaultAllowedLabels = []string{
```
lalitadithya: The default should be defined in the helm chart and not in code.

platform-connectors/pkg/nodemetadata/config.go
```diff
@@ -0,0 +1,168 @@
...
+	"node.kubernetes.io/instance-type",   // Current standard
+	"beta.kubernetes.io/instance-type",   // Legacy
+
+	// GPU-specific labels (NVSentinel domain)
+	"nvidia.com/gpu.present",             // GPU node identification
+	"nvidia.com/gpu.deploy.dcgm",         // DCGM deployment topology
+	"nvidia.com/gpu.deploy.driver",       // Driver deployment topology
+	"nvidia.com/gpu.count",               // Number of GPUs
+	"nvsentinel.dgxc.nvidia.com/dcgm.version",     // DCGM version tracking
+	"nvsentinel.dgxc.nvidia.com/driver.installed", // Driver installation status
```
lalitadithya: We don't need these, can we instead pick the ones that have cuda/driver versions on them?

platform-connectors/pkg/nodemetadata/config.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// Physical topology (if available in your cluster)
+	"rack",                               // Physical rack identifier
+	"datacenter",                         // Datacenter identifier
+	"row",                                // Physical row in datacenter
+	"capacity-tranche",                   // Capacity reservation group
```
lalitadithya: These are likely custom, and we shouldn't add as defaults, user can include this if required.

platform-connectors/pkg/nodemetadata/config.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// Workload classification (useful for cost tracking)
+	"workload-type",                      // Workload classification
```
lalitadithya: Same comment as earlier about the custom labels

platform-connectors/pkg/nodemetadata/processor.go
```diff
@@ -0,0 +1,153 @@
...
+	stopCh    chan struct{}
+}
+
+// NewProcessor creates a new node metadata processor.
+func NewProcessor(ctx context.Context, config *Config) (Processor, error) {
+	if err := config.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid config: %w", err)
+	}
+
+	restConfig, err := rest.InClusterConfig()
```
lalitadithya: We need to build from flags using clientcmd. I suspect we already have another client that is already built with the correct QPS etc, let's use the same.

platform-connectors/pkg/nodemetadata/processor.go
```diff
@@ -0,0 +1,153 @@
...
+				event.Metadata[k] = v
+			}
+		} else {
+			event.Metadata["node.providerID"] = metadata.ProviderID
+		}
+	}
+
+	for _, labelKey := range p.config.AllowedLabels {
+		if labelValue, exists := metadata.Labels[labelKey]; exists {
+			event.Metadata[fmt.Sprintf("node.label.%s", labelKey)] = labelValue
```
lalitadithya: Why do we want the custom prefix `node.label.`, can't we use the node labels as is?

platform-connectors/pkg/nodemetadata/processor_test.go
```diff
@@ -0,0 +1,697 @@
...
+				"topology.kubernetes.io/region": "us-west-2",
+				"node.kubernetes.io/instance-type": "p4d.24xlarge",
+			},
+		},
+		Spec: corev1.NodeSpec{
+			ProviderID: "aws:///us-west-2a/i-1234567890abcdef0",
+		},
+	}
+
+	clientset := fake.NewSimpleClientset(node)
```
lalitadithya: no -- use envtest like it is done in the other packages

platform-connectors/pkg/nodemetadata/provider.go
lalitadithya, line 1: I don't think we this decoding, if decoding is required, it can be done by the consumers of the health events.

platform-connectors/pkg/nodemetadata/processor_test.go
Gyan172004: Okay, so I should use envtest for integration tests. Should I remove these unit tests using fake nodes?

lalitadithya: we can use env test in both unit test and integration test. Anywhere we need access to the kube apis, let's use envtest and not the fake client

platform-connectors/pkg/nodemetadata/provider.go
Gyan172004: In the original issue **#119**, it mentioned:  
> *"Provider ID (bonus points if the provider ID is decoded)"*  

So I had implemented the decoding as an **optional feature**.

Since I’ll be adding the `decodeProviderID` config flag in the Helm chart (defaults to `true`) I wanted to confirm which direction we should take:
**Option A:** Keep it optional (operators can choose to decode or not)  
**Option B:** Remove the decoding entirely (always store the raw `providerID`)

lalitadithya: Ahh, I see. Since the issue was filed we realized that not all IDs can be decoded, for example, OCI's IDs are opaque. So let's drop it for now. I will update the GH issue. Apologies for the confusion

platform-connectors/pkg/nodemetadata/README.md
Gyan172004: Removed the package-level README. The code is now self-documenting and updated in main README.

platform-connectors/pkg/nodemetadata/cache.go
Gyan172004: Agreed and removed. Only kept comments where logic isn't immediately clear.

platform-connectors/pkg/nodemetadata/cache.go
Gyan172004: Agree! Removed cache.go and moved the logic directly into processor.go.

platform-connectors/pkg/nodemetadata/cache_test.go
Gyan172004: Understood, there is no need to test the already-tested OSS package, deleted cache_test.go.

platform-connectors/pkg/nodemetadata/config.go
Gyan172004: Fixed! Removed DefaultAllowedLabels from config.go and moved defaults to values.yaml

platform-connectors/pkg/nodemetadata/config.go
Gyan172004: removed custom labels (rack, datacenter, row, capacity-tranche, workload-type).

platform-connectors/pkg/nodemetadata/config.go
Gyan172004: removed custom labels (rack, datacenter, row, capacity-tranche, workload-type).

platform-connectors/pkg/nodemetadata/processor.go
Gyan172004: Done, NewProcessor now accepts the shared clientset from main.go. Removed InClusterConfig() and all QPS/Burst/Timeout configs.

platform-connectors/pkg/nodemetadata/processor.go
Gyan172004: Removed, Labels are now stored directly as Kubernetes provides them (e.g., "topology.kubernetes.io/zone").

platform-connectors/pkg/nodemetadata/processor_test.go
Gyan172004: Converted all tests to use envtest with testmain setup. Removed all fake.NewSimpleClientset usage.

platform-connectors/pkg/nodemetadata/provider.go
Gyan172004: Understood! Removed provider.go, provider_test.go, and all decoding logic. Now storing raw providerID only.

platform-connectors/pkg/nodemetadata/config.go
Gyan172004: Done, Added the NVIDIA CUDA/driver version labels (checked with glen):
 - "nvidia.com/cuda.driver-version.full"                       - "nvidia.com/cuda.runtime.major"
 - "nvidia.com/cuda.runtime.minor"                             - "nvidia.com/cuda.driver.major"
 - "nvidia.com/cuda.driver.minor"                                 - "nvidia.com/cuda.driver.rev"

lalitadithya: Can you check the official GPU operator docs on their github repo and confirm the labels, please?

README.md
```diff
@@ -42,6 +42,7 @@ helm show chart oci://ghcr.io/nvidia/nvsentinel --version v0.2.0
 - **⚡ Real-time Processing**: Event-driven architecture with immediate fault response
 - **📊 Persistent Storage**: MongoDB-based event store with change streams for real-time updates
 - **🛡️ Graceful Handling**: Coordinated workload eviction with configurable timeouts
+- **🏷️ Metadata Enrichment**: Automatic augmentation of health events with cloud provider and node topology information
```
lalitadithya:
```suggestion
- **🏷️ Metadata Enrichment**: Automatic augmentation of health events with cloud provider and node metadata information
```

platform-connectors/main.go
```diff
@@ -123,7 +125,35 @@ func initializeMongoDBConnector(
...
+		return nil, fmt.Errorf("failed to create node metadata config: %w", err)
+	}
+
+	if !cfg.Enabled {
+		slog.Info("Node metadata augmentation is disabled")
+		return nil, nil
+	}
+
+	if clientset == nil {
+		return nil, fmt.Errorf("node metadata augmentation requires K8s connector to be enabled")
```
lalitadithya: This should not be an error, we should gracefully skip. Also, can we make this check before this function is called since the arg is passed into the funcion?

platform-connectors/pkg/nodemetadata/types.go
lalitadithya, line 1: Let's call this `platform-connectors/pkg/nodemetadata/types.go`

platform-connectors/pkg/nodemetadata/processor.go
```diff
@@ -0,0 +1,157 @@
...
+	stopCh    chan struct{}
+}
+
+// NewProcessor creates a new node metadata processor using an existing Kubernetes client.
+func NewProcessor(ctx context.Context, config *Config, clientset kubernetes.Interface) (Processor, error) {
+	if err := config.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid config: %w", err)
+	}
+
+	if clientset == nil {
```
lalitadithya: Can we validate this in the caller? I think I have already seen a nil check for the clientset, do we need it twice?

platform-connectors/pkg/nodemetadata/processor.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get metadata for node %s: %w", event.NodeName, err)
+	}
+
+	if event.Metadata == nil {
+		event.Metadata = make(map[string]string)
+	}
+
+	if metadata.ProviderID != "" {
+		event.Metadata["node.providerID"] = metadata.ProviderID
```
lalitadithya:
```suggestion
		event.Metadata["providerID"] = metadata.ProviderID
```

platform-connectors/pkg/nodemetadata/kube.go
```diff
@@ -0,0 +1,157 @@
...
+				metadata.Labels[labelKey] = labelValue
+			}
+		}
+	}
+
+	return metadata, nil
+}
+
+// LRU library handles TTL expiration automatically.
+func (p *processor) Start(ctx context.Context) {
```
lalitadithya: I don't follow why we need this function. This function doesn't seem to do much other than printing logs, can you please explain?

platform-connectors/pkg/nodemetadata/processor_test.go
```diff
@@ -0,0 +1,671 @@
...
+	ctx := context.Background()
+	created, err := testClient.CoreV1().Nodes().Create(ctx, node, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatalf("Failed to create test node: %v", err)
+	}
+	return created
+}
+
+// cleanupTestNode deletes a node from the test API server
+func cleanupTestNode(t *testing.T, nodeName string) {
```
lalitadithya:
```suggestion
func deleteTestNode(t *testing.T, nodeName string) {
```

platform-connectors/pkg/nodemetadata/processor_test.go
```diff
@@ -0,0 +1,671 @@
...
+	)
+
+	ctx := context.Background()
+	event := &pb.HealthEvent{
+		NodeName: "test-node",
+		Metadata: make(map[string]string),
+	}
+
+	err := p.AugmentHealthEvent(ctx, event)
+	if err != nil {
```
lalitadithya: Please use the assert library to check for conditions.

Please update everywhere

platform-connectors/pkg/nodemetadata/processor_test.go
```diff
@@ -0,0 +1,671 @@
...
+	p.Stop()
+
+	select {
+	case <-p.stopCh:
+	default:
+		t.Error("expected stopCh to be closed")
+	}
+}
+
+func TestProcessorCachingBehavior(t *testing.T) {
```
lalitadithya: I don't follow this test, what is this supposed to be testing?

platform-connectors/pkg/nodemetadata/processor_test.go
```diff
@@ -0,0 +1,671 @@
...
+	if event.Metadata == nil {
+		t.Error("expected metadata to be initialized")
+	}
+
+	if event.Metadata["node.providerID"] != "aws:///us-west-2a/i-1234567890abcdef0" {
+		t.Errorf("expected provider ID to be set")
+	}
+}
+
+func TestProcessorAugmentHealthEventExistingMetadata(t *testing.T) {
```
lalitadithya: A lot of these tests seem to have the same setup and teardown, can we use table driven tests instead?

platform-connectors/pkg/nodemetadata/kube.go
lalitadithya, line 1: since this processor is specific to kube, let's call it 

platform-connectors/pkg/nodemetadata/kube.go

platform-connectors/pkg/nodemetadata/config_test.go
```diff
@@ -0,0 +1,178 @@
...
+		{
+			name: "allowed labels",
+			cfgMap: map[string]interface{}{
+				"nodeMetadataAllowedLabels": []interface{}{
+					"topology.kubernetes.io/zone",
+					"topology.kubernetes.io/region",
+				},
+			},
+			validate: func(t *testing.T, cfg *Config) {
+				if len(cfg.AllowedLabels) != 2 {
```
lalitadithya: Please use assert to validate the expected conditions for tests

platform-connectors/main.go
```diff
@@ -154,36 +186,42 @@ func initializeConnectors(
...
 		err            error
 	)
 
 	if config["enableK8sPlatformConnector"] == True {
-		k8sRingBuffer, err = initializeK8sConnector(ctx, config, stopCh)
+		k8sRingBuffer, clientset, err = initializeK8sConnector(ctx, config, stopCh)
 		if err != nil {
-			return nil, nil, fmt.Errorf("failed to initialize K8s connector: %w", err)
+			return nil, nil, nil, fmt.Errorf("failed to initialize K8s connector: %w", err)
 		}
```
lalitadithya, line 207: Can we initialize the processor here so that we don't have to pass the clientset everywhere?

Also, can we use a factory pattern here so that it is easier add different processors (e.g., Slurm) in future?

platform-connectors/pkg/nodemetadata/processor_test.go
Gyan172004: wanted to test that cached node metadata matches freshly fetched metadata for the same node. But I agree its a weak assertion. Just checking event1.Metadata == event2.Metadata is obvious, of course they're the same for the same node. The test doesn't actually verify caching behavior (like ensuring the K8s API is only called once). 
 
 **will remove it during the table-driven refactor.**

platform-connectors/pkg/nodemetadata/kube.go
Gyan172004: Originally needed for a custom cache(i was planning to make) that required background cleanup. Now with hashicorp LRU, its not strictly necessary, but I kept it to match the lifecycle pattern used by other connectors in main.go. 
Happy to remove if you think it's overkill.

distros/kubernetes/nvsentinel/values.yaml
```diff
@@ -108,5 +108,52 @@ platformConnector:
...
+
+# Node selector for platform connector pods (overrides global.systemNodeSelector if set)
+nodeSelector: {}
+
+# Tolerations for platform connector pods (merged with global.systemNodeTolerations)
+tolerations: []
+
+# Affinity rules for platform connector pods
+affinity: {}
+
```
lalitadithya: Why is this change needed?

README.md
Gyan172004: done

platform-connectors/main.go
Gyan172004: handled

platform-connectors/pkg/nodemetadata/types.go
Gyan172004: done

platform-connectors/pkg/nodemetadata/processor.go
Gyan172004: yes, done

platform-connectors/pkg/nodemetadata/processor.go
Gyan172004: corrected

platform-connectors/pkg/nodemetadata/processor_test.go
Gyan172004: done

platform-connectors/pkg/nodemetadata/processor_test.go
Gyan172004: done

platform-connectors/pkg/nodemetadata/processor_test.go
Gyan172004: yes done

platform-connectors/pkg/nodemetadata/kube.go
Gyan172004: done

platform-connectors/pkg/nodemetadata/config_test.go
Gyan172004: done

platform-connectors/main.go
Gyan172004: yes, factory pattern is perfect design here
done

distros/kubernetes/nvsentinel/values.yaml
Gyan172004: idk it came while resolving merge conflicts.
removed these changes

platform-connectors/pkg/nodemetadata/kube.go
lalitadithya: It's overkill, let's remove it.

lalitadithya:
DCO is failing, can you please check?

data-models/pkg/model/health_event_extentions.go
```diff
@@ -53,3 +55,144 @@ type HealthEventWithStatus struct {
 	HealthEvent       *protos.HealthEvent `bson:"healthevent,omitempty"`
 	HealthEventStatus HealthEventStatus   `bson:"healtheventstatus"`
 }
+
+// MarshalBSON implements custom BSON marshaling for HealthEventWithStatus.
+// This is necessary because protobuf-generated structs don't have BSON tags,
+// so we need to manually convert the HealthEvent to a BSON-compatible format.
```
lalitadithya: I don't follow this. So far on main we've never had this issue that needs manual marshalling/unmarshalling. The library was smart enough to skip the non-exported. Can you please elaborate on what changed recently that now requires us to make this change?

data-models/pkg/model/health_event_extentions.go
```diff
@@ -53,3 +55,144 @@ type HealthEventWithStatus struct {
...
+	// Marshal the complete document
+	return bson.Marshal(bson.M{
+		"createdAt":         h.CreatedAt,
+		"healthevent":       healthEventDoc,
+		"healtheventstatus": h.HealthEventStatus,
+	})
+}
+
+// UnmarshalBSON implements custom BSON unmarshaling for HealthEventWithStatus.
+// This is the reverse operation - converting BSON back to protobuf struct.
```
lalitadithya: Same comment as earlier

distros/kubernetes/nvsentinel/values-tilt.yaml
```diff
@@ -15,6 +15,7 @@
 global:
   dryRun: false
   kubeVersion: 1.31.0
+  installMongoDB: true
```
lalitadithya: This is not needed with the recent changes on main. This is likely a remanent of a bad merge, can you please check?

distros/kubernetes/nvsentinel/values-tilt.yaml
```diff
@@ -123,3 +124,10 @@ csp-health-monitor:
   logLevel: debug
   quarantineTriggerEngine:
     logLevel: debug
+
+# Enable node metadata enrichment for testing
+platformConnector:
+  mongodbStore:
+    enabled: true  # Enable mongodb-store subchart deployment
```
lalitadithya: Do we need the obvious comment? Can we drop obvious comments everywhere in net new code?

distros/kubernetes/nvsentinel/values.yaml
```diff
@@ -21,54 +21,136 @@ global:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  
+  # Node selector for control plane components (platform-connectors, monitoring services)
+  # Used to schedule critical infrastructure components on specific nodes
```
lalitadithya: We have a values-full that documents all the changes, let's not add the documentation in all values files.

kubelogin-darwin-arm64.zip
lalitadithya, line 1: This should not be checked in, plesae remove

platform-connectors/go.mod
```diff
@@ -6,11 +6,12 @@ toolchain go1.25.3
 
 require (
 	github.com/golang/protobuf v1.5.4
+	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/nvidia/nvsentinel/commons v0.0.0
 	github.com/nvidia/nvsentinel/data-models v0.0.0
 	github.com/prometheus/client_golang v1.23.2
 	github.com/stretchr/testify v1.11.1
-	go.mongodb.org/mongo-driver v1.17.4
+	go.mongodb.org/mongo-driver v1.17.6
```
lalitadithya: Why did we do this upgrade?

platform-connectors/main.go
```diff
@@ -123,7 +132,45 @@ func initializeMongoDBConnector(
...
+		return nil, fmt.Errorf("failed to create node metadata config: %w", err)
+	}
+
+	if !cfg.Enabled {
+		slog.Info("Node metadata augmentation is disabled")
+
+		return nil, nil
+	}
+
+	if clientset == nil {
```
lalitadithya: In what cases would this code path be called, but the clientset be nil?

platform-connectors/main.go
```diff
@@ -305,6 +366,7 @@ func run() error {
 		return nil
 	})
 
-	// Wait for both goroutines to finish
 	return g.Wait()
 }
+
+// Trigger rebuild for MongoDB serialization fix - 1762118253
```
lalitadithya: please remove

platform-connectors/pkg/nodemetadata/factory.go
```diff
@@ -0,0 +1,91 @@
...
+	factory.Register(ProcessorTypeKubernetes, newKubernetesProcessor)
+	
+	return factory
+}
+
+// Register adds a new processor type to the factory.
+// This allows external packages to register custom processor types.
+func (f *ProcessorFactory) Register(pType ProcessorType, creator ProcessorCreator) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
```
lalitadithya: I'm not sure I follow this -- in what places in the code, do we have multiple go routines calling this function?

platform-connectors/pkg/nodemetadata/factory.go
```diff
@@ -0,0 +1,91 @@
...
+
+const (
+	// ProcessorTypeKubernetes creates a processor that fetches metadata from Kubernetes API.
+	ProcessorTypeKubernetes ProcessorType = "kubernetes"
+	// ProcessorTypeSlurm can be added in the future for Slurm-based clusters.
+	// ProcessorTypeSlurm ProcessorType = "slurm"
+)
+
+// ProcessorCreator is a function that creates a processor of a specific type.
+type ProcessorCreator func(ctx context.Context, config *Config, params interface{}) (Processor, error)
```
lalitadithya: I don't follow this, we know what types of processors we have in code, why can't we call the initialize functions when required? Do we need to make it this generic?

platform-connectors/pkg/nodemetadata/kube.go
```diff
@@ -0,0 +1,190 @@
...
+	clientset kubernetes.Interface
+	cache     *expirable.LRU[string, *NodeMetadata]
+	fetchMu   sync.Mutex
+	stopCh    chan struct{}
+}
+
+// newKubernetesProcessor creates a new Kubernetes-based node metadata processor.
+// This function is package-private and called by the factory.
+func newKubernetesProcessor(ctx context.Context, config *Config, params interface{}) (Processor, error) {
+	if config == nil {
```
lalitadithya: When would this be nil?

platform-connectors/pkg/nodemetadata/kube.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil, fmt.Errorf("config cannot be nil")
+	}
+
+	if err := config.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid config: %w", err)
+	}
+
+	clientset, ok := params.(kubernetes.Interface)
```
lalitadithya: No casts like this, please. Let's use the correct types in the function args

platform-connectors/pkg/nodemetadata/kube.go
```diff
@@ -0,0 +1,190 @@
...
+		"type", "kubernetes",
+		"cacheSize", config.CacheSize,
+		"cacheTTL", config.CacheTTL,
+		"allowedLabels", config.AllowedLabels)
+
+	return p, nil
+}
+
+// NewProcessor creates a new node metadata processor using an existing Kubernetes client.
+// Deprecated: Use ProcessorFactory.CreateProcessor instead.
```
lalitadithya: The feature was never merged; there is no reason for backward compatibility. Can you please drop all backward compatibility specific code and make it simpler?

platform-connectors/pkg/nodemetadata/kube.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	for _, labelKey := range p.config.AllowedLabels {
+		if labelValue, exists := metadata.Labels[labelKey]; exists {
+			event.Metadata[labelKey] = labelValue
+		}
+	}
+
+	// Debug logging to verify metadata augmentation
+	slog.Info("✅ Node metadata augmented successfully",
```
lalitadithya: No emojis in logs. Please remove everywhere.

platform-connectors/pkg/nodemetadata/kube_test.go
```diff
@@ -0,0 +1,487 @@
...
+func createTestNode(t *testing.T, node *corev1.Node) {
+	t.Helper()
+	_, err := testClient.CoreV1().Nodes().Create(context.Background(), node, metav1.CreateOptions{})
+	require.NoError(t, err, "failed to create test node")
+}
+
+func deleteTestNode(t *testing.T, nodeName string) {
+	t.Helper()
+	err := testClient.CoreV1().Nodes().Delete(context.Background(), nodeName, metav1.DeleteOptions{})
+	if err != nil {
```
lalitadithya: please use asserts

platform-connectors/pkg/nodemetadata/kube_test.go
```diff
@@ -0,0 +1,487 @@
...
+	ctx := context.Background()
+
+	// First call - cache miss
+	event1 := &pb.HealthEvent{
+		NodeName: "cache-test-node",
+		Metadata: make(map[string]string),
+	}
+	require.NoError(t, p.AugmentHealthEvent(ctx, event1))
+
+	// Second call - cache hit
```
lalitadithya: How do we know it was picked up from cache?

tilt/Tiltfile
lalitadithya, line 1: Why was this deleted?

data-models/pkg/model/health_event_extentions.go
Gyan172004: Its working withour marshalling as well now :)
Removed that part

.gitignore
```diff
@@ -426,6 +426,9 @@ terraform.rc
 digests.txt
 images.json
 
+# Downloaded binaries
+kubelogin*.zip
```
lalitadithya: the gitignore file should only have project specific ignores, let's not add global not specific binaries to this.

data-models/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.25
 toolchain go1.25.3
 
 require (
+	go.mongodb.org/mongo-driver v1.17.4
```
lalitadithya: I'm not sure about this change. The data models should be agnostic of the database

platform-connectors/pkg/nodemetadata/factory.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+
+	"k8s.io/client-go/kubernetes"
+)
+
+// NewProcessor creates a node metadata processor.
+// Currently supports Kubernetes; can be extended for other platforms (e.g., Slurm) in the future.
+func NewProcessor(ctx context.Context, config *Config, clientset kubernetes.Interface) (Processor, error) {
```
lalitadithya: There may have been some miscommunication. When I said let's simplify this, I meant make a simple factory like we have here -- https://github.com/NVIDIA/NVSentinel/blob/main/janitor/pkg/csp/client.go

platform-connectors/pkg/nodemetadata/kube.go
lalitadithya: bump

platform-connectors/pkg/nodemetadata/kube.go
```diff
@@ -0,0 +1,152 @@
...
+	cache     *expirable.LRU[string, *NodeMetadata]
+	fetchMu   sync.Mutex
+}
+
+func newKubernetesProcessor(ctx context.Context, config *Config, clientset kubernetes.Interface) (Processor, error) {
+	if config == nil {
+		return nil, fmt.Errorf("config cannot be nil")
+	}
+
+	if clientset == nil {
```
lalitadithya: Same question as earlier

Gyan172004: Coderabbit mentioned to handle this everywhere.
I will remove it, clientset will always be non nil

Gyan172004: done

platform-connectors/pkg/nodemetadata/factory.go
Gyan172004: done, implementef factory design in switch style

distros/kubernetes/nvsentinel/values-tilt.yaml
```diff
@@ -133,3 +133,10 @@ csp-health-monitor:
   logLevel: debug
   quarantineTriggerEngine:
     logLevel: debug
+
+# Enable node metadata enrichment for testing
+platformConnector:
+  mongodbStore:
+    enabled: true
```
lalitadithya: mongodb is already enabled with the global flag now, please check on main. We only need to enable the nodeMetadata

platform-connectors/main.go
```diff
@@ -76,34 +78,43 @@ func loadConfig(configFilePath string) (map[string]interface{}, error) {
...
 	}
 
 	go k8sConnector.FetchAndProcessHealthMetric(ctx)
 
-	return k8sRingBuffer, nil
+	// Node metadata enrichment is optional - failures are logged but don't abort startup
+	processor, err := initializeNodeMetadataProcessor(ctx, config, clientset)
+	if err != nil {
+		slog.Warn("Failed to initialize node metadata processor, continuing without enrichment", "error", err)
+		processor = nil
```
lalitadithya: If there is no error, the function `initializeNodeMetadataProcessor` should return nil, we don't have to explicitly set it to nil

platform-connectors/main.go
```diff
@@ -154,35 +192,37 @@ func initializeConnectors(
...
-	return k8sRingBuffer, storeConnector, nil
+	return k8sRingBuffer, storeConnector, processor, nil
 }
 
 func cleanupResources(
 	socket string,
 	lis net.Listener,
 	k8sRingBuffer *ringbuffer.RingBuffer,
 	storeConnector *store.MongoDbStoreConnector,
+	processor nodemetadata.Processor,
```
lalitadithya: I don't think this arg is used in this function, can you please check and remove if it is not used?

platform-connectors/pkg/nodemetadata/kube.go
```diff
@@ -0,0 +1,148 @@
...
+
+type processor struct {
+	config    *Config
+	clientset kubernetes.Interface
+	cache     *expirable.LRU[string, *NodeMetadata]
+	fetchMu   sync.Mutex
+}
+
+func newKubernetesProcessor(ctx context.Context, config *Config, clientset kubernetes.Interface) (Processor, error) {
+	if config == nil {
```
lalitadithya: Can we make sure the caller never passed in nil config instead of adding a check here and erroring out?

platform-connectors/pkg/nodemetadata/kube_test.go
```diff
@@ -0,0 +1,471 @@
...
+	p := createTestProcessor(config)
+	ctx := context.Background()
+
+	event1 := &pb.HealthEvent{
+		NodeName: "cache-test-node",
+		Metadata: make(map[string]string),
+	}
+	require.NoError(t, p.AugmentHealthEvent(ctx, event1))
+
+	// Second call for same node - verify cache returns consistent results
```
lalitadithya: How do we know the cache is used?

platform-connectors/pkg/nodemetadata/kube_test.go
```diff
@@ -0,0 +1,471 @@
...
+			assert.NoError(t, err)
+			assert.Equal(t, "aws:///us-west-2a/i-concurrent123", event.Metadata["providerID"])
+			assert.Equal(t, "test-value", event.Metadata["test-label"])
+		}()
+	}
+
+	wg.Wait()
+}
+
+func TestProcessorMultipleNodesEnrichment(t *testing.T) {
```
lalitadithya: Can this be merged into the other table-driven test function?

lalitadithya (commented):
looks much better now, please rebase.

platform-connectors/pkg/nodemetadata/kube_test.go
Gyan172004: This was just to check consistency.
I will add to test cache as well, delete previous result and verify no more api call is made.

platform-connectors/pkg/nodemetadata/kube_test.go
Gyan172004: Merged into table-driven test as 'multiple nodes enrichment' test case.

platform-connectors/pkg/nodemetadata/kube.go
Gyan172004: Removed the nil config check from newKubernetesProcessor. The caller (initializeNodeMetadataProcessor) creates the config via NewConfigFromMap before passing it, guaranteeing it's never nil.

platform-connectors/main.go
Gyan172004: yes its not used, removed

platform-connectors/main.go
Gyan172004: removed the redundant assignment.

distros/kubernetes/nvsentinel/values-tilt.yaml
Gyan172004: Removed. Now only nodeMetadata.enabled under platformConnector.

platform-connectors/pkg/nodemetadata/kube.go
```diff
@@ -0,0 +1,157 @@
...
+
+	metadata, err := p.fetchNodeMetadata(ctx, nodeName)
+	if err != nil {
+		return nil, err
+	}
+
+	p.cache.Add(nodeName, metadata)
+	return metadata, nil
+}
+
```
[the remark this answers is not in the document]
Gyan172004: The expirable.LRU documentation explicitly states it 'implements a thread-safe LRU with expirable entries' (see go doc). The cache handles its own internal locking, so i think no additional mutex is needed around Get/Add calls. The existing fetchMu prevents duplicate API calls, which is separate from cache thread-safety.

platform-connectors/pkg/nodemetadata/processor.go
```diff
@@ -0,0 +1,157 @@
...
+		stopCh:    make(chan struct{}),
+	}
+
+	slog.Info("Node metadata processor initialized",
+		"cacheSize", config.CacheSize,
+		"cacheTTL", config.CacheTTL,
+		"allowedLabels", config.AllowedLabels)
+
+	return p, nil
+}
```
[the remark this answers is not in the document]
Gyan172004: Removed per earlier feedback to ensure the caller guarantees non-nil config instead of adding defensive checks here.

platform-connectors/main.go
```diff
@@ -123,7 +125,35 @@ func initializeMongoDBConnector(
...
+	}
+
+	processor, err := nodemetadata.NewProcessor(ctx, cfg, clientset)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create node metadata processor: %w", err)
+	}
+
+	slog.Info("Node metadata processor initialized successfully")
+	return processor, nil
+}
```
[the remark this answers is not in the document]
Gyan172004: 1.Already implemented. The code logs a warning and returns nil on processor initialization failure, allowing the service to continue without enrichment

2.done

lalitadithya:
@Gyan172004 , can you please include https://github.com/NVIDIA/NVSentinel/issues/119#issuecomment-3484128843 in this PR?

lalitadithya (approved):

PR #280, opened 2025-11-05
feat: Add ability to cancel breakfix pipeline with manual uncordon

### LINKED ISSUE #141: [Bug]: Add ability to cancel breakfix pipeline manually
### Prerequisites

## Manual Node Uncordon Not Propagated to Node Drainer and Fault Remediation Modules

When an operator manually uncordons a node (using `kubectl uncordon`) that was previously cordoned by the Fault Quarantine (FQ) module, the uncordon event is only registered by FQ but not propagated to downstream modules (Node Drainer and Fault Remediation). This can lead to stale state and incorrect behavior in the remediation pipeline.

### Current System Behavior

The normal event flow is:
1. Health Monitor detects a fatal health event
2. Platform Connector forwards event to MongoDB
3. Modules consume events in order: **Fault Quarantine → Node Drainer → Fault Remediation**
4. Fault Remediation creates a maintenance CR (e.g., RebootNode)
5. After remediation, Health Monitor sends a healthy event
6. Fault Quarantine uncordons the node upon receiving the healthy event

### Problem

The FQ module has a [callback](https://github.com/NVIDIA/NVSentinel/blob/v0.2.0/fault-quarantine-module/pkg/reconciler/reconciler.go#L170) that detects manual uncordon operations and updates its own state (removes taint, clears `quarantineHealthEvent` and `quarantineHealthEventIsCordoned` annotations, adds `quarantinedNodeUncordonedManually` annotation). However, this information is **not propagated** to Node Drainer (ND) or Fault Remediation (FR) modules if the Health Monitor doesn't send a healthy event.

#### Short-term Fix
Add callbacks in Node Drainer and Fault Remediation modules (similar to FQ's callback) to detect and handle manual uncordon events independently.

**Pros:**
- Quick to implement
- Each module maintains its own state consistency

**Cons:**
- Duplicates uncordon detection logic across modules
- Doesn't address the root cause of missing event propagation

#### Long-term Fix
Modify the event stream architecture to naturally handle manual uncordon events:
- When a manual uncordon, generate an unquarantine event in MongoDB
- Fault handling modules (FQ, ND, FR) consume this event and update their state accordingly
- Ensures all modules stay synchronized through the standard event pipeline

**Pros:**
- Centralized event handling
- Maintains architectural consistency
- Easier to maintain and extend

**Cons:**
- Requires changes to event schema and flow
- More complex implementation

### Impact
- **Severity:** Medium-High
- **Frequency:** Occurs whenever operators manually intervene with uncordon operations
- **Affected Components:** Node Drainer, Fault Remediation

### Component

Fault Management

#### Scenario 1: Node Drainer Stuck Waiting

1. Health Monitor sends a fatal health event
2. FQ cordons the node
3. Node Drainer starts waiting for pods to finish (assuming `AllowCompletion` mode)
4. Operator manually uncordons the node using `kubectl uncordon <node>`
5. FQ registers the uncordon and updates its state
6. **Problem:** Node Drainer continues waiting indefinitely for pods to complete, unaware that the node was manually uncordoned

#### Scenario 2: Fault Remediation Annotation Not Cleared

1. Health Monitor sends a fatal health event
2. FQ cordons the node
3. ND drains the node successfully
4. FR creates a RebootNode CR and the node is remediated
5. Health Monitor doesn't send a healthy event for some reason
6. Operator manually uncordons the node using `kubectl uncordon <node>`
7. FQ registers the manual uncordon, but FR doesn't receive an unquarantine event
8. **Problem:** [Annotation cleanup code](https://github.com/NVIDIA/NVSentinel/blob/main/fault-remediation-module/pkg/reconciler/reconciler.go#L119) is not executed, leaving stale `latestFaultRemediationState` annotation on the node
9. Later, Health Monitor sends another fatal event
10. Node is cordoned and drained again
11. When FR [checks for existing CR status](https://github.com/NVIDIA/NVSentinel/blob/main/fault-remediation-module/pkg/reconciler/reconciler.go#L387), it finds the stale annotation indicating remediation is already in progress
12. **Result:** FR skips creating a new maintenance CR, and the node is never remediated

### Expected Behavior

When a node is manually uncordoned by an operator:
- FQ should detect the manual uncordon
- ND should stop waiting/draining operations for that node
- FR should clear remediation state annotations
- All modules should be consistent with the node's actual state

### Actual Behavior

- Only FQ detects and handles manual uncordon
- ND and FR remain unaware, leading to stuck operations or stale state
- No mechanism exists to propagate manual uncordon events through the pipeline

### Environment

- NVSentinel version: v0.1.0 / 1.167.1
- Kubernetes version: 1.31.1
- Deployment method: Argo

### Logs/Output

```
I1023 11:41:02.381627       1 reconciler.go:93] Event received: map[_id:map[_data:82..04] clusterTime:{T:1761219662 I:1} documentKey:map[_id:ObjectID("68..c8")] fullDocument:map[_id:ObjectID("68..c8") createdAt:1761219482208 healthevent:map[agent:syslog-health-monitor checkname:SysLogsXIDError componentclass:GPU drainoverrides:<nil> entitiesimpacted:[map[entitytype:PCI entityvalue:0000:03:00] map[entitytype:GPUID entityvalue:GPU-455-ffc457bff834]] errorcode:[109] generatedtimestamp:map[nanos:207323127 seconds:1761219482] isfatal:true ishealthy:false message:ROBUST_CHANNEL_CTXSW_TIMEOUT_ERROR metadata:<nil> nodename:[redacted]-2206 quarantineoverrides:<nil> recommendedaction:20 version:1] healtheventstatus:map[faultremediated:<nil> nodequarantined:Quarantined userpodsevictionstatus:map[status:Succeeded]]] ns:map[coll:HealthEvents db:HealthEventsDatabase] operationType:update updateDescription:map[removedFields:[] truncatedArrays:[] updatedFields:map[healtheventstatus.userpodsevictionstatus:map[status:Succeeded]]] wallTime:1761219662379]
I1023 11:41:02.390753       1 reconciler.go:415] Found existing CR maintenance-[redacted]-w-2206-68f731bc07936fda07aceb08 for node [redacted]-w-2206 group restart with status Succeeded
I1023 11:41:02.390762       1 reconciler.go:339] Skipping event for node [redacted]-w-2206 - CR maintenance-[redacted]-w-2206-68f731bc07936fda07aceb08 is Succeeded
I1023 11:41:02.390766       1 reconciler.go:258] Skipping event for node [redacted]-w-2206 due to existing CR maintenance-[redacted]-w-2206-68f731bc07936fda07aceb08
```

KaivalyaMDabhadkar:
## Type of Change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Recognizes a new "Cancelled" quarantine status and invokes cancellation flows to stop drains, clear remediation state, and notify the reconciler.

* **New Metrics / Docs**
  * Added metrics for manual uncordons and cancelled drain events; documentation updated to include them.

* **Bug Fixes**
  * Improved cancellation propagation so labels, gauges, and state are consistently cleared and errors are tracked.

* **Tests**
  * Expanded e2e/integration tests for cancellation/remediation; one legacy manual-uncordon test removed.


fault-quarantine/pkg/mongodb/event_watcher.go
```diff
@@ -242,3 +251,66 @@ func (w *EventWatcher) updateNodeQuarantineStatus(
 
 	return nil
 }
+
+func (w *EventWatcher) CancelLatestQuarantiningEvents(
```
lalitadithya, line 257: Hmm.. don't we have the object id in the annotation we store for the node? Can we use that? If we don't have it, can we add the id and make use of that here?

KaivalyaMDabhadkar: Is it necessary to do that though? The current state of the node can be determined purely by looking at the sequence of latest mongodb events on the node right?

lalitadithya: But it is a cleaner flow right? It is a single query and you already get the node object in your event. You don't have to make multiple queries for mongo

KaivalyaMDabhadkar: We also need to mark all the subsequent fatal events as cancelled as well which occurred on the node after the first cancelled event to maintain consistency. So we would anyways need to make an update query to mongodb

KaivalyaMDabhadkar: > But it is a cleaner flow right? It is a single query and you already get the node object in your event. You don't have to make multiple queries for mongo


We would need to store the mongo document object id separately in the annotations for that as a field for each health event (we are only storing the health event data currently in the annotation and not the associated mongodb doc data). And even after we do store the object Id, we would still need to make an updateMany call to mongo to cancel the subsequent "AlreadyQuarantined" events as well

lalitadithya: > We also need to mark all the subsequent fatal events as cancelled as well which occurred on the node after the first cancelled event to maintain consistency. So we would anyways need to make an update query to mongodb

Are you saying the ones that we marked as "AlreadyQuarantined"? If we have the ID, we can do > ID and then status AlreadyQuarantined and bulk update right?

> We would need to store the mongo document object id separately in the annotations for that as a field for each health event (we are only storing the health event data currently in the annotation and not the associated mongodb doc data). And even after we do store the object Id, we would still need to make an updateMany call to mongo to cancel the subsequent "AlreadyQuarantined" events as well

I see, I guess the question is if we have the ID, can we fire a single update many?

lalitadithya: Right now, it looks like we first get the data from mongo, do some process and then update based on the timestamp. So, can we make this simpler if we had the ID stored?

KaivalyaMDabhadkar: We can do that, but I think that will make the annotation management code complicated without the adequate payoff in terms of performance gain (which we would get by avoiding the findOne query which would be very fast in this case anyways as there is an index on `createdAt`). Also, rather than storing the object Id state in the annotations and using that as another reference for the event, I think it will be easier and more consistent to directly query the primary source of truth and fetch it from there, what do you think?

KaivalyaMDabhadkar: > Are you saying the ones that we marked as "AlreadyQuarantined"? If we have the ID, we can do > ID and then status AlreadyQuarantined and bulk update right?

Yes we are currently doing that in the code, but the difference is we are fetching the ID by first making a findOne query rather than from the annotation

lalitadithya: Hmm.. do we have sufficient checks to make sure we don't update events in mongodb that we have already remediated? I think we do, but please confirm. The scenario that I'm looking for is:
1. Fault occurs
2. Quarantine
3. Remediated
4. Unquarantine
5. User cordon
6. User does something
7. User uncordon

When 7 happens, we should never come to this flow.

KaivalyaMDabhadkar: Yes, we are safe from this scenario because when 7 occurs, this returns an `UnQuarantined` status event:
```
	filter := bson.M{
		"healthevent.nodename": nodeName,
		"healtheventstatus.nodequarantined": bson.M{
			"$in": []model.Status{model.Quarantined, model.UnQuarantined},
		},
	}
```
which we skip with this check and thus take no action:

```
	if *latestEvent.HealthEventStatus.NodeQuarantined != model.Quarantined {
		slog.Info("No latest quarantining event found for node, no events to cancel", "node", nodeName)
		return nil
	}
```

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1166,6 +1168,16 @@ func (r *Reconciler) handleManualUncordon(nodeName string) error {
 		return err
 	}
 
+	if err := r.eventWatcher.CancelLatestQuarantiningEvents(ctx, nodeName); err != nil {
+		slog.Error("Failed to cancel latest quarantining events for manually uncordoned node",
+			"node", nodeName,
+			"error", err)
+		metrics.ProcessingErrors.WithLabelValues("mongodb_cancelled_update_error").Inc()
+
+		return err
```
lalitadithya: please wrap the error

tests/fault_management_test.go
```diff
@@ -266,6 +266,460 @@ func TestNodeDeletedDuringDrain(t *testing.T) {
 	testEnv.Test(t, feature.Feature())
 }
 
+func TestManualUncordonPropagation(t *testing.T) {
```
XRFXLP, line 269: can we remove [this](https://github.com/NVIDIA/NVSentinel/blob/main/tests/fault_quarantine_test.go#L66) test if this is already checking FQ ack-ing uncordon ?

node-drainer/pkg/mongodb/event_watcher.go
```diff
@@ -153,18 +162,38 @@ func (w *EventWatcher) preprocessAndEnqueueEvent(ctx context.Context, event bson
...
-		slog.Any("event", healthEventWithStatus.HealthEvent),
-		slog.Any("userPodEvictingStatus", healthEventWithStatus.HealthEventStatus.UserPodsEvictionStatus))
-
-	// Extract fullDocument to access the actual document _id
 	document, ok := event["fullDocument"].(bson.M)
 	if !ok {
 		return fmt.Errorf("error extracting fullDocument from event: %+v", event)
 	}
 
+	eventID := document["_id"]
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Not necessary, it will never be nil

node-drainer/pkg/mongodb/event_watcher.go
```diff
@@ -153,18 +162,38 @@ func (w *EventWatcher) preprocessAndEnqueueEvent(ctx context.Context, event bson
...
+	statusPtr := healthEventWithStatus.HealthEventStatus.NodeQuarantined
+
+	if statusPtr != nil && *statusPtr == model.Cancelled {
+		slog.Info("Detected Cancelled event, marking specific event as cancelled (not enqueueing)",
+			"node", nodeName,
+			"eventID", eventID)
+		w.onEventCancelled(eventID)
+
+		return nil
+	}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Not necessary, it is never nil

node-drainer/pkg/mongodb/event_watcher.go
```diff
@@ -153,18 +162,38 @@ func (w *EventWatcher) preprocessAndEnqueueEvent(ctx context.Context, event bson
...
+
+	if statusPtr != nil && *statusPtr == model.UnQuarantined {
+		slog.Info("Detected UnQuarantined event, marking all in-progress events for node as cancelled",
+			"node", nodeName,
+			"eventID", eventID)
+		w.onNodeEventsCancelled(nodeName)
+	}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Not necessary, it is never nil

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -332,3 +349,108 @@ func (r *Reconciler) updateNodeUserPodsEvictedStatus(ctx context.Context, collec
...
+	eventIDs, ok := val.([]interface{})
+	if !ok {
+		return
+	}
+
+	for _, eventID := range eventIDs {
+		r.eventStatusMap.Store(eventID, model.Cancelled)
+		slog.Info("Marked event as cancelled for node", "node", nodeName, "eventID", eventID)
+	}
+}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Not needed

tests/fault_management_test.go
```diff
@@ -266,6 +266,460 @@ func TestNodeDeletedDuringDrain(t *testing.T) {
...
+		for _, podName := range podNames {
+			pod := &v1.Pod{}
+			err := client.Resources().Get(ctx, podName, testCtx.TestNamespace, pod)
+			require.NoError(t, err, "pod %s should still exist after deleteAfterTimeout deadline", podName)
+			assert.Nil(t, pod.DeletionTimestamp, "pod %s should never have been marked for deletion", podName)
+			assert.Equal(t, v1.PodRunning, pod.Status.Phase, "pod %s should still be running", podName)
+		}
+
+		return ctx
+	})
```
lalitadithya, line 592: Can we add a check to see if rebootnode CR was never created just to round out the flow?

tests/fault_management_test.go
```diff
@@ -266,6 +266,460 @@ func TestNodeDeletedDuringDrain(t *testing.T) {
...
+
+	feature.Teardown(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		return helpers.TeardownNodeDrainer(ctx, t, c)
+	})
+
+	testEnv.Test(t, feature.Feature())
+}
+
+func TestHealthyEventWithFaultRemediation(t *testing.T) {
```
lalitadithya: I think we can merge this one with the other tests. Let's aim to consolidate the number of tests since these are e2e.

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -42,6 +42,8 @@ type Reconciler struct {
 	informers           *informers.Informers
 	evaluator           evaluator.DrainEvaluator
 	kubernetesClient    kubernetes.Interface
+	eventStatusMap      sync.Map
+	nodeToEventIDsMap   sync.Map
```
lalitadithya: Let us combine these two into a single map -- map[nodeName][]struct{eventId, status}

lalitadithya (commented):
As we discussed offline:
1. Let's more toughly look at the possible races between drainer and remediation.
2. Merge the cancellation logic with the unquarantine logic: to the drainer module they are the same operation - so we don't need different handling for them

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -332,3 +349,108 @@ func (r *Reconciler) updateNodeUserPodsEvictedStatus(ctx context.Context, collec
...
+	val, _ := r.nodeToEventIDsMap.LoadOrStore(nodeName, []interface{}{})
+
+	eventIDs, ok := val.([]interface{})
+	if !ok {
+		eventIDs = []interface{}{}
+	}
+
+	eventIDs = append(eventIDs, eventID)
+	r.nodeToEventIDsMap.Store(nodeName, eventIDs)
+}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Added a mutex to protect the map

tests/fault_management_test.go
KaivalyaMDabhadkar: Done

tests/fault_management_test.go
KaivalyaMDabhadkar: DOne

tests/fault_management_test.go
KaivalyaMDabhadkar: Done

node-drainer/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: Done

fault-quarantine/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: Done

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -127,9 +127,16 @@ func (r *Reconciler) processEvent(ctx context.Context, event bson.M, watcher Wat
...
+	if nodeQuarantined != nil {
+		if *nodeQuarantined == model.UnQuarantined {
+			r.handleUnquarantineEvent(ctx, nodeName, watcher)
+			return
+		}
+
+		if *nodeQuarantined == model.Cancelled {
+			r.handleCancelledEvent(ctx, nodeName, healthEventWithStatus.HealthEvent, watcher)
+			return
+		}
```
lalitadithya: Does the handling of unquarantine and cancel have to be different? In both cases they have to clear the annotation right?

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -42,6 +42,8 @@ type Reconciler struct {
 	informers           *informers.Informers
 	evaluator           evaluator.DrainEvaluator
 	kubernetesClient    kubernetes.Interface
+	nodeEventsMap       map[string]map[interface{}]model.Status // nodeName → map[eventID]status
```
lalitadithya: two things:
1. can we make a type for `map[interface{}]model.Status` so that it is self documenting?
2. can we convert the eventid to string?

tests/fault_management_test.go
```diff
@@ -283,12 +283,255 @@ func TestNodeRecoveryDuringDrain(t *testing.T) {
...
+			node, err := helpers.GetNodeByName(ctx, client, testCtx.NodeName)
+			if err != nil {
+				return false
+			}
+
+			manualAnnotation, hasManual := node.Annotations["quarantinedNodeUncordonedManually"]
+			_, hasQuarantine := node.Annotations["quarantineHealthEvent"]
+
+			return hasManual && manualAnnotation == "True" && !hasQuarantine
+		}, helpers.EventuallyWaitTimeout, helpers.WaitInterval, "FQ should detect manual uncordon")
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: The value is explicitly set to "True" as a string constant, not using strconv.FormatBool

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -332,3 +350,101 @@ func (r *Reconciler) updateNodeUserPodsEvictedStatus(ctx context.Context, collec
...
+		if !exists {
+			slog.Debug("No in-progress events found for node", "node", nodeName)
+			return
+		}
+
+		for evtID := range eventsMap {
+			eventsMap[evtID] = model.Cancelled
+			slog.Info("Marked event as cancelled for node", "node", nodeName, "eventID", evtID)
+		}
+	}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Applied a fix, please check

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -332,3 +350,121 @@ func (r *Reconciler) updateNodeUserPodsEvictedStatus(ctx context.Context, collec
 
 	return nil
 }
+
+func (r *Reconciler) HandleCancellation(eventID interface{}, nodeName string, status model.Status) {
+	r.nodeEventsMapMu.Lock()
+	defer r.nodeEventsMapMu.Unlock()
+
+	//nolint:exhaustive
```
lalitadithya: Can we add a comment explaining why we added this no lint?

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -144,23 +166,21 @@ func (r *Reconciler) executeSkip(ctx context.Context,
...
 	slog.Info("Skipping event for node", "node", nodeName)
 
-	// Update MongoDB status to StatusSucceeded for healthy events that cancel draining
-	if healthEvent.HealthEventStatus.NodeQuarantined != nil &&
-		*healthEvent.HealthEventStatus.NodeQuarantined == model.UnQuarantined {
-		// Update MongoDB status to StatusSucceeded for healthy events that cancel draining
+	if statusPtr := healthEvent.HealthEventStatus.NodeQuarantined; statusPtr != nil &&
+		*statusPtr == model.UnQuarantined {
 		podsEvictionStatus := &healthEvent.HealthEventStatus.UserPodsEvictionStatus
 		podsEvictionStatus.Status = model.StatusSucceeded
```
lalitadithya, line 178: Shouldn't this be cancelled?

KaivalyaMDabhadkar: No the cancelled events are not enqueued actually, as they are not "new" events, but the unquarantined ones are enqueued. So we only need this check for the unquarantined ones

lalitadithya: yes, yes, but the status I mean. If this is success, won't FR react?

node-drainer/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: Done

node-drainer/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: Done

fault-remediation/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: I had initially kept them different and was specifically updating the group in the annotation which was affected by the cancelled event. But you are right, this is unnecessary, we should clear the entire state in both cases. I merged both the paths into one

node-drainer/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: Yes, this is expected. FR will receive the event with [UnQuarantined, Succeeded] and will clear the FR annotation on the node:
```
					// Watch for unquarantine events (for annotation cleanup)
					bson.D{
						bson.E{Key: "fullDocument.healtheventstatus.nodequarantined", Value: model.UnQuarantined},
						bson.E{Key: "fullDocument.healtheventstatus.userpodsevictionstatus.status", Value: model.StatusSucceeded},
					},
```

KaivalyaMDabhadkar: FR won't create a CR for this event

node-drainer/pkg/reconciler/reconciler_integration_test.go
```diff
@@ -727,6 +733,176 @@ func assertPodsEvicted(t *testing.T, client kubernetes.Interface, ctx context.Co
...
+			return false
+		}
+		_, exists := node.Labels[statemanager.NVSentinelStateLabelKey]
+		return !exists
+	}, 15*time.Second, 500*time.Millisecond, "draining label should be removed after UnQuarantined event")
+
+	pod, err := setup.client.CoreV1().Pods("timeout-test").Get(setup.ctx, "stuck-pod", metav1.GetOptions{})
+	require.NoError(t, err)
+	assert.Nil(t, pod.DeletionTimestamp, "pod should not be deleted after cancellation")
+}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Not needed

node-drainer/pkg/reconciler/reconciler_integration_test.go
```diff
@@ -727,6 +733,176 @@ func assertPodsEvicted(t *testing.T, client kubernetes.Interface, ctx context.Co
...
+		_, exists := node.Labels[statemanager.NVSentinelStateLabelKey]
+		return !exists
+	}, 15*time.Second, 500*time.Millisecond, "draining label should be removed after UnQuarantined")
+
+	pods, err := setup.client.CoreV1().Pods("timeout-test").List(setup.ctx, metav1.ListOptions{})
+	require.NoError(t, err)
+	for _, pod := range pods.Items {
+		assert.Nil(t, pod.DeletionTimestamp, "pod %s should not be deleted after cancellation", pod.Name)
+	}
+}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Not needed

lalitadithya (approved):

PR #297, opened 2025-11-10
docs: Add design doc for kubernetes object monitor

### LINKED ISSUE #192: [Feature]: Add a Kubernetes object health monitor
### Prerequisites

### Feature Summary

Currently to add a new health check a new health monitor should be developed, tested and published. In many Kubernetes there exists problem detectors that modify kubernetes objects to indicate bad hardware. For example, some vendor apply node conditions to inform the user that the hardware they are running on is bad, another case is the classic NotReady and kubelet's standard conditions. We should provide customers with a policy driven way to inform a generic kubernetes health monitor that a node is "bad" and what needs to be done.

This can also provide a bridge for users to try out NVSentinel with their existing checks before migrating to an "NVSentinel compatible check"

### Problem/Use Case

As an operator I want a way for NVSentinel to react to conditions that my existing checks are setting

### Proposed Solution

Develop a new kubernetes health monitor that makes use controller's unmanaged controller and some CEL policies to determine if the a "node" is healthy and publish health events. 

### Component

Health Monitor

XRFXLP:
## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added ADR for a policy-driven Kubernetes Object Monitor that evaluates CEL-based policies against cluster resources and emits HealthEvents and recovery events.
  * Documents configuring resource policies (including time-based predicates), reconciliation behavior, watched resources, integration points, metrics, examples for Node/Event policies, and operational deployment and recovery considerations.


docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,267 @@
...
+
+An example of such configuration would be:
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubernetes-health-monitor
+  namespace: nvsentinel
+data:
+  policies.yaml: |
```
lalitadithya: Can we do TOML like the rest of the system. The TOML can then be mapped to helm values cleanly. Please see how it is done in the quarantine module and let's follow the same pattern.

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,267 @@
...
+    policies:
+      # Node condition-based policy
+      # Node not ready for more than 2 hours
+      - name: GPUNodeNotReady
+        enabled: true
+        resourceType: Node
```
lalitadithya: Don't you also need the group and version too?

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,267 @@
...
+
+The Kubernetes Object Monitor is implemented as a **Deployment** that uses **Kubernetes informers** with a **work-queue** for efficient cluster-wide monitoring. It evaluates CEL-based policies against Kubernetes objects (Nodes, Events) and publishes HealthEvents when policies match.
+
+**Key Components:**
+- **Informer-based watch**: Efficiently monitors Nodes and Events using Kubernetes client-go informers
+- **Work-queue**: Provides retry logic, rate-limiting, and deduplication
+- **CEL Policy Engine**: Evaluates user-defined predicates against Kubernetes objects
+
+
+**Watched Resources:**
```
lalitadithya: This is a great start, can we keep the door open to generic resources as well? Can we add GVK/GVR and use that construct the informers rather than just Nodes and Events? 

For the initial iterations only Nodes and Events will be supported and tested, but in future we want to add a vendor specific CR, then let's keep the option open for that now rather than having to do a refactor for that later.

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,267 @@
...
+- **Nodes**: Primary resource for condition, taint, label, and annotation checks
+- **Events**: Kubernetes events (Node, Pod, etc.) for detecting specific conditions
+
+#### Policy Structure
+
+Policies consist of three key components:
+
+1. **Resource Type**: The Kubernetes resource to watch (`Node` or `Event`)
+2. **Predicate** (CEL expression): Boolean condition evaluated against the resource object
+3. **Time Threshold** (optional): Duration the predicate must remain true before triggering an event
```
lalitadithya: Since we have CEL do we need explicit configuration for time threshold? For example, if we are looking at an object that uses the conditions resource, then we can use the last transition time, for others we can probably use the creation timestamp.

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,267 @@
...
+
+**Time evaluation strategy:**
+- Prefer built-in timestamps (NodeCondition.LastTransitionTime) over tracked timestamps
+- For Events, use event.firstTimestamp and event.count for frequency analysis
+
+#### State Tracking
+
+**State Persistence:**
+
+State is written to disk as JSON at `/var/run/nvsentinel/k8s-monitor-state.json`. State tracks:
```
lalitadithya: do we need to store state? Can't we determine the current state during the initial list of the informer?

docs/designs/011-kubernetes-object-monitor.md
XRFXLP: Assumption here was it would map to node type, but it would be better to have full GVK spec

docs/designs/011-kubernetes-object-monitor.md
XRFXLP: Yeah, sure.

docs/designs/011-kubernetes-object-monitor.md
XRFXLP: Yeah, right.
I added it earlier to track annotation and labels as well but realized it doesn't make sense. Will remove this.

docs/designs/011-kubernetes-object-monitor.md
XRFXLP: If we have CEL evaluates to false => healthy, then yes we don't need this.

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,283 @@
...
+    
+    style K8S fill:#52c41a,stroke:#389e0d,stroke-width:2px,color:#fff
+    style CM fill:#fa8c16,stroke:#d46b08,stroke-width:2px,color:#fff
+    style PC fill:#1890ff,stroke:#096dd9,stroke-width:2px,color:#fff
+    style MONGO fill:#13c2c2,stroke:#08979c,stroke-width:2px,color:#fff
+```
+
+### Key Integration Points
+
+1. **Kubernetes Object Monitor** watches Kubernetes API resources (based on policy GVKs) via informers
```
lalitadithya:
```suggestion
1. Kubernetes Object Monitor watches Kubernetes API resources (based on policy GVKs) via informers
```

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,283 @@
...
+- Informers watch configured GVK resources (e.g., Nodes, Events)
+- When a watched resource changes (create/update/delete), triggers reconciliation
+- For each policy matching the resource's GVK, evaluate the CEL predicate
+
+#### 2. Periodic Reconciliation
+
+- Required for time-based predicates (e.g., "NotReady for more than 2 hours")
+- When a Node condition transitions to NotReady, the predicate evaluates to `false` initially
+- After 2 hours, periodic reconciliation re-evaluates and detects the threshold crossing
+- Configured via controller's resync period (e.g., every 5 minutes)
```
lalitadithya: Since we have a work queue, this will be on the workqueue right? Since the time duration is specified in CEL, how do we know when to requeue and when not to requeue?

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,283 @@
...
+
+Uniform reconciliation for all resources:
+
+1. Retrieve the resource object (from informer cache)
+2. For each policy matching the resource's GVK, evaluate the CEL predicate
+3. If predicate evaluates to `true`:
+   - Publish unhealthy HealthEvent via gRPC
+4. If predicate evaluates to `false`:
+   - Publish healthy/recovery HealthEvent via gRPC
+5. Multiple reconciliations run concurrently (configured via `MaxConcurrentReconciles`)
```
lalitadithya: Do you mean one reconciler per rule? Can we clarify what are we going to reconcile in parallel?

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,283 @@
...
+- Associate the health event with that Node in the HealthEvent payload
+
+
+
+
+### Recovery Event Handling
+
+Recovery is handled uniformly for all resource types:
+
+- When CEL predicate evaluates to `false`, a healthy/recovery HealthEvent is published
```
lalitadithya, line 352: How would this work for ephemeral objects like events? Events are deleted after a TTL so we need to have some way to track that an event that caused an unhealthy event doesn't exist anymore. Maybe we maintain an in-memory list of ephemeral objects that have triggered an unhealthy event and we send a healthy event when the event is deleted, or it has not been seen after X min. The state doesn't have to be written to disk because we should be able to reconstruct the state on the initial re-list.

We can have a configuration of something to the effect of `isEphemeral` and then based on that add the required logic.

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -0,0 +1,283 @@
...
+   - Publish unhealthy HealthEvent via gRPC
+4. If predicate evaluates to `false`:
+   - Publish healthy/recovery HealthEvent via gRPC
+5. Multiple reconciliations run concurrently (configured via `MaxConcurrentReconciles`)
+
+#### Special Handling for Pod Events
+
+For Event resources where `resource.regarding.kind == 'Pod'`:
+- Look up the Pod to determine which Node it's scheduled on
+- Associate the health event with that Node in the HealthEvent payload
```
lalitadithya: Can we externalize this in config somehow? Maybe we have a property like:

```toml
[policies.nodeLookup]
group = "events.k8s.io"  # Event API group (v1 Events)
version = "v1"
kind = "Event"
nameExperssion = "[json path]"
nodeNameExperssion = "[json path]"
```

So, we have two cases:
1. the node name is present in the object itself, then only `nodeNameExperssion` can be specified and the output of the json path expression will be used as the node name
2. the node name is in a different object (single nesting only), then the object will be looked up using GVK with the name in `nameExperssion` and then finally `nodeNameExperssion` will be evaluated on the looked up object for the node name

docs/designs/011-kubernetes-object-monitor.md
XRFXLP: I had the same in the earlier version where there was a fixed grace period associated with the event, and we had state persistence storage.

Now, how not seen after X min is different from not seen after default TTL? Are we going to have something like `X = 5 x TTL`, for events would it be like 3-hours?

I'm considering few sample cases to see what can be done:
```
78s (x1383 over 5h1m)   Warning   FailedCreatePodSandBox   Pod/gpu-feature-discovery-sq47w   Failed to create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox runtime: no runtime for "nvidia" is configured
```

Another would be:
```
  Warning  Failed   19m (x4 over 21m)   kubelet  Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli.real: detection error: nvml error: gpu requires reset: unknown
```

these erroneous pretty are pretty much always present as long as the error exists and it gets repeated.

docs/designs/011-kubernetes-object-monitor.md
XRFXLP: This is kinda hard without losing any generality, earlier we were solving this problem by separating declaration of time threshold in a different var.

docs/designs/011-kubernetes-object-monitor.md
lalitadithya: Hmm, good point. I'll take back `or it has not been seen after X min`. For events specifically though since they get GC'd after an hour, `or it has not been seen after X min` would end up being auto deleted after 1 hour.

So, can we do:
1. on a cold start, if nothing matches send a healthy event
2. on regular operation, on delete of an object we are tracking, send healthy event

docs/designs/011-kubernetes-object-monitor.md
lalitadithya: I see, okay. So then can we have a default resync time for the informer say 10m available via configuration outside of any rules and we don't requeue in the workqueue unless in case of an error. The granularity of health check will be every 10m/15m which I think should be fine and it is something we can document.

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -281,3 +367,29 @@ resource.reason == 'Failed' &&
...
+Controller-level settings are configured via command-line flags in the Deployment:
+
+| Flag | Type | Default | Description |
+|------|------|---------|-------------|
+| `--resync-period` | duration | `5m` | Periodic reconciliation interval for all watched resources |
+| `--max-concurrent-reconciles` | int | `10` | Maximum number of resources to reconcile concurrently |
+| `--policy-config-path` | string | `/etc/nvsentinel/config/policies.toml` | Path to policy configuration file |
+| `--metrics-bind-address` | string | `:8080` | Address to bind Prometheus metrics endpoint |
+| `--health-probe-bind-address` | string | `:8081` | Address to bind health probe endpoints |
+| `--log-level` | string | `info` | Log level (debug, info, warn, error) |
```
lalitadithya: I believe other modules use env vars for log level, can we do the same here?

docs/designs/011-kubernetes-object-monitor.md
```diff
@@ -281,3 +367,29 @@ resource.reason == 'Failed' &&
...
+
+## Metrics
+
+The Kubernetes Object Monitor exposes Prometheus metrics for observability and debugging.
+
+| Metric Name | Type | Labels | Description |
+|------------|------|--------|-------------|
+| `k8s_object_monitor_policy_matches_total` | Counter | `policy_name`, `node`, `resource_kind` | Total number of times a policy matched (predicate=true), indicating detected issues by node |
+| `k8s_object_monitor_active_unhealthy_conditions` | Gauge | `policy_name`, `node`, `resource_kind` | Current number of resources with active unhealthy conditions |
```
lalitadithya: I'm not sure if we need this gauge. We can always get this info using kube-state-metrics or from the mongodb

XRFXLP: I don't have any strong opinion on this but I've found mongodb not query-able from outside of the cluster, and also kube-state-metric won't have the generality and it will be split across multiple metrics.

lalitadithya: > I don't have any strong opinion on this but I've found mongodb not query-able from outside of the cluster

Yes, but we have a work item in this sprint to change that, we are going to continuously export mongodb to a datalake which will be quarriable.

docs/designs/011-kubernetes-object-monitor.md
XRFXLP: handled, ptal

docs/designs/011-kubernetes-object-monitor.md
XRFXLP: handled, ptal

lalitadithya (approved):

PR #333, opened 2025-11-11
fix: csp monitor fixes

tanishagoyal2:
## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary
Tested CSP health monitor on AWS GB200 cluster. Identified below fixes:
1.  updated deployment.yaml file to include correct maintenance-notifier image name
2. updated IAM policy name
3. added quarantineTriggerEngine image name in values.yaml which will be used in the deployment file
4. properly propogating the errors to the calling function and corrected log statements to correctly print the info
5. updated unit test as per the new change

Requested [scheduled maintenance ](https://health.console.aws.amazon.com/health/home?region=us-east-1#/account/event-log?eventID=arn:aws:health:us-east-1::event/EC2/AWS_EC2_INSTANCE_REBOOT_MAINTENANCE_SCHEDULED/AWS_EC2_INSTANCE_REBOOT_MAINTENANCE_SCHEDULED_1762870915256&eventTab=details) from AWS support.
Following are the key observations:
1. The maintenance Event was created
Screenshot 2025-11-11 at 7 57 56 PM

2. The quarantine events were created and nodes were cordoned and node condition was added
Screenshot 2025-11-11 at 7 58 45 PM
Screenshot 2025-11-11 at 7 58 52 PM
Screenshot 2025-11-11 at 8 25 41 PM
 
3. When maintenance was completed, the maintenance-notifier waited for node to come back healthy and then triggered healthy event

Screenshot 2025-11-11 at 9 06 12 PM
Screenshot 2025-11-11 at 9 06 37 PM
Screenshot 2025-11-11 at 9 06 43 PM

## Summary by CodeRabbit

* **New Features**
  * Structured logging across monitors.
  * Node informer to map cloud instance IDs to cluster nodes.

* **Bug Fixes**
  * Aggregated error reporting for health-event processing.
  * More robust handling of missing or malformed event/node data.

* **Chores**
  * Updated service account role annotation.
  * Added sidecar image configuration and adjusted container startup/image pull policy.
  * Dependency updates and new test setup variable.

* **Tests**
  * Tests now run against a control-plane test environment with informer wiring.


health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
```diff
@@ -496,13 +506,25 @@ func (c *AWSClient) processAWSHealthEvent(
...
-	// Process all affected entities for this event
 	for _, entity := range affectedEntities {
-		c.processSingleEntityForEvent(ctx, eventArn, evt, entity, desc, action, nodeMap, eventChan)
+		err := c.processSingleEntityForEvent(ctx, eventArn, evt, entity, desc, action, nodeMap, eventChan)
+		if err != nil {
+			slog.Error("Error processing single entity for event",
+				"eventArn", eventArn,
+				"error", err)
+
+			mu.Lock()
```
KaivalyaMDabhadkar: Why do we need a mutex here? This is not being shared by multiple goroutines right?

health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
KaivalyaMDabhadkar: It should be AWS monitor not GCP

health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
```diff
@@ -746,13 +796,11 @@ func (c *AWSClient) getClusterInstanceNodeMap(ctx context.Context) (map[string]s
 
 		idPart := strings.TrimPrefix(node.Spec.ProviderID, "aws:///")
 
-		// The ID might include the zone, so we need to extract just the instance ID
 		parts := strings.Split(idPart, "/")
-		instanceID := parts[len(parts)-1] // Get the last part which should be the instance ID
+		instanceID := parts[len(parts)-1]
 
-		// Ensure it's a valid EC2 instance ID (i-xxxxxxxxxxxxxxxxx format)
 		if strings.HasPrefix(instanceID, "i-") {
```
KaivalyaMDabhadkar: nit: I think we should keep the explanatory comment above, it helps understanding what we are trying to validate for

health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
KaivalyaMDabhadkar: `getClusterInstanceNodeMap` will make a k8s node list api call each time it is invoked to populate the map, is that necessary to do every poll cycle? Can we just cache the map once during startup and then make api-server calls only for cache misses?

KaivalyaMDabhadkar (changes requested):

health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
tanishagoyal2: I also had same question but then I thought what if some nodes are deleted/added after the health-monitor started then we will have stale data.
But now I am thinking a better way to handle it. I can use informer which listens for delete/update (to handle new nodes added) of node and will update the map according to that.
I'll update the code, thanks

health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
tanishagoyal2: The above comment was not adding much value. The intention is clear from code and var names.

health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
tanishagoyal2: right. I will remove it

health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
tanishagoyal2: will update

health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
```diff
@@ -559,17 +577,105 @@ func (c *AWSClient) getAffectedEntities(
...
+
+	nodeName, instanceID, eventArn := activeEvent.NodeName, activeEvent.ResourceID, activeEvent.Metadata["eventArn"]
+	eventMetadata := eventpkg.EventMetadata{
+		Event:            awsEvent,
+		NodeName:         nodeName,
+		InstanceId:       instanceID,
+		EntityArn:        activeEvent.EventID,
+		Action:           activeEvent.RecommendedAction,
+		EventDescription: activeEvent.Metadata["description"],
+	}
```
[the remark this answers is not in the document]
tanishagoyal2: We are storing entityARN as eventID only. You can confirm that from aws_normalizer.go file also where at line 128 we are assigning entityARN as eventID value.

health-monitors/csp-health-monitor/pkg/csp/aws/informer_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestNodeInformer(t *testing.T) {
```
KaivalyaMDabhadkar: I don't think writing unit tests for the informer will bring us any value, especially tests which use the fake client. This is already a well tested external library. Can we remove this file altogether?

health-monitors/csp-health-monitor/pkg/csp/aws/informer.go
```diff
@@ -0,0 +1,176 @@
...
+	"k8s.io/client-go/tools/cache"
+)
+
+// NodeInformer watches Kubernetes nodes and maintains an up-to-date mapping
+// of AWS instance IDs to node names
+type NodeInformer struct {
+	k8sClient   kubernetes.Interface
+	informer    cache.SharedIndexInformer
+	stopCh      chan struct{}
+	instanceIDs map[string]string
```
KaivalyaMDabhadkar: nit: can we rename this to instanceIdToNodeNameMap for better readability?

health-monitors/csp-health-monitor/pkg/csp/aws/informer.go
```diff
@@ -0,0 +1,176 @@
...
+		return
+	}
+
+	slog.Info("AWS node informer cache synced successfully", "nodesMap", ni.instanceIDs)
+
+	// Log the initial node count
+	ni.mu.RLock()
+	count := len(ni.instanceIDs)
+	ni.mu.RUnlock()
+	slog.Info("AWS node informer ready", "nodeCount", count)
```
KaivalyaMDabhadkar: Why do we need to log this? I don't think this log would bring any value

health-monitors/csp-health-monitor/pkg/csp/aws/informer.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func NewNodeInformer(k8sClient kubernetes.Interface) *NodeInformer {
+	ni := &NodeInformer{
+		k8sClient:   k8sClient,
+		stopCh:      make(chan struct{}),
+		instanceIDs: make(map[string]string),
+	}
+
+	// Use informer factory which works with both real and fake clients
```
KaivalyaMDabhadkar: We can remove this comment

health-monitors/csp-health-monitor/pkg/csp/aws/informer.go
```diff
@@ -0,0 +1,176 @@
...
+			node := obj.(*v1.Node)
+			ni.handleNodeAdd(node)
+		},
+		DeleteFunc: func(obj interface{}) {
+			node := obj.(*v1.Node)
+			ni.handleNodeDelete(node)
+		},
+	})
+	if err != nil {
+		slog.Error("Failed to add event handlers to informer", "error", err)
```
KaivalyaMDabhadkar, line 62: Shouldn't we return early from here if an error occurs? Also I think it would be better to bubble this error up to the caller so that we can handle it there, the function return type can be:

`func NewNodeInformer(k8sClient kubernetes.Interface) (*NodeInformer , error)`

KaivalyaMDabhadkar (changes requested):

health-monitors/csp-health-monitor/pkg/csp/aws/aws.go
```diff
@@ -167,6 +164,11 @@ func NewClient(
 
 	slog.Info("AWS Client: Kubernetes clientset initialized successfully.")
 
+	nodeInformer := NewNodeInformer(k8sClient)
+	nodeInformer.Start(ctx)
+
+	slog.Info("AWS Client: Node informer started successfully")
```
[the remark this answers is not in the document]
tanishagoyal2: but we already have this goroutine which is checking if ctx is cancelled then call the stop function. then It should properly clean up the informer. please check this code in informer.start

go func() {
		<-ctx.Done()
		ni.Stop()
	}()

health-monitors/csp-health-monitor/pkg/csp/aws/informer_test.go
tanishagoyal2: right, removed it

health-monitors/csp-health-monitor/pkg/csp/aws/informer.go
tanishagoyal2: updated

health-monitors/csp-health-monitor/pkg/csp/aws/informer.go
tanishagoyal2: removed it. it was useless

health-monitors/csp-health-monitor/pkg/csp/aws/informer.go
tanishagoyal2: removed

health-monitors/csp-health-monitor/pkg/csp/aws/informer.go
tanishagoyal2: Yeah right, its returning error now

health-monitors/csp-health-monitor/pkg/csp/aws/aws_test.go
```diff
@@ -104,15 +99,24 @@ func createTestClient(t *testing.T) (*AWSClient, *MockAWSHealthClient, *fake.Cli
 	_, err := fakeK8sClient.CoreV1().Nodes().Create(context.Background(), node, metav1.CreateOptions{})
 	assert.NoError(t, err)
 
+	nodeInformer := &NodeInformer{
+		k8sClient: fakeK8sClient,
```
KaivalyaMDabhadkar: Would it be possible to convert this into an env test rather than using a fake client?

tanishagoyal2: sure, let me check

health-monitors/csp-health-monitor/pkg/csp/aws/aws_test.go
```diff
@@ -247,11 +286,18 @@ func TestMultipleAffectedEntities(t *testing.T) {
...
+
+		nodeName := nodeData.name
+		t.Cleanup(func() {
+			_ = k8sClient.CoreV1().Nodes().Delete(context.Background(), nodeName, metav1.DeleteOptions{})
+		})
 	}
 
-	// Setup test data
+	// Wait for informer to sync the new nodes
+	time.Sleep(200 * time.Millisecond)
```
KaivalyaMDabhadkar: Please use `eventually` rather than sleep. Sleep can lead to test flakiness

health-monitors/csp-health-monitor/pkg/csp/aws/aws_test.go
```diff
@@ -429,16 +468,36 @@ func TestCompletedEvent(t *testing.T) {
...
+		Spec: v1.NodeSpec{
+			ProviderID: "aws:///" + testRegion + "/" + testInstanceID2,
+		},
+	}
+	_, err = testK8sClient.CoreV1().Nodes().Create(context.Background(), node3, metav1.CreateOptions{})
+	assert.NoError(t, err)
+	defer testK8sClient.CoreV1().Nodes().Delete(context.Background(), testNodeName2, metav1.DeleteOptions{})
+
+	// Wait for the informer to sync the new nodes
+	time.Sleep(200 * time.Millisecond)
```
KaivalyaMDabhadkar: Use `eventually` rather than sleep as much as possible to reduce test flakiness

health-monitors/csp-health-monitor/pkg/csp/aws/aws_test.go
```diff
@@ -811,18 +849,32 @@ func TestInstanceRebootEvent(t *testing.T) {
...
+		},
+	}
+	_, err := k8sClient.CoreV1().Nodes().Create(context.Background(), nodeIgnored, metav1.CreateOptions{})
+	assert.NoError(t, err)
+	t.Cleanup(func() {
+		_ = k8sClient.CoreV1().Nodes().Delete(context.Background(), testNodeNameIgnored, metav1.DeleteOptions{})
+	})
+
+	// Wait for informer to sync
+	time.Sleep(200 * time.Millisecond)
```
KaivalyaMDabhadkar: Same as the previous comment

health-monitors/csp-health-monitor/pkg/csp/aws/aws_test.go
tanishagoyal2: added eventually now

health-monitors/csp-health-monitor/pkg/csp/aws/aws_test.go
tanishagoyal2: sure, added require.eventually

health-monitors/csp-health-monitor/pkg/csp/aws/informer.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (ni *NodeInformer) Start(ctx context.Context) {
+	slog.Info("Starting AWS node informer")
+
+	go ni.informer.Run(ni.stopCh)
+
+	if !cache.WaitForCacheSync(ni.stopCh, ni.informer.HasSynced) {
+		slog.Error("Failed to sync node informer cache")
+		return
+	}
```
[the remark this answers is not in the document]
tanishagoyal2: handled it now

KaivalyaMDabhadkar (dismissed):

KaivalyaMDabhadkar (approved):

PR #341, opened 2025-11-13
Fix duplicate node-drainer events by sorting pod lists

### LINKED ISSUE #339: [Bug]: Node drainer's events are not ordered causing duplicate Kubernetes events.
### Prerequisites

### Bug Description

If a node has multiple pods running, the node drainer add multiple events with different combinations of the same pod as seen below:

```
  NodeDraining   AwaitingPodCompletion      99s (x2 over 2m19s)  nvsentinel-node-drainer  Waiting for following pods to finish: [lalitadithya/ubuntu-gpu-deployment-64c54f5b4c-2c4c5 alloy/alloy-metrics-55d6fbdbd-vrp2h alloy/dgxc-alloy-gateway-0]
  NodeDraining   AwaitingPodCompletion      39s (x11 over 3m4s)  nvsentinel-node-drainer  Waiting for following pods to finish: [alloy/alloy-gateway-0 alloy/alloy-metrics-55d6fbdbd-vrp2h lalitadithya/ubuntu-gpu-deployment-64c54f5b4c-2c4c5]
  NodeDraining   AwaitingPodCompletion      19s (x9 over 2m54s)  nvsentinel-node-drainer  Waiting for following pods to finish: [lalitadithya/ubuntu-gpu-deployment-64c54f5b4c-2c4c5 alloy/alloy-gateway-0 alloy/dgxc-alloy-metrics-55d6fbdbd-vrp2h]
```

Ideally the node drainer should only add a single event with the list ordered

### Component

Fault Management

### Steps to Reproduce

1. Run multiple pods on a node in different namespaces
2. Configure the node drainer with allow completion
3. Trigger a fatal health event

### Environment

- NVSentinel version: v0.3.0
- Kubernetes version: any
- Deployment method: ArgoCD

### Logs/Output

```
  NodeDraining   AwaitingPodCompletion      99s (x2 over 2m19s)  nvsentinel-node-drainer  Waiting for following pods to finish: [lalitadithya/ubuntu-gpu-deployment-64c54f5b4c-2c4c5 alloy/alloy-metrics-55d6fbdbd-vrp2h alloy/dgxc-alloy-gateway-0]
  NodeDraining   AwaitingPodCompletion      39s (x11 over 3m4s)  nvsentinel-node-drainer  Waiting for following pods to finish: [alloy/alloy-gateway-0 alloy/alloy-metrics-55d6fbdbd-vrp2h lalitadithya/ubuntu-gpu-deployment-64c54f5b4c-2c4c5]
  NodeDraining   AwaitingPodCompletion      19s (x9 over 2m54s)  nvsentinel-node-drainer  Waiting for following pods to finish: [lalitadithya/ubuntu-gpu-deployment-64c54f5b4c-2c4c5 alloy/alloy-gateway-0 alloy/dgxc-alloy-metrics-55d6fbdbd-vrp2h]
```

Gyan172004:
## Summary

**Issue:**  
Issue #339 – The node-drainer emitted multiple Kubernetes events for the same set of pods because the pod list in the event message wasn’t ordered. As a result, Kubernetes treated each permutation of the pod list as a separate event instead of aggregating them.

**Reason:**  
Go’s iteration over maps/slices is non-deterministic. Without sorting the pod names, identical pod sets produced different event message strings, which caused Kubernetes to record duplicate events rather than incrementing counts.

**Solution:**  
Sort the pod names before constructing event messages in both `AllowCompletion` (reconciler.go) and `DeleteAfterTimeout` (informers.go). This ensures the same pod set always generates a consistent event string, allowing Kubernetes to aggregate events correctly.

## Type of Change
- [X] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Pod lists in status messages and node events are now deterministically and alphabetically sorted for consistent, clearer logging and to prevent spurious or duplicate events.

* **Tests**
  * Integration tests updated to run multiple reconciliation steps, validate sorted pod lists, and cover multi-pod scenarios. A per-test reconciliation count was added so tests can simulate repeated processing and verify stable event messages.

### DIFF
--- a/node-drainer/pkg/reconciler/reconciler_integration_test.go
+++ b/node-drainer/pkg/reconciler/reconciler_integration_test.go
@@ -1038,6 +1038,7 @@ func getHistogramCount(t *testing.T, histogram prometheus.Histogram) uint64 {
 	return metric.Histogram.GetSampleCount()
 }
 
+<<<<<<< HEAD
 // TestReconciler_CancelledEventWithOngoingDrain validates that Cancelled events stop ongoing drain operations
 func TestReconciler_CancelledEventWithOngoingDrain(t *testing.T) {
 	setup := setupRequeueTest(t, []config.UserNamespace{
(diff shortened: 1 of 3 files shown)


Gyan172004:
Here in the test, I are following an approach to catch non-deterministic behavior:  
The core problem for issue is that the order we get when iterating a map is not fixed, so each time we loop over it, we might* get a different sequence of items.

Our assumption is that out of 5 attempts, the internal map order will change at least once. The initial assumption was 10, but we saw that iterations were failing within 3–4 tries, so we cut it down to 5.

The test runs in a loop, generating the pod event message each time. It takes the first message as the reference and checks that all subsequent messages match. If any sequence is different, it means the bug is present, and the test fails immediately.

To make sure this was foolproof, I ran this test case in a loop 10,000 times with our sorting code removed, and it resulted in all FAIL. This means there were no false positives and shows that the probability of this test missing the bug is <0.0001.

I chose this iterative comparison because trying to write 2–3 separate tests(which llm was also doing same) is not reliable; the map might give the same order by chance, leading to a false positive. 

**Let me know if this makes sense 
Open to Feedback**

@lalitadithya

node-drainer/pkg/reconciler/reconciler_integration_test.go
```diff
@@ -1020,3 +1020,98 @@ func getHistogramCount(t *testing.T, histogram prometheus.Histogram) uint64 {
 	require.NoError(t, err)
 	return metric.Histogram.GetSampleCount()
 }
+
+// TestReconciler_PodListSortingConsistentEvents verifies that pod list sorting prevents duplicate Kubernetes events by ensuring consistent event messages across multiple reconciliation calls.
+func TestReconciler_PodListSortingConsistentEvents(t *testing.T) {
```
lalitadithya: can we instead update the existing test to cover this scenario -- https://github.com/NVIDIA/NVSentinel/blob/1ac2e83ae8947f7c6cd66cae2c08b699294eeee0/node-drainer/pkg/reconciler/reconciler_integration_test.go#L166

Gyan172004: just to confirm, do you mean I should add the existing test - `AllowCompletion mode waits for pods to complete` logic into my test, or the other way around?

I kept this test seperate as adding this new test into the table-driven tests would require adding the reconciler parameter to the table, which would then need to be added to all test cases.

lalitadithya: What I meant is in the existing test, can we make it wait for multiple pods and always add the functionality to compare the order, i.e., the pod list in the event should be in sorted order

Gyan172004: Got it. I was just worried the order check inside table driven might make the test a bit flaky sometimes.
will that much flackiness be okay?, might be i am overthinking here

lalitadithya: I don't think it should be flaky. we can run the test multiple times to confirm. But if the test is flaky, then it likely indicates that the order is not deterministic which means the fix that we've made is likely incorrect

Gyan172004: > we can run the test multiple times to confirm.

Yes, running it multiple times will work correctly. 

I initially tried to avoid running it multiple times inside the table-driven test because that would require passing the reconciler as a parameter to the validateFunc, which would then need to be added to all test cases in the table.

lalitadithya (approved):

PR #361, opened 2025-11-18
Log collector tilt test

### LINKED ISSUE #122: [Feature]: Support for automated log collection
### Prerequisites

### Feature Summary

NVSentinel automatically perform the recovery action for faults detected. In some cases, an operator will need to perform more deeper investigation on nodes that are failing repeatedly or nodes that are exhibiting new failure patterns. In order to do this, the operator need to have access to bug reports collected before the failure was remediated. 

### Problem/Use Case

Empower an operator with sufficient logs so that they can investigate failures beyond the initial recovery. These logs may also be used in the incident RCCA forums.

### Proposed Solution

Add a simple cluster local nginx based file server that can store the collected logs. The fault remediation module can spin up a job that will run the nvidia-bug-report, vendor provided SOS report and will upload it to the cluster local nginx server. 

### Component

Fault Management

rupalis-nv:
## Summary

Implements mock mode for the log collector to enable faster regression testing in Tilt environments, with both positive (success) and negative (failure) test cases. Also fixes race conditions in smoke tests when running on real nodes.

### 1. Mock Mode Implementation (Positive Test)

**Mock Scripts**: Created `mock-nvidia-bug-report.sh` and `mock-must-gather.sh`

**Entrypoint**: Updated `entrypoint.sh` to use PATH manipulation for mocking tools when `MOCK_MODE=true`

**Dockerfile**: Added mock scripts to container image

**Tilt Config**: Enabled mock mode via `values-tilt.yaml` (`MOCK_MODE: "true"`)

**Test**: Modified `smoke_test.go` to run on real nodes (`GetRealNodeName`) for actual execution validation

### 2. KWOK Failure Injection (Negative Test)

**KWOK Stage**: Created `log-collector-failure-stage.yaml` with node-specific selector for isolated failure injection

**Test Isolation**: New `log_collector_test.go` in separate file, uses KWOK nodes exclusively

**Helpers**: Added `ApplyKwokStageFromFile()`, `DeleteKwokStage()` with defensive cleanup via `t.Cleanup()`

**Node State Cleanup**: Teardown sends healthy event and waits for node uncordon to prevent test interference

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Log-collector: three-way image tag fallback, configurable LOG_COLLECTOR_TIMEOUT, injectable env vars, MOCK_MODE with local mock scripts; incluster file server enabled with logCleanup config.

* **Bug Fixes / Improvements**
  * Clearer local-artifact logging when uploads are absent; tighter upload-directory permissions.

* **Tests**
  * New e2e negative-stage for log-collector failures, KWOK-based scenarios, expanded test helpers, smoke checks validating job completion and uploads.

* **Chores**
  * Local dev build/workflow and Tilt integration updated for log-collector.


tests/helpers/kube.go
```diff
@@ -1577,65 +1775,43 @@
...
+	if err != nil {
+		// Log but don't fail - cleanup should be best-effort
+		t.Logf("Warning: failed to delete KWOK stage (non-fatal): %s", string(output))
+		return
+	}
 
-		return true
-	}, EventuallyWaitTimeout, WaitInterval)
+	t.Logf("KWOK stage deleted successfully")
 }
```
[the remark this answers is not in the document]
rupalis-nv: No linter errors

tests/helpers/kube.go
```diff
@@ -513,6 +544,47 @@ func GetAllNodesNames(ctx context.Context, c klient.Client) ([]string, error) {
...
+		// Skip KWOK nodes
+		if node.Labels["type"] == "kwok" {
+			continue
+		}
+
+		// Categorize real nodes
+		if _, isControlPlane := node.Labels["node-role.kubernetes.io/control-plane"]; isControlPlane {
+			realControlPlane = append(realControlPlane, node.Name)
+		} else {
+			realWorkers = append(realWorkers, node.Name)
```
lalitadithya: Can't we exit early when we find one worker node? Do we need to put them all into a map and then select the first one?

tests/helpers/kube.go
```diff
@@ -513,6 +544,47 @@ func GetAllNodesNames(ctx context.Context, c klient.Client) ([]string, error) {
 	return nodeNames, nil
 }
 
+// GetRealNodeName returns a real (non-KWOK) node name from the cluster.
+// Prefers worker nodes that are schedulable, falls back to any real node.
+func GetRealNodeName(ctx context.Context, c klient.Client) (string, error) {
```
lalitadithya, line 551: Do the nodes have some label (node role maybe) and we can select based on the labels?

tests/helpers/kube.go
```diff
@@ -1429,6 +1501,132 @@ func WaitForNodeConditionWithCheckName(
 	}, EventuallyWaitTimeout, WaitInterval, "node %s should have a condition with check name %s", nodeName, checkName)
 }
 
+// WaitForNodeConditionHealthy waits for the node condition to become healthy
+// (Status=False, Reason=checkName+"IsHealthy").
+// This is needed when testing with real nodes where gpu-health-monitor
+// takes time to update conditions asynchronously.
+func WaitForNodeConditionHealthy(
```
lalitadithya: Can't we use the existing `WaitForNodeConditionWithCheckName` function?

tests/helpers/kube.go
```diff
@@ -1429,6 +1501,132 @@ func WaitForNodeConditionWithCheckName(
...
+
+		return false
+	}, EventuallyWaitTimeout, WaitInterval, "node %s should have healthy condition %s", nodeName, checkName)
+}
+
+// WaitForNodeConditionUnhealthy waits for the node condition to become unhealthy
+// (Status=True, Reason=checkName+"IsNotHealthy").
+// This is needed when testing with real nodes where gpu-health-monitor
+// takes time to update conditions asynchronously.
+func WaitForNodeConditionUnhealthy(
```
lalitadithya: Can't we use the existing `WaitForNodeConditionWithCheckName` function?

distros/kubernetes/nvsentinel/values-tilt.yaml
```diff
@@ -124,6 +124,17 @@ fault-quarantine:
 fault-remediation:
   logLevel: debug
 
+  logCollector:
+    enabled: true
+    image:
+      repository: localhost:5001/ghcr.io_nvidia_nvsentinel_log-collector
```
XRFXLP, line 218: doesn't tilt already handles this?

tests/helpers/kube.go
```diff
@@ -1577,65 +1775,44 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
+func ApplyKwokStageFromFile(ctx context.Context, t *testing.T, stageFilePath string, nodeName ...string) {
+	t.Logf("Applying KWOK stage from %s", stageFilePath)
 
-	require.Eventually(t, func() bool {
-		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			node, err := GetNodeByName(ctx, client, nodeName)
-			if err != nil {
-				return err
-			}
+	// Read the stage file
```
lalitadithya: Do we need the obvious code comments?

tests/helpers/kube.go
```diff
@@ -1577,65 +1775,44 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
+	if len(nodeName) > 0 && nodeName[0] != "" {
+		stageYAML = strings.ReplaceAll(stageYAML, "NODE_NAME_PLACEHOLDER", nodeName[0])
+		t.Logf("Replaced NODE_NAME_PLACEHOLDER with: %s", nodeName[0])
+	}
 
-			for i := range node.Status.Conditions {
-				if node.Status.Conditions[i].Type == conditionType {
-					found = true
+	// Apply via kubectl
+	cmd := exec.CommandContext(ctx, "kubectl", "apply", "-f", "-")
```
lalitadithya: Can use the go-client to apply? execs are always fragile, what if the environment doesn't have kubectl or if it has a version who's output is different?

tests/helpers/kube.go
```diff
@@ -1577,65 +1775,44 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
-					Reason:             "TestCondition",
-					Message:            "Set by test",
-				})
-				modified = true
-			}
-
-			if !modified {
-				return nil
-			}
+	cmd := exec.CommandContext(ctx, "kubectl", "delete", "stage", stageName, "--ignore-not-found=true")
```
lalitadithya: Same comment as earlier

tests/log_collector_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+		// Defensive cleanup of any leftover KWOK Stage
+		t.Logf("Defensive cleanup: Removing any existing KWOK Stage")
+		helpers.DeleteKwokStage(ctx, t, "log-collector-failure")
+
+		// Ensure Stage is deleted even if test panics
+		t.Cleanup(func() {
+			t.Logf("Cleanup: Deleting KWOK Stage")
+			helpers.DeleteKwokStage(ctx, t, "log-collector-failure")
+		})
```
lalitadithya: AFAIK the feature.teardown should already take care of this. Are you seeing some cases where it is not being executed?

tests/helpers/kube.go
```diff
@@ -294,6 +295,36 @@ func WaitForNodesWithLabel(
 	}, EventuallyWaitTimeout, WaitInterval, "all nodes should have label %s=%s", labelKey, expectedValue)
 }
 
+// WaitForNodeLabelNotEqual waits for a single node's label to NOT equal a specific value.
+// This is useful for waiting for state transitions when you don't know the final state.
+func WaitForNodeLabelNotEqual(
+	ctx context.Context, t *testing.T, c klient.Client, nodeName, labelKey, notExpectedValue string,
+) {
```
XRFXLP, line 305: can you add `t.Helper()`

tests/log_collector_test.go
```diff
@@ -0,0 +1,141 @@
...
+		nodeName := ctx.Value(keyNodeName).(string)
+
+		client, err := c.NewClient()
+		assert.NoError(t, err, "failed to create kubernetes client")
+
+		// Log-collector job will fail due to KWOK Stage, causing remediation-failed label
+		t.Logf("Waiting for remediation-failed label on node %s (log-collector will fail)", nodeName)
+		helpers.WaitForNodesWithLabel(ctx, t, client, []string{nodeName}, statemanager.NVSentinelStateLabelKey, string(statemanager.RemediationFailedLabelValue))
+
+		t.Logf("✓ Remediation-failed label set correctly - negative test passed!")
```
lalitadithya: Doesn't the passing of the test imply that the labels are correct, do we need this log?

tests/log_collector_test.go
```diff
@@ -0,0 +1,141 @@
...
+		return ctx
+	})
+
+	feature.Assess("Remediation failed label is set due to log-collector failure", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		nodeName := ctx.Value(keyNodeName).(string)
+
+		client, err := c.NewClient()
+		assert.NoError(t, err, "failed to create kubernetes client")
+
+		// Log-collector job will fail due to KWOK Stage, causing remediation-failed label
```
lalitadithya: I don't follow; the failing of log collection should not cause the remediation to fail.

tests/helpers/kube.go
```diff
@@ -1429,6 +1501,132 @@ func WaitForNodeConditionWithCheckName(
...
 }
 
+// WaitForNodeConditionHealthy waits for the node condition to become healthy
+// (Status=False, Reason=checkName+"IsHealthy").
+// This is needed when testing with real nodes where gpu-health-monitor
+// takes time to update conditions asynchronously.
+func WaitForNodeConditionHealthy(
+	ctx context.Context, t *testing.T, c klient.Client, nodeName, checkName string,
+) {
+	require.Eventually(t, func() bool {
```
XRFXLP, line 1465: can we re-use CheckNodeConditionExists?

tests/helpers/kube.go
XRFXLP: can we re-use WaitForNodeConditionWithCheckName or modify that function itself to fit the usecase above as well?

tests/smoke_test.go
```diff
@@ -48,37 +48,10 @@ func TestFatalHealthEvent(t *testing.T) {
...
-			}
-		}
-
-		if nodeName == "" {
-			nodeName = unusedNodes[0]
-			t.Logf("No uncordoned node in unused subset, using first unused node: %s", nodeName)
-		} else {
-			t.Logf("Selected uncordoned node: %s (from unused nodes starting at index %d)", nodeName, startIdx)
-		}
+		// Use a real (non-KWOK) node for smoke test to validate actual container execution
```
lalitadithya, line 51: We should also validate:

1. the jobs was success
2. the log files were uploaded to the file server

log-collector/entrypoint.sh
```diff
@@ -250,4 +280,10 @@ if [ -n "${UPLOAD_URL_BASE:-}" ]; then
...
 fi
 
-echo "[INFO] Done. Artifacts under ${ARTIFACTS_DIR}"
\ No newline at end of file
+echo "[INFO] Done. Artifacts under ${ARTIFACTS_DIR}"
+
+# Exit with MOCK_EXIT_CODE if in mock mode (for testing failure scenarios)
+if [ "${MOCK_MODE}" = "true" ]; then
+  echo "[MOCK] Exiting with code: ${MOCK_EXIT_CODE}"
+  exit "${MOCK_EXIT_CODE}"
```
lalitadithya: I don't think we are using the mock exit code anymore, so can we drop this?

distros/kubernetes/nvsentinel/charts/fault-remediation/files/log-collector-job.yaml
```diff
@@ -36,7 +36,7 @@ spec:
       {{- end }}
       containers:
         - name: log-collector
-          image: {{ .Values.logCollector.image.repository }}:{{ .Values.global.image.tag | default "latest"  }}
+          image: {{ .Values.logCollector.image.repository }}:{{ .Values.logCollector.image.tag | default .Values.global.image.tag | default "latest"  }}
           imagePullPolicy: {{ .Values.logCollector.image.pullPolicy }}
```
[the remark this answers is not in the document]
rupalis-nv: `.Values.global.image.tag` does exist and is correctly accessible here. 

In Helm, `global` values defined in the parent chart (`distros/kubernetes/nvsentinel/values.yaml` lines 17-18) are automatically inherited by all subcharts. The fault-remediation subchart can access `.Values.global.image.tag` which has the value `"main"`.

The current fallback chain is correct:
1. `.Values.logCollector.image.tag` (log-collector specific override)
2. `.Values.global.image.tag` (inherited from parent chart: `"main"`)
3. `"latest"` (hardcoded default)

log-collector/Tiltfile
```diff
@@ -12,8 +12,13 @@
...
-    dockerfile="./Dockerfile"
+# Build docker image and push to local registry with fixed tag
+# Using local_resource since the image is used in dynamically created jobs
+# Build from repo root since Dockerfile expects that context
+local_resource(
+    "log-collector",
+    "cd .. && docker build -t localhost:5001/ghcr.io_nvidia_nvsentinel_log-collector:latest -f log-collector/Dockerfile . && docker push localhost:5001/ghcr.io_nvidia_nvsentinel_log-collector:latest",
+    deps=["./"],
+    ignore=["./Tiltfile"],
+    labels=["images"]
```
lalitadithya, line 23: docker_build has a tag property -- `extra_tag` -- https://docs.tilt.dev/api.html#api.docker_build -- does that not work as expected?

tests/log_collector_test.go
```diff
@@ -0,0 +1,141 @@
...
+		podTemplate := helpers.NewGPUPodSpec(workloadNamespace, 1)
+		helpers.CreatePodsAndWaitTillRunning(ctx, t, client, []string{nodeName}, podTemplate)
+
+		ctx = context.WithValue(ctx, keyNodeName, nodeName)
+		ctx = context.WithValue(ctx, keyNamespace, workloadNamespace)
+
+		return ctx
+	})
+
+	feature.Assess("Send unsupported health event to trigger log-collector", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
```
XRFXLP: What are we assessing in this `feature.Assess`?

distros/kubernetes/nvsentinel/values-tilt.yaml
rupalis-nv: Good catch! However, we do need this explicit configuration for the log-collector because of how it's deployed:

**Why explicit config is needed:**
1. **Dynamic job creation**: The log-collector runs as a Kubernetes Job that's created dynamically by the fault-remediation controller at runtime (not deployed directly via Helm)
2. **Runtime image reference**: When the controller creates the job, it reads these Helm values to determine which image to use. Tilt can't automatically inject the image reference into dynamically-created jobs
3. **Local registry**: In Tilt, we need the job to pull from `localhost:5001` (Tilt's local registry) rather than trying to pull from the public `ghcr.io` registry

**What Tilt handles vs. what we handle:**
- Tilt handles: Building the image, pushing to local registry, auto-reloading deployments
- Tilt doesn't handle: Updating image references in Jobs created dynamically by controllers at runtime

So this explicit configuration ensures that when a test triggers the log-collector job, it uses the Tilt-built local image rather than trying to pull from the remote registry. 👍

tests/helpers/kube.go
rupalis-nv: Good point! 
But i thought in this way that - we do need to collect all workers to prefer schedulable ones. The logic is:
1. Collect all real workers (schedulable and unschedulable)
2. Prefer schedulable workers first
3. Fall back to unschedulable/control-plane if needed

This ensures we get the best node for testing. If we exit early, we might pick an unschedulable node when a schedulable one exists later in the list.
U can suggest best way i will implement in the same manner

tests/helpers/kube.go
rupalis-nv: Yes! We're already using `node-role.kubernetes.io/control-plane` label (line shown in your comment) to distinguish control-plane from worker nodes. We filter KWOK nodes by checking `type=kwok` label, and then separate real nodes into workers and control-plane based on the role label.
Is there something else u r trying to convey?

lalitadithya: Yes, don't the worker nodes also have a different `node-role.kubernetes.io/control-plane` which is not present in kwok nodes?

tests/helpers/kube.go
rupalis-nv: `WaitForNodeConditionWithCheckName` only checks if a condition exists with a given check name, but doesn't validate the Status (True/False) or Reason fields.

These new functions specifically check:
- Healthy: Status=False AND Reason="<checkName>IsHealthy"
- Unhealthy: Status=True AND Reason="<checkName>IsNotHealthy"

This is critical for the race condition fix - we need to wait for gpu-health-monitor to fully update the condition to the expected state, not just check if the condition exists. Should I refactor the existing function to support these checks, or keep them separate for clarity?
I faced the race condition issue while testing locally so i came up with this solution
U can suggest if u have some better suggestions

tests/helpers/kube.go
lalitadithya: > If we exit early, we might pick an unschedulable node when a schedulable one exists later in the list.

We can check for that in the loop right? Find the first schedulable worker node and return from the function?

tests/helpers/kube.go
lalitadithya: Can we update the existing function to add the required additional validation?

tests/helpers/kube.go
rupalis-nv: KWOK Stages are a CRD and require dynamic YAML application. The go-client would need:
Dynamic client to handle unknown CRDs,YAML parsing and type conversion, GVK discovery
Isn't this adds complexity.

tests/log_collector_test.go
rupalis-nv: Good catch! The `t.Cleanup()` is defensive - it guarantees Stage deletion even if the test panics or exits early before reaching Teardown. This prevents test pollution where a failed test leaves a Stage that affects subsequent tests.

log-collector/Tiltfile
rupalis-nv: I initially tried `extra_tag` but ran into issues with the job template needing the specific repository format (`localhost:5001/ghcr.io_nvidia_nvsentinel_log-collector`). The manual build approach guarantees the exact image reference the dynamically-created job needs. Happy to revisit if there's a cleaner way with `extra_tag` - any suggestions?

lalitadithya: Just for my curiosity, what were the issues? Was it that the repository format was dynamic?

tests/helpers/kube.go
lalitadithya: Hmmm. If we add https://github.com/kubernetes-sigs/kwok/tree/main/pkg/apis to the scheme it is no longer dynamic right?

tests/log_collector_test.go
lalitadithya: But isn't the teardown always executed even if the test errors or panics?

tests/helpers/kube.go
rupalis-nv: Updated to use label selector `type!=kwok` in the List() call to filter out KWOK nodes at the API level rather than in the loop. 
We can't use a positive selector for worker nodes (like `node-role.kubernetes.io/worker`) because that label is optional in many K8s distributions. Worker nodes are identified by the *absence* of the `node-role.kubernetes.io/control-plane` label, which we still check in the loop. We also need to check `node.Spec.Unschedulable` which is a spec field, not a label, so some iteration is still necessary.

tests/helpers/kube.go
rupalis-nv: by importing `github.com/kubernetes-sigs/kwok/tree/main/pkg/apis` and adding KWOK types to the scheme, we can use the typed go-client instead of kubectl exec. but this would add a new dependency to the project (KWOK APIs). 
Should we add this dependency for better type safety, or is the current kubectl approach acceptable for test code? I'm happy to refactor if you think the dependency is worth it for the improved robustness.

tests/log_collector_test.go
rupalis-nv: After analyzing the e2e-framework code, `feature.Teardown()` uses `defer` internally and IS guaranteed to run even if Setup or Assess panic. The `t.Cleanup()` is indeed redundant.
I'll remove the `t.Cleanup()` and keep only:
1. Line 43: Defensive cleanup at the start of Setup (clears leftover state from previous failed tests)
2. Line 113: `feature.Teardown()` handles the normal cleanup path
This simplifies the code while maintaining proper cleanup. The defensive deletion at the start of Setup is still valuable to handle cases where a previous test run was forcefully terminated (SIGKILL) before any cleanup could run.

tests/helpers/kube.go
lalitadithya: the dependency will be for the test code, so it should be fine. I think we can add it.

tests/helpers/kube.go
lalitadithya: > because that label is optional in many K8s distributions

True, but in our kind cluster do we set it. Since this is test, we don't have to solve for every distribution, we just need to make it work with our kind cluster.

tests/log_collector_test.go
lalitadithya: > Defensive cleanup at the start of Setup (clears leftover state from previous failed tests)

I'm not sure we need this. The test command is configured to stop when the first test fails, so there should not be any previous failed tests.

log-collector/Tiltfile
rupalis-nv: The issue is that the log-collector job is dynamically created by the fault-remediation controller at runtime(not deployed via Helm), which creates a challenge for Tilt's automatic image reference rewriting:
Why `docker_build` + `extra_tag` doesn't work:
1. Tilt builds with hash-based tags:
   docker_build('ghcr.io/nvidia/nvsentinel/log-collector', ...)
    → produces: ghcr.io/nvidia/nvsentinel/log-collector:tilt-abc123def
2. Tilt rewrites image references in YAML:
   - Tilt automatically rewrites `ghcr.io/nvidia/nvsentinel/log-collector:latest` → `ghcr.io/nvidia/nvsentinel/log-collector:tilt-abc123def`
   - This works great for static Helm charts/k8s manifests
3. But dynamic job creation breaks this flow:
   - The fault-remediation controller reads `.Values.logCollector.image.repository` from the ConfigMap
   - It creates the job at runtime using Go code: `job.Spec.Template.Spec.Containers[0].Image = repository + ":" + tag`
   - Tilt has no visibility into this dynamic job creation → can't rewrite the image reference
4. The `extra_tag` doesn't solve it:
   extra_tags=['localhost:5001/ghcr.io_nvidia_nvsentinel_log-collector:latest']
      - This would tag the image with the fixed name, BUT
   - Tilt would still only track the base ref (`ghcr.io/nvidia/nvsentinel/log-collector:tilt-abc123def`)
   - The dynamic job would reference `localhost:5001/ghcr.io_nvidia_nvsentinel_log-collector:latest`
   - Tilt wouldn't trigger rebuilds when that image changes (not tracking the extra_tag)
Why `local_resource` works:
- Builds and pushes with the **exact fixed name** the dynamic job expects
- No reliance on Tilt's image rewriting (which doesn't work for runtime-created resources)
- Simple and explicit: "build this image, push it to registry, job will find it"

tests/log_collector_test.go
rupalis-nv: this step is more of a "trigger/action" than an assessment. It just sends the health event and verifies the SendHealthEventsToNodes API call succeeded (doesn't fail).
We could rename this to `feature.Setup()` for the second setup step, or rename it to something like "Trigger health event" to be clearer about its role. What's your preference?

log-collector/Tiltfile
lalitadithya: It looks like only the auto rebuilds will not work right? Everything else should work?

rupalis-nv: yes correct

lalitadithya: Okay, I think instead of doing custom logic, let's use the function provided by Tilt and we can add a comment saying that auto rebuild will not work. What do you think?

rupalis-nv: We can do it
But i am just afraid I can try it once

rupalis-nv: Its risky because even when we do single change in any file it will trigger tilt rebuild and when anyone does run tests it will fail

tests/log_collector_test.go
XRFXLP: can we merge this into the next assess step i.e. assert node is cordoned?

log-collector/Tiltfile
rupalis-nv: I tried this and its not working out 
Why extra_tag Doesn't Solve It
From Tilt docs:
> extra_tag: Additional tags to apply to the built image on the local Docker daemon
extra_tag only tags locally, it does NOT push to the registry. Even if we use:
docker_build('localhost:5001/...', extra_tag='localhost:5001/...:latest')
Tilt will:
Tag locally as :latest
Only push the primary reference with hash: :tilt-abc123def
Jobs can't pull :latest from registry (not there!)

tests/helpers/kube.go
rupalis-nv: Addressed

tests/helpers/kube.go
rupalis-nv: I've updated the implementation to use multiple negative selectors. 
> resources.WithLabelSelector("type!=kwok,!node-role.kubernetes.io/control-plane,!node-role.kubernetes.io/agent")

tests/helpers/kube.go
rupalis-nv: addressed

tests/log_collector_test.go
rupalis-nv: Addressed

tests/log_collector_test.go
rupalis-nv: addressed

tests/helpers/kube.go
rupalis-nv: addressed

tests/helpers/kube.go
rupalis-nv: addressed

tests/helpers/kube.go
rupalis-nv: addressed

tests/helpers/kube.go
rupalis-nv: addressed

tests/log_collector_test.go
rupalis-nv: addressed

tests/smoke_test.go
rupalis-nv: addressed

log-collector/entrypoint.sh
rupalis-nv: addressed

log-collector/Tiltfile
rupalis-nv: Moving forward with local_resource

tests/helpers/kube.go
rupalis-nv: addressed

tests/helpers/kube.go
rupalis-nv: addressed

tests/log_collector_test.go
rupalis-nv: addressed

tests/helpers/kube.go
rupalis-nv: addressed

log-collector/entrypoint.sh
```diff
@@ -248,6 +276,8 @@ if [ -n "${UPLOAD_URL_BASE:-}" ]; then
       echo "[UPLOAD_FAILED] Failed to upload AWS SOS report: $(basename "${AWS_SOS_REPORT}")" >&2
     fi
   fi
+elif [ "${MOCK_MODE}" = "true" ]; then
+  echo "[MOCK] Skipping upload in mock mode - artifacts created locally at ${ARTIFACTS_DIR}"
```
lalitadithya: Why? We should still update the artefacts in mock mode and validate it in tests

tests/helpers/kube.go
```diff
@@ -1399,9 +1453,14 @@ func PortForwardPod(
 }
 
 // WaitForNodeConditionWithCheckName waits for the node to have a condition with the reason as checkName.
+// WaitForNodeConditionWithCheckName waits for a node condition matching the given criteria.
+// Optional parameters (use nil to skip): expectedStatus, expectedReason, expectedMessage.
 func WaitForNodeConditionWithCheckName(
-	ctx context.Context, t *testing.T, c klient.Client, nodeName, checkName, message string,
+	ctx context.Context, t *testing.T, c klient.Client, nodeName, checkName string,
+	expectedStatus *v1.ConditionStatus, expectedReason, expectedMessage *string,
```
lalitadithya: If we don't use pointers, we can just check for empty strings so that all callers don't have to create pointers for simple strings

```suggestion
	expectedStatus *v1.ConditionStatus, expectedReason, expectedMessage string,
```

tests/helpers/kube.go
```diff
@@ -1577,65 +1713,252 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
-					Type:               conditionType,
-					Status:             status,
-					LastTransitionTime: now,
-					LastHeartbeatTime:  now,
-					Reason:             "TestCondition",
-					Message:            "Set by test",
-				})
-				modified = true
-			}
+	t.Logf("✓ KWOK stage %s applied successfully", stage.Name)
```
lalitadithya: If there were no errors, doesn't it already indicate that the stage was applied correctly?

tests/helpers/kube.go
```diff
@@ -1577,65 +1713,252 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
+		if apierrors.IsNotFound(err) {
+			t.Logf("KWOK stage %s already deleted", stageName)
+		} else {
+			t.Logf("Warning: failed to delete KWOK stage %s: %v", stageName, err)
+		}
+
+		return
+	}
+
+	t.Logf("✓ KWOK stage %s deleted successfully", stageName)
```
lalitadithya: Same comment as earlier

tests/helpers/kube.go
```diff
@@ -1577,65 +1713,252 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
+		err := c.Resources().Delete(ctx, &job, func(do *metav1.DeleteOptions) {
+			do.PropagationPolicy = &deletePolicy
 		})
+		if err != nil && !apierrors.IsNotFound(err) {
+			t.Logf("Warning: failed to delete job %s: %v", job.Name, err)
+		}
+	}
+
+	// Wait a bit for jobs to be fully deleted
+	time.Sleep(2 * time.Second)
```
lalitadithya: What if the job doesn't get deleted in 2 seconds? Can we instead use require.Eventually and check for jobs not found?

tests/helpers/kube.go
```diff
@@ -1577,65 +1713,252 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
+	var podList v1.PodList
+
+	err := c.Resources().List(ctx, &podList,
+		resources.WithLabelSelector(fmt.Sprintf("job-name=%s", job.Name)))
+	require.NoError(t, err, "failed to list pods for job %s", job.Name)
+	require.NotEmpty(t, podList.Items, "no pods found for job %s", job.Name)
+
+	// Get pod logs
+	pod := &podList.Items[0]
+	logs, err := getPodLogs(ctx, c, pod.Namespace, pod.Name, "log-collector")
```
lalitadithya: No. No parsing logs. We should instead download the file and check if it was uploaded.

distros/kubernetes/nvsentinel/charts/incluster-file-server/templates/deployment.yaml
```diff
@@ -59,7 +59,7 @@ spec:
         - name: init-perms
           image: "{{ .Values.initImage.repository }}:{{ .Values.initImage.tag }}"
           imagePullPolicy: {{ .Values.initImage.pullPolicy }}
-          command: ["sh", "-lc", "mkdir -p /usr/share/nginx/html/upload /var/cache/nginx /var/run /opt/bitnami/nginx/tmp/client_body /opt/bitnami/nginx/tmp/proxy /opt/bitnami/nginx/tmp/fastcgi /opt/bitnami/nginx/tmp/uwsgi /opt/bitnami/nginx/tmp/scgi && chmod -R g+rwX /usr/share/nginx/html /var/cache/nginx /opt/bitnami/nginx/tmp || true"]
+          command: ["sh", "-lc", "mkdir -p /usr/share/nginx/html/upload /var/cache/nginx /var/run /opt/bitnami/nginx/tmp/client_body /opt/bitnami/nginx/tmp/proxy /opt/bitnami/nginx/tmp/fastcgi /opt/bitnami/nginx/tmp/uwsgi /opt/bitnami/nginx/tmp/scgi && chmod -R 777 /usr/share/nginx/html/upload && chmod -R g+rwX /var/cache/nginx /opt/bitnami/nginx/tmp || true"]
```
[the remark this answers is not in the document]
rupalis-nv: chmod 777 is required because nginx runs as 1001:1001 while the directory is owned by root:101 (fsGroup). Since nginx is neither the owner nor in group 101, it needs "others" write permission to create subdirectories via WebDAV. This is acceptable as the directory is isolated within the pod's PVC and not exposed outside the cluster.

tests/data/log-collector-failure-stage.yaml
```diff
@@ -0,0 +1,55 @@
...
+        started: false
+        state:
+          terminated:
+            exitCode: 1
+            finishedAt: {{ $now }}
+            reason: Error
+            message: "Simulated failure for log-collector testing"
+            startedAt: {{ $now }}
+      phase: Failed
+
```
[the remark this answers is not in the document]
rupalis-nv: addressed

distros/kubernetes/nvsentinel/charts/incluster-file-server/templates/deployment.yaml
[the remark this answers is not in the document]
rupalis-nv: added

tests/helpers/kube.go
```diff
@@ -1577,65 +1712,293 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
+		}
+
+		t.Logf("No log-collector job found for node %s yet", nodeName)
+
+		return false
+	}, EventuallyWaitTimeout, WaitInterval,
+		"log-collector job for node %s should reach status %s", nodeName, expectedStatus)
+
+	return foundJob
+}
```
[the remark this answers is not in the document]
rupalis-nv: The label selector works correctly as-is. Kubernetes automatically propagates labels from spec.template.metadata.labels to the Job's own metadata.labels. When the Job controller creates the Job, it copies the app: log-collector label from the pod template to the Job object itself, making it discoverable by resources.WithLabelSelector("app=log-collector"). No changes needed.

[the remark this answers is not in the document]
rupalis-nv: I've verified this behavior directly in our Kubernetes cluster. When a Job has labels in `spec.template.metadata.labels, Kubernetes automatically propagates them to the Job's metadata.labels.`
Proof:
Created a test Job with label only in pod template:
`spec:  
          template:    
                  metadata: 
                        labels:  
                               my-test-label: propagated`
Result:
`$ kubectl get job test-label-propagation -o jsonpath='{.metadata.labels}'{"my-test-label": "propagated", ...}$ kubectl get jobs -l my-test-label=propagatedNAME                     STATUS    COMPLETIONStest-label-propagation   Running   0/1`
The app=log-collector label from the pod template is automatically added to the Job by the Job controller, making resources.WithLabelSelector("app=log-collector") work correctly. Our tests confirm this works as expected.

tests/helpers/kube.go
rupalis-nv: addressed

log-collector/Tiltfile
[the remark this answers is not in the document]
rupalis-nv: The current configuration is correct. The Dockerfile only copies files from the log-collector/ directory (lines 44-46), so deps=["./"] correctly watches all files that could trigger a rebuild. While the Docker build context is the repository root (required by the Dockerfile's COPY log-collector/... commands), no files outside log-collector/ are actually used in the build. Expanding deps to ["../"] would cause unnecessary rebuilds when unrelated files in the repository change.

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
```diff
@@ -105,3 +106,5 @@ logCollector:
   enableGcpSosCollection: false
   # Enable AWS-specific SOS report collection
   enableAwsSosCollection: false
+  # Timeout in seconds for waiting for log collection job to complete (default: 600s/10min)
+  timeoutSeconds: 600
```
XRFXLP: Should we pass a general duration here? Like `1h`, `10m` or `200s`? Any reason for sticking with seconds?

tests/health_events_analyzer_test.go
```diff
@@ -230,7 +238,9 @@ func TestRepeatedXIDRule(t *testing.T) {
 		}
 
 		message += fmt.Sprintf("ErrorCode:%s GPU:0 Recommended Action=CONTACT_SUPPORT;", helpers.ERRORCODE_31)
-		helpers.WaitForNodeConditionWithCheckName(ctx, t, client, testCtx.NodeName, "RepeatedXidError", message)
+		statusTrue3 := v1.ConditionTrue
```
XRFXLP: should we just declare this as a constant in helpers or somewhere rather than re-declaring everytime?

tests/helpers/kube.go
```diff
@@ -1577,65 +1712,293 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
 
-		return true
-	}, EventuallyWaitTimeout, WaitInterval)
+		return len(remainingJobs.Items) == 0
+	}, EventuallyWaitTimeout, WaitInterval, "log-collector jobs should be deleted")
+
+	t.Logf("Log-collector jobs cleanup completed")
+}
+
+// WaitForLogCollectorJobStatus waits for a log-collector job to reach the specified status.
```
XRFXLP: Can we move this comment to the function below as it talks about that?

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
rupalis-nv: Not any specific reason
my thought process was to keep shorter timeouts
If u suggest  i can pass 10 m?

tests/helpers/kube.go
rupalis-nv: Addressed

tests/health_events_analyzer_test.go
rupalis-nv: I have used inline approach

distros/kubernetes/nvsentinel/charts/fault-remediation/files/log-collector-job.yaml
```diff
@@ -61,6 +61,10 @@ spec:
               value: {{ .Values.logCollector.enableGcpSosCollection | quote }}
             - name: ENABLE_AWS_SOS_COLLECTION
               value: {{ .Values.logCollector.enableAwsSosCollection | quote }}
+            {{- range $key, $value := .Values.logCollector.env }}
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+            {{- end }}
```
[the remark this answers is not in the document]
rupalis-nv: addressed

tests/data/log-collector-failure-stage.yaml
```diff
@@ -0,0 +1,55 @@
...
```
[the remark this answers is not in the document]
rupalis-nv: addressed

tests/log_collector_test.go
```diff
@@ -0,0 +1,135 @@
...
+		t.Logf("Cleaning up old log-collector jobs from previous tests")
+		helpers.DeleteAllLogCollectorJobs(ctx, t, client)
+
+		// Apply KWOK Stage FIRST to ensure it's active before any pods are created
+		helpers.ApplyKwokStageFromFile(ctx, t, client, "data/log-collector-failure-stage.yaml")
+		t.Logf("KWOK Stage applied - ALL log-collector pods will fail during this test")
+
+		// Use a KWOK node for negative test
+		nodeName := helpers.SelectTestNodeFromUnusedPool(ctx, t, client)
+		t.Logf("Selected KWOK node for log-collector failure test: %s", nodeName)
```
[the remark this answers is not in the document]
rupalis-nv: addressed

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
rupalis-nv: @XRFXLP  I have adressed your feedback and made a timeout as 10m by default
And for your concern around must-gather timeout I will be looking for it in the issue -[https://github.com/NVIDIA/NVSentinel/issues/441](https://github.com/NVIDIA/NVSentinel/issues/441)

Please take a look at comments - [https://github.com/NVIDIA/NVSentinel/issues/441#issuecomment-3576812493](https://github.com/NVIDIA/NVSentinel/issues/441#issuecomment-3576812493) and [https://github.com/NVIDIA/NVSentinel/issues/441#issuecomment-3577012237](https://github.com/NVIDIA/NVSentinel/issues/441#issuecomment-3577012237)

tilt/Tiltfile
```diff
@@ -282,3 +287,11 @@ k8s_resource(
     objects=['pod-delete:stage'],
     resource_deps=['kwok', 'kwok-stage-fast']
 )
+
+# Incluster file server - optional component for log collection
+# Make it wait for cert-manager like other resources (even though it doesn't need certs itself,
```
lalitadithya: I don't follow, if it doesn't need the cert manager resources, why does it need to wait for it?

tests/helpers/kube.go
rupalis-nv: The function does check if the file was uploaded by running ls -lh inside the file server pod (lines 1968-1974)
However, it still parses logs (lines 1930-1939) to get the timestamp. Because files are save with timestamp, so to find those files we r using logs to get timestamp and then checking if files are uploaded

tilt/Tiltfile
rupalis-nv: I realized, i added it for safer side 
I tried testing by removing it because its actually not needed and its working fine hence removed

tests/helpers/kube.go
```diff
@@ -1577,65 +1713,269 @@ func VerifyEventsMatchPatterns(t *testing.T, ctx context.Context,
...
-					}
+	stage, ok := obj.(*kwokv1alpha1.Stage)
+	require.True(t, ok, "decoded object is not a KWOK Stage")
 
-					break
-				}
-			}
+	err = c.Resources().Create(ctx, stage)
+	require.NoError(t, err, "failed to create KWOK stage %s", stage.Name)
+}
```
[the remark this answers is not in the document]
rupalis-nv: addressed

tests/log_collector_test.go
```diff
@@ -0,0 +1,142 @@
...
+// 3. Log-collector failure is independent - doesn't set remediation-failed label
+func TestLogCollectorFailure(t *testing.T) {
+	feature := features.New("TestLogCollectorFailure").
+		WithLabel("suite", "smoke")
+
+	feature.Setup(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		workloadNamespace := "logcollector-failure-test"
+
+		client, err := c.NewClient()
+		assert.NoError(t, err, "failed to create kubernetes client")
```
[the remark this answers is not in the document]
rupalis-nv: Addressed

distros/kubernetes/nvsentinel/values-tilt.yaml
```diff
@@ -206,6 +212,15 @@ fault-quarantine:
...
+      tag: latest  
+      pullPolicy: Always 
+    timeout: "10s"  # Short timeout for faster testing (production default: "10m")
+    env:
+      MOCK_MODE: "true"
```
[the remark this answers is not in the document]
rupalis-nv: We can skip comments nothing in values.tilt has so many comments and just implementation

distros/kubernetes/nvsentinel/values-tilt.yaml
```diff
@@ -56,7 +56,7 @@ global:
       enabled: false
 
   inclusterFileServer:
-    enabled: false
+    enabled: true
```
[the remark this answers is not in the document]
rupalis-nv: We can skip comments in values.tilt because whole values.tilt has very less comments

tests/smoke_test.go
```diff
@@ -204,6 +194,14 @@ func TestFatalHealthEvent(t *testing.T) {
 		t.Logf("Waiting for node %s to be uncordoned", nodeName)
 		helpers.WaitForNodesCordonState(ctx, t, client, []string{nodeName}, false)
 
+		// Wait for gpu-health-monitor to update the node condition to healthy
```
XRFXLP: This can cause confusion as GPU health monitor doesn't updates the node condition.

rupalis-nv: Removed all unnecessary comments

lalitadithya (approved):

PR #388, opened 2025-11-21
feat: define proto for GPUs

### LINKED ISSUE #275: [Feature]: Create a generic GPU Health API that any consumer can invoke to determine the health of GPUs
### Prerequisites

### Feature Summary

Currently we don't have an easy way to determine if a GPU is healthy, consumers can either look at mongodb or node conditions. There is a strong desire to have an API that can be invoked by operators or automation alike to react to GPU health changes.

### Problem/Use Case

Make it easier for operator and automation alike to react to GPU health changes

### Proposed Solution

TBD

### Component

Health Monitor

pteranodan:
## Summary
**Request for Comments**

Introduces initial protobuf definitions for a GPU API (see https://github.com/NVIDIA/NVSentinel/issues/275 for additional details)

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Introduced a GPU Service API to query and monitor GPU devices: retrieve single GPU details, list all GPUs, and watch GPU status updates (including status, conditions, and recommended actions).


api/proto/gpu/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,249 @@
...
+  google.protobuf.FieldMask read_mask = 2;
+}
+
+message GetGpuResponse {
+  Gpu gpu = 1;
+}
+
+message ListGpusRequest {
+  // A selector to filter objects by their fields.
+  // Example: "spec.node_name=worker-1"
```
lalitadithya: I don't think we will be able to list GPUs across nodes. Since the API is local to a node, it might be better to use a different example.

api/proto/gpu/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,249 @@
...
+  string type = 1;
+  string status = 2;
+  int64 observed_generation = 3;
+  google.protobuf.Timestamp last_transition_time = 4;
+  string reason = 5;
+  string message = 6;
+}
+
+// GpuConditionType corresponds to the Go type 'v1alpha1.GPUConditionType'.
+enum GpuConditionType {
```
lalitadithya: Can we simplify this, can we instead just have 

```
  GPU_CONDITION_TYPE_UNSPECIFIED = 0;
  GPU_CONDITION_TYPE_READY = 1;
  GPU_CONDITION_TYPE_DEGRADED = 2;
  GPU_CONDITION_TYPE_FAILED = 3;
```

what we need to do to recover the GPU, I think will be covered in the recommended action right?

api/proto/gpu/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,249 @@
...
+
+// GpuReadyReasonType corresponds to the Go type 'v1alpha1.GPUReadyReason'.
+enum GpuReadyReasonType {
+  GPU_READY_REASON_TYPE_UNSPECIFIED = 0;
+  GPU_READY_REASON_TYPE_DRIVER_READY = 1;
+  GPU_READY_REASON_TYPE_DRIVER_INIT_FAILURE = 2;
+  GPU_READY_REASON_TYPE_STATUS_UNKNOWN = 3;
+}
+
+// GpuDegradedReasonType corresponds to the Go type 'v1alpha1.GPUDegradedReason'.
```
lalitadithya: Now that I think about this more, I think we should make these free form strings. These can evolve over time and having an enum might lead to more maintenance overhead over time.

api/proto/gpu/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,249 @@
...
+// NOTE: This schema defines the wire format for the GPU API.
+//
+// For detailed documentation on field behaviors, enum definitions, etc.,
+// please refer to the canonical Go definitions in 'api/v1alpha1/'.
+// =============================================================================
+
+// Gpu represents a single GPU resource.
+message Gpu {
+  string name = 1;
+  string generate_name = 2;
```
lalitadithya: I'm not sure if we want to generate names for the GPU. every GPU has a well defined ID and I think we should just make use of that. It will make things simpler.

api/proto/gpu/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,249 @@
...
+// GpuSpec defines the desired state for a specific GPU.
+message GpuSpec {
+  string id = 1;
+  string node_name = 2;
+}
+
+// GpuStatus defines the observed state of a single GPU.
+message GpuStatus {
+  repeated Condition conditions = 1;
+  repeated GpuRecommendedActionType recommended_actions = 2;
```
lalitadithya: Shouldn't the recommended action be in the condition?

api/proto/gpu/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// Condition mimics k8s.io/apimachinery/pkg/apis/meta/v1.Condition.
+message Condition {
+  string type = 1;
+  string status = 2;
+  int64 observed_generation = 3;
```
lalitadithya: I don't follow this field, can you please explain what we would put in here?

api/proto/gpu/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,249 @@
...
+  GPU_RECOMMENDED_ACTION_TYPE_RESTART_VM = 3;
+  GPU_RECOMMENDED_ACTION_TYPE_REBOOT_NODE = 4;
+  GPU_RECOMMENDED_ACTION_TYPE_RESET_FABRIC = 5;
+  GPU_RECOMMENDED_ACTION_TYPE_RUN_EXTENDED_UTILITY_DIAGNOSTICS = 6;
+  GPU_RECOMMENDED_ACTION_TYPE_RUN_FIELD_DIAGNOSTICS = 7;
+  GPU_RECOMMENDED_ACTION_TYPE_REPORT_ISSUE = 8;
+}
+
+// GpuTaintType corresponds to the Go type 'v1alpha1.GPUTaintType'.
+enum GpuTaintType {
```
lalitadithya: I don't see this being used in the API definition, is that expected?

api/proto/gpu/v1alpha1/gpu.proto
pteranodan: Agreed. I'll remove the enums.

api/proto/gpu/v1alpha1/gpu.proto
pteranodan: Not needed. It was a straight copy from the `k8s.io/apimachinery/pkg/apis/meta/v1.Condition` type to avoid having to include extra dependencies. Removed.

api/proto/gpu/v1alpha1/gpu.proto
pteranodan: Good catch! Removed.

api/proto/gpu/v1alpha1/gpu.proto
pteranodan: Removed all the filters to simplify this initial version of the API per your other comment.

api/device/v1alpha1/conversion.go
```diff
@@ -0,0 +1,206 @@
...
+// limitations under the License.
+
+package v1alpha1
+
+import (
+	"google.golang.org/protobuf/types/known/timestamppb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	pb "github.com/nvidia/nvsentinel/api/gen/go/device/v1alpha1"
+)
```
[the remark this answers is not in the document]
pteranodan: fixed

api/Makefile
```diff
@@ -0,0 +1,144 @@
...
+		--bounding-dirs github.com/nvidia/nvsentinel/api/device/v1alpha1 \
+		--go-header-file="" \
+		--output-file zz_generated.deepcopy.go \
+	  github.com/nvidia/nvsentinel/api/device/v1alpha1
+	@echo "Done."
+
+.PHONY: protos-generate
+protos-generate: $(PROTOC_GEN_GO) $(PROTOC_GEN_GO_GRPC) $(PROTOC_GEN_GRPC_GATEWAY) protos-clean ## Generate Go code from Proto definitions.
+	@[ -d "$(THIRD_PARTY_DIR)" ] || { echo "Error: $(THIRD_PARTY_DIR) not found. Run 'make vendor-protos' from the repo root."; exit 1; }
+	@echo "Cleaning old generated code..."
```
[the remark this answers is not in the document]
pteranodan: fixed

ArangoGutierrez:
++

api/device/v1alpha1/validation.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	if !GPUIDRegex.MatchString(g.Spec.ID) {
+		return fmt.Errorf("spec.id '%s' is invalid: must be a valid NVIDIA GPU UUID (e.g. GPU-1234...)", g.Spec.ID)
+	}
+
+	if g.Name != "" && !NameRegex.MatchString(g.Name) {
+		return fmt.Errorf("name '%s' is invalid: must be a valid DNS-1123 subdomain", g.Name)
+	}
+
+	expectedName, _ := g.ComputeCanonicalName()
```
ArangoGutierrez: is there a valid reason why not to err check here?

api/device/v1alpha1/types.go
```diff
@@ -0,0 +1,72 @@
...
+// +k8s:deepcopy-gen=package
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen=true
+// GPU represents a single GPU resource.
+type GPU struct {
```
ArangoGutierrez: So on the meeting we discussed the naming "device" instead of GPU, whould that make sense here and then change 

```go
type DeviceSpec struct {
	// ID is the Device UUID.
	ID string `json:"id"`
	// DeviceType is the name of the Device [gpu,nic,storage].
	DeviceType string `json:"nodeName"`
	// NodeName is the name of the node where the GPU is located.
	NodeName string `json:"nodeName"`
}
```

Makefile
```diff
@@ -11,6 +11,7 @@ ENVTEST := setup-envtest
 # Variables
 GOPATH ?= $(shell go env GOPATH)
 GO_CACHE_DIR ?= $(shell go env GOCACHE)
+THIRD_PARTY_DIR := third_party
```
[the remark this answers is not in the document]
pteranodan: fixed.

api/device/v1alpha1/validation.go
pteranodan: No. CodeRabbit identified that as well (https://github.com/NVIDIA/NVSentinel/pull/388#discussion_r2550416956). I'll fix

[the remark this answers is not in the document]
pteranodan: fixed.

third_party/google/protobuf/descriptor.proto
```diff
@@ -0,0 +1,1426 @@
...
+// Describes the 'visibility' of a symbol with respect to the proto import
+// system. Symbols can only be imported when the visibility rules do not prevent
+// it (ex: local symbols cannot be imported).  Visibility modifiers can only set
+// on `message` and `enum` as they are the only types available to be referenced
+// from other files.
+enum SymbolVisibility {
+  VISIBILITY_UNSET = 0;
+  VISIBILITY_LOCAL = 1;
+  VISIBILITY_EXPORT = 2;
+}
```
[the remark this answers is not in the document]
pteranodan: removed.

api/device/v1alpha1/types.go
ArangoGutierrez: @dims WDYT

ArangoGutierrez: @pteranodan and I discussed this offline , we can resolve this comment

api/proto/gpu/v1alpha1/gpu.proto
pteranodan: Removed.

ArangoGutierrez (approved):

api/proto/gpu/v1alpha1/gpu.proto
pteranodan: enums removed.

ArangoGutierrez (approved):
LGTM

.versions.yaml
```diff
@@ -46,13 +46,16 @@ go_tools:
...
+  deepcopy_gen: 'v0.30.0'
 
 # Protocol Buffers / gRPC
 protobuf:
   protobuf: 'v33.0'
   protoc_gen_go: 'v1.36.10'
   protoc_gen_go_grpc: 'v1.5.1'
+  protoc_gen_grpc_gateway: 'v2.20.0'
   grpcio_tools: '1.75.1'
+  googleapis: '67782cb57b1fd81a4f49edef8243acb02843bf75'
```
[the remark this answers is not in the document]
pteranodan: outdated.

api/proto/gpu/v1alpha1/gpu.proto
pteranodan: I kept them separate to distinguish between state and remediation. A Condition describes what is happening to the device, while recommended_actions describes what to do.

api/proto/device/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,171 @@
...
+import "google/api/annotations.proto";
+import "google/api/http.proto";
+import "google/protobuf/timestamp.proto";
+
+// ==========================================
+// Resource Definitions
+// ==========================================
+
+// Gpu represents a single GPU resource in Kubernetes.
+message Gpu {
```
pteranodan, line 30: I think we need to add `resource_version` here so Watch events are valid and strictly ordered.

api/proto/device/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,171 @@
...
+  // spec defines the identity and desired attributes of the GPU resource.
+  GpuSpec spec = 2;
+
+  // status contains the most recently observed state of the GPU resource.
+  // This data may lag slightly behind the actual on-device state.
+  GpuStatus status = 3;
+}
+
+// GpuList is a collection of GPU resources.
+message GpuList {
```
pteranodan, line 47: We should also add `resource_version` to the list response; without it, clients can't safely transition from listing items to watching them.

api/proto/device/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,171 @@
...
+import "google/protobuf/timestamp.proto";
+
+// ==========================================
+// Resource Definitions
+// ==========================================
+
+// Gpu represents a single GPU resource in Kubernetes.
+message Gpu {
+  // name is the unique logical identifier of the GPU resource.
+  string name = 1;
```
lalitadithya, line 36: Should we instead use an id here? GPU/devices in general don't have names, so maybe using an id here would be more appropriate.

api/proto/device/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,171 @@
...
+message GpuList {
+  // items is the list of GPU resources.
+  repeated Gpu items = 1;
+}
+
+// GpuSpec describes the identity and desired attributes of the GPU.
+message GpuSpec {
+  // id is the physical hardware UUID of this GPU.
+  // Format: 'GPU-<hex-string>' (e.g., 'GPU-a1b2c3d4-e5f6-a7b8-c9d0-e1f2a3b4c5d6').
+  string id = 1;
```
lalitadithya: Ah, I see you have the ID here, I think we should put this outside of the spec since this will be unique identifier for the GPUs.

api/proto/device/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,171 @@
...
+
+  // node_name is the name of the host node where this GPU is physically attached.
+  string node_name = 2;
```
lalitadithya: Since the API is going to be node local, I'm not sure we want to include the node name. Should we instead start with an empty spec and we can add properties to it as we make progress?

api/proto/device/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// GpuStatus defines the observed state of the GPU.
+message GpuStatus {
+  // conditions represent the observations of this GPU's current state.
+  repeated Condition conditions = 1;
+
+  // recommended_actions is a list of suggested remediation steps to resolve negative conditions.
```
lalitadithya: Should we just provide a single recommendation instead of list? For example, the recommendation can be "highest priority" recommendation. What I mean is that if we have an error that needs a gpu reset and another that needs a reboot, we can just provide the recommendation for reboot

third_party/google/api/annotations.proto
lalitadithya, line 1: I'm not quite sure why we need this, can you please help me understand the use case for this proto?

third_party/google/api/field_behavior.proto
lalitadithya, line 1: I'm not quite sure why we need this, can you please help me understand the use case for this proto?

third_party/google/api/http.proto
lalitadithya, line 1: I'm not quite sure why we need this, can you please help me understand the use case for this proto?

api/proto/device/v1alpha1/gpu.proto
lalitadithya: Hmm. I think it depends. If we send the complete status on changes then we might not need the resource version. This might be a simpler option to implement as well

api/proto/device/v1alpha1/gpu.proto
lalitadithya: Yeah, I feel it depends on the nature of the watch as well. If the watch sends only partial updates, then yeah, it will need to be ordered, and we need to maintain versions/token to allow transitions. But if the watch always returns the complete state, we may not need this.

api/proto/device/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,171 @@
...
+
+import "google/api/annotations.proto";
+import "google/api/http.proto";
+import "google/protobuf/timestamp.proto";
+
+// ==========================================
+// Resource Definitions
+// ==========================================
+
+// Gpu represents a single GPU resource in Kubernetes.
```
ArangoGutierrez: should we remove the Kubernetes wording? this API could be also used by SLURM and other similar tools

api/proto/device/v1alpha1/gpu.proto
ArangoGutierrez: Agree look at go-nvml implementation of a Device

https://github.com/NVIDIA/go-nvml/blob/main/pkg/nvml/device.go#L64

```go
// GpuInstanceInfo includes an interface type for Device instead of nvmlDevice
type GpuInstanceInfo struct {
	Device    Device
	Id        uint32
	ProfileId uint32
	Placement GpuInstancePlacement
}
```

api/proto/device/v1alpha1/gpu.proto
ArangoGutierrez: agree, see my comment above

api/proto/device/v1alpha1/gpu.proto
ArangoGutierrez: where will you get the data for this? This variable is not relevant.

api/proto/device/v1alpha1/gpu.proto
ArangoGutierrez: agree, a "list" of recommendations can be confusing, most on what the reconciler using the API should do

api/proto/device/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,171 @@
...
+
+  // message is a human-readable message indicating details about the transition.
+  string message = 5;
+}
+
+// ==========================================
+// Service Definition
+// ==========================================
+
+// GpuService provides an API for managing and observing GPU resources.
```
ArangoGutierrez: I do think we should stick to watching for now.  If we want to propose this API to be a management API it should be after we move it to it's own repo, not while it's living on the NVsentinel repo.

Also I think we should further discuss the extra layers and steps required for this API to be more than just for health monitoring.

third_party/google/api/field_behavior.proto
pteranodan: That was leftover from an earlier iteration which included a 'field_mask' field. Missed cleaning this up when it was removed. Will remove.

third_party/google/api/http.proto
pteranodan: We need to download `http.proto` because it is a transitive dependency. The protobuf definition imports `annotations.proto` to define REST endpoints, and `annotations.proto` internally imports `http.proto` to define the underlying rules.

lalitadithya: I don't quite follow; I thought we were only going to expose this over UDS right? So, I don't think we will need to have `http.proto`. If we were to expose this as HTTP service over TLS then I think we'd need that right?

api/proto/device/v1alpha1/gpu.proto
pteranodan: We can go without a resource version for v1, but it leaves a gap. If a GPU changes between the List and Watch calls, the client misses it. They stay with old data until the next update.

api/proto/device/v1alpha1/gpu.proto
pteranodan: We can go without a resource version for v1, but it leaves a gap. If a GPU changes between the List and Watch calls, the client misses it. They stay with old data until the next update.

api/proto/device/v1alpha1/gpu.proto
pteranodan: I am fine removing `node_name` for now and assuming node-local context to keep v1 minimal.

However, it is worth noting that any tool consuming this API (like a CLI or a multi-node dashboard) will immediately lose context of *where* this GPU exists. Including `node_name` makes the resource self-describing, which simplifies upstream consumption, but we can always add it back in a future iteration.

api/proto/device/v1alpha1/gpu.proto
pteranodan: I agree, a list introduces unnecessary ambiguity.

api/proto/device/v1alpha1/gpu.proto
pteranodan: That may be a valid use case, though the primary design intent remains Kubernetes. I updated the comment to reference the Kubernetes Resource Model structure which accurately describes the schema without strictly limiting the environment.

api/proto/device/v1alpha1/gpu.proto
pteranodan: I updated the comment to "GpuService provides an API for observing GPU resources."

third_party/google/api/annotations.proto
pteranodan: The protobuf definition imports `annotations.proto` to define REST endpoints.

api/proto/device/v1alpha1/gpu.proto
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// GpuList is a collection of GPU resources.
+message GpuList {
+  // items is the list of GPU resources.
+  repeated Gpu items = 1;
+}
+
+// GpuSpec describes the identity and desired attributes of the GPU.
+message GpuSpec {
```
ArangoGutierrez, line 53: Should we take some inspiration from https://github.com/NVIDIA/go-nvml/blob/main/pkg/nvml/device.go#L64

```go
// GpuInstanceInfo includes an interface type for Device instead of nvmlDevice
type GpuInstanceInfo struct {
	Device    Device
	Id        uint32
	ProfileId uint32
	Placement GpuInstancePlacement
}
```

third_party/google/api/http.proto
pteranodan: Context: gRPC-gateway was included to satisfy a requirement for a dual HTTP+JSON and gRPC interface.

I initially used custom annotations out of familiarity, but I see now that we could use [generate_unbound_methods](https://github.com/grpc-ecosystem/grpc-gateway?tab=readme-ov-file#1-using-the-default-mapping) to avoid the third-party proto dependencies entirely.

That said, I believe it is best to simply remove the HTTP+JSON interface from v1 to keep the initial scope minimal. Since adding it later (via annotations or default mapping) is a non-breaking change for gRPC clients, we can easily re-introduce the REST interface in a future iteration when needed.

Thoughts, @lalitadithya?

third_party/google/api/annotations.proto
pteranodan: Combining threads. See https://github.com/NVIDIA/NVSentinel/pull/388#discussion_r2581504625

api/proto/device/v1alpha1/gpu.proto
pteranodan: I prefer to keep the v1 Spec minimal and focused on identity (UUID), adding those specific attributes in future iterations as we expand support.

third_party/google/api/http.proto
lalitadithya: I agree. Let's start simple with just gRPC.

api/proto/device/v1alpha1/gpu.proto
pteranodan: FYI, I also renamed `spec.id` to `spec.uuid` to make it self-documenting and avoid confusion.

third_party/google/api/http.proto
pteranodan: Done.

api/proto/device/v1alpha1/gpu.proto
pteranodan: @ArangoGutierrez see https://github.com/NVIDIA/NVSentinel/pull/388#discussion_r2581709163 regarding adding additional attributes in v1.

api/proto/device/v1alpha1/gpu.proto
pteranodan: I renamed `spec.id` to `spec.uuid` to make it self-documenting and avoid confusion.

api/proto/device/v1alpha1/gpu.proto
ArangoGutierrez: cool

api/proto/device/v1alpha1/gpu.proto
pteranodan: I see the appeal of flattening the structure, but I lean towards keeping `name` (logical) separate from `spec.uuid` (physical) to maintain flexibility. It gives us the freedom to handle awkward or evolving identifier formats without breaking the API contract, preventing us from painting ourselves into a corner later.

api/proto/device/v1alpha1/gpu.proto
pteranodan: @lalitadithya see https://github.com/NVIDIA/NVSentinel/pull/388#discussion_r2581794235 regarding the placement of the identifier. 

_Resolving this to consolidate the conversation._

ArangoGutierrez (approved):
Ship fast, ship often

Let' get this in and start iterating on it, but merging is good for the overall project 

LGTM

lalitadithya (approved):

PR #447, opened 2025-11-25
docs: add design doc for audit logging (Issue #274)

### LINKED ISSUE #274: [Feature]: Add support for audit logs
### Prerequisites

### Feature Summary

Currently to find what operations NVSentinel has taken, we need to comb though the logs to understand what changes have been made. This is easy on smaller cluster with verbose logging, but not easy on larger cluster. We need a append only ledger of all the write changes made by NVSentinel

### Problem/Use Case

Make it easier to understand all the changes that are happening in the cluster.

### Proposed Solution

Write a commons package for audit logging that pushes the audit logs to MongoDB as a append only collection and it also add Kubernetes events where applicable. 

### Component

Core services

rupalis-nv:
## Summary

Design doc for audit logging feature

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added an ADR describing per-pod, file-based JSON audit logging for write operations with automatic capture of outgoing REST/gRPC requests, environment-driven configuration (enablement, log location, optional request-body capture, rotation), guidance for deployment/rollout, and recommendations for log collection/shipping and operational considerations with examples and integration patterns.

### DIFF
--- a/docs/designs/016-audit-logging.md
+++ b/docs/designs/016-audit-logging.md
@@ -1,0 +1,698 @@
+# ADR-016: Audit Logging for NVSentinel Write Operations
+
+## Problem
+
+NVSentinel performs critical write operations on Kubernetes and cloud infrastructure. Operators need an immutable record of every operation for:
+
+- **Accountability**: Prove what NVSentinel did or did not do on a cluster
+- **Compliance**: Audit trails for infrastructure changes
+- **Debugging**: Trace exact API calls when investigating issues
+
+### Audit Logs vs Metrics
+
+Metrics show aggregated behavior (`nodes_cordoned_total: 42`). Audit logs capture every individual API call with timestamp, URL, method, and optionally request body - details that would cause cardinality explosion in metrics.
+
+## Solution
+
+Implement a file-based audit logging system that:
+
+1. **Automatically captures write operations** via HTTP RoundTripper wrapping client-go and CSP SDKs
+2. **Writes structured JSON logs to a file** (`/var/log/nvsentinel/{POD_NAME}-audit.log`)
+3. **Leverages existing log collection infrastructure** (Fluentd/Fluent Bit/Alloy) for shipping logs to any destination
+4. **Uses existing logging patterns** - new `commons/pkg/auditlogger/` package with direct JSON encoding
+
+This approach follows the same pattern used by Kubernetes itself (`/var/log/kubernetes/audit.log`) and other operators.
+
+### Architecture Overview
+
+```mermaid
+graph TB
+    subgraph NVSentinel_Components["NVSentinel Components"]
+        FR[Fault Remediation]
+        FQ[Fault Quarantine]
+        ND[Node Drainer]
+        JN[Janitor]
+        LB[Labeler]
+        PC[Platform Connector]
+    end
+    
+    subgraph Auditing_Layer["Auditing Layer"]
+        RT[Auditing RoundTripper]
+        AL["Audit Logger (commons/pkg/auditlogger)"]
+    end
+    
+    subgraph File_Output["File Output"]
+        FILE["audit.log"]
+    end
+    
+    subgraph Operator_Infrastructure["Operator Infrastructure"]
+        FB["Fluentd / Fluent Bit / Alloy"]
+        DEST["Any Destination (Elasticsearch, S3, Splunk, Loki)"]
+    end
+    
+    FR --> RT
+    FQ --> RT
+    ND --> RT
+    JN --> RT
+    LB --> RT
+    PC --> RT
+    
+    RT --> AL
+    AL --> FILE
+    FILE --> FB
+    FB --> DEST
+```
+
+### Key Design Principles
+
+1. **Automatic Capture**: HTTP RoundTripper intercepts all write operations
+2. **File-Based Output**: Structured JSON to file, operators ship via their log collection infrastructure
+3. **Minimal Dependencies**: No external services required for audit logging itself
+
+## Audited Components and Operations
+
+We audit calls made to **external systems** only (Kubernetes API and CSPs), not internal datastore calls.
+
+| Component | Operation | Description | HTTP Method |
+|-----------|-----------|-------------|-------------|
+| **fault-quarantine** | Cordon/Taint nodes | Quarantine unhealthy nodes | PUT |
+| **node-drainer** | Evict pods, Force delete | Evict/delete pods from unhealthy nodes | POST, DELETE |
+| **fault-remediation** | Create maintenance CR, Update annotations | Create RebootNode/TerminateNode CR | POST, PATCH |
+| **janitor** | Update CR, CSP calls | Execute reboot/terminate via cloud provider API | PUT, POST |
+| **labeler** | Update node labels | Update node labels (DCGM version, driver, kata) | PUT |
+| **platform-connector** | Create/Update node events and conditions | K8s Events and Node Conditions | POST, PATCH |
+
+### Not Audited (Read-Only Components)
+
+- Health Monitors (GPU, Syslog, Kubernetes Object Monitor) - only detect and report
+- MongoDB Change Streams - read operations only
+
+## Data Model
+
+### AuditEntry Structure
+
+```go
+type AuditEntry struct {
+    Timestamp   string `json:"timestamp"`
+    Component   string `json:"component"`
+    Method      string `json:"method"`
+    URL         string `json:"url"`
+    RequestBody string `json:"requestBody,omitempty"`
+}
+```
+
+The `RequestBody` field is optional and only included when `AUDIT_LOG_REQUEST_BODY=true`. With method, URL, and request body, the specific operation can be inferred.
+
+### Audit Log Example
+
+**Default:**
+```json
+{"timestamp":"2025-01-15T10:30:45Z","component":"fault-quarantine","method":"PUT","url":"/api/v1/nodes/gpu-node-01"}
+```
+
+**With body enabled (`AUDIT_LOG_REQUEST_BODY=true`):**
+
+**fault-quarantine operations:**
+```json
+// Cordon node
+{"timestamp":"2025-01-15T10:30:45Z","component":"fault-quarantine","method":"PUT","url":"/api/v1/nodes/gpu-node-01","requestBody":"{\"spec\":{\"unschedulable\":true}}"}
+
+// Uncordon node
+{"timestamp":"2025-01-15T10:30:46Z","component":"fault-quarantine","method":"PUT","url":"/api/v1/nodes/gpu-node-01","requestBody":"{\"spec\":{\"unschedulable\":false}}"}
+
+// Add taint
+{"timestamp":"2025-01-15T10:30:47Z","component":"fault-quarantine","method":"PUT","url":"/api/v1/nodes/gpu-node-01","requestBody":"{\"spec\":{\"taints\":[{\"key\":\"nvsentinel.nvidia.com/unhealthy\",\"effect\":\"NoSchedule\"}]}}"}
+
+// Update circuit breaker ConfigMap
+{"timestamp":"2025-01-15T10:30:48Z","component":"fault-quarantine","method":"PUT","url":"/api/v1/namespaces/nvsentinel/configmaps/circuit-breaker","requestBody":"{\"data\":{\"status\":\"open\"}}"}
+```
+
+**node-drainer operations:**
+```json
+// Evict pod
+{"timestamp":"2025-01-15T10:31:00Z","component":"node-drainer","method":"POST","url":"/api/v1/namespaces/default/pods/my-pod/eviction","requestBody":"{\"apiVersion\":\"policy/v1\",\"kind\":\"Eviction\",\"metadata\":{\"name\":\"my-pod\"}}"}
+
+// Force delete pod (after timeout)
+{"timestamp":"2025-01-15T10:31:01Z","component":"node-drainer","method":"DELETE","url":"/api/v1/namespaces/default/pods/my-pod?gracePeriodSeconds=0"}
+
+// Update node label (drain status)
+{"timestamp":"2025-01-15T10:31:02Z","component":"node-drainer","method":"PATCH","url":"/api/v1/nodes/gpu-node-01","requestBody":"{\"metadata\":{\"labels\":{\"nvsentinel.nvidia.com/state\":\"draining\"}}}"}
+```
+
+**fault-remediation operations:**
+```json
+// Create RebootNode CR
+{"timestamp":"2025-01-15T10:32:00Z","component":"fault-remediation","method":"POST","url":"/apis/janitor.dgx.nvidia.com/v1alpha1/rebootnodes","requestBody":"{\"metadata\":{\"name\":\"maintenance-gpu-node-01-abc123\"},\"spec\":{\"nodeName\":\"gpu-node-01\"}}"}
+
+// Create TerminateNode CR
+{"timestamp":"2025-01-15T10:32:01Z","component":"fault-remediation","method":"POST","url":"/apis/janitor.dgx.nvidia.com/v1alpha1/terminatenodes","requestBody":"{\"metadata\":{\"name\":\"maintenance-gpu-node-01-def456\"},\"spec\":{\"nodeName\":\"gpu-node-01\"}}"}
+
+// Update node annotation (remediation state)
+{"timestamp":"2025-01-15T10:32:02Z","component":"fault-remediation","method":"PATCH","url":"/api/v1/nodes/gpu-node-01","requestBody":"{\"metadata\":{\"annotations\":{\"nvsentinel.nvidia.com/remediation-state\":\"{...}\"}}}"}
+```
+
+**janitor operations:**
+```json
+// Update CR with finalizer
+{"timestamp":"2025-01-15T10:33:00Z","component":"janitor","method":"PUT","url":"/apis/janitor.dgx.nvidia.com/v1alpha1/rebootnodes/maintenance-gpu-node-01-abc123"}
+
+// Update CR status
+{"timestamp":"2025-01-15T10:33:01Z","component":"janitor","method":"PUT","url":"/apis/janitor.dgx.nvidia.com/v1alpha1/rebootnodes/maintenance-gpu-node-01-abc123/status"}
+
+// CSP: AWS Reboot
+{"timestamp":"2025-01-15T10:33:02Z","component":"janitor","method":"POST","url":"https://ec2.amazonaws.com/?Action=RebootInstances&InstanceId=i-1234567890abcdef0"}
+
+// CSP: GCP Stop+Start (reboot)
+{"timestamp":"2025-01-15T10:33:03Z","component":"janitor","method":"POST","url":"https://compute.googleapis.com/compute/v1/projects/my-project/zones/us-central1-a/instances/gpu-node-01/stop"}
+
+// CSP: Azure Restart
+{"timestamp":"2025-01-15T10:33:04Z","component":"janitor","method":"POST","url":"https://management.azure.com/subscriptions/.../virtualMachines/gpu-node-01/restart"}
+
+// CSP: OCI Instance Action
+{"timestamp":"2025-01-15T10:33:05Z","component":"janitor","method":"POST","url":"https://iaas.us-phoenix-1.oraclecloud.com/.../instances/ocid1.../action?action=RESET"}
+```
+
+**labeler operations:**
+```json
+// Update DCGM version label
+{"timestamp":"2025-01-15T10:34:00Z","component":"labeler","method":"PUT","url":"/api/v1/nodes/gpu-node-01","requestBody":"{\"metadata\":{\"labels\":{\"nvidia.com/dcgm.version\":\"4.x\"}}}"}
+
+// Update driver installed label
+{"timestamp":"2025-01-15T10:34:01Z","component":"labeler","method":"PUT","url":"/api/v1/nodes/gpu-node-01","requestBody":"{\"metadata\":{\"labels\":{\"nvidia.com/gpu.driver.installed\":\"true\"}}}"}
+
+// Update kata enabled label
+{"timestamp":"2025-01-15T10:34:02Z","component":"labeler","method":"PUT","url":"/api/v1/nodes/gpu-node-01","requestBody":"{\"metadata\":{\"labels\":{\"nvidia.com/kata.enabled\":\"true\"}}}"}
+```
+
+**platform-connector operations:**
+```json
+// Create node event
+{"timestamp":"2025-01-15T10:35:00Z","component":"platform-connector","method":"POST","url":"/api/v1/namespaces/default/events","requestBody":"{\"involvedObject\":{\"kind\":\"Node\",\"name\":\"gpu-node-01\"},\"reason\":\"GPUError\",\"message\":\"...\"}"}
+
+// Update node conditions
+{"timestamp":"2025-01-15T10:35:01Z","component":"platform-connector","method":"PATCH","url":"/api/v1/nodes/gpu-node-01/status"}
+```
+
+With method + URL + body, the specific operation can be inferred.
+
+## Implementation
+
+### Package Structure
+
+```text
+commons/
+└── pkg/
+    ├── logger/           # Existing - writes to stderr
+    │   └── logger.go
+    │
+    └── auditlogger/      # NEW - writes to file
+        ├── auditlogger.go    # Main audit logger (direct JSON encoding)
+        └── roundtripper.go   # HTTP RoundTripper for automatic capture
+```
+
+### Audit Logger Package
+
+```go
+package auditlogger
+
+import (
+    "encoding/json"
+    "fmt"
+    "os"
+    "path/filepath"
+    "strconv"
+    "time"
+
+    "gopkg.in/natefinch/lumberjack.v2"
+)
+
+const (
+    DefaultAuditLogDir     = "/var/log/nvsentinel"
+    EnvAuditLogBasePath    = "AUDIT_LOG_BASE_PATH"
+    EnvAuditEnabled        = "AUDIT_ENABLED"
+    EnvAuditLogRequestBody = "AUDIT_LOG_REQUEST_BODY"
+    EnvPodName             = "POD_NAME"
+    EnvAuditLogMaxSize     = "AUDIT_LOG_MAX_SIZE_MB"
+    EnvAuditLogMaxBackups  = "AUDIT_LOG_MAX_BACKUPS"
+    EnvAuditLogMaxAge      = "AUDIT_LOG_MAX_AGE_DAYS"
+    EnvAuditLogCompress    = "AUDIT_LOG_COMPRESS"
+)
+
+type AuditEntry struct {
+    Timestamp   string `json:"timestamp"`
+    Component   string `json:"component"`
+    Method      string `json:"method"`
+    URL         string `json:"url"`
+    RequestBody string `json:"requestBody,omitempty"`
+}
+
+var (
+    auditLogger *lumberjack.Logger
+    component   string
+)
+
+func getEnvInt(key string, defaultVal int) int {
+    if v := os.Getenv(key); v != "" {
+        if i, err := strconv.Atoi(v); err == nil {
+            return i
+        }
+    }
+    return defaultVal
+}
+
+func getEnvBool(key string, defaultVal bool) bool {
+    if v := os.Getenv(key); v != "" {
+        if b, err := strconv.ParseBool(v); err == nil {
+            return b
+        }
+    }
+    return defaultVal
+}
+
+func InitAuditLogger(comp string) error {
+    if comp == "" {
+        return fmt.Errorf("component name cannot be empty")
+    }
+    
+    if os.Getenv(EnvAuditEnabled) == "false" {
+        return nil
+    }
+    
+    component = comp
+    
+    baseDir := os.Getenv(EnvAuditLogBasePath)
+    if baseDir == "" {
+        baseDir = DefaultAuditLogDir
+    }
+    
+    podName := os.Getenv(EnvPodName)
+    if podName == "" {
+        podName = component
+    }
+    if podName == "" {
+        return fmt.Errorf("unable to determine pod name: POD_NAME env var and component are both empty")
+    }
+    
+    logPath := filepath.Join(baseDir, podName+"-audit.log")
+    
+    if err := os.MkdirAll(baseDir, 0700); err != nil {
+        return err
+    }
+    
+    auditLogger = &lumberjack.Logger{
+        Filename:   logPath,
+        MaxSize:    getEnvInt(EnvAuditLogMaxSize, 100),
+        MaxBackups: getEnvInt(EnvAuditLogMaxBackups, 7),
+        MaxAge:     getEnvInt(EnvAuditLogMaxAge, 30),
+        Compress:   getEnvBool(EnvAuditLogCompress, true),
+    }
+    
+    return nil
+}
+
+func CloseAuditLogger() error {
+    if auditLogger != nil {
+        return auditLogger.Close()
+    }
+    return nil
+}
+
+func Log(method, url, body string) {
+    if auditLogger == nil {
+        return
+    }
+    
+    entry := AuditEntry{
+        Timestamp: time.Now().UTC().Format(time.RFC3339),
+        Component: component,
+        Method:    method,
+        URL:       url,
+    }
+    
+    if os.Getenv(EnvAuditLogRequestBody) == "true" && body != "" {
+        entry.RequestBody = body
+    }
+    
+    if err := json.NewEncoder(auditLogger).Encode(entry); err != nil {
+        fmt.Fprintf(os.Stderr, "audit: failed to write entry: %v\n", err)
+    }
+}
+```
+
+### Log Rotation
+
+The `InitAuditLogger` function uses [lumberjack](https://github.com/natefinch/lumberjack) for built-in log rotation with the following settings:
+
+**Output format:**
+```json
+{"timestamp":"2025-11-27T19:29:26Z","component":"fault-quarantine","method":"PUT","url":"/api/v1/nodes/gpu-node-01"}
+```
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| `MaxSize` | 100 MB | Max size before rotation |
+| `MaxBackups` | 7 | Number of old log files to keep |
+| `MaxAge` | 30 days | Max age of old log files |
+| `Compress` | true | Compress rotated files with gzip |
+
+### HTTP RoundTripper for Automatic Capture
+
+```go
+package auditlogger
+
+import (
+    "bytes"
+    "io"
+    "net/http"
+    "os"
+)
+
+type AuditingRoundTripper struct {
+    delegate http.RoundTripper
+}
+
+func NewAuditingRoundTripper(delegate http.RoundTripper) *AuditingRoundTripper {
+    return &AuditingRoundTripper{delegate: delegate}
+}
+
+func (rt *AuditingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+    if isWriteMethod(req.Method) {
+        body := ""
+        if os.Getenv(EnvAuditLogRequestBody) == "true" && req.Body != nil {
+            bodyBytes, _ := io.ReadAll(req.Body)
+            req.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+            body = string(bodyBytes)
+        }
+        Log(req.Method, req.URL.String(), body)
+    }
+    return rt.delegate.RoundTrip(req)
+}
+
+func isWriteMethod(method string) bool {
+    return method == "POST" || method == "PUT" || method == "PATCH" || method == "DELETE"
+}
+```
+
+**Limitation**: HTTP RoundTripper only works for REST/HTTP-based clients. For gRPC clients, use gRPC interceptors.
+
+### gRPC Interceptor
+
+> **Note**: This section is included as a reference for future implementation. Currently, all NVSentinel CSP clients use REST (e.g., `NewInstancesRESTClient()` for GCP), so the HTTP RoundTripper covers all use cases. When CSP modules move to gRPC clients (e.g., GCP protobuf clients), this interceptor pattern will be used.
+>
+> This is **not** for internal NVSentinel communication (e.g., health monitors → platform connector). We only audit external API calls to Kubernetes and CSPs.
+
+For gRPC clients, use interceptors to audit write operations:
+
+```go
+import (
+    "context"
+    "encoding/json"
+    "os"
+    "strings"
+
+    "google.golang.org/grpc"
+    "github.com/NVIDIA/NVSentinel/commons/pkg/auditlogger"
+)
+
+func AuditingUnaryInterceptor() grpc.UnaryClientInterceptor {
+    return func(ctx context.Context, method string, req, reply interface{},
+        cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+        
+        if isWriteRPC(method) {
+            body := ""
+            if os.Getenv(auditlogger.EnvAuditLogRequestBody) == "true" {
+                if b, err := json.Marshal(req); err == nil {
+                    body = string(b)
+                }
+            }
+            auditlogger.Log("gRPC", method, body)
+        }
+        
+        return invoker(ctx, method, req, reply, cc, opts...)
+    }
+}
+
+func isWriteRPC(method string) bool {
+    parts := strings.Split(method, "/")
+    rpcName := parts[len(parts)-1]
+    
+    writePrefixes := []string{"Create", "Update", "Delete", "Patch", "Put", "Replace", "Insert", "Remove"}
+    for _, prefix := range writePrefixes {
+        if strings.HasPrefix(rpcName, prefix) {
+            return true
+        }
+    }
+    return false
+}
+
+// Usage:
+conn, _ := grpc.Dial(target, grpc.WithUnaryInterceptor(AuditingUnaryInterceptor()))
+```
+
+### Integration Pattern
+
+#### Kubernetes Client Integration
+
+```go
+func main() {
+    if err := auditlogger.InitAuditLogger("fault-quarantine"); err != nil {
+        slog.Warn("Failed to initialize audit logger", "error", err)
+    }
+    defer auditlogger.CloseAuditLogger()
+    
+    restConfig, _ := rest.InClusterConfig()
+    restConfig.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+        return auditlogger.NewAuditingRoundTripper(rt)
+    })
+    
+    clientset, _ := kubernetes.NewForConfig(restConfig)
+    // ...
+}
+
+func (c *Client) QuarantineNode(nodeName string) error {
+    _, err := c.clientset.CoreV1().Nodes().Update(...)
+    return err
+    // Logged: {"timestamp":"...","component":"fault-quarantine","method":"PUT","url":"/api/v1/nodes/gpu-node-01"}
+}
+```
+
+#### CSP SDK Integration
+
+For CSP SDKs, wrap the existing transport to preserve SDK-configured settings.
+
+**Helper function:**
+
+```go
+func wrapTransport(client *http.Client) {
+    existing := client.Transport
+    if existing == nil {
+        existing = http.DefaultTransport
+    }
+    client.Transport = auditlogger.NewAuditingRoundTripper(existing)
+}
+```
+
+**AWS SDK:**
+
+```go
+func NewAWSEC2Client(ctx context.Context) (*ec2.Client, error) {
+    cfg, err := config.LoadDefaultConfig(ctx)
+    if err != nil {
+        return nil, err
+    }
+
+    httpClient := cfg.HTTPClient
+    if httpClient == nil {
+        httpClient = &http.Client{}
+    }
+    wrapTransport(httpClient)
+    cfg.HTTPClient = httpClient
+
+    return ec2.NewFromConfig(cfg), nil
+}
+```
+
+**GCP SDK:**
+
+```go
+func NewGCPComputeClient(ctx context.Context) (*compute.InstancesRESTClient, error) {
+    httpClient := &http.Client{}
+    wrapTransport(httpClient)
+
+    return compute.NewInstancesRESTClient(ctx, option.WithHTTPClient(httpClient))
+}
+```
+
+**Azure SDK:**
+
+```go
+func NewAzureVMClient(ctx context.Context, subscriptionID string) (*armcompute.VirtualMachinesClient, error) {
+    cred, err := azidentity.NewDefaultAzureCredential(nil)
+    if err != nil {
+        return nil, err
+    }
+
+    transport := auditlogger.NewAuditingRoundTripper(http.DefaultTransport)
+
+    return armcompute.NewVirtualMachinesClient(subscriptionID, cred, &arm.ClientOptions{
+        ClientOptions: azcore.ClientOptions{
+            Transport: &http.Client{Transport: transport},
+        },
+    })
+}
+```
+
+**OCI SDK:**
+
+```go
+func NewOCIComputeClient(ctx context.Context) (core.ComputeClient, error) {
+    configProvider, err := auth.OkeWorkloadIdentityConfigurationProvider()
+    if err != nil {
+        return core.ComputeClient{}, err
+    }
+
+    computeClient, err := core.NewComputeClientWithConfigurationProvider(configProvider)
+    if err != nil {
+        return core.ComputeClient{}, err
+    }
+
+    wrapTransport(computeClient.HTTPClient)
+    return computeClient, nil
+}
+```
+
+## Configuration
+
+### Environment Variables
+
+| Variable | Type | Default | Description |
+|----------|------|---------|-------------|
+| `AUDIT_ENABLED` | bool | `true` | Master switch for audit logging |
+| `AUDIT_LOG_REQUEST_BODY` | bool | `false` | Include request body in audit logs (for debugging operations) |
+| `AUDIT_LOG_BASE_PATH` | string | `/var/log/nvsentinel/` | Base directory for audit log files |
+| `POD_NAME` | string | - | Pod name for unique log file per pod (injected from downward API) |
+| `AUDIT_LOG_MAX_SIZE_MB` | int | `100` | Max size in MB before rotation |
+| `AUDIT_LOG_MAX_BACKUPS` | int | `7` | Number of old log files to keep |
+| `AUDIT_LOG_MAX_AGE_DAYS` | int | `30` | Max age in days of old log files |
+| `AUDIT_LOG_COMPRESS` | bool | `true` | Compress rotated files with gzip |
+
+### Deployment Configuration
+
+```yaml
+spec:
+  containers:
+    - name: fault-quarantine
+      env:
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+      volumeMounts:
+        - name: audit-logs
+          mountPath: /var/log/nvsentinel
+  volumes:
+    - name: audit-logs
+      hostPath:
+        path: /var/log/nvsentinel
+        type: DirectoryOrCreate
+```
+
+## Log Collection (Operator Responsibility)
+
+### Example: Fluent Bit Configuration
+
+```ini
+[INPUT]
+    Name              tail
+    Path              /var/log/nvsentinel/*-audit.log
+    Tag               nvsentinel.audit
+    Parser            json
+
+[OUTPUT]
+    Name              es
+    Match             nvsentinel.audit
+    Host              elasticsearch.logging.svc
+    Index             nvsentinel-audit
+```
+
+### Example: Fluentd Configuration
+
+```text
+<source>
+  @type tail
+  path /var/log/nvsentinel/*-audit.log
+  pos_file /var/log/fluentd/nvsentinel-audit.pos
+  tag nvsentinel.audit
+  <parse>
+    @type json
+  </parse>
+</source>
+
+<match nvsentinel.audit>
+  @type elasticsearch
+  host elasticsearch.logging.svc
+  index_name nvsentinel-audit
+</match>
+```
+
+### Documentation
+
+- Audit logs: `/var/log/nvsentinel/{POD_NAME}-audit.log`
+- Format: JSON, one entry per line
+- Rotation: 100MB max, 7 backups, 30 days retention (via lumberjack)
+- Log shipping: Operator responsibility (Fluentd, Fluent Bit, Alloy, etc.)
+
+## Rationale
+
+- **Follows industry patterns**: Kubernetes writes audit logs to files; operators do the same
+- **Automatic capture**: HTTP RoundTripper eliminates risk of missing audit logs
+- **Minimal log format**: Just timestamp, component, method, URL
+- **Operator flexibility**: Operators choose their own log destination
+- **No infrastructure dependencies**: NVSentinel doesn't manage log shipping
+
+## Consequences
+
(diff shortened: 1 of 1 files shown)


docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,456 @@
...
+| `AUDIT_ENABLED` | bool | `true` | Master switch for audit logging |
+| `AUDIT_MONGODB_ENABLED` | bool | `true` | Enable MongoDB audit persistence |
+| `AUDIT_K8S_EVENTS_ENABLED` | bool | `true` | Enable Kubernetes events creation |
+| `AUDIT_MONGODB_URI` | string | (fallback: `MONGODB_URI`) | MongoDB connection URI |
+| `AUDIT_MONGODB_DATABASE` | string | (fallback: `MONGODB_DATABASE_NAME`) | MongoDB database name |
+| `AUDIT_MONGODB_COLLECTION` | string | `auditlogs` | MongoDB collection name |
+| `MONGODB_CLIENT_CERT_PATH` | string | - | Path to MongoDB client certificate |
+| `MONGODB_CLIENT_KEY_PATH` | string | - | Path to MongoDB client key |
+| `MONGODB_CA_CERT_PATH` | string | - | Path to MongoDB CA certificate |
+
```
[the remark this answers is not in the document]
rupalis-nv: addressed

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,479 @@
...
+4. **Visibility**: Cluster operators have no single view of all changes NVSentinel has made
+
+NVSentinel performs critical write operations on Kubernetes resources (tainting nodes, evicting pods, creating maintenance CRs, rebooting/terminating nodes). These operations need to be tracked in an append-only ledger for compliance, debugging, and operational visibility.
+
+## Solution
+
+Implement a commons audit logging package that:
+
+1. **Pushes audit logs to MongoDB** as an append-only collection
+2. **Creates Kubernetes Events** for visibility within the cluster
```
lalitadithya: It should be generic. The system should be able to push events to any sink, mongodb is just one possible sink, kube is another and so on.

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,479 @@
...
+            "operation", auditLog.Operation,
+            "resource", auditLog.ResourceName)
+    }
+}
+```
+
+#### 2. Graceful Error Handling
+
+```go
+func (l *Logger) processAuditLog(auditLog *AuditLog) {
```
lalitadithya: Can we follow OCP here? If we have a third sink say google cloud logging, then we have to keep adding to this function.

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,479 @@
...
+
+// In operation method
+func (c *Client) QuarantineNode(nodeName string) error {
+    // Perform the actual operation
+    err := c.k8sClient.CoreV1().Nodes().Patch(...)
+    if err != nil {
+        return err
+    }
+    
+    // Log audit (fire and forget)
```
lalitadithya: This is very invasive in the sense that we have to add this change everywhere we have client-go code. If we add new calls, but we forge to add this then audit logs will be missed. Can we instead think of a middleware or some kind of custom http client that will automatically audit log all the changes being made?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,479 @@
...
+3. **Queryability**: Logs are not structured for efficient querying by time, node, operation type, etc.
+4. **Visibility**: Cluster operators have no single view of all changes NVSentinel has made
+
+NVSentinel performs critical write operations on Kubernetes resources (tainting nodes, evicting pods, creating maintenance CRs, rebooting/terminating nodes). These operations need to be tracked in an append-only ledger for compliance, debugging, and operational visibility.
+
+## Solution
+
+Implement a commons audit logging package that:
+
+1. **Pushes audit logs to MongoDB** as an append-only collection
```
lalitadithya: We should say audit to the datastore. All the mongodb code has been move to the datastore behind an interface so we should update the document so mention that datastore is a possible option.

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,479 @@
...
+        BG->>K8s: CreateEvent()
+        K8s-->>BG: Success/Error
+    end
+    
+    Note over BG: Log error if failed,<br/>don't propagate to main
+```
+
+### Key Implementation Details
+
+#### 1. Non-Blocking Log Method
```
lalitadithya: How are we handling pod restarts? Say an API call was made and then pod restarted, would the audit log be dropped in that case?

rupalis-nv: can we make sure we are writing synchronously after API call and before returing?

lalitadithya: Possibly, but the audit will happen in a different thread right? are you suggesting that the audit thread will flush all the events and then exit?

rupalis-nv: We have 2 options

- Write audit log synchronously AFTER the API call completes, BEFORE returning.
- Audit logs queued to async buffer. On SIGTERM, flush all pending logs before exit.

Suggest me which can be good design decision we will follow that

rupalis-nv: With option 1 - we will be having some latency due to sync call
With option2 - we might loose logs during OOM kill

docs/designs/013-audit-logging.md
rupalis-nv: Please take a look

docs/designs/013-audit-logging.md
rupalis-nv: Please take a look

docs/designs/013-audit-logging.md
rupalis-nv: Please take a look

docs/designs/013-audit-logging.md
rupalis-nv: Please take a look

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,527 @@
...
+    style RT fill:#fa8c16,stroke:#d46b08,stroke-width:2px,color:#fff
+    style SI fill:#52c41a,stroke:#389e0d,stroke-width:2px,color:#fff
+    style DS fill:#13c2c2,stroke:#08979c,stroke-width:2px,color:#fff
+    style K8S fill:#1890ff,stroke:#096dd9,stroke-width:2px,color:#fff
+```
+
+### Key Design Principles
+
+1. **Open/Closed Principle**: New sinks can be added without modifying existing code
+2. **Automatic Capture**: HTTP round-tripper intercepts all write operations - no manual audit calls needed
```
lalitadithya: Can we confirm that this is supported by all the clients we are using -- AWS, GCP, Azure, OCI and Kube?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,527 @@
...
+    ResourceName string `json:"resourceName" bson:"resourceName"`
+
+    // ResourceNamespace is the namespace of the resource (empty for cluster-scoped)
+    ResourceNamespace string `json:"resourceNamespace,omitempty" bson:"resourceNamespace,omitempty"`
+
+    // NodeName is the name of the node if the operation is node-related
+    NodeName string `json:"nodeName,omitempty" bson:"nodeName,omitempty"`
+
+    // Description is a human-readable description of what was done
+    Description string `json:"description" bson:"description"`
```
lalitadithya: If we are using a HTTP round trip based solution how would get the description?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,527 @@
...
+
+    // Details contains additional operation-specific details
+    Details map[string]interface{} `json:"details,omitempty" bson:"details,omitempty"`
```
lalitadithya: If we are using a HTTP round trip based, what would this contain?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,527 @@
...
+
+    // HealthEventID is the ID of the health event that triggered this operation
+    HealthEventID string `json:"healthEventId,omitempty" bson:"healthEventId,omitempty"`
```
lalitadithya: Can you please add some details on this will be added? Specifically in the case of Janitor

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,527 @@
...
+type OperationType string
+const (
+    OperationCreate OperationType = "CREATE"
+    OperationUpdate OperationType = "UPDATE"
+    OperationPatch  OperationType = "PATCH"
+    OperationDelete OperationType = "DELETE"
+)
+
+type ResourceType string
+const (
```
lalitadithya: If we are using const, what about cloud resources?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,527 @@
...
+        rt.auditRequest(req, resp)
+    }
+    
+    return resp, err
+}
+
+func (rt *AuditingRoundTripper) auditRequest(req *http.Request, resp *http.Response) {
+    // Parse K8s API path to extract resource info
+    // e.g., /api/v1/nodes/gpu-node-01 -> ResourceType: Node, ResourceName: gpu-node-01
+    resourceInfo := parseK8sAPIPath(req.URL.Path)
```
lalitadithya: How is this parsing done? What if the API structure changes? Also how would this work for cloud resources? More broadly, what is the use case for parsing?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,527 @@
...
+}
+
+func isWriteMethod(method string) bool {
+    return method == "POST" || method == "PUT" || method == "PATCH" || method == "DELETE"
+}
+```
+
+### Integration Pattern
+
+Components wrap their Kubernetes client with the auditing transport:
```
lalitadithya: Can you also add how this will be done for the CSP clients?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,527 @@
...
+    // Just perform the operation - auditing happens automatically
+    _, err := c.clientset.CoreV1().Nodes().Patch(
+        context.Background(),
+        nodeName,
+        types.StrategicMergePatchType,
+        patchBytes,
+        metav1.PatchOptions{},
+    )
+    return err
+    // ✅ Audit log is automatically captured by the RoundTripper!
```
lalitadithya: No Emojis please

docs/designs/013-audit-logging.md
lalitadithya: Option 1 will be a no go, what if the sink is not available at that time, it will just end up blocking all the operations. Can we see how other solutions like fluent handle this?

rupalis-nv: Fluent- they use a disk-based buffer
1. Write audit log to local disk first (fast, ~1ms)
2. Background thread reads from disk and sends to sinks
3. On restart, resume from last position - no data loss
We need PersistentVolume in this case

docs/designs/013-audit-logging.md
rupalis-nv: yes its supported by all

docs/designs/013-audit-logging.md
lalitadithya: Hmm, can't we use https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/?

docs/designs/013-audit-logging.md
lalitadithya: okay, can you add pusedo code like you have done for kube?

docs/designs/013-audit-logging.md
rupalis-nv: With RoundTripper, we only have access to:
HTTP Method: PATCH
URL Path: /api/v1/nodes/gpu-node-01
Request Body: {"spec":{"taints":[...]}}

Can we use both roundtripper and log?RoundTripper captures basics, optional Log() for enrichment?

docs/designs/013-audit-logging.md
rupalis-nv: `{
  "details": {
    "httpMethod": "PATCH",
    "apiPath": "/api/v1/nodes/gpu-node-01",
    "statusCode": 200,
    "requestBody": {
      "spec": {
        "taints": [
          {"key": "nvsentinel.nvidia.com/quarantine", "effect": "NoSchedule"}
        ]
      }
    }
  }
}`

lalitadithya: That is in the case of patch right? what about in cases where we are issuing an update/create/delete? Would that contain the entire body, wouldn't that make it too large?

docs/designs/013-audit-logging.md
lalitadithya: If we use two methods, then we will have to do some coordination between them which will be complex. Plus, with log we have to remember to keep all the call points updated. Maybe the information provided by the round tripper is sufficient. Do we have use cases where we need more information?

docs/designs/013-audit-logging.md
rupalis-nv: Create and update might be bit large
Like if we have Reboot node CR-

```
{
  "details": {
    "crKind": "RebootNode",
    "crName": "maintenance-gpu-node-01-abc123",
    "targetNode": "gpu-node-01",
    "force": false,
    "healthEventId": "abc123def456"
  }
}
```

rupalis-nv: this will be for Create CR case

docs/designs/013-audit-logging.md
rupalis-nv: yes we can use ephemeral volumes
But -Pod reschedule - emptyDir lost
Node failure - emptyDir lost
Container OOM/crash - resume from WAL
Container restart  - data preserved
SIGTERM (drain, scale)  - flush before exit

docs/designs/013-audit-logging.md
rupalis-nv: https://github.com/NVIDIA/NVSentinel/commit/2ba4b3d970dab60bcf961df089bccb7fa4c1a86e#diff-7a1565c42cb5d244557efde900f3b14d6059c7bec6c693be2934e7f6e2ac65a7R4-R354

docs/designs/013-audit-logging.md
rupalis-nv: Yes for most of the cases info provided by round tripper is sufficient
What did NVSentinel change?- Roundtripper can cover
When was this node tainted? - Roundtripper can cover
What taint/label was applied?-- Roundtripper can cover
Show all pod evictions- Roundtripper can cover
Correlate with health event -  HealthEventID - Can extract from CR name- Parse HealthEventID from CR name/annotations in request body
GPU XID error 79- Round tripper wont cover- its stored in health events

docs/designs/013-audit-logging.md
rupalis-nv: https://github.com/NVIDIA/NVSentinel/pull/447/files#diff-7a1565c42cb5d244557efde900f3b14d6059c7bec6c693be2934e7f6e2ac65a7R457

docs/designs/013-audit-logging.md
rupalis-nv: Can we use generic cloud resource type
`const (    
                  ResourceCloudInstance  ResourceType = "CloudInstance"
 )`
Then use Details field for CSP-specific info:
`{  "resourceType": "CloudInstance",  "details": {    "csp": "aws",    "instanceId": "i-1234567890abcdef0"  }}
`

docs/designs/013-audit-logging.md
rupalis-nv: For different cloud providers all are having different structure, we will be having URL parsing complexity

Now this is drawback with RoundTripper
Should we reconsider?
Or we can look for any workaround?

docs/designs/013-audit-logging.md
rupalis-nv: added

docs/designs/013-audit-logging.md
rupalis-nv: addressed

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,613 @@
...
+- **Storage growth**: Audit logs will accumulate over time
+- **Best-effort**: Audit logs may be dropped if buffer is full or sinks are unavailable
+- **RoundTripper complexity**: Parsing K8s API paths requires careful handling
+
+### Mitigations
+
+- **TTL indexes**: Configure datastore TTL for automatic old log cleanup
+- **Buffer sizing**: Configurable buffer size (default 1000) handles burst scenarios
+- **Graceful degradation**: If one sink fails, others continue working
+- **Path parsing**: Well-tested regex patterns for K8s API path extraction
```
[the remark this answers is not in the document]
rupalis-nv: addressed

lalitadithya:
I have been thinking about this a lot over the course of the day and this looks like we are essentially duplicating the functionality of what we have in fluentbit/fluentd/alloy. So, I'm thinking we start very very simple and start just write the audit logs in a structured format to a file in a host path say /var/log/nvsentinel/audit.log. We can then recommend customers make use of fluentbit/fluentd/alloy to push the logs to any sink they choose. This I think will make our implementation also simple as we just need to use some existing logger to write a file. This will also allow us to audit the platform connector. 

What do you think?

docs/designs/013-audit-logging.md
lalitadithya: Why do we want to parse the URL? Can't we just log the URL as is?

docs/designs/013-audit-logging.md
lalitadithya: Right, but this will require a lot of parsing, do we really need that?

docs/designs/013-audit-logging.md
lalitadithya: Yeah, it will be large, do we need all the info? Maybe just the HTTP method is enough? We could consider logging the body based on a flag possibly

docs/designs/013-audit-logging.md
lalitadithya: Right please see -- https://github.com/NVIDIA/NVSentinel/pull/447#issuecomment-3582523317. I'm thinking something much more simpler, just host path

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,735 @@
...
+Janitor reconciles CR               HealthEventID in name!
+       ↓
+RoundTripper intercepts API calls
+       ↓
+Extracts HealthEventID from CR name
+```
+
+#### API Call Captured by RoundTripper:
+```
+URL: /apis/janitor.dgx.nvidia.com/v1alpha1/rebootnodes/maintenance-gpu-node-01-abc123def456
```
lalitadithya: But how is this matched with the cloud API call? This API call is made by the fault remediation module right?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,735 @@
...
+}
+
+func isWriteMethod(method string) bool {
+    return method == "POST" || method == "PUT" || method == "PATCH" || method == "DELETE"
+}
+
+// extractHealthEventID extracts the health event ID from maintenance CR names
+// CR name format: maintenance-<nodeName>-<healthEventID>
+// Example: maintenance-gpu-node-01-abc123def456
+func extractHealthEventID(resourceInfo ResourceInfo) string {
```
lalitadithya: I'm not a fan of this. We can change the CR format in future and this will break logging. We should try and see if there is a more structured approach if we really need this. I suspect we won't

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,735 @@
...
+```
+
+#### GCP SDK
+```go
+func NewGCPClient(ctx context.Context, logger *auditlog.Logger) (*compute.InstancesClient, error) {
+    auditingHTTPClient := &http.Client{
+        Transport: auditlog.NewAuditingRoundTripper(http.DefaultTransport, logger, "janitor"),
+    }
+
+    client, err := compute.NewInstancesRESTClient(ctx,
```
lalitadithya: Doesn't GCP SDK also offer protobuf based clients? What would we do in those cases?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,735 @@
...
+    if err != nil {
+        return nil, err
+    }
+
+    computeClient, err := core.NewComputeClientWithConfigurationProvider(configProvider)
+    if err != nil {
+        return nil, err
+    }
+    
+    computeClient.HTTPClient = auditingHTTPClient  // Inject auditing client
```
lalitadithya: Since we are replacing the client after making the `core.NewComputeClientWithConfigurationProvider`, how do we make sure we are not wiping the HTTP configuration that was set by `core.NewComputeClientWithConfigurationProvider`?

docs/designs/013-audit-logging.md
rupalis-nv: In our simplified design, we're NOT logging the request body - just the metadata:
HTTP method (PATCH, CREATE, DELETE)
Request path (/api/v1/nodes/gpu-node-01)
Resource type/name (parsed from path)
Response status code
This keeps audit logs small and fast. If we need body logging for debugging, we can add a flag (AUDIT_LOG_BODY=true) as a future enhancement.

will this be oky?

lalitadithya: Yes, that is fine. 

I'd say we don't do any parsing not even the "Resource type/name (parsed from path)"

lalitadithya: How would you get the response status code? Do we need that, can we just audit that the calls have been made?

docs/designs/013-audit-logging.md
rupalis-nv: Correct, these are two separate API calls:
1. fault-remediation → K8s API → Creates RebootNode CR
2. janitor → CSP API → Actually reboots the node
Both are logged separately:
- Log 1: POST /apis/janitor.../rebootnodes/maintenance-gpu-node-01-abc123
- Log 2: POST https://ec2.amazonaws.com/?Action=RebootInstances...
Correlation is via:
- Same node name in both logs
- HealthEventID (abc123) present in both
- Timestamps are close together
Since we're logging the raw URL (no parsing), both K8s and CSP calls are captured the same way by RoundTripper.

lalitadithya: yes, since they are two separate API calls, they will be logged in different lines and we can't automatically correlate them right?

docs/designs/013-audit-logging.md
rupalis-nv: Makes sense
Simplest audit log which i can think of-
 >{
>   "timestamp": "2025-01-15T10:30:45Z",
>   "component": "fault-quarantine",
>   "method": "PATCH",
>   "url": "/api/v1/nodes/gpu-node-01"
> }
>

lalitadithya: Yeah, I think that is good enough to start with, we can iterate on it and add more info if required

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,458 @@
...
+4. **Visibility**: Cluster operators have no single view of all changes NVSentinel has made
+
+NVSentinel performs critical write operations on Kubernetes resources (tainting nodes, evicting pods, creating maintenance CRs, rebooting/terminating nodes). These operations need to be tracked in an append-only ledger for compliance, debugging, and operational visibility.
+
+## Solution
+
+Implement a file-based audit logging system that:
+
+1. **Automatically captures write operations** via HTTP RoundTripper wrapping client-go and CSP SDKs
+2. **Writes structured JSON logs to a file** (`/var/log/nvsentinel/audit.log`)
```
lalitadithya: Since it is possible for multiple components to run on the same node, maybe let's do `/var/log/nvsentinel/{POD_NAME}-audit.log

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,458 @@
...
+        return err
+    }
+    
+    // Open file for append
+    file, err := os.OpenFile(logPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+    if err != nil {
+        return err
+    }
+    
+    // Create JSON handler writing to file
```
lalitadithya: Can you add some details here on what the default log rotation strategy would be and how users can configure it?

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,458 @@
...
+        - name: audit-logs
+          mountPath: /var/log/nvsentinel
+  volumes:
+    - name: audit-logs
+      hostPath:
+        path: /var/log/nvsentinel
+        type: DirectoryOrCreate
+```
+
+## Log Collection (Customer Responsibility)
```
lalitadithya:
```suggestion
## Log Collection (Operator Responsibility)
```

docs/designs/013-audit-logging.md
rupalis-nv: Sure

docs/designs/016-audit-logging.md
rupalis-nv: Sure

docs/designs/013-audit-logging.md
rupalis-nv: Done

docs/designs/013-audit-logging.md
```diff
@@ -0,0 +1,458 @@
...
+    if err != nil {
+        return err
+    }
+    
+    // Create JSON handler writing to file
+    auditLogger = slog.New(slog.NewJSONHandler(file, &slog.HandlerOptions{}))
+    auditLogger = auditLogger.With("component", component)
+    
+    return nil
+}
```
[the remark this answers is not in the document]
rupalis-nv: Addressed

docs/designs/013-audit-logging.md
rupalis-nv: This is simplified now

docs/designs/016-audit-logging.md
lalitadithya, line 1: Can you make this number 16 -- `docs/designs/016-audit-logging.md`

docs/designs/013-audit-logging.md
rupalis-nv: NVSentinel uses NewInstancesRESTClient() - so RoundTripper works. If we switch to gRPC in future, we'd need direct calls.

lalitadithya: Hmmm... Can we document that? Doesn't the GRPC client have similar functionality?

docs/designs/016-audit-logging.md
rupalis-nv: done

docs/designs/013-audit-logging.md
rupalis-nv: gRPC clients have similar functionality called interceptors
I've documented the limitation in the design doc. If we switch to gRPC clients in the future, we have two options:
1.. gRPC Interceptor - similar to RoundTripper, intercepts all gRPC calls
2. Direct auditlogger.Log() calls - simpler but manual

rupalis-nv: Limitation: HTTP RoundTripper only works for REST/HTTP-based clients. If CSP SDKs use gRPC/protobuf clients (e.g., compute.NewInstancesClient() instead of compute.NewInstancesRESTClient()), RoundTripper cannot intercept those calls. Currently, NVSentinel uses REST clients for all CSP operations, so this is not an issue. If gRPC clients are used in the future, those components would need to call auditlogger.Log() directly.

THis i have documented

lalitadithya: Can we try gRPC Interceptor and ensure it works?

lalitadithya: >  If gRPC clients are used in the future, those components would need to call auditlogger.Log() directly.

If gRPC Interceptor works then we can just use that right?

docs/designs/013-audit-logging.md
rupalis-nv: yes 
Automatic corelation needs URL parsing i think so

docs/designs/013-audit-logging.md
rupalis-nv: Yes
I have tried to document it PTAL and suggest if you think of better approach

docs/designs/013-audit-logging.md
lalitadithya: okay, then I don't let's do that now. Let's keep it simple for now

rupalis-nv: Should i document coorelation also?
Do we really need coorelation?

lalitadithya: we don't. I meant we don't want co-relation now

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,532 @@
...
+## Audited Components and Operations
+
+| Component | Operation | Description | HTTP Method |
+|-----------|-----------|-------------|-------------|
+| **fault-quarantine** | Cordon/Taint nodes | Quarantine unhealthy nodes | PUT |
+| **node-drainer** | Evict pods | Evict pods from unhealthy nodes | POST |
+| **fault-remediation** | Create RebootNode/TerminateNode CR | Create maintenance custom resources | POST |
+| **janitor** | Reboot/Terminate via CSP | Execute reboot/terminate via cloud provider API | POST |
+| **labeler** | Update node labels | Update node labels (DCGM version, driver, kata) | PUT |
+| **platform-connector** | Ingest health events | Write health events to datastore | N/A (direct call) |
```
lalitadithya: I don't think we have to audit the calls to the data store since that is a component we own. Can we instead audit just calls made to external system -- kube and CSPs?

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,532 @@
...
+    auditLogger.Info("audit",
+        "method", method,
+        "url", url,
+    )
+}
+```
+
+### Log Rotation
+
+NVSentinel does not implement built-in log rotation. Operators should configure log rotation using standard Linux tools.
```
lalitadithya: Can we instead use a logger that provides log rotation? AFAIK a combination of lumberjack and logrus should do the trick. I may be wrong, but there should also be a way in which we can plug slog into that, please do a small POC and confirm though.

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,532 @@
...
+        component: component,
+    }
+}
+
+func (rt *AuditingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+    // Execute the actual request
+    resp, err := rt.delegate.RoundTrip(req)
+    
+    // Only audit write operations (POST, PUT, PATCH, DELETE)
+    if isWriteMethod(req.Method) {
```
lalitadithya, line 379: Shouldn't this be done before the round trip?

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,532 @@
...
+  @type elasticsearch
+  host elasticsearch.logging.svc
+  index_name nvsentinel-audit
+</match>
+```
+
+### Documentation Note
+
+Document in NVSentinel docs:
+- Audit logs are written to `/var/log/nvsentinel/audit.log`
```
lalitadithya: This has changed now right?

docs/designs/016-audit-logging.md
rupalis-nv: yes lumberjack and slog works together
Slog- for logging 
lumberjack -for rotation

docs/designs/016-audit-logging.md
rupalis-nv: for platform connector we just audit node node events and node conditions?

lalitadithya: yes, I mean if we are using the same round tripper it should just work right?

docs/designs/016-audit-logging.md
rupalis-nv: yes

docs/designs/016-audit-logging.md
rupalis-nv: yes yes

docs/designs/016-audit-logging.md
rupalis-nv: yes updated

docs/designs/013-audit-logging.md
rupalis-nv: Updated all CSP SDK examples to wrap the existing transport instead of replacing it.

docs/designs/013-audit-logging.md
rupalis-nv: okay

docs/designs/013-audit-logging.md
rupalis-nv: New simple implementation has no resourcetype at all

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,553 @@
...
+
+// No audit code needed in operation methods - automatic!
+func (c *Client) QuarantineNode(nodeName string) error {
+    _, err := c.clientset.CoreV1().Nodes().Patch(...)
+    return err
+    // Automatically logged: {"method":"PATCH","url":"/api/v1/nodes/gpu-node-01"}
+}
+```
+
+#### CSP SDK Integration
```
lalitadithya, line 479: I like the way we had it before where we had specific examples per CSP, can we bring that back? It was specific and when we implement, we will know what needs to be done.

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,557 @@
...
+func (rt *AuditingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+    // Audit write operations BEFORE the request (capture intent)
+    if isWriteMethod(req.Method) {
+        // Log method and URL as-is - no parsing!
+        Log(req.Method, req.URL.String())
+    }
+    
+    // Execute the actual request
+    return rt.delegate.RoundTrip(req)
+}
```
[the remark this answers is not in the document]
rupalis-nv: addressed

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,553 @@
...
+    MaxBackups: 7,    // number of backups
+    MaxAge:     30,   // days
+    Compress:   true, // compress rotated files
+}
+
+// Use lumberjack as slog output
+auditLogger = slog.New(slog.NewJSONHandler(lumberjackLogger, nil))
+```
+
+**Output format (verified via POC):**
```
lalitadithya:
```suggestion
**Output format:**
```

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,553 @@
...
+```json
+{"time":"2025-11-27T19:29:26Z","level":"INFO","msg":"audit","component":"fault-quarantine","method":"PATCH","url":"/api/v1/nodes/gpu-node-01"}
```
lalitadithya: Can we drop the `msg` and `level`? Does the library have that option?

lalitadithya: The struct above says timestamp, but here we have time. Can we pick on option and stick to it for consistency?

docs/designs/013-audit-logging.md
rupalis-nv: In current design -We just log the URL as-is. No parsing

docs/designs/013-audit-logging.md
lalitadithya: It seems like we removed the CSP specific examples, can we bring that back?

docs/designs/016-audit-logging.md
rupalis-nv: Sure i picked timestamp

rupalis-nv: yes

docs/designs/016-audit-logging.md
rupalis-nv: updated

docs/designs/016-audit-logging.md
rupalis-nv: Updated

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,533 @@
...
+    }
+    
+    file, err := os.OpenFile(logPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+    if err != nil {
+        return err
+    }
+    auditWriter = file
+    
+    return nil
+}
```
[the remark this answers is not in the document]
rupalis-nv: addressed

docs/designs/013-audit-logging.md
rupalis-nv: PTAL

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,525 @@
...
+    
+    entry := AuditEntry{
+        Timestamp: time.Now().UTC().Format(time.RFC3339),
+        Component: component,
+        Method:    method,
+        URL:       url,
+    }
+    
+    auditMu.Lock()
+    defer auditMu.Unlock()
```
lalitadithya: Why do we need this lock? Won't the logger take care of it?

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,525 @@
...
+    json.NewEncoder(auditWriter).Encode(entry)
```
lalitadithya: This encodes the entry, but where are we calling the audit logger?

rupalis-nv: addreseed

docs/designs/016-audit-logging.md
rupalis-nv: yes correct

docs/designs/016-audit-logging.md
lalitadithya: I'm not following, can you please walk me though the change? I'm probably missing something

rupalis-nv: So the current code is - 

var (
    auditLogger *lumberjack.Logger   // The file logger
    component   string
)

func Log(method, url string) {
    if auditLogger == nil {
        return
    }
    
    entry := AuditEntry{
        Timestamp: time.Now().UTC().Format(time.RFC3339),
        Component: component,
        Method:    method,
        URL:       url,
    }
    
    json.NewEncoder(auditLogger).Encode(entry)
}

How it works:
auditLogger is *lumberjack.Logger (initialized in InitAuditLogger)
lumberjack.Logger implements io.Writer interface
json.NewEncoder(auditLogger) creates a JSON encoder that writes to the file
.Encode(entry) serializes AuditEntry to JSON and writes it to the file
Output in /var/log/nvsentinel/{POD_NAME}-audit.log:

> {"timestamp":"2025-01-15T10:30:45Z","component":"fault-quarantine","method":"PATCH","url":"/api/v1/nodes/gpu-node-01"}

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,520 @@
...
+    Component string `json:"component"`
+    Method    string `json:"method"`
+    URL       string `json:"url"`
+}
+```
+
+### Audit Log Example
+
+```json
+{"timestamp":"2025-01-15T10:30:45Z","component":"fault-quarantine","method":"PATCH","url":"/api/v1/nodes/gpu-node-01"}
```
XRFXLP: How do we know if the node is cordoned or tainted or unquarantined or labeled?

rupalis-nv: For getting specific operation- cordoned or tainted or unquarantined or labeled we need to parse request body 
but for now lalit said to only use URL  and dont complicate 
Do you think its necessary to get the action?

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,520 @@
...
+    return func(ctx context.Context, method string, req, reply interface{},
+        cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+        auditlogger.Log("gRPC", method)
+        return invoker(ctx, method, req, reply, cc, opts...)
+    }
+}
+
+// Usage:
+conn, _ := grpc.Dial(target, grpc.WithUnaryInterceptor(AuditingUnaryInterceptor()))
+```
```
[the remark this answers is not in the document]
rupalis-nv: addressed

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,520 @@
...
+    entry := AuditEntry{
+        Timestamp: time.Now().UTC().Format(time.RFC3339),
+        Component: component,
+        Method:    method,
+        URL:       url,
+    }
+    
+    json.NewEncoder(auditLogger).Encode(entry)
+}
+```
```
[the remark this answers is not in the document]
rupalis-nv: addresed

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,520 @@
...
+        podName := os.Getenv(EnvPodName)
+        if podName == "" {
+            podName = component
+        }
+        logPath = filepath.Join(DefaultAuditLogDir, podName+"-audit.log")
+    }
+    
+    if err := os.MkdirAll(filepath.Dir(logPath), 0755); err != nil {
+        return err
+    }
```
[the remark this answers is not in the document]
rupalis-nv: addressed

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,520 @@
...
+    component = comp
+    
+    logPath := os.Getenv(EnvAuditLogPath)
+    if logPath == "" {
+        podName := os.Getenv(EnvPodName)
+        if podName == "" {
+            podName = component
+        }
+        logPath = filepath.Join(DefaultAuditLogDir, podName+"-audit.log")
+    }
```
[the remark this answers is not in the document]
rupalis-nv: addressed

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,520 @@
...
+    auditLogger *lumberjack.Logger
+    component   string
+)
+
+func InitAuditLogger(comp string) error {
+    if os.Getenv(EnvAuditEnabled) == "false" {
+        return nil
+    }
+    
+    component = comp
```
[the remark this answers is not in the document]
rupalis-nv: addressed

docs/designs/016-audit-logging.md
XRFXLP: Is that a follow up task? Do we have GH issue for that?

rupalis-nv: No but its not requirement

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,560 @@
+# ADR-016: Audit Logging for NVSentinel Write Operations
+
+## Problem
+
+Currently, to understand what operations NVSentinel has taken, users must comb through component logs scattered across multiple pods. This approach has several limitations:
```
XRFXLP: Why metrics cannot solve this problem?

docs/designs/016-audit-logging.md
XRFXLP: Looks like the solution proposed is not solving the problem that is described in the problem statement above.

Problem starts with:
> Currently, to understand what operations NVSentinel has taken

But we are not tracking this the `operations` itself

docs/designs/016-audit-logging.md
rupalis-nv: Metrics show 100 nodes tainted today but not which nodes, when, in what order
Metrics might only tell number of operations ... but with audit log we can check numbers with node name and timestamp

XRFXLP: We can check node name in metric along with timestamp, no?

We've this metric: https://github.com/NVIDIA/NVSentinel/blob/main/fault-quarantine/pkg/metrics/metrics.go#L74 that also contains the node name information.

rupalis-nv: Hmm...
But one question do operators who are using nvsentinel can use this metrics?

XRFXLP: anyone using nvsentinel can use any nvsentinel metrics.

Why do you think they would not be able to use though?

rupalis-nv: but those grafana dashboards are for internal na?

docs/designs/016-audit-logging.md
rupalis-nv: For operations we considered as API calls

docs/designs/016-audit-logging.md
lalitadithya: I see, okay.

docs/designs/016-audit-logging.md
rupalis-nv: The current design captures "what API call was made" not "what operation was performed".

XRFXLP: Do we debug what HTTP call (POST | PATCH | DELETE | CREATE) was made by nvsentinel? Or do we concern ourself with higher level operational stuffs like cordon/drain/taint/reboot?

docs/designs/016-audit-logging.md
XRFXLP: Why would that matter? Even in this solution, we're taking liberty of assuming any arbitrary audit log sink. Grafana is just another sink, right?

lalitadithya: Metrics work to understand how the system is behaving, but they are not an audit for every operation that is performed by NVSentinel. You are right in that metrics work for cases like node cordoned, uncordoned etc, but we want to know every change that is being by NVSentinel, not just cordons. Changes could be anything from applying labels, annotations to making CSP API calls. 

The point of an audit is to be able to defend the operations that NVSentinel did or did not do on a cluster and also prove that NVSentinel did or did not do something that is/was claimed. We could have metric as an audit log, but they will result in cardinality explosion when we start to add things like URL, HTTP method and the request body. We could aggregate the events into a metrics, but they would mean we will end up losing some information.

docs/designs/016-audit-logging.md
rupalis-nv: @lalitadithya should we reconsider on our approach on HTTP only tracking?
Or is it out of requirement as if now?

docs/designs/016-audit-logging.md
XRFXLP: Can we make this motivation clear in the doc? because the current problem statement looks like a way to uninvent metrics.

docs/designs/016-audit-logging.md
lalitadithya: The goal here is to debug every HTTP/gRPC call that was made by NVSentinel, what operation was performed can be inferred from the audit log. It was mentioned in one of the comments, but maybe not added to the document -- we have a flag that can be turned on to log the request body along with the HTTP metadata. When you have a combination of request method, url and request body you should be able to fairly easily say what operation was done. 

At the end of the day, auditing should be done at the lowest level possible so that every change that is made is captured and not just the operational ones.

docs/designs/016-audit-logging.md
lalitadithya: okay, that is fair. @rupalis-nv , can you please update the problem statement to reflect what we discussed in this thread?

docs/designs/016-audit-logging.md
XRFXLP: Okay.
Can we update the doc/struct/log examples?

rupalis-nv: Sure

docs/designs/016-audit-logging.md
rupalis-nv: PTAL, do Let me know if it needs more detail.

docs/designs/016-audit-logging.md
rupalis-nv: PTAL, I have added operation example from all modules ...do let me know if i should reduce some examples and only keep imp one

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,676 @@
...
+type AuditEntry struct {
+    Timestamp   string `json:"timestamp"`
+    Component   string `json:"component"`
+    Method      string `json:"method"`
+    URL         string `json:"url"`
+    RequestBody string `json:"requestBody,omitempty"`
+}
+```
+
+The `RequestBody` field is optional and only included when `AUDIT_LOG_BODY=true`. With method, URL, and request body, the specific operation can be inferred.
```
lalitadithya:
```suggestion
The `RequestBody` field is optional and only included when `AUDIT_LOG_REQUEST_BODY=true`. With method, URL, and request body, the specific operation can be inferred.
```

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,676 @@
...
+func isWriteMethod(method string) bool {
+    return method == "POST" || method == "PUT" || method == "PATCH" || method == "DELETE"
+}
+```
+
+**Limitation**: HTTP RoundTripper only works for REST/HTTP-based clients. For gRPC clients, use gRPC interceptors (see below).
+
+### gRPC Interceptor
+
+For gRPC clients, use interceptors to audit write operations:
```
XRFXLP: Do we use this for health monitors talking to platform connector?

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,676 @@
...
+        if strings.HasPrefix(rpcName, prefix) {
+            return true
+        }
+    }
+    return false
+}
+
+// Usage:
+conn, _ := grpc.Dial(target, grpc.WithUnaryInterceptor(AuditingUnaryInterceptor()))
+```
```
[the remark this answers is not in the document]
rupalis-nv: addressed

docs/designs/016-audit-logging.md
rupalis-nv: No, the gRPC interceptor is not for health monitors → platform connector.

docs/designs/016-audit-logging.md
rupalis-nv: addresed

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,669 @@
...
+
+    "gopkg.in/natefinch/lumberjack.v2"
+)
+
+const (
+    DefaultAuditLogDir  = "/var/log/nvsentinel"
+    EnvAuditLogPath     = "AUDIT_LOG_PATH"
+    EnvAuditEnabled     = "AUDIT_ENABLED"
+    EnvAuditLogRequestBody     = "AUDIT_LOG_REQUEST_BODY"
+    EnvPodName          = "POD_NAME"
```
lalitadithya: nit:

```suggestion
    DefaultAuditLogDir              = "/var/log/nvsentinel"
    EnvAuditLogPath                  = "AUDIT_LOG_PATH"
    EnvAuditEnabled                  = "AUDIT_ENABLED"
    EnvAuditLogRequestBody    = "AUDIT_LOG_REQUEST_BODY"
    EnvPodName                        = "POD_NAME"
```

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,669 @@
...
+    if err := os.MkdirAll(filepath.Dir(logPath), 0700); err != nil {
+        return err
+    }
+    
+    auditLogger = &lumberjack.Logger{
+        Filename:   logPath,
+        MaxSize:    100,
+        MaxBackups: 7,
+        MaxAge:     30,
+        Compress:   true,
```
lalitadithya: These would be exposed as flags so that the end user can configure them as required right? We will have defaults, but if they need changes they have the option to do so.

Would these be flags or envvars for the end user?

docs/designs/016-audit-logging.md
```diff
@@ -0,0 +1,669 @@
...
+
+## Configuration
+
+### Environment Variables
+
+| Variable | Type | Default | Description |
+|----------|------|---------|-------------|
+| `AUDIT_ENABLED` | bool | `true` | Master switch for audit logging |
+| `AUDIT_LOG_REQUEST_BODY` | bool | `false` | Include request body in audit logs (for debugging operations) |
+| `AUDIT_LOG_PATH` | string | `/var/log/nvsentinel/{POD_NAME}-audit.log` | Path to audit log file (overrides default) |
```
lalitadithya: Can we instead say `AUDIT_LOG_BASE_PATH` and only accept `false`?

docs/designs/016-audit-logging.md
rupalis-nv: Better if we keep them as env vars

docs/designs/016-audit-logging.md
rupalis-nv: yes we can

rupalis-nv: but this is path na? how come it take `false`

lalitadithya: copy paste error, I meant `/var/log/nvsentinel/`

docs/designs/016-audit-logging.md
XRFXLP: What it is for then?

docs/designs/016-audit-logging.md
rupalis-nv: u mean 

> | `AUDIT_LOG_BASE_PATH` | string | `/var/log/nvsentinel/` | Base directory for audit log files |

docs/designs/016-audit-logging.md
rupalis-nv: The gRPC interceptor is for CSP SDKs that use gRPC instead of REST.
For example, GCP offers both REST and protobuf (gRPC) clients. Currently, NVSentinel uses NewInstancesRESTClient() (REST), but if we switch to gRPC clients in the future, we'd use this interceptor.

docs/designs/016-audit-logging.md
lalitadithya: yes

docs/designs/016-audit-logging.md
XRFXLP: that means we are not going to implement this currently?

rupalis-nv: hmm ,,yes

XRFXLP: can we clarify this in docs or remove it altogether?

rupalis-nv: should i remove this detailed explanation and just document this-
The gRPC interceptor is for CSP SDKs that use gRPC instead of REST.
For example, GCP offers both REST and protobuf (gRPC) clients. Currently, NVSentinel uses NewInstancesRESTClient() (REST), but if we switch to gRPC clients in the future, we'd use this interceptor.

docs/designs/016-audit-logging.md
rupalis-nv: okiesss done

docs/designs/016-audit-logging.md
rupalis-nv: PTAL

lalitadithya: let's not remove it. Let's clarify it in the doc so that when we do implement it, we have a reference point. We can say that this method will not be used for the health monitors, but we will use this in future when module X/Y/Z moves to using gRPC. This section is intended as a POC that the audit logging works for gRPC as well. Something to that effect

rupalis-nv: Hmm... but its not coming up as result of this design doc 
I have made this comment
**Note on gRPC**: HTTP RoundTripper only works for REST/HTTP-based clients. For CSP SDKs that use gRPC instead of REST, gRPC interceptors would be needed. For example, GCP offers both REST and protobuf (gRPC) clients. Currently, NVSentinel uses `NewInstancesRESTClient()` (REST), so RoundTripper covers all current use cases. If we switch to gRPC clients in the future, we'd implement gRPC interceptors.


If you suggest i can bring back and add this comment?

rupalis-nv: PTAL

lalitadithya: I see the comment you have added in your latest commit. There might have been a race when I added the GH comment, and you pushed a commit. You can ignore https://github.com/NVIDIA/NVSentinel/pull/447#discussion_r2570866519

rupalis-nv: hmm...can you check latest file? if you are okay with it?

lalitadithya (approved):

PR #449, opened 2025-11-25
use ctrl-runtime for fault-remediation

### LINKED ISSUE #359: [Feature]: Use Common Controller/Event Watcher Framework across controllers
### Prerequisites

### Feature Summary

Currently, most of the actors on the health events like fault-quarantine, fault-remediation, node-drainer etc all have similar EventWatcher needs. Each has their own slightly different implementation. We should solidified to a single shared EventWatcher/Reconciler/Controller thing that each of the modules can reuse. 

### Problem/Use Case

Useful for maintainability and readability.

### Proposed Solution

Common shared eventwatcher package. It looks like one already exists in store-client but we just need to move to using it. So maybe this work is in progress already? But creating this issue for tracking.

### Component

Multiple Components

ivelichkovich:
## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [X] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [X] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Optional controller-runtime management mode with manager integration and leader-election support
  * Controller-runtime-aware metrics and an additional health port when enabled

* **Configuration**
  * New ctrlRuntimeEnabled flag (default: false), healthPort (default: 9440), and leader-election flags; dev tilt values enable ctrl-runtime
  * RBAC expanded to allow leases and event creation/patching

* **Bug Fixes**
  * Fixed typo in a test name

* **Chores**
  * Expanded indirect dependencies across multiple modules

* **Tests**
  * Tests and e2e scaffolding updated to run with manager-enabled flow and new reconciler wiring


fault-remediation/main.go
```diff
@@ -87,14 +101,17 @@ func run() error {
...
 		return nil
 	})
-
-	g.Go(func() error {
-		return components.Reconciler.Start(gCtx)
-	})
-
-	return g.Wait()
+	slog.Info("Starting controller runtime controller")
+	//TODO: make sure informers setup correctly
```
ivelichkovich: Planning to update the k8s clients to use the built in ctrl runtime clients from mgr

fault-remediation/pkg/initializer/init.go
```diff
@@ -92,6 +109,67 @@ func InitializeAll(ctx context.Context, params InitializationParams) (*Component
...
+
+	watcherInstance, err := watcher.CreateChangeStreamWatcher(ctx, ds, watcherConfig)
+	if err != nil {
+		return nil, fmt.Errorf("error initializing change stream watcher: %w", err)
+	}
+
+	// Get the HealthEventStore for document operations
+	healthEventStore := ds.HealthEventStore()
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
```
ivelichkovich: I'll add a flag for leader election and make sure we use this built in metrics server replacing the manually setup. Later once we have leader election we should probably default this component to running with 3 replicas with leader election enabled.

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -1,42 +1,32 @@
-// Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ivelichkovich, line 1: I'll fix license and imports

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -64,96 +46,75 @@ type HealthEventData struct {
...
 	}
-
-	return nil
 }
 
-// processEvent handles a single event from the watcher
-func (r *Reconciler) processEvent(ctx context.Context, eventWithToken datastore.EventWithToken,
-	watcherInstance datastore.ChangeStreamWatcher, healthEventStore datastore.HealthEventStore) {
+func (r *FaultRemediationReconciler) Reconcile(ctx context.Context, event *datastore.EventWithToken) (ctrl.Result, error) {
+	//TODO: may not need these metrics anymore if we have some defaults from ctrl runtime
```
ivelichkovich: we'll probably want to leave metrics as is on first iteration for backwards compat then we can see what if any we can deprecate for the built in ctrl-runtime metrics

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -226,7 +187,8 @@ func (r *Reconciler) performRemediation(ctx context.Context, healthEventWithStat
 	success := false
 	crName := ""
 
-	for i := 1; i <= r.Config.UpdateMaxRetries; i++ {
```
ivelichkovich, line 230: So I'd like to remove the config/args for retries and retry delay here and just return an error. If an error is returned from Reconcile it automatically handles retries with backoff.

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -373,7 +335,7 @@ func (r *Reconciler) updateNodeRemediatedStatus(ctx context.Context, healthEvent
 	}
 
 	// Use the healthEventStore to update the status with retries
-	for i := 1; i <= r.Config.UpdateMaxRetries; i++ {
+	for i := 1; i <= r.config.UpdateMaxRetries; i++ {
```
ivelichkovich, line 388: same here, just return error and let ctrl-runtime handle retries

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -502,3 +464,71 @@ func (r *Reconciler) parseHealthEvent(eventWithToken datastore.EventWithToken,
...
+
+	return reconciler.
+		Named("fault-remediation-controller").
+		WatchesRawSource(
+			src,
+		).
+		Complete(r)
+}
+
+// TODO: make this part of the datastore interface?
```
ivelichkovich: maybe not part of interface but at least should be reusable so each controller doesn't have to re-implement this. 

I'll leave a lot of stuff in this MR potentially less generic than it could be and then if we like this pattern and want to apply it to other controllers can focus and reusing code at that point

Makefile
```diff
@@ -819,7 +819,7 @@ e2e-test-ci: tilt-ci ## Run end-to-end test suite in CI mode
 # Tilt end-to-end test target
 .PHONY: e2e-test
 e2e-test:
-	$(MAKE) -C dev tilt-up
+	#$(MAKE) -C dev tilt-up
```
lalitadithya: friendly reminder to uncomment this before the final commit

fault-remediation/main.go
```diff
@@ -48,64 +74,138 @@ func main() {
...
-		}
+	//TODO: add actual readiness check for connection to datasource and caches synced
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		slog.Error("Unable to set up ready check", "error", err)
+		return err
+	}
 
-		return nil
-	})
+	err = components.FaultRemediationReconciler.SetupWithManager(mgr)
```
lalitadithya: Since the changes is very nicely structured, do you think it would be possible to add this behind a feature flag? We can enable the flag in CI, allow it run for some time and then we can graduate it to default? What do you think?

fault-remediation/pkg/initializer/init.go
```diff
@@ -64,6 +66,7 @@ func InitializeAll(ctx context.Context, params InitializationParams) (*Component
 		slog.Info("Log collector enabled")
 	}
 
+	//TODO: replace this with manager client
```
lalitadithya, line 70: AFAIK the manager client is cached, so we may want to think about the effects of the cache. Since we read the state of the remediation flow in our equivalence logic and we may end up operating on older information.

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -17,26 +17,30 @@ package reconciler
...
-	"sync"
 	"time"
 
 	"github.com/nvidia/nvsentinel/commons/pkg/eventutil"
 	"github.com/nvidia/nvsentinel/commons/pkg/statemanager"
 	"github.com/nvidia/nvsentinel/data-models/pkg/model"
 	"github.com/nvidia/nvsentinel/data-models/pkg/protos"
 	"github.com/nvidia/nvsentinel/fault-remediation/pkg/common"
-	"github.com/nvidia/nvsentinel/store-client/pkg/client"
+	nvclient "github.com/nvidia/nvsentinel/store-client/pkg/client"
```
lalitadithya: nit: 

```suggestion
	nvstoreclient "github.com/nvidia/nvsentinel/store-client/pkg/client"
```

fault-remediation/pkg/reconciler/reconciler.go
lalitadithya: Yes, 100% agreed.

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -64,96 +60,79 @@ type HealthEventData struct {
...
 		eventHandlingDuration.Observe(time.Since(start).Seconds())
 	}()
 
 	totalEventsReceived.Inc()
 
-	healthEventWithStatus, err := r.parseHealthEvent(eventWithToken, watcherInstance)
+	healthEventWithStatus, err := r.parseHealthEvent(*event, r.watcher)
 	if err != nil {
-		return
+		//TODO: error handling?
```
lalitadithya: This is the case when the health event is not understandable, so we should just log the event and drop it so that we don't run into a hot loop

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -64,96 +60,79 @@ type HealthEventData struct {
...
+		return ctrl.Result{}, nil
 	}
 
 	// Safety checks for nil pointers
 	if healthEventWithStatus.HealthEvent == nil {
+		//TODO: error handling?
```
lalitadithya: This was a defensive check; this ideally should not happen. So, in this case too we want to drop the event with a warning log

fault-remediation/pkg/reconciler/reconciler.go
lalitadithya: yes, agreed! Let's do that, it will simplify the logic certainly

fault-remediation/pkg/reconciler/reconciler.go
lalitadithya: yes, let's do it.

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -502,3 +489,77 @@ func (r *Reconciler) parseHealthEvent(eventWithToken datastore.EventWithToken,
...
+// SetupWithManager wires the reconciler + mongo stream
+func (r *FaultRemediationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	ctx := context.Background()
+
+	r.watcher.Start(ctx)
+
+	reconciler := builder.TypedControllerManagedBy[*datastore.EventWithToken](mgr)
+	typedCh := AdaptEvents(ctx, r.watcher.Events())
+
+	//TODO: can we handle mark processsed ctrl runtime native?
```
lalitadithya: I have been thinking about this as well. I'm not sure if controller runtime has hooks that can allow us to do that. I will also do some digging to see if we can do something like that. It would be really cool if we can do that.

fault-remediation/main.go
ivelichkovich: Yeah that sounds good

fault-remediation/pkg/initializer/init.go
ivelichkovich: Sure I'll just leave that out for now to reduce scope of this change and try to get multiple smaller PRs in, instead of a single large one

fault-remediation/pkg/reconciler/reconciler.go
ivelichkovich: Yeah will see if I can get this in a follow up as well

fault-remediation/main.go
ivelichkovich: It makes it a little messy with some of the error handling examples we've discussed on this MR. Some of the controller logic will actually become different when using ctrl-runtime vs the current patterns. For example, returning an error in `performRemediation` instead of manually doing retries. It's a little tricky to make both co-exist cleanly.

Will try to just get the controller-runtime mode in first and then we can figure out how to iterate and refactoring some of the logic patterns. There's some slight changes in behavior to make it follow the usual ctrl-runtime patterns so it'll be good to break that up into separate smaller pieces

fault-remediation/pkg/reconciler/reconciler.go
ivelichkovich: Will try to get this in a follow up to breakup the PR if that sounds okay.

fault-remediation/pkg/reconciler/reconciler.go
ivelichkovich: same as other retries comment

fault-remediation/main.go
lalitadithya: okay, that makes sense.

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -329,37 +309,46 @@ func (r *Reconciler) handleRemediationEvent(
...
 
 	eventsProcessed.WithLabelValues(CRStatusCreated, nodeName).Inc()
 
-	if err := watcherInstance.MarkProcessed(ctx, eventWithToken.ResumeToken); err != nil {
+	if err = watcherInstance.MarkProcessed(ctx, eventWithToken.ResumeToken); err != nil {
 		processingErrors.WithLabelValues("mark_processed_error", nodeName).Inc()
 		slog.Error("Error updating resume token", "error", err)
+		return ctrl.Result{}, nil
 	}
+	return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
ivelichkovich: This is to retain current behavior, later will follow up with error handling

distros/kubernetes/nvsentinel/values-tilt.yaml
```diff
@@ -21,6 +21,8 @@ global:
   # - values-tilt-postgresql.yaml (when USE_POSTGRESQL=1)
   # This ensures proper configuration for both database backends
 
+  ctrlRuntimeEnabled: false
```
ivelichkovich: once tests pass with this disabled I'll do another run with it enabled and leave it enabled

fault-remediation/pkg/initializer/init.go
joejulian: @lalitadithya as an aside, reconciling on state metadata is an antipattern. You should always be reconciling actual measured state and reflecting that in the resource state.

lalitadithya:
> > hey @ivelichkovich , can you please complete the DCO -- https://github.com/NVIDIA/NVSentinel/pull/449/checks?check_run_id=57068729072 so that we can trigger the e2e tests on the PR for some feedback?
> 
> Can I see progress or logs to make sure its running anywhere?

yes, you be able to see them in the checks page. I need to do the "ok to test" for it start.

.github/workflows/lint-test.yml
```diff
@@ -15,6 +15,9 @@
 name: Lint and Test
 
 on:
+  pull_request:
+    branches:
+      - '*'
```
lalitadithya: this isn't allow by our security policy, can you please undo this change?

fault-remediation/pkg/reconciler/metrics.go
```diff
@@ -26,28 +27,30 @@ const (
 )
 
 var (
+	//TODO: evaluate and remove redundant metrics with ctrl-runtime defaults
+
 	// Event Processing Metrics
-	totalEventsReceived = promauto.NewCounter(
+	totalEventsReceived = promauto.With(crmetrics.Registry).NewCounter(
```
lalitadithya, line 33: Since we have `With(crmetrics.Registry)` would the metrics still behave the same way with it disabled?

.github/workflows/lint-test.yml
ivelichkovich: yes

fault-remediation/pkg/reconciler/metrics.go
ivelichkovich: Yeah I validated the metrics, that's why I update the server to use 

```
server.WithPrometheusMetricsCtrlRuntime()
```

fault-remediation/main.go
```diff
@@ -89,26 +146,140 @@ func run() error {
...
+
+	flag.StringVar(&kubeconfigPath, "kubeconfig-path", "", "path to kubeconfig file")
+
+	flag.StringVar(&tomlConfigPath, "config-path", "/etc/config/config.toml",
 		"path where the fault remediation config file is present")
 
-	dryRun = flag.Bool("dry-run", false, "flag to run fault remediation module in dry-run mode")
+	flag.BoolVar(&dryRun, "dry-run", false, "flag to run fault remediation module in dry-run mode.")
+
+	flag.BoolVar(
```
richardcase, line 246: Is there a specific reason to support both modes of operation (i.e. with and without controller-runtime)?

It would seem cleaner to just support controller-runtime, but i suspect i'm missing the context.

lalitadithya: it's just that for the first iteration we want to able to switch back to the old flow if we start to see something odd. I agree that we shouldn't see many issues, but just a safety measure so that we don't have to revert commits or work under the gun for fixes. 

We can remove the flag fairly quickly once we have had the flow soke and run through several e2e tests on main in CI and also in people's non prod envs.

richardcase: Makes sense, thanks @lalitadithya 👍

ivelichkovich: I have some ideas for the next commit how we can make the old mode work more like ctrl-runtime with really low effort which will allow us to start refactoring some of the logic while retaining both modes in the short term as well

joejulian (commented):
This looks great! Thanks Igor.

lalitadithya (approved):

PR #457, opened 2025-11-27
docs: add design for node drain extensibility

### LINKED ISSUE #351: [Feature]: Add support for custom drain
### Prerequisites

### Feature Summary

Currently NVSentinel provides the standard 3 types of drains which works really well for 80% of the use cases. But in complex environments that make use of meta schedulers or in environments where coordination with users is required for draining pods, the current approaches are not sufficient. We need to provide the ability for customers to be able to plugin custom drain logic into NVSentinel without creating forks. 

As a part of this, also provide an exemplar plugin using Slinky as an example.

### Problem/Use Case

Two use cases:

1. Support for Slinky 
2. Support for environments that need user coordination for drains 

### Proposed Solution

Define and API for node drain in protobuf and allow adding a side car to the node drainer. Based on configuration, call the side car API to perform the drain

### Component

Fault Management

XRFXLP:
## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added an Architecture Decision Record describing a CRD-based custom node-drain extensibility design. Users can opt into a custom drain workflow that renders template-driven DrainRequests, configure a customDrain schema, and supply templates via configuration. Includes sample templates, a controller example, rationale, consequences, alternatives, and testing guidance. The design enforces choosing either built-in or custom drain mode (not both).


docs/designs/014-custom-drain-extensibility.md
```diff
@@ -0,0 +1,250 @@
...
+```go
+func (r *Reconciler) executeCustomDrain(...) error {
+    // Create CR
+    crName, err := r.customDrainClient.CreateDrainCR(ctx, templateData)
+    // Store crName in nodeEventsMap or database
+    return nil  // Return immediately, informer will trigger next reconcile
+}
+```
+
+Assumption: We will either have node drainer supported modes (Immediate, DeleteAfterTimeout, AllowCompletion) or the custom drain mode defined above, not both.
```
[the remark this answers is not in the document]
XRFXLP: the drain modes are not specific to a node, it is cluster-wide config.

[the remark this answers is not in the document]
XRFXLP: fixed

docs/designs/015-custom-drain-extensibility.md
```diff
@@ -0,0 +1,250 @@
+# ADR-015: Behavior — Node Drain Extensibility
+
+## Context
+
+Current flow of node drainer essentially runs `kubectl drain` and when exactly a node will be drained in a cluster is configurable by configmap using `mode` which are: `Immediate`, `DeleteAfterTimeout` and `AllowCompletion`. These three modes essentially covers the entire lifecycle of workload. This works in most of the cases but there is no coordination with scheduler present in the cluster in this process. We need a way to extend the functionality of node drain to make it aware of the ecosystem around scheduler present in the cluster.
```
lalitadithya:
```suggestion
Current flow of node drainer essentially runs `kubectl drain` and when exactly a node will be drained in a cluster is configurable by configmap using `mode` which are: `Immediate`, `DeleteAfterTimeout` and `AllowCompletion`. These three modes essentially covers the entire lifecycle of workload. This works in most of the cases but there are some cases where additional coordination with scheduler (or meta scheduler) present in the cluster in required. Hence, we need a way to extend the functionality of node drain to make it aware of the ecosystem around scheduler present in the cluster.
```

docs/designs/015-custom-drain-extensibility.md
```diff
@@ -0,0 +1,250 @@
...
+    H --> I{Status Phase?}
+    I -->|Complete| J[Proceed with Pod Eviction]
+    I -->|Failed| K[Error/Retry]
+    I -->|Pending/InProgress| H
+```
+
+### Template Variables
+
+```go
+type TemplateData struct {
```
lalitadithya, line 33: Can this instead be:

```suggestion
type TemplateData struct {
  HealthEvent   HealthEevent
  PodsToDrain   map[string][]string
}
```

docs/designs/015-custom-drain-extensibility.md
```diff
@@ -0,0 +1,250 @@
...
+
+### Configuration (TOML)
+
+```toml
+[customDrain]
+    enabled = false
+    templateMountPath = "/etc/drain-templates"
+    templateFileName = "drain-template.yaml"
+    namespace = "nvsentinel"
+    apiGroup = "slurm.slinkyproject.io"
```
lalitadithya: Let's position the language in the design doc as the node drainer module providing an examplar CRD which can be used as is by external controllers or users can bring in their own

```suggestion
    apiGroup = "nvsentinel.nvidia.com"
```

docs/designs/015-custom-drain-extensibility.md
```diff
@@ -0,0 +1,250 @@
...
+    // 4. Return true if complete, false otherwise
+}
+```
+
+**3. Reconciler** (`reconciler/reconciler.go`):
+```go
+func (r *Reconciler) executeCustomDrain(...) error {
+    // Create CR
+    crName, err := r.customDrainClient.CreateDrainCR(ctx, templateData)
+    // Store crName in nodeEventsMap or database
```
lalitadithya: If we store it in `nodeEventsMap` we will loose the state on pod restart right? Would it be simpler to just use a deterministic standardized name for the resource so that we can easily look it up?

docs/designs/015-custom-drain-extensibility.md
```diff
@@ -0,0 +1,250 @@
...
+
+    timeout = "3600s"
+
+    # Status watching (follows Kubernetes standard: InProgress, Complete, Failed)
+    statusConditionType = "Complete"
+    statusConditionStatus = "True"
+
+[customDrain.context]
+    partition = "gpu"
+    nodeSet = "gpu-workers"
```
lalitadithya: Why do we need this? I don't see it being used in the templates

docs/designs/015-custom-drain-extensibility.md
XRFXLP: Using name:
```
 name: drain-{{ .NodeName }}-{{ .HealthEventID }}
```
now

and since reconciler is aware of health event ID, this should be easy to track

docs/designs/015-custom-drain-extensibility.md
XRFXLP: This was intended to name `customDrain.Metadata` where we would pipe both metadata from platform connector and from this configmap to the CR. These were supposed to contains information about identification of node groups, but on second thought I realized that this information can moved to bridge controller of CR.

Removed this now.

docs/designs/015-custom-drain-extensibility.md
```diff
@@ -0,0 +1,237 @@
...
+    statusConditionStatus = "True"
+```
+
+### Template Example
+
+```yaml
+apiVersion: drain.example.com/v1alpha1
+kind: DrainRequest
+metadata:
+    name: drain-{{ .HealthEvent.NodeName }}-{{ .HealthEvent.EventID }}
```
lalitadithya: How do we enforce the name in the templates since this is user provided? Would we have some validation? Can you please add some info about that?

docs/designs/015-custom-drain-extensibility.md
```diff
@@ -0,0 +1,237 @@
...
+    {{- end }}
+
+    metadata:
+    {{- range $key, $value := .HealthEvent.Metadata }}
+        {{ $key }}: {{ $value }}
+    {{- end }}
+
+    reason: "{{ .HealthEvent.Message }}"
+
+    podsToDrain:
```
lalitadithya, line 90: Would this be updated if new pods show up on the node as we are draining or will a new CR be created?

docs/designs/015-custom-drain-extensibility.md
```diff
@@ -0,0 +1,237 @@
...
+    // 2. Parse YAML to unstructured
+    // 3. Create CR via dynamic client
+    // 4. Return CR name
+}
+
+func (c *Client) GetCRStatus(ctx context.Context, crName string) (bool, error) {
+    // 1. Get CR via dynamic client
+    // 2. Extract status.conditions array
+    // 3. Check if condition type matches config and status is True
+    // 4. Return true if complete, false otherwise
```
lalitadithya, line 146: Once the CR is deemed to be complete, should we delete it similar to how the `TokenRequest` works? This will prevent an unnecessary build-up of CRs

docs/designs/015-custom-drain-extensibility.md
XRFXLP: Assumption here is that node will be cordoned once the event arrive to ND => no new pods will land

docs/designs/015-custom-drain-extensibility.md
XRFXLP: I thought the user is going to provide only the GVK and we're free to populate metadata.name.

docs/designs/015-custom-drain-extensibility.md
XRFXLP: + the namespaces map will only contain name other than defined in systemNamspaces field

docs/designs/015-custom-drain-extensibility.md
lalitadithya: yes, that is correct. To enforce that, let's not put the name in the template. Right now, to someone how is reading the doc it gives the impression that they can define the name.

docs/designs/015-custom-drain-extensibility.md
lalitadithya: Okay, let's call that assumption out in the document

docs/designs/015-custom-drain-extensibility.md
```diff
@@ -0,0 +1,273 @@
...
+    apiGroup = "nvsentinel.nvidia.com"
+    version = "v1alpha1"
+    kind = "SlurmDrainRequest"
+
+    timeout = "3600s"
+
+    # Status watching (follows Kubernetes standard: DrainInProgress, DrainComplete, DrainFailed)
+    statusConditionType = "DrainComplete"
+    statusConditionStatus = "True"
+```
```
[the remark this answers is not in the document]
XRFXLP: using 1 gvk now

lalitadithya (approved):

PR #483, opened 2025-12-04
fix: add gpu uuid in cloud event data to showcase in dashboard

nitz2407:
## Summary

Added Gpu uuid inside cloud event to support below use case:

Given a GPU id, how do we know if the same GPU was in other clusters/nodes

can't use entitiesImpacted list since it has other fields also i.e.  GPU and PCI. So if try to show the data in dashboard, all 3  fields are coming(GPU, PCI and GPU_UUID) since part of same list, that will create confusion.

## Testing

Tested on azure dev cluster and health events are coming with gpu uuid successfully

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [x] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: Dashboard

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * GPU UUID is now extracted from impacted entities and included in event data when present.

* **Tests**
  * Added tests covering events with and without GPU UUIDs to validate extraction and exposure.


event-exporter/pkg/transformer/cloudevents.go
```diff
@@ -82,6 +89,10 @@ func ToCloudEvent(event *pb.HealthEvent, metadata map[string]string) (*CloudEven
 		}
 	}
 
+	if gpuUUID != "" {
+		healthEventData["gpuUUID"] = gpuUUID
```
lalitadithya, line 93: This will result in duplicate information in the same health event. I don't agree with this change. The information is already present in the entities impacted. The dashboarding tools should make use of that. Changing the data model to fit dashboards isn't the right approach imo.

PR #485, opened 2025-12-04
feat: restrict node condition message length to 1024Bytes(1KB)

### LINKED ISSUE #470: [Feature]: Clean up node condition length
### Prerequisites

### Feature Summary

With @pteranodan : The length of the node condition messages can be quite long (>38k characters).  Example:

```
 SysLogsXIDError                                           True    Fri, 21 Nov 2025 17:18:33 -0500   Fri, 21 Nov 2025 17:01:02 -0500   SysLogsXIDErrorIsNotHealthy           ErrorCode:145.0h PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec NVLINK_RLW_ERROR; Resolution: WORKFLOW_NVLINK5_ERR Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec ROBUST_CHANNEL_PREEMPTIVE_REMOVAL; Resolution: WORKFLOW_XID_45 Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec ROBUST_CHANNEL_PREEMPTIVE_REMOVAL; Resolution: WORKFLOW_XID_45 Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec ROBUST_CHANNEL_PREEMPTIVE_REMOVAL; Resolution: WORKFLOW_XID_45 Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec ROBUST_CHANNEL_PREEMPTIVE_REMOVAL; Resolution: WORKFLOW_XID_45 Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec ROBUST_CHANNEL_PREEMPTIVE_REMOVAL; Resolution: WORKFLOW_XID_45 Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec ROBUST_CHANNEL_PREEMPTIVE_REMOVAL; Resolution: WORKFLOW_XID_45 Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec ROBUST_CHANNEL_PREEMPTIVE_REMOVAL; Resolution: WORKFLOW_XID_45 Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec ROBUST_CHANNEL_PREEMPTIVE_REMOVAL; Resolution: WORKFLOW_XID_45 Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec ROBUST_CHANNEL_PREEMPTIVE_REMOVAL; Resolution: WORKFLOW_XID_45 Recommended Action=CONTACT_SUPPORT;ErrorCode:45 PCI:0000:29:00 GPU:GPU-8614c5d9-371d-1d8a-9bab-78d0434427ec
```
...continues on for 38,000+ chars.

### Problem/Use Case

This makes the node error conditions easier to read

### Proposed Solution

We should limit the length of a single error description to `1024` characters or something (ex: `512`? `2048`?)

### Component

Health Monitor

deesharma24:
## Summary

restrict node condition message field to be configurable and have been set to 1024 by default which can be changed as per need

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Configurable maximum node condition message length for the Kubernetes connector (default: 1024 bytes); long messages are truncated with "..." and always end with a trailing semicolon.
  * Validation added to ensure a positive length value.

* **Documentation**
  * Configuration docs and examples updated to include the new setting under the Kubernetes connector.

* **Tests**
  * Added/updated tests covering truncation behavior, duplicate XID handling, and flood/truncation scenarios.

### DIFF
--- a/platform-connectors/pkg/connectors/kubernetes/process_node_events.go
+++ b/platform-connectors/pkg/connectors/kubernetes/process_node_events.go
@@ -140,7 +140,7 @@ func (r *K8sConnector) updateNodeConditions(ctx context.Context, healthEvents []
 			}
 
 			if len(messages) > 0 {
-				matchedCondition.Message = fmt.Sprintf("%s;", strings.Join(messages, ";"))
+				matchedCondition.Message = r.truncateNodeConditionMessage(messages)
 				matchedCondition.Status = corev1.ConditionTrue
 				matchedCondition.Reason = r.updateHealthEventReason(events[len(events)-1].CheckName, false)
 			} else {
@@ -547,4 +547,43 @@ func isKubernetesStringError(errStr string) bool {
 	}
 
 	return false
+}
+
+// truncateNodeConditionMessage builds the node condition message while respecting the max node condition
+// message length. It preserves complete error entries and adds a truncation indicator if needed.
+func (r *K8sConnector) truncateNodeConditionMessage(messages []string) string {
+	if len(messages) == 0 {
+		return NoHealthFailureMsg
+	}
+
+	truncationSuffix := "..."
+	maxLen := int(r.maxNodeConditionMessageLength) - len(truncationSuffix)
+
+	var result strings.Builder
+	truncated := false
+
+	for i, msg := range messages {
+		var newEntry string
+		if i == 0 {
+			newEntry = msg
+		} else {
+			newEntry = ";" + msg
+		}
+
+		// Check if adding this entry would exceed the limit (+1 for trailing semicolon)
+		if result.Len()+len(newEntry)+1 > maxLen {
+			truncated = true
+			break
+		}
+
+		result.WriteString(newEntry)
+	}
+
+	result.WriteString(";")
+
+	if truncated {
+		result.WriteString(truncationSuffix)
+	}
+
+	return result.String()
 }
--- a/platform-connectors/pkg/connectors/kubernetes/process_node_events.go
+++ b/platform-connectors/pkg/connectors/kubernetes/process_node_events.go
@@ -552,10 +552,6 @@ func isKubernetesStringError(errStr string) bool {
 // truncateNodeConditionMessage builds the node condition message while respecting the max node condition
 // message length. It preserves complete error entries and adds a truncation indicator if needed.
 func (r *K8sConnector) truncateNodeConditionMessage(messages []string) string {
-	if len(messages) == 0 {
-		return NoHealthFailureMsg
-	}
-
 	truncationSuffix := "..."
 	maxLen := int(r.maxNodeConditionMessageLength) - len(truncationSuffix)
 
--- a/platform-connectors/pkg/connectors/kubernetes/process_node_events.go
+++ b/platform-connectors/pkg/connectors/kubernetes/process_node_events.go
@@ -556,6 +556,7 @@ func (r *K8sConnector) truncateNodeConditionMessage(messages []string) string {
 	maxLen := int(r.maxNodeConditionMessageLength) - len(truncationSuffix)
 
 	var result strings.Builder
+
 	truncated := false
 
 	for i, msg := range messages {
--- a/tests/syslog_health_monitor_test.go
+++ b/tests/syslog_health_monitor_test.go
@@ -94,18 +94,12 @@ func TestSyslogHealthMonitorXIDDetection(t *testing.T) {
 			"kernel: [16450076.435595] NVRM: Xid (PCI:0001:00:00): 31, Graphics SM Warp Exception on (GPC 1, TPC 0, SM 0): Misaligned Address",
 			"kernel: [16450076.435584] NVRM: Xid (PCI:0000:19:00): 62, 32260b5e 000154b0 00000000 2026da96 202b5626 202b5832 202b5872 202b58be",
 			"kernel: [16450076.435595] NVRM: Xid (PCI:0000:19:00): 45, pid=2864945, name=kit, Ch 0000000b",
-			"kernel: [16450076.436857] NVRM: Xid (PCI:0000:19:00): 45, pid=2864945, name=kit, Ch 0000000c",
-			"kernel: [16450076.449750] NVRM: Xid (PCI:0000:19:00): 45, pid=2864945, name=kit, Ch 0000000d",
-			"kernel: [16450076.463693] NVRM: Xid (PCI:0000:19:00): 45, pid=2864945, name=kit, Ch 0000000e",
 		}
 
 		expectedSequencePatterns := []string{
 			`ErrorCode:119 PCI:0002:00:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0002:00:00\): 119.*?Recommended Action=COMPONENT_RESET`,
 			`ErrorCode:79 PCI:0001:00:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0001:00:00\): 79.*?Recommended Action=RESTART_BM`,
 			`ErrorCode:62 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 62.*?Recommended Action=COMPONENT_RESET`,
-			`ErrorCode:45 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 45.*?Recommended Action=CONTACT_SUPPORT`,
-			`ErrorCode:45 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 45.*?Recommended Action=CONTACT_SUPPORT`,
-			`ErrorCode:45 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 45.*?Recommended Action=CONTACT_SUPPORT`,
 			`ErrorCode:45 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 45.*?Recommended Action=CONTACT_SUPPORT`,
 		}
 		expectedEventPatterns := []string{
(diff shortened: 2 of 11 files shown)


platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -548,3 +548,42 @@ func isKubernetesStringError(errStr string) bool {
 
 	return false
 }
+
+// truncateNodeConditionMessage builds the node condition message while respecting the max node condition
+// message length. It preserves complete error entries and adds a truncation indicator if needed.
+func (r *K8sConnector) truncateNodeConditionMessage(messages []string) string {
+	if len(messages) == 0 {
```
lalitadithya: When would `len(messages)` be 0?

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -548,3 +548,42 @@ func isKubernetesStringError(errStr string) bool {
 
 	return false
 }
+
+// truncateNodeConditionMessage builds the node condition message while respecting the max node condition
+// message length. It preserves complete error entries and adds a truncation indicator if needed.
+func (r *K8sConnector) truncateNodeConditionMessage(messages []string) string {
```
lalitadithya, line 554: Can this be simpler where we do a string join and then get the first X chars based on length?

tests/syslog_health_monitor_test.go
```diff
@@ -94,19 +94,13 @@ func TestSyslogHealthMonitorXIDDetection(t *testing.T) {
 			"kernel: [16450076.435595] NVRM: Xid (PCI:0001:00:00): 31, Graphics SM Warp Exception on (GPC 1, TPC 0, SM 0): Misaligned Address",
 			"kernel: [16450076.435584] NVRM: Xid (PCI:0000:19:00): 62, 32260b5e 000154b0 00000000 2026da96 202b5626 202b5832 202b5872 202b58be",
 			"kernel: [16450076.435595] NVRM: Xid (PCI:0000:19:00): 45, pid=2864945, name=kit, Ch 0000000b",
-			"kernel: [16450076.436857] NVRM: Xid (PCI:0000:19:00): 45, pid=2864945, name=kit, Ch 0000000c",
-			"kernel: [16450076.449750] NVRM: Xid (PCI:0000:19:00): 45, pid=2864945, name=kit, Ch 0000000d",
-			"kernel: [16450076.463693] NVRM: Xid (PCI:0000:19:00): 45, pid=2864945, name=kit, Ch 0000000e",
```
lalitadithya, line 99: Why did we remove these?

tests/syslog_health_monitor_test.go
```diff
@@ -94,19 +94,13 @@ func TestSyslogHealthMonitorXIDDetection(t *testing.T) {
...
 		}
 
 		expectedSequencePatterns := []string{
 			`ErrorCode:119 PCI:0002:00:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0002:00:00\): 119.*?Recommended Action=COMPONENT_RESET`,
 			`ErrorCode:79 PCI:0001:00:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0001:00:00\): 79.*?Recommended Action=RESTART_BM`,
 			`ErrorCode:62 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 62.*?Recommended Action=COMPONENT_RESET`,
 			`ErrorCode:45 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 45.*?Recommended Action=CONTACT_SUPPORT`,
-			`ErrorCode:45 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 45.*?Recommended Action=CONTACT_SUPPORT`,
-			`ErrorCode:45 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 45.*?Recommended Action=CONTACT_SUPPORT`,
-			`ErrorCode:45 PCI:0000:19:00 GPU_UUID:GPU-[0-9a-fA-F-]+ kernel:.*?NVRM: Xid \(PCI:0000:19:00\): 45.*?Recommended Action=CONTACT_SUPPORT`,
```
lalitadithya, line 109: Why did we remove these?

tests/syslog_health_monitor_test.go
```diff
@@ -434,3 +428,160 @@ func TestSyslogHealthMonitorSXIDDetection(t *testing.T) {
...
+				"SysLogsXIDError", "SysLogsXIDErrorIsNotHealthy")
+			if err != nil || condition == nil {
+				t.Logf("Condition not found yet: %v", err)
+				return false
+			}
+
+			messageLen := len(condition.Message)
+			t.Logf("Node condition message length: %d characters", messageLen)
+
+			// Verify message length is within the 1KB limit
```
lalitadithya: Isn't the code self explanatory, do we need the comment?

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
deesharma24: Rather than the first X characters, shouldn't we do it like that where if it crossing the 1024 characters limit after adding a new message to existing messages, then we should not add new message at all. That way, it would be much clearer else it might be possible that only a prefix of new message is attached and some part is truncated which might be misunderstood. That's the reason I added .... so that we need there was more to this message but it got truncated because of length

tests/syslog_health_monitor_test.go
deesharma24: because it is crossing the 1024 characters(1kb) limit and from now onwards, we are not supported it unless someone update the values files in helm chart

tests/syslog_health_monitor_test.go
deesharma24: because of above . it was input string and since I   deleted input, the corresponding output also needs to be deleted

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
lalitadithya: At the end of day in both cases, the user is not getting the full information. If they need the full information, they need to look at the database, so I'm wondering if we need the additional complexity since the end user experience will remain the same regardless

tests/syslog_health_monitor_test.go
lalitadithya: Right, but what was this test case testing before? If we remove these lines, won't it cause a silent regression?

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
deesharma24: it's not same.right? showing a partial message  which is inconclusive(as it is not complete) will create more confusion. I get that one can take a look at database to read it complete but at least this way whatever the user is seeing, he/she can try to make sense out of it than the former(partial message)

tests/syslog_health_monitor_test.go
deesharma24: I took at look at the testcase. it will not cause regression because either you add 5 unique kernel logs related to XID or 8, it's shouldn't make a difference as everyone of this log is unique. if it would have same, then I agree, there could be regression

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
deesharma24: removed

tests/syslog_health_monitor_test.go
deesharma24: removed

tests/syslog_health_monitor_test.go
lalitadithya: I think the intent of the test case is the in the real world we have multiple same XIDs happen on the same GPU so we want to test for that here to ensure the system is processing all the XIDs and not dropping it. I'd recommend we either increase the limit or update the test cases by re-ordering the lines so that this is covered

deesharma24: added a new testcase where we have duplicates xid and it doesn't cross 1kb threshold.  That is better than increasing 1kb threshold

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
lalitadithya: Hmm, given the number of XIDs that happen in a burst I think practically it will be the same. What we have is fine for now, resolving this thread.

tests/syslog_health_monitor_test.go
```diff
@@ -185,6 +179,170 @@ func TestSyslogHealthMonitorXIDDetection(t *testing.T) {
 	testEnv.Test(t, feature.Feature())
 }
 
+// TestSyslogHealthMonitorXIDFloodWithDuplicates tests burst XID injection
+func TestSyslogHealthMonitorXIDFloodWithDuplicates(t *testing.T) {
```
lalitadithya: Can we merge this into the truncation test cases? So we have a ton of XID like how we have in the read world, we can check that the first few XIDs show up correctly and then the rest are truncated, what do you think?

deesharma24: fair point. updated the tests to reflect the same.

tests/syslog_health_monitor_test.go
```diff
@@ -434,3 +428,160 @@ func TestSyslogHealthMonitorSXIDDetection(t *testing.T) {
...
+		err = helpers.RemoveNodeManagedByNVSentinelLabel(ctx, client, nodeName)
+		if err != nil {
+			t.Logf("Warning: failed to remove ManagedByNVSentinel label: %v", err)
+		}
+
+		return ctx
+	})
+
+	testEnv.Test(t, feature.Feature())
+}
```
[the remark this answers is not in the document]
deesharma24: removed that.

lalitadithya (approved):

PR #499, opened 2025-12-08
docs: Add health event property overrides design doc

### LINKED ISSUE #462: [Feature]: Provide the ability to override recommendations, fatality of health events
### Prerequisites

### Feature Summary

Currently the health monitors publish health events with all the required information populated which works great for majority of use cases. In some cases, operators may need to override some properties of health event such as `isFatal`, `isHealthy`, `recommendedAction` and so on. Currently it is possible to disable quarantine using CEL, but the health events show up in the Kubernetes node object and events, and this gets confusing to the end user. 

For example, in some environments operators may have disabled remediation for an XID, but the XID will show up on the node object confusing end users who think that something is wrong with the node.

### Problem/Use Case

As an operator I want to override properties of some health events based on my environments without having to fork the health monitoring components.

  

### Proposed Solution

We have 2 options:

- Make every health monitor expose some sort of override options
- Override the health events in the platform connector

The second option is a bit more scalable as it will work for any health monitor both existing and new.

### Component

Health Monitor

XRFXLP:
## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added ADR describing a configuration-driven, rule-based health-event property override system: priority/first-match-wins rule evaluation, match criteria with array-overlap semantics, and supported override fields for v1 (isFatal, recommendedAction). Documents architecture and integration flow, configuration examples (Helm values), observability (metrics/logs), testing guidance, scope, consequences, mitigations, and alternatives.


docs/designs/021-health-event-property-overrides.md
```diff
@@ -0,0 +1,187 @@
...
+
+**Dynamic recommendation changes**: During an incident or maintenance window, you need to quickly change what NVSentinel recommends. Maybe you want all GPU errors to trigger support escalation instead of automatic remediation. Or your organization has specific runbooks that don't match the default recommendations. Right now, you'd need to modify health monitor code and redeploy—too slow for operational response.
+
+### Current Limitations
+
+Today's workarounds fall short:
+
+- CEL rules in fault-quarantine control *remediation* (quarantine or not) but can't change the event properties themselves
+- You can skip quarantine, but the node still shows a fatal error—the disconnect confuses users
```
lalitadithya: nit: 
```suggestion
- You can skip quarantine, but the node still shows a fatal error — the disconnect confuses users
```

docs/designs/021-health-event-property-overrides.md
```diff
@@ -0,0 +1,187 @@
...
+```yaml
+platformConnector:
+  healthEventOverrides:
+    enabled: true
+    rules:
+      # Suppress XID 109 - known to be non-critical in this environment
+      - match:
+          agent: "syslog-health-monitor"
+          errorCode: 
+            - "109"
```
lalitadithya: Can we instead use CEL for the matching? We use CEL in all other places, let's do the same here as well

docs/designs/021-health-event-property-overrides.md
```diff
@@ -0,0 +1,187 @@
...
+
+#### Observability
+
+Prometheus metrics track overrides applied and evaluated. Logs record each override application with event details.
+
+## Consequences
+
+### Positive
+- Suppress errors and adjust recommendations via configuration (no code/redeployment)
+- Solves issue #462 confusion (K8s shows correct severity)
```
lalitadithya: Isn't this a given? Since the design doc is for the issue, it should solve the issue 😅

docs/designs/021-health-event-property-overrides.md
```diff
@@ -0,0 +1,187 @@
...
+- Full observability (logs, metrics, original values in metadata)
+- Clear documentation with warnings
+
+## Alternatives Considered
+
+### 1. CEL-Based Overrides
+
+Use CEL expressions for both matching and overrides.
+
+Rejected: Overkill for simple field matching
```
lalitadithya: I don't quite agree. I think the flexibility it provides might be required otherwise we may end up over time implementing support for various operators. Concretely if someone wants to say, "all errors expect A, B, C are non-fatal", it is easy to do with CEL, but with our custom engine we will need to implement not in, or the user has to enumerate all the errors. 

Moreover, we use CEL in other modules, so it will be helpful to remain consistent

docs/designs/021-health-event-property-overrides.md
```diff
@@ -0,0 +1,187 @@
...
+Rejected: 
+- Code changes required in every health monitor (GPU, CSP, Syslog, K8s Object)
+- Decentralized configuration - operators must configure overrides in 4+ different places
+- Inconsistent implementations (GPU is Python, others Go)
+- Maintenance burden - bug fixes replicated across monitors
+
+
+## Notes
+
+- Scope: Only `isFatal` and `recommendedAction` are overridable in v1
```
lalitadithya: We should also include `isHealthy` as well.

XRFXLP: added, ptal

docs/designs/021-health-event-property-overrides.md
XRFXLP: yeah right, but nonetheless it is one of the use case. Removed thought, ptal

docs/designs/021-health-event-property-overrides.md
XRFXLP: yeah, I was going back and forth with CEL and fixed pattern based matching. Using CEL now ,PTAL.

lalitadithya (approved):

PR #532, opened 2025-12-09
feat: enhance gpu_health_monitor logging format compatible with others nvsentinel component

### LINKED ISSUE #479: [Feature]: Add structured logging for GPU health monitor
### Prerequisites

### Feature Summary

All other modules make use of structured logging except GPU health monitor, it should also be updated to use structured logging for consistency

### Problem/Use Case

Consistent way for logging for all modules

### Proposed Solution

Find and use the slog equivalent in Python

### Component

Health Monitor

deesharma24:
nvsentinel component

## Summary

github Issue => https://github.com/NVIDIA/NVSentinel/issues/479

make gpu_health_monitor logging format to be compatible with other nvsentinel components logging format which is json. used the structlog package for this

```
{"event": "Updated cache for key GpuPowerWatch|GPU|7 with value CachedEntityState(isFatal=False, isHealthy=True)", "level": "info", "timestamp": "2025-12-15T11:41:35.826558Z", "module": "gpu-health-monitor", "version": "0.5.0.dev0"}
{"event": "Updated cache for key GpuSmWatch|GPU|0 with value CachedEntityState(isFatal=False, isHealthy=True)", "level": "info", "timestamp": "2025-12-15T11:41:35.826630Z", "module": "gpu-health-monitor", "version": "0.5.0.dev0"}
{"event": "Updated cache for key GpuSmWatch|GPU|1 with value CachedEntityState(isFatal=False, isHealthy=True)", "level": "info", "timestamp": "2025-12-15T11:41:35.826717Z", "module": "gpu-health-monitor", "version": "0.5.0.dev0"}
{"event": "Updated cache for key GpuSmWatch|GPU|2 with value CachedEntityState(isFatal=False, isHealthy=True)", "level": "info", "timestamp": "2025-12-15T11:41:35.826795Z", "module": "gpu-health-monitor", "version": "0.5.0.dev0"}
{"event": "Updated cache for key GpuSmWatch|GPU|3 with value CachedEntityState(isFatal=False, isHealthy=True)", "level": "info", "timestamp": "2025-12-15T11:41:35.826873Z", "module": "gpu-health-monitor", "version": "0.5.0.dev0"}
{"event": "Updated cache for key GpuSmWatch|GPU|4 with value CachedEntityState(isFatal=False, isHealthy=True)", "level": "info", "timestamp": "2025-12-15T11:41:35.826956Z", "module": "gpu-health-monitor", "version": "0.5.0.dev0"}
```

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Structured JSON logging with configurable level (CLI or LOG_LEVEL) and automatic module/version fields.
  * Container builds accept VERSION, GIT_COMMIT and BUILD_DATE build-time args; package version set during build and exposed in image metadata.

* **Chores**
  * Added structlog dependency to enable structured logging.

* **Tests**
  * Added tests validating JSON log format and log-level filtering.


deesharma24:
@lalitadithya  I checked it. There are few points. below is the comparison between my format(which is exactly same as how we are doing it in other go modules) and structlog format

`Your Current Format (Go slog compatible)
{  "time": "2025-12-09T15:37:51.105805+05:30",  "level": "INFO",  "msg": "User logged in",  "module": "gpu-health-monitor",  "version": "v1.0.0",  "user_id": "12345"}



structlog Default Format
{  "event": "User logged in",  "level": "info",  "timestamp": "2025-12-09T10:07:51.105805Z",  "logger": "gpu-health-monitor",  "user_id": "12345"}
`

if you take a look, in our modules, we called the module name as **module** whereas structlog called it as **logger**.So, I would anyway need to write custom wrapper on top of it to use it. so, Still I would need to write code for that. That way my logger module is simpler.

2. It's is not that complicated. The part that I like is , it is following the exact same design philosphy as golang logger module which we have present here(https://github.com/NVIDIA/NVSentinel/tree/main/commons/pkg/logger). I have created only a similar python version of that which in future other modules can also use. So, we are consistent that way

lalitadithya:
>  we called the module name as module whereas structlog called it as logger.

I think that is fine, we don't have to worry about that. IMO the added complexity for a name change isn't worth it. The goal here is to have structured logging as a philosophy, we don't have to match the match exactly want go's slog does. 

> It's is not that complicated. The part that I like is , it is following the exact same design philosphy as golang logger module

I beg to differ. I think it is fairly complex. The reason why we have a common library for go is because we have multiple modules that are written in go. We have a single one in Python, so the added complexity of maintaining a python package is an overhead. Furthermore, for a python package to be really a package they need to be published to a pypi repo (AFIK), you can't use them as a library like go allows you to do with go mod.

health-monitors/gpu-health-monitor/gpu_health_monitor/cli.py
```diff
@@ -95,11 +96,12 @@ def cli(
...
     dcgm_health_conditions_categorization_mapping_config = config["DCGMHealthConditionsCategorizationMapping"]
-    log.basicConfig(format=logging_config["LogFormat"], datefmt=logging_config["DateTimeFormat"])
-    if verbose:
-        log.getLogger().setLevel(log.DEBUG)
-    else:
-        log.getLogger().setLevel(log.INFO)
+
+    # Initialize structured JSON logging (matches Go slog format)
+    # Version should be injected at build time; "dev" is the default for local development
+    version = os.getenv("VERSION", "dev")
```
lalitadithya: if we read from envvar won't it mean that the user can spoof this as runtime by setting the env var to something else?

health-monitors/gpu-health-monitor/gpu_health_monitor/cli.py
```diff
@@ -95,11 +96,12 @@ def cli(
 
     dcgm_errors_info_dict: dict[str, str] = {}
     dcgm_health_conditions_categorization_mapping_config = config["DCGMHealthConditionsCategorizationMapping"]
-    log.basicConfig(format=logging_config["LogFormat"], datefmt=logging_config["DateTimeFormat"])
-    if verbose:
-        log.getLogger().setLevel(log.DEBUG)
-    else:
-        log.getLogger().setLevel(log.INFO)
+
+    # Initialize structured JSON logging (matches Go slog format)
```
lalitadithya: nit: 
```suggestion
    # Initialize structured JSON logging
```

health-monitors/gpu-health-monitor/gpu_health_monitor/logger.py
```diff
@@ -0,0 +1,111 @@
...
+    # Configure standard library root logger
+    root_logger = logging.getLogger()
+    root_logger.setLevel(log_level)
+
+    # Remove all existing handlers
+    for handler in root_logger.handlers[:]:
+        root_logger.removeHandler(handler)
+
+    # Create handler that writes to stderr
+    handler = logging.StreamHandler(sys.stderr)
```
lalitadithya, line 91: shouldn't it go to stdout?

deesharma24: the advantage of using stderr is it is unbuffered so logs will appear immediately as is inline with golang logic 

https://github.com/NVIDIA/NVSentinel/blob/main/commons/pkg/logger/logger.go#L57

health-monitors/gpu-health-monitor/gpu_health_monitor/cli.py
deesharma24: @lalitadithya This is get trickier. I tried to move this version to the compile time in poetry and peotry is not allowing it and asking for a pep 440 format string (https://python-poetry.org/docs/pyproject/#version). In our case, it will not work because our release are like(v0.1.0,v0.2.0 etc) and has prefix v due to which it is not PEP 440 format. 

So, this leaves us with this option only. now, even in the worst case user spoof this , what is the harm? it is just a version only and is used only in logging. 

let me know what's your opinion on this.

lalitadithya: > So, this leaves us with this option only. now, even in the worst case user spoof this , what is the harm? it is just a version only and is used only in logging.

It will make debugging very hard if we can't rely on the version field. 

Since our versions will always be prefixed with a v, can't we drop the first character?

deesharma24: are you saying if we drop the v prefix completely in gpu-health-monitor  prod version log storing  , that is also okay?

lalitadithya: sure, does it work in a way that can't be spoofed?

deesharma24: yes.updated the code to skip using the env variable and directly read it from poetry which can be read  at compile time and can't be modified at runtime.

health-monitors/gpu-health-monitor/Dockerfile
```diff
@@ -47,8 +55,15 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
...
+ARG VERSION
+ARG GIT_COMMIT
+ARG BUILD_DATE
+
 ENV PYTHONPATH=${PYTHONPATH} \
-    PYTHONUNBUFFERED=1
+    PYTHONUNBUFFERED=1 \
+    VERSION=${VERSION} \
+    GIT_COMMIT=${GIT_COMMIT} \
+    BUILD_DATE=${BUILD_DATE}
```
lalitadithya: With the recent changes, we don't need these vars at runtime right?

deesharma24: right.updated the dockerfile for the same

lalitadithya (approved):

PR #544, opened 2025-12-11
fix(remediation): use ctrl-runtime business logic

ivelichkovich:
## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ X ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ X ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added remediation state tracking on nodes for multi-step recovery workflows.
  * Introduced orchestrated log collection jobs for fault diagnostics.
  * Enhanced multi-template remediation with dry-run support.

* **Bug Fixes**
  * Improved error handling and retry logic for state management operations.

* **Improvements**
  * Unified internal architecture for more reliable fault remediation.
  * Added HTTP request auditing for better observability.
  * Enhanced metrics collection for remediation and log collector operations.


fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -197,63 +197,47 @@ func (r *FaultRemediationReconciler) runLogCollector(ctx context.Context, health
...
 
 	// Update state to "remediating"
-	_, err := r.config.StateManager.UpdateNVSentinelStateNodeLabel(ctx,
+	_, err := r.Config.StateManager.UpdateNVSentinelStateNodeLabel(ctx,
 		healthEventWithStatus.HealthEvent.NodeName,
 		statemanager.RemediatingLabelValue, false)
 	if err != nil {
 		slog.Error("Error updating node label to remediating", "error", err)
 		processingErrors.WithLabelValues("label_update_error", nodeName).Inc()
+		return "", err
```
ivelichkovich: so want to call this out:

this will have some behavior changes, now it'll pretty much always return an error when it experiences an error so there's cases where reconciliation will then fully retry rather than ignoring the error and continuing the process

fault-remediation/main.go
```diff
@@ -166,7 +166,17 @@ func setupNonCtrlRuntimeManaged(ctx context.Context, components *initializer.Com
 
 		for event := range components.FaultRemediationReconciler.Watcher.Events() {
 			slog.Info("Event received", "event", event)
-			_, _ = components.FaultRemediationReconciler.Reconcile(gCtx, &event)
+			for i := 1; i <= components.FaultRemediationReconciler.Config.UpdateMaxRetries; i++ {
```
ivelichkovich: so ctrl-runtime pattern is to throw error to retry and then there's built in exponential backoff and all that so updating the "legacy" reconciler to also retry on errors so that the underlying code can rely on throwing errors to trigger retries while still supporting both modes

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -454,8 +440,10 @@ func (r *FaultRemediationReconciler) checkExistingCRStatus(
 
 	slog.Info("CR completed or failed, allowing retry", "node", nodeName, "crName", groupState.MaintenanceCR)
 
-	if err := r.annotationManager.RemoveGroupFromState(ctx, nodeName, group); err != nil {
+	//TODO: this also has built in client retry on the api request
```
ivelichkovich: Yeah so it looks like some of the managers like annotationManager have built in retries. I'm planning to have a flag set on the manager struct that tells it to retry or not so that we don't double up on retries when following this "error for retry" pattern

fault-remediation/pkg/reconciler/reconciler.go
lalitadithya: can we wrap this error like we have done in the other places

lalitadithya: makes sense

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -336,27 +318,30 @@ func (r *FaultRemediationReconciler) handleRemediationEvent(
 
 	// Run log collector only when we're about to create a new CR
 	// This prevents duplicate log-collector jobs when multiple events arrive for the same node
-	_ = r.runLogCollector(ctx, healthEvent)
-
-	nodeRemediatedStatus, _, err := r.performRemediation(ctx, healthEventWithStatus)
+	err = r.runLogCollector(ctx, healthEvent)
```
lalitadithya: We actually don't want to retry on this one. If the log collection fails, that is fine, we can proceed with remediation. There can be cases when the log collection will never to completion due to the node being in a very bad shape. If we retry on this then we may end up in an infinite loop.

fault-remediation/pkg/reconciler/reconciler.go
lalitadithya: makes sense

fault-remediation/main.go
lalitadithya: make sense

lalitadithya (commented):
overall, I like the direction! Let's keep pushing forward on this

fault-remediation/pkg/reconciler/reconciler.go
ivelichkovich: Okay, I have some future refactors planned for the log collector anyway but for now I'll make it not throw an error if the log collector job fails to complete in time but still throw an error and retry if there's i.e. a failure on the api request to create the job resource.

Eventually I want to get the flow of the log collector to be something like this:

1. Creates log collector Job (retry on error here)
3. Requeues event to be re-processed after X min, this allows other events to be processed
4. Checks if job finished, requeue if not (until timeout or complete)
5. Eventually after completed or the timeout has passed, continue on

ivelichkovich: This would likely come later once we're able to remove the "legacy" reconciler mode because it would depend on the ctrl-runtime queuing functionality

lalitadithya: okay, that makes sense.

fault-remediation/pkg/annotation/annotation.go
```diff
@@ -0,0 +1,169 @@
...
+		}, node, nil
+	}
+
+	var state RemediationStateAnnotation
+	if err = json.Unmarshal([]byte(annotationValue), &state); err != nil {
+		slog.Error("Failed to unmarshal annotation", "node", nodeName, "error", err)
+		// Return empty state if unmarshal fails
+		return &RemediationStateAnnotation{
+			EquivalenceGroups: make(map[string]EquivalenceGroupState),
+		}, node, nil
```
yavinash007, line 73: nit: do you think returning an error might be useful ?

fault-remediation/pkg/reconciler/reconciler.go
ivelichkovich: so this one doesn't have an initial error to join together with the updateLabels error

fault-remediation/pkg/reconciler/reconciler.go
ivelichkovich: I ended up doing this in this PR, created a ctrl runtime version of each of the interfaces so it seemed to fit here

commons/pkg/statemanager/statemanager.go
```diff
@@ -292,6 +295,98 @@ func (manager *stateManager) UpdateNVSentinelStateNodeLabel(ctx context.Context,
...
+		"node", nodeName)
+
+	// Return validation error AFTER successful label update
+	// This allows callers to emit error metrics while the label reflects reality
+	if validationErr != nil {
+		return true, validationErr
+	}
+
+	return true, err
+}
```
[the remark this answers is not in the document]
ivelichkovich: This will return an error and ctrlruntime will then retry

fault-remediation/pkg/remediation/remediation.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+
+	if len(existingJobs.Items) == 0 {
+		err = c.client.Create(ctx, job)
+		if err != nil {
+			return batchv1.Job{}, ctrl.Result{}, err
+		}
+		// if created, requeue to check status later
+		return batchv1.Job{}, ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+	}
```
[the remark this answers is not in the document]
ivelichkovich: if that happens it will just retry.

commons/pkg/statemanager/statemanager.go
```diff
@@ -292,6 +295,98 @@ func (manager *stateManager) UpdateNVSentinelStateNodeLabel(ctx context.Context,
...
+
+// UpdateNVSentinelStateNodeLabel will update the given node to the given value for the dgxc.nvidia.com/nvsentinel-state
+// label or it will remove the given label if removeStateLabel is true.
+func (manager *ctrlRuntimeStateManager) UpdateNVSentinelStateNodeLabel(
+	ctx context.Context,
+	nodeName string,
+	newStateLabelValue NVSentinelStateLabelValue,
+	removeStateLabel bool,
+) (bool, error) {
+	node := &corev1.Node{}
```
lalitadithya: Instead of duplicating the logic, can we instead get a Kubernetes client from controller runtime manager and use that:

```
client, err := kubernetes.NewForConfig(mgr.GetConfig())
```

It is not ideal, but it will allow us to keep the logic in a single function, and we don't have to make bug fixes in two places. Once we have all the modules cut over, we can do a final refactor to use the controller-runtime client, what do you think?

fault-remediation/pkg/annotation/annotation.go
lalitadithya, line 1: Same comment as earlier about converting clients so that we don't have duplicated logic

fault-remediation/pkg/remediation/remediation.go
```diff
@@ -0,0 +1,467 @@
...
+		return "", err
+	}
+
+	slog.Info(fmt.Sprintf("Generated YAML: %s", buf.String()))
+
+	// Convert YAML to unstructured
+	var obj map[string]any
+	if err = yaml.Unmarshal(buf.Bytes(), &obj); err != nil {
+		slog.Error("Failed to unmarshal YAML", "error", err)
+		return "", err
```
lalitadithya: Can we wrap errors please?

fault-remediation/pkg/remediation/deprecated_remediation.go
lalitadithya, line 1: I'm wondering if we should just bite the bullet and go all in on controller-runtime for this module... I'm not really a fan of all the duplication we have due to use of "kubenertes" client v/s controller-runtime client. TBH, it is much more than what I had initially anticipated. I'm worried that bug fixes will become complex till we get rid of one of the flows. What do you think?

ivelichkovich: Yeah agreed I think that would be much cleaner at this point

commons/pkg/statemanager/statemanager.go
ivelichkovich: Yeah that's a good call

ivelichkovich: Part of the reason was because the retries here would conflict with the built in ctrl runtime retries when an error is returned but that's probably worth the tradeoff of not having the duplicate logic

fault-remediation/pkg/annotation/annotation.go
ivelichkovich: ended up consolidating on ctrl-runtime

fault-remediation/pkg/annotation/annotation.go
ivelichkovich: It could be but when it's malformed it ends up returning empty and because there's no error subsequent functions can keep going and fix the malformed json i.e. in UpdateRemediationState

fault-remediation/pkg/annotation/annotation.go
```diff
@@ -0,0 +1,166 @@
...
+		return fmt.Errorf("failed to patch node annotation: %w", err)
+	}
+
+	slog.Info("Updated remediation state annotation for node",
+		"node", nodeName,
+		"group", group,
+		"crName", crName)
+
+	return nil
+}
```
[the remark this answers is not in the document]
ivelichkovich: this should return an error and make ctrl runtime requeue and retry

fault-remediation/pkg/remediation/remediation.go
```diff
@@ -0,0 +1,592 @@
...
+			fmt.Errorf("expecting zero or one log collector job per event per node, found %v", existingJobs.Items)
+	}
+
+	if len(existingJobs.Items) == 0 {
+		err = c.client.Create(ctx, job)
+		if err != nil {
+			return batchv1.Job{}, ctrl.Result{}, err
+		}
+		// if created, requeue to check status later
+		return batchv1.Job{}, ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
KaivalyaMDabhadkar: nit: Can we declare common config like `10 * time.Second` as constants in the package and use them?

fault-remediation/main.go
```diff
@@ -235,20 +181,13 @@ func parseFlags() {
 			" (otherwise metrics and health are on same port).",
 	)
 
-	flag.StringVar(&kubeconfigPath, "kubeconfig-path", "", "path to kubeconfig file")
+	flag.StringVar(&kubeconfigPath, "kubeconfig", "", "path to kubeconfig file")
```
KaivalyaMDabhadkar: Since we are using `GetConfigOrDie()` above to get the kubeconfig, do we need this separate flag to pass the kubeconfig value?

ivelichkovich: good call, that var is automatically setup by ctrl-runtime

fault-remediation/pkg/annotation/annotation.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	var state RemediationStateAnnotation
+	if err = json.Unmarshal([]byte(annotationValue), &state); err != nil {
+		slog.Error("Failed to unmarshal annotation", "node", nodeName, "error", err)
+		// Return empty state if unmarshal fails
+		return &RemediationStateAnnotation{
+			EquivalenceGroups: make(map[string]EquivalenceGroupState),
+		}, node, nil
+	}
```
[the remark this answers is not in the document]
ivelichkovich: we aren't returning an error here on purpose, if it is corrupt we are treating it as empty rather than erroring out. This way if some manual action or something makes it corrupt workflows on the node won't get stuck

fault-remediation/pkg/annotation/annotation.go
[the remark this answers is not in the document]
ivelichkovich: this should be fine... I'll update the code anyway but if we have a silent overwrite it means two things are trying to act on the same part of the node at the same time in different ways which we shouldn't have happening anyway

fault-remediation/pkg/annotation/annotation.go
```diff
@@ -0,0 +1,166 @@
...
+	delete(node.Annotations, AnnotationKey)
+
+	if err = m.client.Patch(ctx, node, client.MergeFrom(patchBase)); err != nil {
+		return fmt.Errorf("failed to patch node annotation: %w", err)
+	}
+
+	slog.Info("Cleared remediation state annotation for node", "node", nodeName)
+
+	return nil
+}
```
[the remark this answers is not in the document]
ivelichkovich: this should be fine... I'll update the code anyway but if we have a silent overwrite it means two things are trying to act on the same part of the node at the same time in different ways which we shouldn't have happening anyway

fault-remediation/pkg/annotation/annotation.go
```diff
@@ -0,0 +1,166 @@
...
+	node.Annotations[AnnotationKey] = string(stateJSON)
+
+	if err = m.client.Patch(ctx, node, client.MergeFrom(patchBase)); err != nil {
+		return fmt.Errorf("failed to patch node annotation: %w", err)
+	}
+
+	slog.Info("Removed group from remediation state for node", "node", nodeName, "group", group)
+
+	return nil
+}
```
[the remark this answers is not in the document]
ivelichkovich: this should be fine... I'll update the code anyway but if we have a silent overwrite it means two things are trying to act on the same part of the node at the same time in different ways which we shouldn't have happening anyway

fault-remediation/pkg/remediation/remediation.go
```diff
@@ -0,0 +1,592 @@
...
+
+			metrics.LogCollectorJobs.WithLabelValues(nodeName, "success").Inc()
+			metrics.LogCollectorJobDuration.WithLabelValues(nodeName, "success").Observe(duration)
+		}
+
+		return true, nil
+	}
+
+	return false, nil
+}
```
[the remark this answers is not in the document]
ivelichkovich: The job should never be complete and not have a startTime

fault-remediation/pkg/remediation/remediation.go
```diff
@@ -0,0 +1,583 @@
...
+		err = c.client.Create(ctx, job)
+		if err != nil {
+			return batchv1.Job{}, ctrl.Result{}, err
+		}
+		// if created, requeue to check status later
+		return batchv1.Job{}, ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	return existingJobs.Items[0], ctrl.Result{}, nil
+}
```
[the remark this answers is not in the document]
ivelichkovich: yes job is namespace scoped

KaivalyaMDabhadkar (approved):

PR #597, opened 2025-12-17
fix: allow deleteaftertimout priority to be higher than allow-completion in node-drainer

### LINKED ISSUE #593: [Bug]: Node Drainer DeleteAfterTimeout mode never triggers when AllowCompletion namespaces exist on same node
### Prerequisites

### Bug Description

In the investigation for #386 I think I hit a bug. Please correct me if I misunderstand how this works!

When a node has pods from both AllowCompletion AND DeleteAfterTimeout namespaces, the DeleteAfterTimeout behavior is never triggered. Pods in DeleteAfterTimeout namespaces are never force-deleted, even after the configured timeout period.

**Expected:** DeleteAfterTimeout pods should be force-deleted after `deleteAfterTimeoutMinutes` regardless of AllowCompletion pods on the same node.

**Actual:** DeleteAfterTimeout pods remain indefinitely because the `getAction()` function returns early when AllowCompletion pods exist, preventing the DeleteAfterTimeout logic from ever executing.

### Component

Other

### Steps to Reproduce

1. Configure node-drainer with mixed eviction modes:ml
```
   deleteAfterTimeoutMinutes = 2
   [[userNamespaces]]
   name = "test-immediate"
   mode = "Immediate"
   [[userNamespaces]]
   name = "test-allow-completion"
   mode = "AllowCompletion"
   [[userNamespaces]]
   name = "test-delete-timeout"
   mode = "DeleteAfterTimeout"
```

2. Deploy pods to all three namespaces on the same nodes

3. Trigger node cordon via NVSentinel (fatal health event)

4. Observe:
   - test-immediate pods: evicted immediately ✅
   - test-allow-completion pods: waiting (correct) ✅
   - test-delete-timeout pods: should force-delete after 2 min, but never do ❌

5. Check node-drainer logs - action is always "CheckCompletion", never "EvictWithTimeout"

### Environment

**Environment**
- Version: v0.4.0 (also confirmed in v0.5.0 and main branch)
- Kubernetes version: 1.29
- Deployment method: Helm
- Cluster size: 1500 nodes

### Logs/Output

**Logs**
Node drainer logs show only CheckCompletion action, never EvictWithTimeout:
```
{"level":"INFO","msg":"Evaluated action for node","node":"ip-100-64-128-15.ec2.internal","action":"CheckCompletion"}
{"level":"INFO","msg":"Pods still running on node, requeueing for later check","node":"ip-100-64-128-15.ec2.internal","remainingPods":["test-allow-completion/inference-sim-597cb57655-j9s5r","test-delete-timeout/inference-sim-6564f68667-z97ps"]}
```

### Suspected root cause 
in `node-drainer/pkg/evaluator/evaluator.go` lines 157-169:
The getAction() function processes modes sequentially with early returns:
   1. Immediate → works
   2. AllowCompletion → if pods exist, returns ActionCheckCompletion (BLOCKS step 3)
   3. DeleteAfterTimeout → NEVER REACHED when step 2 returns

```
func (e *NodeDrainEvaluator) getAction(ctx context.Context, ns namespaces, nodeName string) *DrainActionResult {
	// Step 1: Immediate - works fine
	if len(ns.immediateEvictionNamespaces) > 0 {
		// ... evicts immediately
	}

	// Step 2: AllowCompletion - THE BUG IS HERE
	if len(ns.allowCompletionNamespaces) > 0 {
		action := e.handleAllowCompletionNamespaces(ns, nodeName)
		if action != nil {
			return action  // ← EARLY RETURN blocks Step 3!
		}
	}

	// Step 3: DeleteAfterTimeout - NEVER REACHED
	if len(ns.deleteAfterTimeoutNamespaces) > 0 {
		action := e.handleDeleteAfterTimeoutNamespaces(ns, nodeName)
		// This code never runs when AllowCompletion pods exist
	}
}
```

The fix requires processing all eviction modes independently, not sequentially.

deesharma24:
## Summary

Issue => https://github.com/NVIDIA/NVSentinel/issues/593

Currently, node-drainer works in running different modes(allowCompletion,deleteAfterTimeout,Immediate) sequentially which means and run in the following order Immediate => allowCompletion => deleteAfterTimeout.this means if there are pods in a namespace are in allowCompletion mode, then even if there are pods in some other namespace which are set to deleteAfterTimeout and their timeout expires, still they will not be deleted because there are pods in allowCompletion mode in some other namespace which are pending and waiting for to be completed. While this doesn't create overall problem from the node drainer point of view because we need all kinds of user pods from all namespaces to be deleted in order to effectively say node is drained, but the behavior has been impacted.

To mitigate this,  giving deleteAfterTimeout more priority than AllowCompletion so that they can be deleted once timeout expires.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Bug Fixes / Improvements**
  * Eviction ordering now prioritizes timeout-based evictions (DeleteAfterTimeout) over allow-completion so timeout pods are processed first while completion pods wait.

* **Tests**
  * Added an integration test and updated end-to-end tests for mixed-mode scenarios, verifying timeout-first behavior, phased verification steps, and final manual completion flow.
  * Adjusted test timing and phase sequencing to reflect the new priority.

* **Documentation**
  * Updated comments and phase log messages to clarify the new ordering and test flow.


lalitadithya:
> in this change, both AllowCompletion and DeleteAfterTimeout mode will run in parallel and whichever one is completed first

I'm not sure if we need this complexity. By definition, `AllowCompletion` can never finish before `DeleteAfterTimeout`, so can we instead flip the order so that `DeleteAfterTimeout` is processed first and then `AllowCompletion`. This will avoid any parallel processing

tests/node_drainer_test.go
```diff
@@ -182,3 +183,106 @@ func TestNodeDrainerEvictionModes(t *testing.T) {
 
 	testEnv.Test(t, feature.Feature())
 }
+
+// TestNodeDrainerMixedEvictionPriority verifies that DeleteAfterTimeout pods are
+// force-deleted even when AllowCompletion pods exist on the same node.
+// This ensures that pods with deadlines (DeleteAfterTimeout) are processed first
+// and not blocked by pods waiting for natural completion (AllowCompletion).
+func TestNodeDrainerMixedEvictionPriority(t *testing.T) {
```
lalitadithya: Can we merge this with the existing `TestNodeDrainerEvictionModes`? That already has the phases for the various modes.

node-drainer/pkg/reconciler/reconciler_integration_test.go
```diff
@@ -448,6 +448,36 @@ func TestReconciler_ProcessEvent(t *testing.T) {
...
+			expectedNodeLabel: ptr.To(string(statemanager.DrainingLabelValue)),
+			validateFunc: func(t *testing.T, client kubernetes.Interface, ctx context.Context, nodeName string, err error) {
+				// When both modes have pods, DeleteAfterTimeout should be processed first
+				// The error should indicate timeout eviction (waiting for timeout), NOT completion check
+				assert.Error(t, err)
+				assert.Contains(t, err.Error(), "timeout eviction",
+					"DeleteAfterTimeout should be processed before AllowCompletion")
+				// Should NOT contain "AwaitingPodCompletion" which is the AllowCompletion mode message
+				assert.NotContains(t, err.Error(), "AwaitingPodCompletion",
+					"AllowCompletion should NOT be processed when DeleteAfterTimeout has pods")
```
lalitadithya: Can we also assert that allow competition event comes in after the pods have been deleted post timeout?

tests/node_drainer_test.go
deesharma24: done

node-drainer/pkg/reconciler/reconciler_integration_test.go
deesharma24: done

node-drainer/pkg/reconciler/reconciler_integration_test.go
```diff
@@ -448,6 +448,54 @@ func TestReconciler_ProcessEvent(t *testing.T) {
...
+
+				// Simulate DeleteAfterTimeout completion by deleting the timeout pod
+				t.Log("Simulating DeleteAfterTimeout completion - deleting timeout-pod")
+				err = client.CoreV1().Pods("timeout-test").Delete(ctx, "timeout-pod", metav1.DeleteOptions{
+					GracePeriodSeconds: ptr.To(int64(0)),
+				})
+				require.NoError(t, err)
+
+				// Wait for pod to be fully deleted
+				require.Eventually(t, func() bool {
```
lalitadithya: I don't understand the test, we are deleting the pods in test and we are validated that they are deleted in the same test?

tests/node_drainer_test.go
```diff
@@ -128,16 +129,28 @@ func TestNodeDrainerEvictionModes(t *testing.T) {
...
+		t.Log("Phase 3: Waiting for deleteAfterTimeout to expire (~60s)")
+		// The deleteAfterTimeoutMinutes is set to 1 minute in nd-all-modes.yaml
+		// Adding 10s buffer to account for processing time
+		time.Sleep(70 * time.Second)
 
 		t.Log("Phase 3: DeleteAfterTimeout pods force-deleted after timeout")
+		// This verifies that DeleteAfterTimeout is processed before AllowCompletion (not blocked by it)
 		helpers.WaitForPodsDeleted(ctx, t, client, "delete-timeout-test", deleteTimeoutPods)
 
+		t.Log("Phase 3: Verify AllowCompletion pods are still waiting (priority verification)")
```
lalitadithya: Should this be phase 4? It looks like we have multiple phase 3s?

deesharma24: > Should this be phase 4? It looks like we have multiple phase 3s?

I have made it simpler to user 4,5,6

node-drainer/pkg/reconciler/reconciler_integration_test.go
deesharma24: Since this is mixed-eviction, created a separate test to target that. That is much simpler because we need a requeue mechnism.

lalitadithya (approved):

PR #603, opened 2025-12-17
feat(fault-remediation): add multi-template support

zydee3:
## Summary

This PR adds support for multiple remediation templates based on recommended action types to enable different CR templates for different fault classes. 

### Key Changes
  - Added action-specific template configs via `RemediationActions` map
  - Updated template selection to be based on the recommended action type
  - Added support for cluster-scoped and namespaced maintenance resources

### Goals
  - Add support for multiple CR based on recommended action type
  - Add support for injecting deployment-specific metadata while remaining vendor-neutral

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [X] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Per-action remediation model with multi-template support; per-action templates and actions now configurable.
  * New log-collector options and configurable retry settings for maintenance updates.

* **Improvements**
  * Startup configuration validation and structured logging.
  * RBAC generation adapts to configured actions/resources; node annotations now record which action created a remediation resource.
  
* **Tests**
  * Added validation, equivalence-group, integration, and template-related tests.


distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
```diff
@@ -17,6 +17,13 @@ replicaCount: 1
 
 logLevel: info
 
+# Global configuration (typically provided by parent chart)
+global:
+  auditLogging:
+    enabled: false
+  datastore:
+    provider: ""
```
lalitadithya: I don't think we need this; the parent chart should provide these. Are you seeing places where helm rendering is failing due to lack of these values?

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
```diff
@@ -43,46 +50,59 @@ podAnnotations: {}
...
-  template: |
-    apiVersion: janitor.dgxc.nvidia.com/v1alpha1
-    kind: RebootNode
-    metadata:
-      name: maintenance-{{ .NodeName }}-{{ .HealthEventID }}
-    spec:
-      nodeName: {{ .NodeName }}
+  # Per-action remediation definitions
+  # Key is the RecommendedAction string (e.g., "COMPONENT_RESET", "RESTART_BM") 
+  actions: {}
```
lalitadithya, line 51: Can we add in the configuration for reboots by default? I believe the tests will fail without that

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -447,7 +447,9 @@ func (r *FaultRemediationReconciler) checkExistingCRStatus(
 		return true, "", nil
 	}
 
-	shouldSkip := statusChecker.ShouldSkipCRCreation(ctx, groupState.MaintenanceCR)
+	// Convert RecommendedAction to string for action lookup
+	actionName := healthEvent.RecommendedAction.String()
+	shouldSkip := statusChecker.ShouldSkipCRCreation(ctx, actionName, groupState.MaintenanceCR)
```
KaivalyaMDabhadkar: Should we also make changes to `checkExistingCRStatus`? We are using equivalence groups there for deduplication and creating different CRDs for different actions will have implications there

fault-remediation/pkg/reconciler/remediation.go
```diff
@@ -71,16 +75,26 @@ type FaultRemediationClient struct {
...
+	
+	// CRD routing metadata (populated from MaintenanceResource)
+	ApiGroup              string
+	Version               string
+	Kind                  string
+	Namespace             string
+	
+	// Generic metadata hooks for extensibility
+	ExtraLabels           map[string]string
+	ExtraAnnotations      map[string]string
```
KaivalyaMDabhadkar: Where are we populating these in the code?

fault-remediation/pkg/crstatus/checker.go
```diff
@@ -66,16 +73,16 @@ func (c *CRStatusChecker) ShouldSkipCRCreation(ctx context.Context, crName strin
 		return false
 	}
 
-	resource, err := c.dynamicClient.Resource(mapping.Resource).Get(ctx, crName, metav1.GetOptions{})
+	cr, err := c.dynamicClient.Resource(mapping.Resource).Get(ctx, crName, metav1.GetOptions{})
```
KaivalyaMDabhadkar: We need to make sure that if the CR is configured at the namespace level, we specify the namespace in this call to fetch the resource as well

fault-remediation/pkg/config/config.go
```diff
@@ -14,16 +14,22 @@
...
 	Version               string `toml:"version"`
 	ApiGroup              string `toml:"apiGroup"`
 	Kind                  string `toml:"kind"`
 	CompleteConditionType string `toml:"completeConditionType"`
+	
+	// Scope determines if the resource is cluster-scoped or namespaced
+	Scope string `toml:"scope"` // "Cluster" or "Namespaced"
+	
+	// Template file path relative to the mount path
+	TemplateFile string `toml:"templateFile"`
```
KaivalyaMDabhadkar: Can we put some sort of validation for ensuring that the `TemplateFile` references that we configure here always exist in the `templates` map as well?

distros/kubernetes/nvsentinel/charts/fault-remediation/templates/clusterrole.yaml
```diff
@@ -18,16 +18,36 @@ metadata:
...
+  {{- range $resources }}
   - {{ . }}
   {{- end }}
   verbs:
   - get
   - list
   - create
+  - update
+  - patch
+  - delete
```
KaivalyaMDabhadkar: Do we need the delete permission?

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
zydee3: I added example actions in the e2e tests so CI would not fail (tests also pass locally). I intentionally avoided enabling reboot remediation by default, since this is likely environment-specific. My assumption was that users should explicitly opt into the reboot behavior.

If you think a safe default reboot action is preferred for the chart, I’m happy to add it. What do you think?

fault-remediation/pkg/reconciler/reconciler.go
zydee3: This is a good point. I am not sure at the moment. I would like to hear your thoughts.

As you've mentioned, different actions can now create different CRD types from different operators, which creates complexity in the deduplication logic since we need to check the status of different CRD types within the same equivalence group. 

An alternative approach would be to shift deduplication to the CRD level. This allows us to group by the actual remediation class rather than the recommended action. For example, all reboot class CRDs would deduplicate against each other regardless of which specific action created them, while different remediation classes like GPU resets versus node reboots could be treated independently. 

I like the CRD-level deduplication because it would:
- align the deduplication logic more naturally with the CRD design 
- simplify the cross-CRD checking requirements
- would simplify the fault-remediator implementation
- be semantically intuitive: "don't run multiple Reboot CRDs on the same node"
- potentially improve operational granularity (?). GPU reset and node reboot are actually different operations so maybe they should be allowed to run in parallel?

My main concern is operational safety since the current equivalence groups encode domain knowledge about what operations may conflict. Honestly though, maybe I am being naive here, but I think that if a CRD is properly, then types should already represent the right granularity.

fault-remediation/pkg/reconciler/remediation.go
zydee3: Removed. This was apart of an initial implementation design that was reverted because we wanted to contribute vendor-neutral changes.

fault-remediation/pkg/config/config.go
zydee3: Done.

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
lalitadithya: I think the safe default is reboot. Plus, also for backwards compatibility since we already had reboot specified as default, we can continue to keep it

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
```diff
@@ -43,46 +43,55 @@ podAnnotations: {}
...
+    # Example configuration:
+    # COMPONENT_RESET:
+    #   apiGroup: "janitor.dgxc.nvidia.com"
+    #   version: "v1alpha1" 
+    #   kind: "RebootNode"
+    #   scope: "Cluster"  # "Cluster" or "Namespaced"
+    #   completeConditionType: "NodeReady"
+    #   templateFile: "nvidia-reboot.yaml"
+    # RESTART_BM:
```
miguelvr, line 67:
```suggestion
    # RESTART_VM:
```

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
miguelvr: @lalitadithya I would argue that a safe default should work in every installation. However, the existing reboot CRD only works on specific cloud providers.

The way that helm defaults work makes it extra hard to disable it / change it. It is not possible to remove the key completely if you want to disable reboots.

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
```diff
@@ -43,46 +43,55 @@ podAnnotations: {}
...
+  # Key is the RecommendedAction string (e.g., "COMPONENT_RESET", "RESTART_BM") 
+  actions: {}
+    # Example configuration:
+    # COMPONENT_RESET:
+    #   apiGroup: "janitor.dgxc.nvidia.com"
+    #   version: "v1alpha1" 
+    #   kind: "RebootNode"
+    #   scope: "Cluster"  # "Cluster" or "Namespaced"
+    #   completeConditionType: "NodeReady"
+    #   templateFile: "nvidia-reboot.yaml"
```
miguelvr: should this just be called `templateRef` instead?

fault-remediation/pkg/reconciler/reconciler.go
miguelvr: my 2cents: we should not assume that "CRD does action X" for the implementation. CR creation should be a reaction to an event emitted by NVSentinel, and the fault-remediation component should simply be an orchestrator.

fault-remediation/pkg/reconciler/remediation.go
```diff
@@ -176,46 +211,94 @@ func (c *FaultRemediationClient) GetStatusChecker() *crstatus.CRStatusChecker {
...
 	healthEvent := healthEventData.HealthEvent
 	healthEventID := healthEventData.ID
 
+	// Convert RecommendedAction to string for lookup
+	recommendedActionName := healthEvent.RecommendedAction.String()
+	
+	// Select appropriate maintenance resource configuration and template
+	var maintenanceResource config.MaintenanceResource
+	var selectedTemplate *template.Template
+	var actionKey string
```
miguelvr: nit: group variables


```suggestion
	var (
		maintenanceResource config.MaintenanceResource
		selectedTemplate 	*template.Template
		actionKey 			string
	)
```

fault-remediation/pkg/reconciler/remediation.go
```diff
@@ -176,46 +211,94 @@ func (c *FaultRemediationClient) GetStatusChecker() *crstatus.CRStatusChecker {
...
+			"node", healthEvent.NodeName,
+			"availableActions", func() []string {
+				actions := make([]string, 0, len(c.remediationConfig.RemediationActions))
+				for action := range c.remediationConfig.RemediationActions {
+					actions = append(actions, action)
+				}
+				return actions
+			}())
+		return false, ""
+	}
```
miguelvr: nit: for improved readability write this as:

```golang
resource, exists := ...
if !exists {
	...
	return false, ""
}

// assume exists from this point on
```

avoiding the `else` block

fault-remediation/pkg/reconciler/remediation.go
```diff
@@ -176,46 +211,94 @@ func (c *FaultRemediationClient) GetStatusChecker() *crstatus.CRStatusChecker {
...
+
 	// Generate CR name
 	crName := fmt.Sprintf("maintenance-%s-%s", healthEvent.NodeName, healthEventID)
 
 	// Skip custom resource creation if dry-run is enabled
 	if len(c.dryRunMode) > 0 {
-		log.Printf("DRY-RUN: Skipping custom resource creation for node %s", healthEvent.NodeName)
+		log.Printf("DRY-RUN: Skipping custom resource creation for node %s using template %s", 
+			healthEvent.NodeName, actionKey)
```
miguelvr: nit: I'm guessing this should use `slog.Info`, not `log.Printf`

fault-remediation/pkg/reconciler/remediation.go
```diff
@@ -176,46 +211,94 @@ func (c *FaultRemediationClient) GetStatusChecker() *crstatus.CRStatusChecker {
...
-
 		return false, ""
 	}
 
-	log.Printf("Creating RebootNode CR for node: %s (UID: %s)", healthEvent.NodeName, node.UID)
-	c.templateData.NodeName = healthEvent.NodeName
-	c.templateData.RecommendedAction = healthEvent.RecommendedAction
-	c.templateData.HealthEventID = healthEventID
+	log.Printf("Creating maintenance CR for node: %s using template %s (UID: %s)", 
+		healthEvent.NodeName, actionKey, node.UID)
```
miguelvr: same here: use slog

fault-remediation/pkg/reconciler/remediation.go
```diff
@@ -176,46 +211,94 @@ func (c *FaultRemediationClient) GetStatusChecker() *crstatus.CRStatusChecker {
...
-		slog.Error("Failed to execute maintenance template", "error", err)
+	if err := selectedTemplate.Execute(&buf, templateData); err != nil {
+		slog.Error("Failed to execute maintenance template", 
+			"template", actionKey,
+			"error", err)
 		return false, ""
 	}
 
-	log.Printf("Generated YAML: %s", buf.String())
+	log.Printf("Generated YAML using template %s: %s", actionKey, buf.String())
```
miguelvr: another one :D

fault-remediation/pkg/reconciler/remediation.go
```diff
@@ -252,16 +336,26 @@ func (c *FaultRemediationClient) CreateMaintenanceResource(
...
+	
 	if err != nil {
 		return c.handleCreateCRError(ctx, err, crName, healthEvent)
 	}
 
 	// Get the actual name of the created CR
 	actualCRName := createdCR.GetName()
-	log.Printf("Created Maintenance CR %s successfully for node %s", actualCRName, healthEvent.NodeName)
+	log.Printf("Created Maintenance CR %s successfully for node %s using template %s", 
+		actualCRName, healthEvent.NodeName, actionKey)
```
miguelvr: and another

fault-remediation/pkg/reconciler/remediation_test.go
```diff
@@ -221,17 +222,30 @@ func TestNewK8sClient(t *testing.T) {
...
+				RemediationActions: map[string]config.MaintenanceResource{
+					"RESTART_BM": {
+						Namespace:             "dgxc-janitor",
+						Version:               "v1alpha1",
+						ApiGroup:              "janitor.dgxc.nvidia.com",
+						Kind:                  "RebootNode",
+						CompleteConditionType: "NodeReady",
+						TemplateFile:          "rebootnode-template.yaml",
+					},
+					"COMPONENT_RESET": {
```
miguelvr: nit: we should use the enum values for these keys

fault-remediation/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: @miguelvr Yes ideally I agree that we should have been perfectly decoupling the CR creation logic in the fault remediation module from the actual "effective" actions they take on the node or the GPU, but in practice we have seen that there are sometimes multiple events emitted from the same node in quick succession which have the same effective remediation action on the node.  For example, 2 different health events, one with `RESTART_BM` and the other with  `COMPONENT_RESET` actions can be emitted a few seconds or minutes apart on the node and both map to rebooting the node in effect. So I think we do need to do deduplication in the FR module based on the effective actions on the node to avoid unnecessary downtime of the node for redundant events. Thus we need to encode the domain knowledge about the equivalence group mapping somewhere in FR. I agree with @zydee3 's suggestion of moving the information about equivalence groups to the CR level instead, I think that will indeed make the deduplication code path much easier and also the configuration of equivalence groups more extensible. Probably adding the `equivalenceGroup` field in the maintenance actions config itself would be the cleanest approach in my opinion. So in the `values.yaml` we can have this for example:

```
maintenance:
  actions:
    COMPONENT_RESET:
      apiGroup: "remediation.example.com"
      version: "v1alpha1"
      kind: "RebootNode"
      equivalenceGroup: "restart"  # Moved from code to config
      templateFile: "nvidia-reboot.yaml"
    RESTART_BM:
      apiGroup: "remediation.example.com"
      version: "v1alpha1"
      kind: "RestartNode"
      equivalenceGroup: "restart"  #  Same group as above
      templateFile: "namespaced-restart.yaml"
```
      
To implement this cleanly, we would also store the name of the action that created the CR in the node annotation. This ensures that even if actions in the same group use different CRDs, FR knows exactly which apigroup/Kind to use for the status check of an existing CR. And then use the `equivalenceGroup` field to deduplicate. What are your thoughts about this?

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
KaivalyaMDabhadkar: This cannot be renamed unfortunately as we have an XID/SXID catalog that we need to conform to: https://docs.nvidia.com/deploy/xid-errors/analyzing-xid-catalog.html#:~:text=YES-,RESTART_BM,-CONTACT_SUPPORT

fault-remediation/pkg/reconciler/remediation.go
KaivalyaMDabhadkar: +1

fault-remediation/pkg/reconciler/reconciler.go
zydee3: I've been thinking more about @miguelvr 's recommendation, and I agree with the approach. The FR module should focus solely on CR creation without handling deduplication logic.

The main issue with FR-level deduplication is that it requires encoding domain knowledge about each CR type and their actual "effective" actions. This creates tight coupling and makes the system harder to maintain as we add new remediation/response types.

Moving deduplication to the controllers makes more sense (to me) because:
1. each controller will understand their specific operational constraints and resource states better than a generalized orchestrator (FR module).
2. some operations need deduplication (node restarts) while others may not (we have some potential use cases for this).
3. as Miguel suggested, FR becomes a simple event -> CR orchestrator, while operational safety stays with domain informed controllers. This is also useful for remaining vendor-neutral.

Ultimately, this approach still provides deduplication where/when needed, but pushes the responsibility downstream to the components that actually understand the runtime.

What are your thoughts on this? Effectively, for this PR, we would just remove deduplication. The original maintenance CR will be effectively deduplicated at the janitor level instead of the FR level.

miguelvr: I think both approaches are valid as long as the "equivalenceGroup" is a user defined string and NVSentinel does not require specific values.

In any case, there can't be a duplicated CR if one already exists with the same name. So the check can simply be if one exists (assuming we don't use generatedName).

I thought about this and imposing the deduplication on the controller/operator will make implementing a component substantially more complex, since de duplicating CRs is not something controllers usually do because CR reconciliation should be independent.

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
zydee3: yeah, I think `templateFile` is kind of ambiguous.  though, `templateRef` could also be misleading because it implies a reference to an external resource (maybe just me though lol). 

since the config map entries get mounted as files and NVSentinel accesses them as filenames, `templateFileName` is more explicit about what this field contains.

fault-remediation/pkg/reconciler/remediation.go
zydee3: good catch!

fault-remediation/pkg/reconciler/reconciler.go
zydee3: It seems I misunderstood the original intentions. Although I see value in deduplicating at the controller level, Miguel's reasoning definitely makes sense. I will implement the cr-level deduplication as originally appealed.

KaivalyaMDabhadkar: > I thought about this and imposing the deduplication on the controller/operator will make implementing a component substantially more complex, since de duplicating CRs is not something controllers usually do because CR reconciliation should be independent.

I agree. If the users can configure the equivalence groups explicitly through the configuration, it would be best to do the deduplication in the fault remediation module itself

distros/kubernetes/nvsentinel/charts/fault-remediation/values.yaml
KaivalyaMDabhadkar: > I would argue that a safe default should work in every installation. However, the existing reboot CRD only works on specific cloud providers.

That's a good point, but I think changing the default remediation action should be part of a separate PR/issue and outside the scope of the current one. We would need to check and discuss good CSP-independent alternatives to setting reboot as the safe default. For the purpose of this PR, we can keep this behavior unchanged for now.

fault-remediation/pkg/reconciler/reconciler.go
natherz97: @zydee3 I want to make sure the approach we take will satisfy the use-case for supporting both GPU resets and reboots. We have these requirements:
- A GPUReset equivalence group is a GPU UUID + node pair because each GPU UUID will need a dedicated equivalence group. A string passed from the config just for "reset" like what we're currently doing for "reboot" wouldn't encode this.
   - We need additional information from the health event, specifically the impacted entities UUID to construct the equivalence group.
- A GPUReset is in the RebootNode group but a RebootNode is not in the GPUReset group. We don't want to execute a GPUReset after a RebootNode completes because the RebootNode will result in a healthy event which uncordons the node.
   - We need to encode relationships between equivalence groups, specifically when a different group can pre-empt another.

Let me know what your thoughts are about these 2 cases. I'm working on implementing this logic as part of the GPU reset support and was adding this inside fault-remediation rather than in the config.

zydee3: Thanks @natherz97 for the feedback. I agree these are real requirements for GPU reset support.

I agree templating equivalence group names would be an improvement, but I would prefer to keep this PR focused and address that refactor in a follow up PR. This PR is already quite complex.

I am also concerned the second point pushes fault remediation beyond orchestration and into policy and scheduling. If we add pre emption rules and cross group relationships here, fault remediation will become the hardest component to extend as we add more actions and operators.

I think the pre emption logic should live close to the execution semantics, in the controller, or in a dedicated module or deployment that owns these rules.

KaivalyaMDabhadkar (dismissed):

KaivalyaMDabhadkar (dismissed):

KaivalyaMDabhadkar (approved):

Issue #390, opened 2025-11-21
[Feature]: Add the ability to dry run health checks


lalitadithya:
### Prerequisites

- [x] I searched existing issues

### Feature Summary

We want the ability to run new health checks in dry mode to allow operators to be able to evaluate the health checks in their environment without impacting their cluster. Currently the only way to dry run a health check is to disable it from the FQ rules, but this will still update the node conditions / events which can result in confusion for the end users, and we'd like to avoid that.

### Problem/Use Case

Ability to run health checks in dry run mode so that will not write/update any resource on the cluster. 

### Proposed Solution

Introduce a dry run / audit flag in the health event. If this flag is set, then the health event should only be persisted to the data store and exported as metrics. All modules should be default drop all health events that have the dry run / audit flag set irrespective of their configuration. 

### Component

Health Monitor

PR #608, opened 2025-12-22
feat: add processingStrategy field in health events structure

tanishagoyal2:
## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

Issue: https://github.com/NVIDIA/NVSentinel/issues/390

## Testing
1. Completed testing on dev cluster 'nvs-dgxc-k8s-oci-lhr-dev3'
2. I have updated image of platform-connector and fault-quarantine, that means all health monitors were running in old mode where no processingStrategy field was applied by health-monitors
3. Inserted XID event 13 on GPU node using kernal log message
4. Saw that platform connector automatically updated processingStrategy to EXECUTE_REMEDIATION and same was stored in db

Screenshot 2026-01-08 at 9 43 18 PM

Platform connector logs

Screenshot 2026-01-08 at 9 44 30 PM

5. Fault quarantine also received the event and started processing the events
6.  Updated the circuit breaker cm to mark the breaker as TRIPPED
7. Inserted some event without processingStrategy

Screenshot 2026-01-08 at 9 54 45 PM

8. Updated the circuit breaker to CLOSED and restarted fault -quarantine
9. It picked up old inserted events without processingStrategy and processed them

Screenshot 2026-01-08 at 9 57 05 PM

## Release Notes

* **New Features**
  * Health events now support processing strategies: execute remediation actions or store for monitoring only.
  * System intelligently filters and routes events based on their assigned strategy.

* **Tests**
  * Added comprehensive test coverage for processing strategy handling across components.

* **Documentation**
  * Updated design documentation covering processing strategy implementation.


platform-connectors/pkg/connectors/kubernetes/k8s_platform_connector_test.go
```diff
@@ -1388,6 +1388,206 @@ func TestUpdateNodeConditions_ErrorHandling(t *testing.T) {
...
+					ComponentClass:     "GPU",
+					RecommendedAction:  protos.RecommendedAction_CONTACT_SUPPORT,
+					Message:            "XID 79: GPU has fallen off the bus",
+					NodeName:           "store-only-test-node",
+					ProcessingStrategy: protos.ProcessingStrategy_EXECUTE_REMEDIATION,
+				},
+			},
+			expectNodeConditions:   true,
+			expectKubernetesEvents: false,
+			expectedConditionCount: 1,
```
lalitadithya: Can we make the test case "strong"? That is, check specifically for the condition that is present in the health event?

platform-connectors/pkg/connectors/kubernetes/k8s_platform_connector_test.go
```diff
@@ -1388,6 +1388,206 @@ func TestUpdateNodeConditions_ErrorHandling(t *testing.T) {
 		})
 	}
 }
+func TestProcessHealthEvents_StoreOnlyStrategy(t *testing.T) {
```
lalitadithya, line 1391: All the test cases seem to have expectedEventCount set to 0, is that expected? Since the health events can be an arrary don't we have to check that if some the health events must generate events, they should?

store-client/pkg/client/postgresql_pipeline_builder.go
```diff
@@ -115,6 +116,21 @@ func (b *PostgreSQLPipelineBuilder) BuildAllHealthEventInsertsPipeline() datasto
...
+// BuildProcessableHealthEventInsertsPipeline creates a pipeline that watches for health event inserts
+// with processingStrategy=EXECUTE_REMEDIATION
+func (b *PostgreSQLPipelineBuilder) BuildProcessableHealthEventInsertsPipeline() datastore.Pipeline {
+	return datastore.ToPipeline(
+		datastore.D(
+			datastore.E("$match", datastore.D(
+				datastore.E("operationType", datastore.D(
+					datastore.E("$in", datastore.A("insert")),
+				)),
+				datastore.E("fullDocument.healthevent.processingstrategy", int32(protos.ProcessingStrategy_EXECUTE_REMEDIATION)),
```
lalitadithya: If we are using fullDocument here, do we need the new column the schema?

tests/platform-connector_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+type PlatformConnectorTestContext struct {
+	NodeName        string
+	ConfigMapBackup []byte
+	TestNamespace   string
+}
+
+func TestPlatformConnectorWithProcessingStrategy(t *testing.T) {
```
lalitadithya: Can we merge this one with the other end to end test? Since we are testing the feature for processing strategy, we can combine the checks for kube objects and not being cordoned, it will save some time in the test case execution

tanishagoyal2: with which test can we combine it? Can you please point me to that file?

tanishagoyal2: Updated tests

store-client/pkg/client/postgresql_pipeline_builder.go
tanishagoyal2: removed column in postrgres schema

platform-connectors/pkg/connectors/kubernetes/k8s_platform_connector_test.go
tanishagoyal2: updated the tests to check expected condition

platform-connectors/pkg/connectors/kubernetes/k8s_platform_connector_test.go
tanishagoyal2: merged platform connector test with fault quarantine test. You can check TestFaultQuarantineWithProcessingStrategy test

data-models/protobufs/health_event.proto
```diff
@@ -29,6 +29,15 @@ message HealthEvents {
   repeated HealthEvent events = 2;
 }
 
+// ProcessingStrategy defines how downstream modules should handle the event.
+// EXECUTE_REMEDIATION: normal behavior; downstream modules may update cluster state.
+// STORE_ONLY: observability-only behavior; event should be persisted/exported but should not modify cluster resources.
```
lalitadithya, line 35: What does UNSPEFICIFIED do? Would it trigger an error in the platform connector?

lalitadithya (commented):
Can you add some details on how are we going to handle backwards compatibility? What about monitors that do not send the field, how will they be processed?

platform-connectors/pkg/connectors/kubernetes/k8s_platform_connector_test.go
tanishagoyal2: I have made changes for this also. instead of checking the count it checking the expected condition and event

expectedConditionType:  "GpuXidError",
			expectedEventType:      "",

data-models/protobufs/health_event.proto
tanishagoyal2: I had to add UNSPECIFIED. Its not going to do anything and won't be used. The problem was enum values should start with 0 and when I was keeping EXECUTE_REMEDIATION as 0 then in postrgres JSON was not storing processingStartegy as JSON serializer ignores 0 or null value to store.

lalitadithya: Sure, but we need to define what that means though else it will be confusing for the users of the API

tanishagoyal2: Yeah, right. I'll update the comment

data-models/protobufs/health_event.proto
```diff
@@ -66,6 +76,7 @@ message HealthEvent {
   string nodeName = 13;
   BehaviourOverrides quarantineOverrides = 14;
   BehaviourOverrides drainOverrides = 15;
+  ProcessingStrategy processingStrategy = 16;
```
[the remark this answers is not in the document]
tanishagoyal2: I don't think we need to add comment for every field we add like drainOverrides and quarantineOverrides also doesn't have comment added. The purpose of this field has been explained in the comment of processingStrategy enum defined proto file

data-models/protobufs/health_event.proto
lalitadithya: This comment says it is not used, but in the design document it says it defaults to execute remediation. Can you please clarify which is the expected behaviour?

tests/fault_quarantine_test.go
lalitadithya, line 1: How are we testing backwards compatibility here specifically:

1. Events that are already present in the DB without the processing strategy
2. Older health monitors that do not specify the field

platform-connectors/pkg/server/platform_connector_server.go
```diff
@@ -54,6 +54,13 @@ func (p *PlatformConnectorServer) HealthEventOccurredV1(ctx context.Context,
 
 	healthEventsReceived.Add(float64(len(he.Events)))
 
+	// Custom monitors that don't set processingStrategy will default to EXECUTE_REMEDIATION.
```
lalitadithya, line 57: Can you please add a unit test for this?

tests/fault_quarantine_test.go
tanishagoyal2: It difficult to test. Because we have made changes in the platform-connector to override the processingStrategy if is UNSPECIFIED to EXECUTE_REMEDIATION. Tilt will always be up with these platform-connector changes.
So, the only way to test it is to open mongodb connection and to insert events directly there (to simulate old events that we will have after upgrade) but we were against of inserting the events to db.

 Can you suggest me if there is another way to do it?

lalitadithya: Won't the test be for a health event to be sent without the field? The platform connector will override it with the correct value, and we can check if the node is cordoned and so on.

tanishagoyal2: That we are already checking with pre existing tests, because currectly all healthevents are sent with no processingStrategy defined (the function NewHealthEvent doesn't have any processingStrategy defined)

lalitadithya: but we have updated `tests/data/fatal-health-event.json` so won't the existing tests also be updated indirectly?

tanishagoyal2: understood, I have to revert those json file changes

lalitadithya (approved):

PR #611, opened 2025-12-22
feat: add event handling strategy changes in gpu health monitor

tanishagoyal2:
## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Testing

1. Tested on dev cluster nvs-dgxc-k8s-oci-lhr-dev3 by installing nvsentinel from main changes
2. Updated gpu-health-monitor image with this branch changes and running gpu-monitor in STORE_ONLY strategy
3. Injected gpu inforom error
4. Event was created with STORE_ONLY strategy

Screenshot 2026-01-12 at 9 03 00 PM

5. Event exporter also exported event with STORE_ONLY strategy
Screenshot 2026-01-12 at 9 06 40 PM

6. Nodes were not cordoned and node condition was not applied

## Summary by CodeRabbit

* **New Features**
  * Add configurable GPU health event processing strategy (EXECUTE_REMEDIATION default or STORE_ONLY) via CLI and deployment values; strategy is embedded in emitted health events.

* **Tests**
  * Updated/added tests to verify propagation of STORE_ONLY and EXECUTE_REMEDIATION across multiple health-event scenarios and connectivity transitions; new end-to-end coverage for store-only path.

* **Chores**
  * Deployment manifests and test harness updated to accept and pass the new strategy flag.


tests/gpu_health_monitor_test.go
```diff
@@ -401,3 +409,113 @@ func TestGPUHealthMonitorDCGMConnectionError(t *testing.T) {
...
+		require.NoError(t, err, "failed to restore GPU health monitor daemon set")
+
+		restConfig := client.RESTConfig()
+
+		podNameVal := ctx.Value(keyGpuHealthMonitorPodName)
+		if podNameVal == nil {
+			t.Log("Skipping teardown: podName not set (setup likely failed early)")
+			return ctx
+		}
+		podName := podNameVal.(string)
```
tanishagoyal2: its not possible that nodename will be null as we are doing its value setup in setup stage. so we dont have to add nil check for it

tests/helpers/kube.go
```diff
@@ -2228,3 +2228,250 @@ func VerifyLogFilesUploaded(ctx context.Context, t *testing.T, c klient.Client,
...
+	})
+	if err != nil {
+		return err
+	}
+
+	t.Logf("Waiting for daemonset %s/%s rollout to complete", NVSentinelNamespace, daemonsetName)
+	waitForDaemonSetRollout(ctx, t, client, daemonsetName)
+
+	t.Logf("Waiting 10 seconds for daemonset pods to start")
+	time.Sleep(10 * time.Second)
```
lalitadithya: Doesn't the daemonset rollout wait already take care of this? Do we need an additional sleep?

tests/gpu_health_monitor_test.go
```diff
@@ -646,3 +653,113 @@ func TestGpuNvlinkWatchSemicolonMessageParsing(t *testing.T) {
...
+
+	feature.Teardown(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		client, err := c.NewClient()
+		require.NoError(t, err, "failed to create kubernetes client")
+
+		nodeName := ctx.Value(keyNodeName).(string)
+
+		err = helpers.RemoveDaemonSetArgs(ctx, t, client, "gpu-health-monitor-dcgm-4.x", "gpu-health-monitor", map[string]string{
+			"--processing-strategy": "EXECUTE_REMEDIATION"})
+		require.NoError(t, err, "failed to restore GPU health monitor daemon set")
```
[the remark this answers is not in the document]
tanishagoyal2: updated the code. Please check

tests/helpers/kube.go
tanishagoyal2: right, I have removed it

tests/gpu_health_monitor_test.go
```diff
@@ -646,3 +654,113 @@ func TestGpuNvlinkWatchSemicolonMessageParsing(t *testing.T) {
...
+
+		testNodeName = gpuHealthMonitorPod.Spec.NodeName
+		gpuHealthMonitorPodName = gpuHealthMonitorPod.Name
+		t.Logf("Using GPU health monitor pod: %s on node: %s", gpuHealthMonitorPodName, testNodeName)
+
+		metadata := helpers.CreateTestMetadata(testNodeName)
+		helpers.InjectMetadata(t, ctx, client, helpers.NVSentinelNamespace, testNodeName, metadata)
+
+		t.Logf("Setting ManagedByNVSentinel=false on node %s", testNodeName)
+		err = helpers.SetNodeManagedByNVSentinel(ctx, client, testNodeName, false)
```
XRFXLP: It doesn't make sense to set this label and then assert the quarantine state. If you set this label then the node will not be touched irrespective of health monitor mode.

tests/gpu_health_monitor_test.go
```diff
@@ -646,3 +654,113 @@ func TestGpuNvlinkWatchSemicolonMessageParsing(t *testing.T) {
...
+		helpers.RestoreDaemonSetArgs(ctx, t, client, GPUHealthMonitorDaemonSetName, GPUHealthMonitorContainerName, originalArgs)
+
+		restConfig := client.RESTConfig()
+
+		podNameVal := ctx.Value(keyGpuHealthMonitorPodName)
+		if podNameVal == nil {
+			t.Log("Skipping teardown: podName not set (setup likely failed early)")
+			return ctx
+		}
+		podName := podNameVal.(string)
```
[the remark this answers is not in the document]
tanishagoyal2: We don't have to check for nil value as during the setup we make sure that these values are not nil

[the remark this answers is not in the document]
tanishagoyal2: Can we resolve this comment?

tests/gpu_health_monitor_test.go
tanishagoyal2: You are right, I have removed this addition of label in here

tests/syslog_health_monitor_test.go
```diff
@@ -447,8 +448,26 @@ func TestSyslogHealthMonitorStoreOnlyStrategy(t *testing.T) {
...
+						condition.Type, condition.Reason, condition.Status, condition.Message)
+
+					return true
+				}
+			}
+
+			t.Logf("Node %s correctly does not have a condition with check name '%s'", nodeName, "SysLogsXIDError")
+
+			return false
+		}, 20*time.Second, helpers.WaitInterval, "node %s should NOT have a condition with check name %s", nodeName, "SysLogsXIDError")
```
tanishagoyal2, line 473: I had to made this changes because syslog monitor process journal messages in every 15seconds of interval and before this change we were immediately checking for node condition not present, by that time syslog monitor was not even publishing event in db so the check was useless.

With this change the timeout set for 20 seconds which gives syslog monitor enough time to process the messages and insert events in db

lalitadithya: Can we add a comment in code explaining this so that we remember why we made this change?

tanishagoyal2: sure

tests/gpu_health_monitor_test.go
```diff
@@ -32,12 +32,20 @@ import (
...
+	dcgmServiceName               = "nvidia-dcgm"
+	dcgmOriginalPort              = 5555
+	dcgmBrokenPort                = 1555
+	GPUHealthMonitorContainerName = "gpu-health-monitor"
+	GPUHealthMonitorDaemonSetName = "gpu-health-monitor-dcgm-4.x"
+)
+
+const (
+	keyGpuHealthMonitorPodName      contextKey = "gpuHealthMonitorPodName"
+	keyOriginalDaemonSet            contextKey = "originalDaemonSet"
```
XRFXLP: Looks like we not using const anywhere, can we remove it?

tests/syslog_health_monitor_test.go
XRFXLP: is the only change 10 second -> 20 second timeout? Do you think adding actual sleep would be better as it matches the sleep in syslog health monitor? Or how about configurable timeout?

tanishagoyal2: Yes, the only change is timeout seconds. We are against of adding sleep so decided to go with adding it this way. We can make it configurable may be later when we have similar usecase for any other test. Currently we needed it at this place only so I felt it will be better to add require.eventually with 20sec instead of making it configurable.

tests/gpu_health_monitor_test.go
tanishagoyal2: removed it

XRFXLP (approved):

PR #640, opened 2025-12-23
WIP - k8s API for healtheventwithstatus model

yavinash007:
## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Introduced HealthEvent and HealthStatus Kubernetes CRDs and rich HealthEvent API (spec/status, enums, entities) for Kubernetes-native event storage and remediation.

* **Tests**
  * Added comprehensive unit tests ensuring bidirectional conversion and field parity between internal models, CRDs, and protobufs.

* **Chores**
  * Added CRD/deepcopy generation support and pinned controller-gen tooling; updated Go module dependencies and Make targets.

* **Documentation**
  * Added ADR detailing the Kubernetes-backed datastore design and operational considerations.


data-models/Makefile
```diff
@@ -37,10 +37,56 @@ DOCKER_EXTRA_ARGS :=
 include ../make/common.mk
 include ../make/go.mk
 
+# Version of controller-gen to use for generating CRD deepcopy, client, etc.
+CONTROLLER_GEN_VERSION := v0.17.2
```
ivelichkovich: v0.20.0 is latest, if we're going to pin we should pickup the newest one

ivelichkovich: also, janitor has a controller-gen as well. It's using latest rather than a pinned version though. Checkout how the main makfile in the root dir is handling versions: https://github.com/NVIDIA/NVSentinel/blob/82e71803c9ea7efade4f715e2d11c453dda0e755/Makefile#L24 

We should probably put controller-gen there with that same pattern since we're using it in multiple modules now

data-models/api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,34 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "data-models.dgxc.nvidia.com", Version: "v1alpha1"}
```
ivelichkovich: I was thinking maybe we do `healthevents.dgxc.nvidia.com` but open to ideas here too. Just because health-events is a little more descriptive than data-models

data-models/api/v1alpha1/healthstatus_types.go
```diff
@@ -0,0 +1,135 @@
...
+	AlreadyDrained   Status = "AlreadyDrained"
+
+	UnQuarantined      Status = "UnQuarantined"
+	Quarantined        Status = "Quarantined"
+	AlreadyQuarantined Status = "AlreadyQuarantined"
+	Cancelled          Status = "Cancelled"
+)
+
+// HealthEventSpec defines the desired state of HealthStatus
+type HealthEventSpec struct {
```
ivelichkovich: We don't want to manage two copies of the same struct just to make it a kubernetes type.

Can we point directly to the objects that already exist? So it would look something like this

```
// +kubebuilder:object:root=true
// +kubebuilder:subresource:status
type HealthEvent struct {
  Spec HealthEvent // this is the existing health event object
  Status HealthEventStatus // this is the existing status object
}
```

data-models/api/v1alpha1/healthstatus_types.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// HealthEventSpec defines the desired state of HealthStatus
+type HealthEventSpec struct {
+	// Unique identifier for the health event
+	// +kubebuilder:validation:Required
+	EventID string `json:"eventID"`
+
+	// Node associated with this health event
+	// +kubebuilder:validation:Required
```
ivelichkovich: This still looks like we're redefining the types, we ideally don't want to maintain two copies of the same types and keep them in sync

data-models/pkg/model/health_event_extentions.go
```diff
@@ -38,8 +38,8 @@ const (
 )
 
 type OperationStatus struct {
-	Status  Status `bson:"status"`
-	Message string `bson:"message,omitempty"`
+	Status  Status `bson:"status" json:"status,omitempty"`
+	Message string `bson:"message,omitempty" json:"message,omitempty"`
 }
 
 type HealthEventStatus struct {
```
ivelichkovich: I was thinking we could just stuff this exact object to be the status for example

data-models/api/v1alpha1/healthstatus_types.go
```diff
@@ -0,0 +1,118 @@
...
+	NodeName string `json:"nodeName"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+type HealthStatus struct {
```
ivelichkovich: so my thought was this struct would look like this:

```
// +kubebuilder:object:root=true
// +kubebuilder:subresource:status
type HealthStatus struct {
  metav1.TypeMeta   `json:",inline"`
  metav1.ObjectMeta `json:"metadata,omitempty"`

  Spec   model.HealthEvent   `json:"spec,omitempty"`
  Status model.HealthEventStatus `json:"status,omitempty"`
}
```

So you would not define your own spec and status objects we would just use the existing ones.

There's some implications for api-versioning going forward if we want to adhere to best practices that we should discuss with the NVIDIA folks if we do it this way but it seems like a clean way to share the object that the other datasources use.

ivelichkovich:
Discussed offline, it'll likely be cleaner to have separate structs than munging the proto and the k8s type. Then we should make sure to have testing to ensure the structs stay in-sync.

One idea is if we can get rid of the `omitempty` (which doesn't seem critical) on the proto types json tags then we can write a unit test to initialize the proto, write it to json, convert it to crd type, write that to json and compare. They should match.

data-models/pkg/conversion/healthevent_spec_contract.go
```diff
@@ -0,0 +1,31 @@
...
+	"RecommendedAction": "recommendedAction",
+	"IsFatal":           "isFatal",
+	"IsHealthy":         "isHealthy",
+}
+
+// ProtoObservationOnlyFields are proto fields that are explicitly
+// NOT part of the CRD Spec contract.
+//
+// These are informational / telemetry / status-only fields.
+var ProtoObservationOnlyFields = map[string]struct{}{
```
ivelichkovich: we probably want some of this data in the CRD, is there any reason why we're skipping all of these? I think EntitiesImpacted for example is important for GPU reset to work.

data-models/pkg/model/health_event_extentions.go
```diff
@@ -38,16 +39,16 @@ const (
...
-	NodeQuarantined        *Status         `bson:"nodequarantined" json:"nodequarantined,omitempty"`
-	UserPodsEvictionStatus OperationStatus `bson:"userpodsevictionstatus" json:"userpodsevictionstatus"`
-	FaultRemediated        *bool           `bson:"faultremediated" json:"faultremediated,omitempty"`
+	NodeQuarantined        *Status         `bson:"nodeQuarantined" json:"nodeQuarantined,omitempty"`
+	UserPodsEvictionStatus OperationStatus `bson:"userpodsEvictionStatus" json:"userPodsEvictionStatus"`
+	FaultRemediated        *bool           `bson:"faultRemediated" json:"faultRemediated,omitempty"`
```
ivelichkovich: lets not change these if it's not needed

data-models/api/v1alpha1/healthstatus_types.go
yavinash007: spoke offline

data-models/pkg/conversion/healthevent_spec_contract.go
yavinash007: Added all, its was the initial version

data-models/pkg/conversion/healthevent_conversion_test.go
```diff
@@ -0,0 +1,334 @@
...
+			for sub := range exportedFieldPathsRecursive(ft, path, nil) {
+				out[sub] = struct{}{}
+			}
+		}
+	}
+
+	return out
+}
+
+func assertRecursiveFieldParity(t *testing.T, leftName string, leftType reflect.Type, rightName string, rightType reflect.Type, ignoreTopLevel map[string]struct{}) {
```
ivelichkovich, line 86: Maybe just add some comments throughout the function to help explain what's going on, just because reflection/recursion can sometimes be a little tough to follow

data-models/Makefile
```diff
@@ -37,10 +37,35 @@ DOCKER_EXTRA_ARGS :=
...
+.PHONY: generate
+generate: ## Generate CRDs and move them to Helm chart directory
+	@echo "Generating CRDs for $(API_DIR)..."
+	@# Install controller-gen if not present
+	@which controller-gen > /dev/null || (echo "Installing controller-gen..." && \
+		go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION))
+	@# Generate deepcopy files for API types
+	go run sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION) object paths=./$(API_DIR)
+	@# Generate CRDs directly into API_DIR
+	go run sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION) \
```
richardcase, line 59: Should we also be generating the RBAC manifests as well?

data-models/api/v1alpha1/healthevent_types.go
```diff
@@ -0,0 +1,165 @@
...
+	StatusFailed         Status = "Failed"
+	StatusSucceeded      Status = "Succeeded"
+	StatusAlreadyDrained Status = "AlreadyDrained"
+
+	StatusUnQuarantined      Status = "UnQuarantined"
+	StatusQuarantined        Status = "Quarantined"
+	StatusAlreadyQuarantined Status = "AlreadyQuarantined"
+	StatusCancelled          Status = "Cancelled"
+)
+
```
richardcase, line 23: Is there any reason why the values for the enum types are different between Status and RecommendedAction/ProcessingStrategy?

data-models/api/v1alpha1/healthevent_types.go
```diff
@@ -0,0 +1,165 @@
...
+const (
+	ProcessingStrategyUnspecified        ProcessingStrategy = "UNSPECIFIED"
+	ProcessingStrategyExecuteRemediation ProcessingStrategy = "EXECUTE_REMEDIATION"
+	ProcessingStrategyStoreOnly          ProcessingStrategy = "STORE_ONLY"
+)
+
+// OperationStatus represents the status of a sub-operation
+type OperationStatus struct {
+	// +kubebuilder:validation:Required
```
richardcase, line 51: It may be worth getting the [Kubernetes API Linter](https://github.com/kubernetes-sigs/kube-api-linter) (KAL) running in the project as early as possible. The reason is that it analyses the custom resources against the Kubernetes API Guidelines and gives a bunch of recommendations.

For example, one of the linters ensures that every field is marked optional or required. It's easier to make the changes now as the API is being defined instead of down the road where it may even require an API version bump.

data-models/api/v1alpha1/healthevent_types.go
```diff
@@ -0,0 +1,165 @@
...
+	Status Status `json:"status,omitempty"`
+
+	Message string `json:"message,omitempty"`
```
richardcase, line 54: Its good to mark strings with min/max length but KAL would highlight this.

data-models/api/v1alpha1/healthevent_types.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Recommended remediation action
+	RecommendedAction RecommendedAction `json:"recommendedAction,omitempty"`
+
+	// Error codes
+	ErrorCode []string `json:"errorCode,omitempty"`
+
+	// Entities impacted
+	EntitiesImpacted []*Entity `json:"entitiesImpacted,omitempty"`
```
richardcase, line 125: For slices, it's good to add the kubebuilder markers to aid SSA. (KAL would highlight this)

data-models/api/v1alpha1/healthevent_types.go
```diff
@@ -0,0 +1,165 @@
...
+	NodeQuarantined *Status `json:"nodeQuarantined,omitempty"`
+
+	// Status of user pod eviction
+	UserPodsEvictionStatus OperationStatus `json:"userPodsEvictionStatus"`
+
+	// Whether the fault has been remediated
+	FaultRemediated *bool `json:"faultRemediated,omitempty"`
+
+	// Timestamp of the last remediation attempt
+	LastRemediationTimestamp *metav1.Time `json:"lastRemediationTimestamp,omitempty"`
```
richardcase, line 70: Some projects use somethign like this:

```go
// Timestamp of the last remediation attempt
LastRemediationTimestamp metav1.Time `json:"lastRemediationTimestamp,omitempty,omitzero"`
```

I believe there are some issues with using the k8s fuzzer with metav1.Time. But this will really only be a problem if have another API version and do roundtrip testing.

data-models/api/v1alpha1/healthevent_types.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Behaviour overrides for drain
+	DrainOverrides *BehaviourOverrides `json:"drainOverrides,omitempty"`
+
+	// ProcessingStrategy
+	ProcessingStrategy ProcessingStrategy `json:"processingStrategy,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
```
richardcase, line 147: It may be useful to also add:
- "kubebuilder:resource:path" - so that users can do something like `kubectl get he"
- Multiple "kubebuilder:printcolumn:name" - again useful to expose info when doing kubectl get

data-models/api/v1alpha1/healthevent_types.go
```diff
@@ -0,0 +1,165 @@
...
+type OperationStatus struct {
+	// +kubebuilder:validation:Required
+	Status Status `json:"status,omitempty"`
+
+	Message string `json:"message,omitempty"`
+}
+
+// HealthEvent Status
+// HealthEventStatus defines the observed state of HealthEvent
+type HealthEventStatus struct {
```
richardcase, line 59: Do you think you'll be adding `Conditions` in the future?

data-models/api/v1alpha1/healthevent_types.go
richardcase: Guessing its because they match the proto?

richardcase (commented):
My biggest piece of feedback is that it would be good to adopt KAL so that it lints the `HealthEvent` against the Kubernetes API Guidelines.....its easier to enable now than later on.

data-models/api/v1alpha1/healthevent_types.go
lalitadithya, line 1: Since we already have the types definitions in the proto, should consider using a code generation plugin such as https://github.com/yandex/protoc-gen-crd to generate the CRDs and possibly the required go types?

distros/kubernetes/nvsentinel/crds/healthevents.dgxc.nvidia.com_healthevents.yaml
lalitadithya, line 1: Should we put the crds in a sub chart of it's own so that we can easily disable it based on a helm value?

data-models/api/v1alpha1/healthevent_types.go
yavinash007: Yeah, thanks for the suggestion @lalitadithya — I tried it out locally and it works well.

One thing to consider is that HealthEventStatus is currently not part of the proto file. We may want to move it into the proto as well, alongside HealthEvent, to keep the model consistent.

I’ve captured this consideration in the design doc too. Let me know what you think:
https://github.com/NVIDIA/NVSentinel/pull/787

lalitadithya:
closing as stale, please feel free to reopen anytime. Thank you!

PR #662, opened 2026-01-06
docs: Add design doc for preflight check

### LINKED ISSUE #658: [Feature]: Support for preflight/prolog checks
### Prerequisites

### Feature Summary

As a part of Kubernetes 1.35, support for gang scheduling was introduced -- https://kubernetes.io/blog/2025/12/29/kubernetes-v1-35-introducing-workload-aware-scheduling/. As a part of this support, the pod spec now has a reference to the gang that the pod is a part of:

```
apiVersion: v1
kind: Pod
metadata:
  name: worker-0
  namespace: some-ns
spec:
  workloadRef:
    name: training-job-workload
    podGroup: workers
```

Using `workloadRef` NVSentinel should be able to inject an init container that can run a variety of checks including but not limited to:
- DCGM diag
- NCCL loopback
- NCCL all-reduce across a gang
- Pluggable third party checks

### Problem/Use Case

As an operator, I want to run preflight checks before a pod starts that allow me to run active health checks to give the user jobs a higher chance of success

### Proposed Solution

TBD

### Component

Health Monitor

XRFXLP:
## Summary

Preview: [ADR-026: Feature — Preflight Checks](https://github.com/XRFXLP/NVSentinel/blob/xrfxlp/658-design/docs/designs/026-preflight-checks.md#adr-026-feature--preflight-checks-via-init-container-injection)

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive Preflight Checks design covering webhook-based injection of per-check init containers and Helm configuration options.
  * Documented gang-aware coordination for multi-pod checks, NCCL topology handling, and rank/bootstrap workflows.
  * Described GPU and network resource detection, DRA integration, namespace/webhook policies, topology mapping, failure handling, events/metrics, and RBAC.


docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,375 @@
...
+
+## Context
+
+GPU failures during training waste compute time. Running diagnostics before the workload starts catches bad GPUs early.
+
+Kubernetes 1.35 introduced `spec.workloadRef` for gang scheduling. Preflight can use `workloadRef` to discover peer pods and run gang-wide checks (NCCL all-reduce).
+
+### Distinction from Health Monitors
+
+NVSentinel already has health monitors (GPU Health Monitor, Syslog Health Monitor) that detect GPU issues. This is different:
```
natherz97, line 11: Did we consider running these tests post-remediation while a node is still cordoned? We're tying our preflight check ability to customer workloads which may not have comprehensive coverage on all GPUs (specifically if their pod is requesting a subset of GPUs on a given node). We could delay fault-quarantine-module uncordoning a node until the corresponding health monitor sends a healthy event + this preflight check runs while the node is still cordoned. We'd have to put more thought into how this would work for multi-node tests but this would be a simpler approach for the single-node dcgm-diag + nccl-loopback checks while also ensuring that we run these preflight checks on any unhealthy GPUs we've identified without modifying customer workloads.

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,375 @@
...
+    env:
+      - name: PREFLIGHT_CHECKS
+        value: "dcgm-diag,nccl-loopback"
+      - name: DCGM_DIAG_LEVEL
+        value: "1"
+      - name: CHECK_TIMEOUT
+        value: "300s"
+      - name: GANG_TIMEOUT
+        value: "600s"
+    resources:
```
natherz97, line 192: Will the NCCL tests require that the initContainer has access to InfiniBand-related device plugins?

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,375 @@
...
+      limits:
+        nvidia.com/gpu: 8  # Copied from main container
```
natherz97: There may be more complex scenarios where other initContainers are requesting GPUs and there might be multiple primary containers all requesting GPUs. My understanding is that the scheduler will ensure a pod is placed on a node that's the max(largest initContainer resource request, sum of all primary containers). Wouldn't we want to take this maximum value to ensure we're capturing all GPUs that the pod may use across either its init or primary containers?

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,375 @@
...
+  enabled: false  # Opt-in
+  
+  checks:
+    - dcgm-diag
+    - nccl-loopback
+    # - nccl-allreduce  # Enable for gang workloads
+  
+  dcgmDiagLevel: 1       # 1 (quick, ~30s) or 2 (medium, ~2-3min)
+  checkTimeout: "300s"   # Per-check timeout
+  gangTimeout: "600s"    # Gang coordination timeout
```
natherz97, line 669: Will the overall timeout for the nccl-allreduce test be 900s? How do we communicate this to customers who may be sensitive with pod start-up times?

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,375 @@
...
+- Platform Connector overrides can modify values
+- Webhook mounts same volumes (GPU, DCGM socket, Platform Connector socket)
+
+### Configuration
+
+Configured at deployment time via Helm values. No per-workload annotations.
+
+### Gang Coordination
+
+For gang-wide checks like `nccl-allreduce`, pods discover peers using `workloadRef`:
```
KaivalyaMDabhadkar: Implementation Note: AFAIK the NCCL all-reduce test on the cluster can disrupt the other workloads already running on the cluster by using the network bandwidth (when the mutating webhook injects an init container into a new workload pod, some older workload pods might already be running on the cluster). We would need to ensure that we run the all-reduce test in a lightweight manner so as to minimize the impact on existing workloads.

docs/designs/026-preflight-checks.md
neerajnv: I think it depends on the test and the hardware where we want to run the test. For hw that does not have nvlink, we likely need IB/Eth plugins. For running all reduce utilising NVLink on H100, we don't need IB.

docs/designs/026-preflight-checks.md
neerajnv: this may not be true depending on the hw. for example on H100, running all reduce utlising nvlink should not disrupt other workloads running on other nodes. In case of GB200, the bandwidth may not be a concern when using nvlink.

docs/designs/026-preflight-checks.md
XRFXLP: Yeah, we've plans for adding post-remediation tests orthogonal to the preflight check, and that would be a different feature.

docs/designs/026-preflight-checks.md
XRFXLP: yeah, we will need that.

docs/designs/026-preflight-checks.md
XRFXLP: Good point, will consider maximum value for device plugin use case, and union for DRA.

docs/designs/026-preflight-checks.md
lalitadithya: Let's file a github issue for this and link it here so that we don't lose track of this.

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,654 @@
...
+
+GPU failures during training waste compute time. Running diagnostics before the workload starts catches bad GPUs early.
+
+Kubernetes 1.35 introduced `spec.workloadRef` for gang scheduling. Preflight can use `workloadRef` to discover peer pods and run gang-wide checks (NCCL all-reduce).
+
+### Distinction from Health Monitors
+
+NVSentinel already has health monitors (GPU Health Monitor, Syslog Health Monitor) that detect GPU issues. This is different:
+
+| | Health Monitors | Preflight Checks |
```
lalitadithya: Can we format the markdown table -- http://markdowntable.com/ -- so that it is easy to read?

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,654 @@
...
+|-|-----------------|------------------|
+| When | Continuous (DaemonSet) | Once at pod start (init container) |
+| Check type | Passive (health watches, syslog parsing) | Active diagnostics (DCGM diag) |
+| Detects | Failures as they occur (XID errors, ECC, thermal) | Latent issues before starting |
+| NCCL tests | No | Yes |
```
lalitadithya: This is invasive tests / active health checks, right? Not just NCCL, it could be dcgm diag, nemotron etc

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,654 @@
...
+      - name: MY_POD_IP
+        valueFrom:
+          fieldRef:
+            fieldPath: status.podIP
+    resources:
+      limits:
+        nvidia.com/gpu: 8          # Max across all containers
+        nvidia.com/mlnxnics: 4     # Max across all containers (if NCCL enabled)
+    securityContext:
+      privileged: true             # DCGM diag
```
lalitadithya: The DCGM diag should connect to the DCGM API, not sure why we need privileged here. From what I understand since we request for the GPU resources, the required libraries should already be mounted in without requiring `privileged: true`

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,654 @@
...
+### Components
+
+```
+preflight/
+├── injector/                       # Webhook (Deployment)
+│   └── pkg/
+│       ├── webhook/                # Admission handler
+│       └── injection/              # Pod mutation + DRA detection
+│
+└── checker/                        # Init container image
```
lalitadithya: Can we keep the preflight checks outside preflight folder much like hoe we have health monitors? Further should we be using a single init container based on the type of check, i.e., one init container for DGCM and a different one for NCCL and so on.

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,654 @@
...
+
+```mermaid
+sequenceDiagram
+    participant W as Webhook
+    participant P0 as Pod 0 Init
+    participant P1 as Pod 1 Init
+    participant API as Kube API
+    participant CM as ConfigMap
+
+    Note over W: First pod in gang
```
lalitadithya: How would we determine which is the "first pod", if the user is using something like a deployment the pods are not going to be ordered.

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,654 @@
...
+    W->>API: Create ConfigMap (expected=2, peers="")
+    
+    P0->>API: Patch ConfigMap: add pod-0:10.0.1.5
```
lalitadithya: This would mean we would have to provide additional RBAC to the job, right? Do we have an option where we don't have to provide additional RBAC to the job? I'm thinking maybe something like a watcher/controller that updates the config map and the pods just have to read from the file system.

I was also thinking we have the Kubernetes lease object, but that would again need additional RBAC to be added to the pod.

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,654 @@
...
+**isFatal determination**:
+- DCGM diag `FAIL` → `isFatal: true`
+- DCGM diag `WARN` → `isFatal: false`
+- NCCL hardware errors (`SYSTEM_ERROR`, `INTERNAL_ERROR`, `REMOTE_ERROR`) → `isFatal: true`
+- NCCL timeout/config errors → `isFatal: false`
+
+### Integration with Node Drainer
+
+Preflight failures quarantine nodes without draining. Rationale:
+- Workload never started → no pods to evict
```
lalitadithya: What if there are other workloads running on the node? Those would have to be evicted right?

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,654 @@
...
+```
+
+**Flow:**
+1. Preflight fails → HealthEvent with `isFatal: true`
+2. Platform Connector applies override → `drainOverrides.skip: true`
+3. Node drainer sees `skip: true` → quarantines node (taint), skips drain
+4. Fault Remediation runs based on `recommendedAction` (EUD, support ticket, etc.)
+5. Remediation succeeds → taint removed → node back in rotation
+
+Gang members on other nodes timeout after `gangTimeout`, fail with `isFatal: false` (coordination failure, not hardware), no quarantine.
```
lalitadithya: If a NCCL test fails on the leader that doesn't indicate that the leader has a problem, right? It is possible that the problem is on the worker and the only way to identify which worker has a problem is to run more deeper diagnostics on all the nodes that failed preflight.

lalitadithya:
Can we add some details on the checks that are made such as the NCCL command that will be run, DCGM command that will be run and so on?

docs/designs/026-preflight-checks.md
XRFXLP: https://github.com/NVIDIA/NVSentinel/issues/695

docs/designs/026-preflight-checks.md
XRFXLP: No, that is kinda borderline. Suppose the GPU on which the preflight was broken but the rest of the GPUs on which existing workload are running fine, then we don't need to drain the existing workload.

docs/designs/026-preflight-checks.md
XRFXLP: We've this natural functional buildup:  dcgmi diag → NCCL loopback → NCCL all‑reduce, right?

docs/designs/026-preflight-checks.md
XRFXLP: We can just sort the pod names and pick up the first one.

docs/designs/026-preflight-checks.md
lalitadithya: Right, then instead of skip, should we consider allow completion instead so that we don't trigger remediation before the node is evacuated?

docs/designs/026-preflight-checks.md
lalitadithya: Correct, but isn't it possible for NCCL all-reduce to fail even though NCCL loopback passes?

XRFXLP: but that would be indicative more of a fabric issue rather than node level issue.

docs/designs/026-preflight-checks.md
XRFXLP: you mean for preflight check somehow changing the drain policy to allow completion? Or just keep AllowCompletion as the mode for all the namespace concerned?

docs/designs/026-preflight-checks.md
lalitadithya: Yes, correct, in that case we can't use any of nodes that is correct to the fabric so we should need to cordon them off

docs/designs/026-preflight-checks.md
lalitadithya: Yeah, we should have some kind of an override so that the breakfix system doesn't start acting on a node that is being used by the customer

XRFXLP: Okay.

But adding node drainer override just for this narrow use case would be an overkill. If we start thinking about possibility of running preflight check on a shared node there would be an obvious interference with the existing workload in the case NCCL tests.

Moreover, do we have training jobs that share a node? IIRC it is usually the inference jobs which share the node.

lalitadithya: We could have single GPU training workloads or smaller models that can run train using 2/4 GPUs. It is not as frequent, I agree, but something to keep in mind while we design the solution so that we aren't caught by surprise when we run into the issue. 

On a shared node we would likely not run NCCL, it will likely be only GPU based ones like DCGM. Now that I think about it more, maybe we should just leave it to what the user has configured for drain. At the end of the day, if we do no preflight the GPU will fail during training and at that time, we will end up using whatever configuration the user has specified, so maybe we don't need any special handling at all.

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,654 @@
+# ADR-026: Feature — Preflight Checks via Init Container Injection
+
+## Context
+
+GPU failures during training waste compute time. Running diagnostics before the workload starts catches bad GPUs early.
+
+Kubernetes 1.35 introduced `spec.workloadRef` for gang scheduling. Preflight can use `workloadRef` to discover peer pods and run gang-wide checks (NCCL all-reduce).
```
ivelichkovich: What if people are using a different mechanism for gang scheduling like volcano/kueue?

XRFXLP: We can keep this mechanism of gang identification pluggable, and only implement native in v1. How does that sound?

XRFXLP: this can be abstracted in `corev1.Pod → { black box } → []PeerInfo`.
For native:
```
Pod → workloadRef → list pods with same workloadRef → []PeerInfo
```
For volcano:
```
Pod → volcano.sh/pod-group annotation → list pods with same annotation → []PeerInfo
```

XRFXLP: Added this in doc, PTAL

docs/designs/026-preflight-checks.md
XRFXLP: That would be the behaviour.

docs/designs/026-preflight-checks.md
XRFXLP: ack, removed special handling.

docs/designs/026-preflight-checks.md
XRFXLP: Added this repsonsbility to PF controller.

docs/designs/026-preflight-checks.md
XRFXLP: done

docs/designs/026-preflight-checks.md
XRFXLP: updated.

docs/designs/026-preflight-checks.md
XRFXLP: handled

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,705 @@
...
+    resources: ["pods"]
+    verbs: ["get", "list", "watch"]
+  # Additional rules based on gang discoverer:
+  # workloadRef: scheduling.k8s.io/workloads (get)
+  # Volcano: scheduling.volcano.sh/podgroups (get)
+  # Kueue: kueue.x-k8s.io/workloads (get)
+```
+
+Controller only touches ConfigMaps with `preflight-` prefix (enforced by code).
+
```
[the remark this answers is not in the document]
XRFXLP: the strategy is code-enforcement as stated.

docs/designs/026-preflight-checks.md
```diff
@@ -27,26 +26,43 @@ Implement a MutatingAdmissionWebhook that injects preflight check init container
...
+│   └── pkg/
+│       ├── webhook/
+│       └── injection/
+│
+├── controller/
+│   └── pkg/
+│       ├── gang/
+│       └── coordination/
+│
+├── dcgm-diag/
```
lalitadithya, line 48: I was thinking the structure would be something like 

```
/preflight-checks/dcgm-diag
/preflight-checks/nccl-loopback
/preflight/controller # if we are using controller runtime we can use the same binary for both controller and webhook
```

docs/designs/026-preflight-checks.md
```diff
@@ -153,94 +162,184 @@ initContainers:
...
+
+Init container remains unprivileged; hostengine performs diagnostics.
+
+### NCCL Loopback
+
+Tests intra-node GPU-to-GPU communication (NVLink/PCIe paths) without network.
+
+**How it works:**
+1. Init container runs `all_reduce_perf` (from nccl-tests) with all allocated GPUs
+2. Command: `all_reduce_perf -b 8 -e 256M -f 2 -g <num_gpus>`
```
lalitadithya, line 263: IIRC we need to run two versions of this command. One that check GPU - NVLink - GPU and one more that does GPU - IB - GPU. The former will validation the connectivity using NVLink the latter will validate GPU to IB connectivity

docs/designs/026-preflight-checks.md
```diff
@@ -153,94 +162,184 @@ initContainers:
...
-    P0->>API: Poll until len(peers) == expected
-    P1->>API: Poll until len(peers) == expected
-    
+    C->>API: Create/Update ConfigMap (expected=2, peers="")
+    C->>API: Update ConfigMap: add pod-0:10.0.1.5
+    C->>API: Update ConfigMap: add pod-1:10.0.1.6
+    C->>API: Update ConfigMap: set nccl_unique_id
+
+    P0->>API: Read ConfigMap until len(peers) == expected
+    P1->>API: Read ConfigMap until len(peers) == expected
```
lalitadithya: Can we instead mount the configmap into the container? Kubelet will automatically update the file as the data in the configmap is updated, there will be some latency, yes, but I think that is something we can tolerate, this way the checks themselves don't need to talk to the kube api at all

docs/designs/026-preflight-checks.md
XRFXLP: but we've nccl loopback for GPU - NVLink - GPU, right?

lalitadithya: Yes, `all_reduce_perf -b 8 -e 256M -f 2 -g <num_gpus>` does the GPU - NVLink - GPU, we need one more version of this for GPU - NIC - GPU

docs/designs/026-preflight-checks.md
ivelichkovich: lgtm

docs/designs/026-preflight-checks.md
ivelichkovich: How long is this delay? My understanding is it could be as long as 2min. For me personally I would rather have my init container talk to apiserver than take a 2 min delay in pod scheduling.

This does bring in extra complexity though because the roles need to be managed and we'd have to watch the configmap so maybe it's a future enhancement.

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,695 @@
...
+## Resource detection and injection
+
+### Detection logic
+
+1. Extended resources (device plugins): check `resources.limits`/`resources.requests` for configured names (e.g. `nvidia.com/gpu`)
+2. DRA: check `spec.resourceClaims`, resolve claim/template, match `deviceClassName` against configured list
+
+### Injected init containers (sketch)
+
+One init container per enabled check:
```
ivelichkovich: I'm assuming these would be prepended to any init containers already part of the pod? Maybe good to state that

docs/designs/026-preflight-checks.md
XRFXLP: I was thinking about the ways we can do nccl all reduce:
1. First option is to give user pods exec permission to run mpirun
2. Second option would be to give controller the exec permission to mpirun

Both seems too permissive.

XRFXLP: Went with third option of using pytorch backed NCCL test without MPI.

docs/designs/026-preflight-checks.md
XRFXLP: done

docs/designs/026-preflight-checks.md
```diff
@@ -0,0 +1,769 @@
...
+Init container remains unprivileged; hostengine performs diagnostics.
+
+### NCCL Loopback
+
+Tests intra-node GPU-to-GPU communication (NVLink/PCIe paths) without network.
+
+**How it works:**
+1. Init container runs `all_reduce_perf` (from nccl-tests) with all allocated GPUs
+2. Command: `all_reduce_perf -b 8 -e 256M -f 2 -g <num_gpus>`
+3. Validates bandwidth meets threshold set in Helm values
```
deesharma24, line 264: how will we know about the  mininum bandwidth? I believe that will change depending on the gpu(h100,gb200 etc)? is that already written somewhere?

deesharma24: as discussed, this will kept minimal to values such as 100gb/s for start and later will finetune depending on expirence.

deesharma24 (approved):
LGTM

lalitadithya (approved):

PR #670, opened 2026-01-09
fix: Retry on failures writing into MongoDB

suket22:
## Summary

Going through the platform connectors, I noticed that we drop health events if there's a failure writing into MongoDB. I think there's value in retrying across a few seconds when writing to MongoDB before we bail. This ensures that if there's a primary failover happening in the underlying data layer or a brief networking interruption, we're resilient to it. Fwiw, I don't think `ring_buffer.go` is an actual implementation of a ring buffer - it seems to be a thin wrapper around workqueues and effectively unbounded so you could still run into memory issues, so that's worth addressing too at a later point. 
* I've added tests that verify that we attempt to rewrite to MongoDB if there's any failure, and drop items after we've exhausted retries so there's no unbounded memory growth. 
* I let the retry mechanism be configurable so unit tests aren't slowed down when simulating multiple failures. 

I think there's value in also making the same changes for the K8s connector since that's [got the same issue IMO](https://github.com/NVIDIA/NVSentinel/blob/main/platform-connectors/pkg/connectors/kubernetes/k8s_connector.go#L104-L106) but I'll wait for feedback in this PR before doing that too. 

In the long term, I think we should do the following - 
*  The platform connectors should actually implement a ring buffer so they're not overwhelmed with a large number of findings reported by an underlying health monitor. 
* We could persist the health events to a WAL in conjunction with the in-memory ring buffer. That way we don't lose in-memory data on a pod restart etc.

I've added in some unit tests, but wasn't able to test this against an actual cluster yet. I'm doing that concurrently but figured I could get feedback in the interim. 

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: Platform Connectors

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Configurable retry/backoff for store writes, exposed via configuration and Helm values.
  * Ring buffer supports configurable retry timing and a rate-limited enqueue path.

* **Bug Fixes**
  * Improved resilience: failed health event writes are retried with backoff up to the configured limit, then dropped to prevent infinite requeues; logs now include retry limits and outcomes.

* **Tests**
  * Added tests covering retry, eventual success after retries, and drop-after-max-retries scenarios.


platform-connectors/pkg/connectors/store/store_connector.go
```diff
@@ -32,6 +32,13 @@ import (
 	"github.com/nvidia/nvsentinel/store-client/pkg/factory"
 )
 
+const (
+	// MaxRetries defines the maximum number of retry attempts for MongoDB writes
+	// before dropping the event. This prevents unbounded memory growth while
+	// providing resilience against transient failures.
+	MaxRetries = 3
```
lalitadithya: Should we expose this as a config value so that we have the option to update it if required?

suket22: Yeah make sense to me. I moved this into the helm values and read it off the ConfigMap, which is how I think we're wiring in all the config values but lmk if that's not right

lalitadithya (approved):

PR #683, opened 2026-01-13
feat: add partial drain support to node-drainer

natherz97:
## Summary

Related design doc for GPU reset: https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/020-nvsentinel-gpu-reset.md.

#### Unhealthy HealthEvents eligible for partial draining

This PR adds partial drain functionality to the node-drainer-module. A partial drain occurs if the recommended action is COMPONENT_RESET and the given unhealthy HealthEvent has an impacted entity which supports partial draining. Currently, the node-drainer will execute partial drains against nodes which have a COMPONENT_RESET recommended action and have a GPU_UUID impacted entity. If the recommended action is COMPONENT_RESET but the given HealthEvent does not include a supported entity for partial drain, we will fail the drain. For all other recommended actions, we will proceed with a full drain. 

To avoid complexity, we choosing to fail the drain and are not falling back to a full drain if a HealthEvent has COMPONENT_RESET but is missing its GPU_UUID. It is the responsibility of each health-monitor to ensure that any COMPONENT_RESET event includes a support impacted entity and any fallback behavior should be implemented in the corresponding health-monitor. This prevents the node-drainer and fault-remediation modules from needing to duplicate this fallback functionality to replace COMPONENT_RESET with RESTART_VM. 

Note that the syslog-health-monitor has logic to fallback to RESTART_VM from COMPONENT_RESET if it cannot lookup the GPU UUID ([ref](https://github.com/NVIDIA/NVSentinel/blob/main/health-monitors/syslog-health-monitor/pkg/xid/xid_handler.go#L227)).

#### How pods are mapped to impacted entities

The metadata-collector adds devices with resource names for entities that support partial drains as a pod annotation to
allow the node-drainer to look up the set of pods which are leveraging that entity. This is configured in
pod_device_annotation.go which provides a map from impacted entities supporting partial drain to a list of device
resource names corresponding to that entity type.
```
	"GPU_UUID": {
		"nvidia.com/gpu",
		"nvidia.com/pgpu",
	},
```
For example, the metadata-collector will add the following device annotation to any pod that is assigned GPU devices:
```
	annotations:
	  dgxc.nvidia.com/devices: '{"devices":{"nvidia.com/gpu":["GPU-123"]}}'
```
When the node-drainer receives an unhealthy HealthEvent for COMPONENT_RESET with GPU_UUID = GPU-123, we will only drain pods with a device annotation that includes GPU-123:
```
	{
	  healthevent: {
	    version: Long('1'),
	    agent: 'syslog-health-monitor',
	    componentclass: 'GPU',
	    checkname: 'SysLogsXIDError',
	    isfatal: true,
	    ishealthy: false,
	    message: 'ROBUST_CHANNEL_CTXSW_TIMEOUT_ERROR',
	    recommendedaction: 2,
	    errorcode: [
	      '48'
	    ],
	    entitiesimpacted: [
	      {
	        entitytype: 'PCI',
	        entityvalue: '0000:03:00'
	      },
	      {
	        entitytype: 'GPU_UUID',
	        entityvalue: 'GPU-123'
	      }
	    ],
...
```
#### When are drains allowed to be skipped?

A full drain can be skipped if a previous full drain completed against the node after it was most-recently quarantined. Additionally, a partial drain can be skipped if either a previous full drain or a previous partial drain against the same impacted entity completed against the node after it was most-recently quarantined.

To discover the set of unhealthy HealthEvents which may allow a drain to be skipped, we will fetch the quarantineHealthEvent annotation for the current node. Each HealthEvent on the annotation includes the object ID which can be used to query the backing DB for an up-to-date full HealthEventWithStatus object which includes its current drain status.

Historically, there has been a difference in how each module looks up historical events (the initial unhealthy event which resulted in quarantining or any subsequent unhealthy events after the quarantine has occurred) while they are processing a new event for the same node.
- fault-quarantine: leverages the quarantineHealthEvent annotation on the node object
- node-drainer: directly queries MongoDB
- fault-remediation: leverages the latestFaultRemediationState annotation on the node object

We have an invariant that the any node which is added to the quarantineHealthEvent annotation will be processed by the node-drainer module. As a result, all historical events that could result in draining being skipped will be included on this annotation. The annotation only stores the HealthEvent and not the full HealthEventWithStatus (which would fall out of date as the event is processed by each module). As a result, we're adding the objectID to the HealthEvent which will allow the node-drainer to efficiently lookup historical events which will allow draining to be skipped. The new logic follows:
- fault-quarantine: leverages the quarantineHealthEvent annotation on the node object
- node-drainer: leverages the quarantineHealthEvent annotation and fetches the latest state from MongoDB for each event
- fault-remediation: leverages the latestFaultRemediationState annotation on the node object

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [X] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Release Notes

* **New Features**
  * Added Partial Drain eviction mode targeting only pods affected by specific GPU faults, reducing unnecessary pod evictions
  * New per-namespace `partialDrainEnabled` configuration option for selective partial drain usage

* **Enhancements**
  * Improved health event tracking with enhanced annotations and audit support
  * Better MongoDB ObjectID handling for event queries

* **Documentation**
  * Updated configuration and eviction mode guidance with partial drain behavior and requirements

* **Tests**
  * Comprehensive integration tests for partial drain scenarios


node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -772,7 +780,7 @@ func (r *Reconciler) executeCustomDrain(ctx context.Context, action *evaluator.D
 	podsToDrain := make(map[string][]string)
 
 	for _, ns := range action.Namespaces {
-		pods, err := r.informers.FindEvictablePodsInNamespaceAndNode(ns, nodeName)
+		pods, err := r.informers.FindEvictablePodsInNamespaceAndNode(ns, nodeName, nil)
```
lalitadithya: Shouldn't this also pass in `partialDrainEntity`? Any reason why we have nil here?

node-drainer/pkg/evaluator/evaluator.go
```diff
@@ -371,70 +406,237 @@ func getEventID(ctx context.Context, database queue.DataStore, nodeName string)
...
+	// and from json rather than try to manually extract our fields.
+	healthEventBytes, err := json.Marshal(healthEventWithStatus.HealthEvent)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to marshal health event for node %s: %w", nodeName, err)
+	}
+
+	var healthEvent protos.HealthEvent
+	if err := json.Unmarshal(healthEventBytes, &healthEvent); err != nil {
+		return nil, nil, fmt.Errorf("failed to unmarshal health event for node %s: %w", nodeName, err)
+	}
```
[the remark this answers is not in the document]
natherz97: The tests in evaluator_integration_test.go covers this round-trip JSON conversion.

node-drainer/pkg/reconciler/reconciler.go
natherz97: Good callout! Regular drains for AllowCompletion, DeleteAfterTimeout, and ImmediateEviction check the pods needing draining in both the Evaluator and in the Reconciler so we're passing the partialDrainEntity to FindEvictablePodsInNamespaceAndNode in both locations. For custom drains, we're only looking up the pods needing draining in the Reconciler and in the Evaluator, we only check the CR status rather than the pod status. 

I had a misunderstanding that the CR for custom draining would need to know if we're executing a partial drain or full drain and the corresponding controller would need to understand this, but it looks like we're passing the set of pods needing drained as input to the CR. I fixed this to pass the partialDrainEntity here so custom draining will also support partial draining.

tests/node_drainer_test.go
```diff
@@ -195,3 +195,169 @@ func TestNodeDrainerEvictionModes(t *testing.T) {
...
+
+		deleteTimeoutPods = helpers.CreatePodsFromTemplate(newCtx, t, client, "data/busybox-pods.yaml", testCtx.NodeName, "delete-timeout-test")
+		deleteTimeoutPodsWithImpactedGPU = helpers.CreatePodsFromTemplate(newCtx, t, client, "data/busybox-pod-with-devices.yaml", testCtx.NodeName, "delete-timeout-test")
+
+		helpers.WaitForPodsRunning(newCtx, t, client, "delete-timeout-test", deleteTimeoutPods)
+		helpers.WaitForPodsRunning(newCtx, t, client, "delete-timeout-test", deleteTimeoutPodsWithImpactedGPU)
+
+		return newCtx
+	})
+	// Since SetupNodeDrainerTest will override the Tilt configmap, we need to ensure that the Helm chart value
```
XRFXLP: nit: should we move this to `feature.Setup` step? Also this is not ensuring but validating, right?

tests/node_drainer_test.go
```diff
@@ -195,3 +195,169 @@ func TestNodeDrainerEvictionModes(t *testing.T) {
...
+	// correctly set the value.
+	feature.Assess("Ensure partial drain enabled in original configmap", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		t.Log("Phase 1: node-drainer configmap has partialDrainEnabled: true")
+		require.Contains(t, string(testCtx.ConfigMapBackup), "partialDrainEnabled = true")
+		return ctx
+	})
+	feature.Assess("Execute partial drain", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
```
XRFXLP, line 230: IIUC: We've two pods here one is runing on bad GPU and other one is not, former is expected to be drained. Can this be tested with Immediate mode for node drain?

natherz97: Yes, that understanding is correct. This e2e test leverages DeleteAfterTimeout only but the reconciler_integration_test.go file has comprehensive tests for all modes where we confirm partial drain functionality works for DeleteAfterTimeout, ImmediateEviction, AllowCompletion: https://github.com/NVIDIA/NVSentinel/blob/3a43a2a41deb0418a233c26804be4fa2dbd2aa6c/node-drainer/pkg/reconciler/reconciler_integration_test.go#L396. 

I'll update the e2e test to cover both DeleteAfterTimeout + ImmediateEviction.

tests/node_drainer_test.go
natherz97: Moved this check to feature.Setup and yes this is only validating. The node-drainer-test overwrites the node-drainer configmap as part of the test. I wanted to make sure that the Helm chart is correctly sets the    partialDrainEnabled = true value in the configmap so I added this validation.

tests/node_drainer_test.go
XRFXLP: No, I mean if we do ImmediateEviction then test would be slightly faster as we don't have to wait 1 minute for DeleteAfterTimeout.

natherz97: Got it, test has been updated to use only ImmediateEviction.

XRFXLP (approved):

PR #669, opened 2026-01-09
feat: add GPU reset support to syslog-health-monitor

natherz97:
## Summary

Related design doc for GPU reset: https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/020-nvsentinel-gpu-reset.md.

This PR adds functionality to the syslog-health-monitor to send unhealthy HealthEvents that will result in a GPU reset as asll we to send healthy HealthEvents after a GPU reset has successfully occurred. In addition to the PCI, we will always add the GPU UUID as an impacted entity if it is available from either dmesg or from the metadata-collector. The COMPONENT_RESET remediation action requires the PCI and GPU UUID are available in the initial unhealthy event we're sending. Additionally, the corresponding healthy event triggered after the COMPONENT_RESET requires the same PCI and GPU UUID impacted entities are included as the initial event. As a result, we will only permit the COMPONENT_RESET action if the GPU UUID was sourced from the metadata-collector to ensure that the same impacted entities can be fetched after a reset occurs. If the GPU UUID does not exist or is sourced from dmesg, we will still include it as an impacted entity but override the remediation action from COMPONENT_RESET to RESTART_VM.

Unhealthy event generation:

1. XID 48 error occurs in syslog which includes the PCI 0000:03:00:
```
Xid (PCI:0000:03:00): 48, pid=91237, name=nv-hostengine, Ch 00000076, errorString CTX SWITCH TIMEOUT, Info 0x3c046
```
2. Using the metadata-collector, look up the corresponding GPU UUID for PCI 0000:03:00 which is GPU-455d8f70-2051-db6c-0430-ffc457bff834
3. Include this PCI and GPU UUID in the list of impacted entities in our unhealthy HealthEvent with the COMPONENT_RESET remediation action.

Healthy event generation:

1. GPU reset occurs in syslog which includes the GPU UUID:
```
GPU reset executed: GPU-455d8f70-2051-db6c-0430-ffc457bff834
```
2. Using the metadata-collector, look up the corresponding PCI for the given GPU UUID.
3. Include this PCI and GPU UUID in the list of impacted entities in our healthy HealthEvent.

Implementation notes:
[...]

PR #725, opened 2026-01-22
feat: add GPU reset support to gpu-health-monitor

natherz97:
## Summary
Related design doc for GPU reset: https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/020-nvsentinel-gpu-reset.md.

This PR adds GPU reset support to the gpu-health-monitor. The corresponding changes to the syslog-health-monitor was previously implemented in https://github.com/NVIDIA/NVSentinel/pull/669. This PR enforces that any unhealthy HealthEvent which includes the COMPONENT_RESET remediation must include the GPU_UUID and PCI as impacted entities. Sending an event with COMPONENT_RESET that is missing the GPU_UUID impacted entity will result in a failed partial drain in node-drainer (as well as a failed remediation in fault-remediation). As a result, we are checking that the GPU_UUID (along with the PCI) can be read from the MetadataReader and are falling back to the RESTART_VM action if it is not present on the event.

Note that entity-specific HealthEvents require an exact match for the set of impacted entities between the initial unhealthy event and the eventual healthy event which clears the event in fault-quarantine. To ensure that there's a consistent view of impacted entities between healthy and unhealthy events, we will only send unhealthy HealthEvents for COMPONENT_RESET which include the GPU index, PCI, and GPU_UUID (and the corresponding HealthyEvent will include all of these as long as there's no failure extracting the PCI or GPU_UUID from the MetadataReader).

## Unhealthy events
If the MetadataReader is working when we send unhealthy events, we will include these impacted entities and support these actions:
- ImpactedEntities: GPU index, GPU_UUID, PCI
- RecommendedAction: any

If the MetadataReader is not working when we send unhealthy events, we will include these impacted entities and support these actions:
- ImpactedEntities: GPU index
- RecommendedAction: COMPONENT_RESET will be overridden to RESTART_VM, otherwise don't modify the original recommended action from dcgmerrorsmapping.csv

## Healthy events
We will send healthy events with whatever impacted entities are available. If the MetadataReader is working when we send healthy events:
- ImpactedEntities: GPU index, GPU_UUID, PCI

If the MetadataReader is not working when we send healthy events:
- ImpactedEntities: GPU index

## Comparing to syslog-health-monitor
Similarities between gpu-health-monitor + syslog-health-monitor:

1. The list of ImpactedEntities should include all available information. While GPUResets can be accomplished without the PCI or GPU index, we do not want to remove entities to simplify our unhealthy/healthy event generation. In summary, the gpu-health-monitor will include whatever is available out of GPU index, GPU_UUID, and PCI and the syslog-health-monitor will include GPU_UUID and PCI in both unhealthy and healthy events.
2. We will not send COMPONENT_RESET events that will result in a failed drain (or failed remediation) if the event is missing the GPU_UUID (and PCI). If this condition is detected, both health-monitors will override the recommended action to RESTART_VM.

Differences between gpu-health-monitor + syslog-health-monitor:
1. The gpu-health-monitor includes an additional impacted entity for the GPU index in healthy + unhealthy events.
2. If the GPU_UUID + PCI are available when an unhealthy event is generated but are not available later when a healthy event needs generated, both health-monitors will fail to clear the condition for COMPONENT_RESETS. However, the gpu-health-monitor will still send a healthy HealthEvent after DCGM reports a healthy condition (even if it only includes the GPU index) since we do not know what the recommended action might be for any previous unhealthy events which may need to be cleared for non-COMPONENT_RESET actions. This is different from the syslog-health-monitor when sending healthy events for COMPONENT_RESET because it generates the event from a GPU reset syslog line. There's no point in sending the HealthEvent when it's missing the GPU_UUID + PCI because we can be certain that it is only targeting a previous COMPONENT_RESET request + the impacted entities won't match.

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [X] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * AWS health monitoring now recommends VM restart when reset guidance references both component and reset.
  * GPU monitoring downgrades component-reset recommendations to VM restart when GPU metadata (PCI/UUID) is missing.

* **Tests**
  * Added test coverage verifying recommendation overrides across multi-GPU scenarios and varying metadata availability, using temporary metadata fixtures.

### DIFF
--- a/health-monitors/gpu-health-monitor/gpu_health_monitor/platform_connector/platform_connector.py
+++ b/health-monitors/gpu-health-monitor/gpu_health_monitor/platform_connector/platform_connector.py
@@ -176,6 +176,8 @@ class PlatformConnectorEventProcessor(dcgmtypes.CallbackInterface):
                                 platformconnector_pb2.Entity(entityType="GPU_UUID", entityValue=gpu_uuid)
                             )
 
+                        entities_impacted_supports_component_reset = pci_address and gpu_uuid
+
                         key = self._build_cache_key(check_name, entity.entityType, entity.entityValue)
                         isFatal = False
                         isHealthy = True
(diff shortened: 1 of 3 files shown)


health-monitors/gpu-health-monitor/gpu_health_monitor/platform_connector/platform_connector.py
```diff
@@ -176,6 +177,9 @@ def health_event_occurred(self, health_details: dict[str, dcgmtypes.HealthDetail
                                 platformconnector_pb2.Entity(entityType="GPU_UUID", entityValue=gpu_uuid)
                             )
 
+                        if pci_address and gpu_uuid:
+                            entities_impacted_supports_component_reset = True
```
lalitadithya: nit:

```suggestion
                            entities_impacted_supports_component_reset = pci_address and gpu_uuid
```

lalitadithya (dismissed):
minor nit otherwise lgtm

health-monitors/gpu-health-monitor/gpu_health_monitor/platform_connector/platform_connector.py
natherz97: Good style suggestion, fixed!

lalitadithya (approved):

PR #737, opened 2026-01-23
Add distributed node locking to prevent concurrent maintenance operations

jtschelling:
## Summary

This PR implements Kubernetes Lease-based distributed locking to prevent concurrent maintenance operations from operating on the same node simultaneously. This critical enhancement ensures safe coordination when multiple RebootNode or TerminateNode CRs target the same node.

### New Package: `pkg/distributedlock`

**File**: `nodelock.go` (~200 lines)
- `NodeLock` interface with `LockNode()` and `CheckUnlock()` methods
- Lease-based implementation using `coordination.k8s.io/v1` Leases
- Automatic cleanup via Kubernetes garbage collection (OwnerReferences)
- Comprehensive error handling with metrics emission

### Controller Integration

**RebootNode Controller** (`pkg/controller/rebootnode_controller.go`):
- Acquire lock before reconciliation begins
- Release lock after CompletionTime is set
- Requeue if lock cannot be acquired

**TerminateNode Controller** (`pkg/controller/terminatenode_controller.go`):
- Same locking pattern as RebootNode
- Ensures node termination operations are serialized

### Metrics

Added lock/unlock action types to `pkg/metrics/metrics.go`:
- `ActionTypeLock` - Lock acquisition failures
- `ActionTypeUnlock` - Lock release failures

### RBAC

Updated both controllers with lease permissions:
```go
// +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;delete
```

## How It Works

1. **Lock Acquisition**: Before reconciling, controller tries to create a Lease with the node name in the `nvsentinel-system` namespace
2. **Lock Hold**: If successful, reconciliation proceeds; the Lease remains until CompletionTime is set
3. **Lock Release**: After CompletionTime is set, the next reconcile deletes the Lease
4. **Automatic Cleanup**: OwnerReferences ensure Leases are deleted if the maintenance CR is deleted

## Concurrency Handling

- **Same Node, Different CRs**: If `RebootNode-A` holds a lock on `node-1`, `RebootNode-B` targeting the same node will requeue until the lock is released
- **Different Nodes**: Operations on different nodes proceed in parallel without interference
- **Cross-Controller**: RebootNode and TerminateNode share the same lock namespace, preventing concurrent operations across controller types

## Testing

- ✅ All 57 tests pass
- ✅ Controller-specific tests verify locking behavior
- ✅ Test fixtures properly set GVK for Lease OwnerReferences
- ✅ Lock acquisition, holding, and release tested
- ✅ Cleanup on CR deletion verified

## Performance Impact

- Minimal: One additional API call per reconcile loop (Lease GET/CREATE/DELETE)
- No polling or watches on Leases
- Efficient requeue mechanism when lock unavailable

## Summary by CodeRabbit

* **New Features**
  * Per-node distributed locking added to maintenance controllers to serialize operations; lock namespace configurable via POD_NAMESPACE; lock/unlock actions emit metrics.
  * Controllers now refresh status after reconciliation and perform unlock checks when work is complete.

* **Tests**
  * Extensive unit and integration tests covering lock acquisition, release, ownership scenarios, error paths, and metrics validation.

* **Chores**
  * RBAC updated to allow lease management; deployment updated to supply POD_NAMESPACE.


janitor/pkg/controller/rebootnode_controller.go
```diff
@@ -292,6 +319,10 @@ func (r *RebootNodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	r.grpcConn = conn
 	r.CSPClient = cspv1alpha1.NewCSPProviderServiceClient(r.grpcConn)
 
+	// Initialize NodeLock for distributed locking across maintenance operations
+	// TODO: Make namespace configurable via environment variable or config
+	r.NodeLock = distributedlock.NewNodeLock(mgr.GetClient(), "nvsentinel-system")
```
natherz97: Should we construct this in main.go so that we can directly reference the namespace from the config? Or should we wait until {{ .Release.Namespace }} is added to the config first?

distros/kubernetes/nvsentinel/charts/janitor/templates/clusterrole.yaml
```diff
@@ -35,6 +35,16 @@ rules:
   - create
   - patch
   - update
+- apiGroups:
```
natherz97, line 38: Are the RBAC resources generated by running ```make manifests```?

jtschelling: no this was by hand. the make manifests target wasnt populating this, i didnt look into it

janitor/pkg/controller/rebootnode_controller.go
jtschelling: i changed this to reference an environment variable i load into the pod with the downward api. for our use case here i dont think theres any need for systems other than janitor to read the lease, so it should be fine for the lease to always reside in the same ns as the janitor pod

natherz97: This is how the fault-quarantine-module maintains the circuit breaker configmap so this sounds reasonable.

distros/kubernetes/nvsentinel/charts/janitor/templates/clusterrole.yaml
natherz97: This is consistent with how NVSentinel RBAC is maintained manually but we might want to automate this for Janitor or remove the controller-runtime RBAC comments if they're not utilized.

natherz97:
LGTM

lalitadithya (approved):

Issue #357, opened 2025-11-17
[Feature]: Connecting to software based components ex. Network/GPU Operator


vasylbruchko:
### Prerequisites

- [x] I searched existing issues

### Feature Summary

Checks to validate software components including Networking/GPU operator are running successfully and take action if they are not

### Problem/Use Case

In a typical K8s deployments, for the node to fully be operational, it is required to run Network Operator which then exposes the NVSwitch components to the GPU Operator. If either of these components are not running successfully, the node will not be available to run additional workloads. Restarting the operators may be the correct action to take depending on the errors of the operators. 

### Proposed Solution

Add ability to monitor the behavior of these and other operators which will dictate if the GPU node is ready to take workloads. 
Add ability to restart the operators. 

### Component

Fault Management

deesharma24:
@vasylbruchko   I believe kubernetes-object-monitor is exactly for this use case only where we can watch for pods in network-operator and gpu-operator namespace and if they are not in running state, a health-event can be published on the node.  Can you take a look at this doc https://github.com/NVIDIA/NVSentinel/blob/main/docs/kubernetes-object-monitor.md and see whether it fits the requirement you have or your usecase is different.

PR #751, opened 2026-01-27
feat : update kubernetes-object-monitor component support of monitoring of pods in gpu-operator and network-operator namespace

deesharma24:
Issue => https://github.com/NVIDIA/NVSentinel/issues/357

Summary:-

This PR is related to kubernetes-object-monitor in order to monitor daemonset pods on the node and if those are unhealthy the  NVSentinel should detect those nodes and publish them as fatal health event and the moment the pods are recovered and are in running state, then the corresponding healthy event is published from the k8s-object-monitor.

1. Add to update the clusterrole to monitor the pods as it was broken during testing. 
2. created a doc to monitor for daemonsetspods which are part of gpu-operator namespace and network-operator namespace.

Example Node condition for gpu-operator pod tracking:-

` gpu-operator-pod-health      True    Mon, 02 Feb 2026 16:57:46 +0530   Mon, 02 Feb 2026 16:40:45 +0530   gpu-operator-pod-healthIsNotHealthy   ErrorCode:GPU_OPERATOR_POD_UNHEALTHY DaemonSet:gpu-operator/test-invalid-image-ds GPU Operator DaemonSet pod has not reached Running state within 30 seconds Recommended Action=CONTACT_SUPPORT;ErrorCode:GPU_OPERATOR_POD_UNHEALTHY DaemonSet:gpu-operator/test-invalid-image-ds-2 GPU Operator DaemonSet pod has not reached Running state within 30 seconds Recommended Action=CONTACT_SUPPORT;`

The above example represents 2 daemonsets pods are not able to run in the expected time and when these pods gets into healthy state, it will look like 

`  gpu-operator-pod-health      False   Mon, 02 Feb 2026 17:10:03 +0530   Mon, 02 Feb 2026 17:10:03 +0530   gpu-operator-pod-healthIsHealthy      No Health Failures`

PR changes:-

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Health events now include the impacted resource (type and namespace/name) and maintain per-resource state for DaemonSet pods (GPU & network operators).
  * Optional short resyncPeriod for faster detection in test runs.

* **Documentation**
  * Added detailed monitoring guide with configuration, policy examples, thresholds, resync guidance, and troubleshooting.

* **Tests**
  * New DaemonSet/Pod failure and recovery tests, expanded test utilities, and more explicit annotation checks.

* **Permissions**
  * RBAC generation refined to skip redundant node entries.


docs/monitoring-critical-operators.md
```diff
@@ -0,0 +1,82 @@
+# Monitoring Critical Operators (pods in gpu-operator & network-operator namespace)
+
+## Overview
+For an NVIDIA GPU cluster to function correctly, critical infrastructure components under **gpu-operator** and **neetwork-operator** namespace must be healthy. If these operators fail, the underlying hardware cannot be utilized effectively.
```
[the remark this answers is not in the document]
deesharma24: corrected

docs/monitoring-critical-operators.md
```diff
@@ -0,0 +1,82 @@
+# Monitoring Critical Operators (pods in gpu-operator & network-operator namespace)
+
+## Overview
+For an NVIDIA GPU cluster to function correctly, critical infrastructure components under **gpu-operator** and **network-operator** namespace must be healthy. If these operators fail, the underlying hardware cannot be utilized effectively.
+
+NVSentinel provides a built-in mechanism to monitor these operators and report health events when their pods are not running correctly.
+
+## Prerequisites
+*   **Kubernetes Object Monitor**: Ensure the `kubernetes-object-monitor` component is enabled in your NVSentinel installation.
```
lalitadithya: How can this be enabled? Can we document the helm values that should be enabled?

docs/monitoring-critical-operators.md
```diff
@@ -0,0 +1,82 @@
...
+  resource:
+    group: ""
+    version: v1
+    kind: Pod
+  predicate:
+    # Trigger event if:
+    # 1. Pod is in gpu-operator namespace
+    # 2. Pod has been assigned to a node (nodeName is set)
+    # 3. Pod is NOT Running AND NOT Succeeded
+    expression: |
```
lalitadithya: There are two types of pods in the namespace -- deployments and daemonsets. In the case of deployments, should we cordon nodes since a deployment failure doesn't indicate a node failure?

docs/monitoring-critical-operators.md
```diff
@@ -0,0 +1,82 @@
...
+    componentClass: Software
+    isFatal: true
+    message: "Network Operator pod is not in Running state"
+    recommendedAction: CONTACT_SUPPORT
+    errorCode:
+      - NETWORK_OPERATOR_UNHEALTHY
+```
+
+## Troubleshooting
+If you receive these alerts:
```
lalitadithya: I don't follow, what alerts are these?

docs/monitoring-critical-operators.md
deesharma24: you meant first we need to enable kubernetes-object-monitor, then only this is supported ???

docs/monitoring-critical-operators.md
deesharma24: Ideally, both of these(deployments/daemonsets) shouldn't indicate a node failure.it needs to be investigated by user whether there is some issue with s/w or node has gone bad. that's why we have contact_support in recommended_action. nodes will be cordoned irrespective of that.Isn't it?

docs/monitoring-critical-operators.md
deesharma24: maybe alert is not the right word. what I meant is when these health-events are published and one needs to investigate it, they can do the following steps. I will change "Alert" to "Investigatory Action"

docs/monitoring-critical-operators.md
lalitadithya: yes, without enabling it the config will not work right?

docs/monitoring-critical-operators.md
lalitadithya: But a deployment can move to different nodes, so would that mean we will keep cordoning and uncording nodes in a cluster?

Daemonsets are typically driver and node level components so failures in them can mean the node is not healthy. If neither deployments nor daemonsets indicate node failures, then we should not be cordoning nodes.

deesharma24: #> But a deployment can move to different nodes, so would that mean we will keep cordoning and uncording nodes in a cluster?
I am thinking from what is our objective. Even if deployment fails on one of these nodes, then it will overall impact the gpu-operator/network-operator readiness. So, we should monitor it. Even if a deployment moves from one place to another, it's not like it will move indefinately and keep on moving. may be one/twice(I agree that we can't gurantee it).even if it is moving indefinately, user still have the right to know there is something wrong with the deployment.

deesharma24: One silly question- Why will the deployment move? Unless , it is stuck in pending stage which we are not monitoring, and if the node crashed and deployments needs to be moved, then also, it will go back to pending stage if it is not able to find correct node.

lalitadithya: I agree we should monitor it, but should we cordon the node though? Say the deployment is failing due to a cloud networking issue, or say the deployment is failing due to a user configuration error then would cordoning the node help the operators? 

> it's not like it will move indefinately and keep on moving. may be one/twice

It depends right, we cordon the node then it moves to a different node and fails, then we cordon that node and then it moves to a different node and fails there and so on.

docs/monitoring-critical-operators.md
deesharma24: updated the doc

docs/monitoring-critical-operators.md
lalitadithya: >  Why will the deployment move?

Deployments can be evicted due to presence of taints, de-scheudlers, a controller that attempting to reconcile it and so on.

deesharma24: should we keep the deployment health event non-fatal. It will not harm the system but user can still monitor it?it only helps.

lalitadithya: But if the GPU operator deployment is crashing then it will not install the driver, so it is actually a fatal error. 

I don't know, I feel like we are bending the current solution to fit a use case.

deesharma24: Should kubernetes-object-monitor have an enhancment if the  criteria(health-event) for which node has been cordoned is not longer holds, then it should uncordon the node. I don't know how it will implemented but there is no harm in exploring? 

For the current case, let's keep it as fatal only. once we have more feedback from the community, then we can see if we need to tweak this or not.

lalitadithya: > Should kubernetes-object-monitor have an enhancment if the criteria(health-event) for which node has been cordoned is not longer holds, then it should uncordon the node. I don't know how it will implemented but there is no harm in exploring?

I think the broader question we need to answer is should NVSentinel monitor errors that don't directly impact nodes. We should have a pro/con analysis and then determine what we think is the best path forward is. 

> For the current case, let's keep it as fatal only. once we have more feedback from the community, then we can see if we need to tweak this or not.

Sure, but then we are right back should we cordon nodes, so I think we have to solve the problem wholistically.

deesharma24: As discussed over call, updated the policies to  only include daemonsets and ignore other types of resources.

docs/monitoring-critical-operators.md
```diff
@@ -0,0 +1,92 @@
...
+        group: ""
+        version: v1
+        kind: Pod
+      predicate:
+        expression: |
+          resource.metadata.namespace == 'network-operator' && 
+          has(resource.spec.nodeName) && resource.spec.nodeName != "" &&
+          resource.status.phase != 'Running' && 
+          resource.status.phase != 'Succeeded' &&
+          resource.status.phase != 'Completed' &&
```
XRFXLP: IIRC we don't have this. PTAL: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

deesharma24: updated that in doc.

health-monitors/kubernetes-object-monitor/pkg/config/types.go
```diff
@@ -23,9 +23,33 @@ type Policy struct {
...
 
+// TrackingLevel defines how resources are tracked for state management
+type TrackingLevel string
+
+const (
+	// TrackingLevelResource tracks by individual resource name (default)
+	// stateKey = policyName/namespace/resourceName
+	TrackingLevelResource TrackingLevel = "resource"
+
+	// TrackingLevelOwner tracks by owner reference (for DaemonSet/ReplicaSet pods)
```
lalitadithya: I don't think we should consider `ReplicaSets` at all. Replicasets can move between nodes and a failure in a replicaset doesn't indicate that we have a node failure.

health-monitors/kubernetes-object-monitor/pkg/config/types.go
```diff
@@ -23,9 +23,33 @@ type Policy struct {
...
+	// stateKey = policyName/namespace/ownerKind/ownerName/nodeName
+	// When a pod is deleted, the node stays cordoned until:
```
lalitadithya: I'm not sure if we want all this additional complexity, can't we simply uncordon the node when the pod is deleted?

health-monitors/kubernetes-object-monitor/pkg/config/types.go
deesharma24: There is cleanup needed wrt comments. I am only monitoring from daemonset point of view and skipping others. will update the comment.

health-monitors/kubernetes-object-monitor/pkg/config/types.go
deesharma24: if the user deletes the pod and k8s-object-monitor uncordons it and it took a while to schedule it again on the same node and in between that job get scheduled in that node, then it will become a problem. then k8s-object-monitor will again cordon it and the cycle continues.

lalitadithya: Hmm... but if the node remains cordoned, AFAIK the daemonset pod will not be scheduled on it unless it has the required tolerations, right? So, this might be a chicken-egg problem. What are the default tolerations for pods scheduled by GPU operator and network operator?

deesharma24: nvidia-driver-daemonset-8lpgf
  - dedicated: Exists
  - nvidia.com/gpu: Exists
  - node.kubernetes.io/not-ready: Exists
  - node.kubernetes.io/unreachable: Exists
  - node.kubernetes.io/disk-pressure: Exists
  - node.kubernetes.io/memory-pressure: Exists
  - node.kubernetes.io/pid-pressure: Exists
  - node.kubernetes.io/unschedulable: Exists


@lalitadithya This is the toleration I am seeing. I  have checked different CSP cluster namely nvcf-dgxc-k8s-oci-nrt-prd7, 
nvcf-dgxc-k8s-aws-use1-prd7,nvcf-dgxc-k8s-gcp-usc1-prd3,  nvcf-dgxc-k8s-forge-az50-prd1, nvcf-dgxc-k8s-oci-ord-prd2

lalitadithya: What about network operator? Also is that the default configuration if we helm install the application or is it something we have added? More broadly, what are the default set of tolerations that are applied to a daemonset pod

deesharma24: network-operator also have toleration for node.kubernetes.io/unschedulable: Exists.

kube-multus-ds-s4cqf
  - dedicated: Exists
  - components.gke.io/gke-managed-components: Exists
  - CriticalAddonsOnly: Exists
  - nvidia.com/gpu: Exists
  - nvidia.com/gpu: Exists
  - node.kubernetes.io/not-ready: Exists
  - node.kubernetes.io/unreachable: Exists
  - node.kubernetes.io/disk-pressure: Exists
  - node.kubernetes.io/memory-pressure: Exists
  - node.kubernetes.io/pid-pressure: Exists
  - node.kubernetes.io/unschedulable: Exists
  - node.kubernetes.io/network-unavailable: Exists
  
 Only one oci cluster nvcf-dgxc-k8s-oci-nrt-prd7 have this network-operator pods.
 
 Here, https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#taints-and-tolerations , it is mentioned that k8s itself added toleration for   - node.kubernetes.io/unschedulable in the daemonset. So, ideally we should be good

lalitadithya: Okay, cool. So, we won't have a chicken-egg problem which is good. 

But I'm still not sold on the complexity that this is introducing. Is the only reason why we need the owner level tracking is for when a pod is deleted, node will get uncordoned?

deesharma24: having an owner of a pod helps in multiple scenarios.
it is also helping in the below cases:-

1. node is cordoned and then user decided to delete the daemonset(for whatever reason), then also tracking key exists as application don't know whether pod will come back or not come back. We can handle it but I think it will not be very clean.
2. user  updates DaemonSet to exclude certain nodes via nodeSelector and then this node is excluded but again k8s-object-monitor doesn't know whether to uncordon or not.


I think in general, when a pod is deleted , it comes with a different name(the last few characters changes). Without tracking via owner, it will be really difficult to keep track of whether it is a replacement pod or a fresh one.

I feel this is making our life simpler rather than complex.

lalitadithya: > node is cordoned and then user decided to delete the daemonset

If the daemonset is deleted, the pod will also be deleted, so we don't need to have any special handling for that case. We can uncordon on pod delete.

> user updates DaemonSet to exclude certain nodes via nodeSelector

In this case too, the pod will be deleted from the node, so we can just uncordon on pod delete. 

> I think in general, when a pod is deleted , it comes with a different name(the last few characters changes). Without tracking via owner, it will be really difficult to keep track of whether it is a replacement pod or a fresh one.

Yeah, the pod does come with a different name, but I'm thinking if we should keep track of it at all. If we consider how metric and alarming would handle this: if a pod is deleted which was originally crashing the alarm would be resolved since the object that caused the failure no longer exists. So, I'm wondering why do we need to do something more complex/different here? Also, if the user really wanted custom logic, they could always build a more complex health monitor, I think, right?

> I feel this is making our life simpler rather than complex.

I don't think so. The code has caching, imports kube's scheduling helpers and has forks in the logic based on object type which I increase the cognitive complexity required to trace though the feature.

deesharma24: > If the daemonset is deleted, the pod will also be deleted, so we don't need to have any special handling for that case. We can uncordon on pod delete.

how will we decide whether a replacement pod will be coming or not. that's where our cycle usecase was coming. without owner tracking, we don't about about fate of last pod.

> Yeah, the pod does come with a different name, but I'm thinking if we should keep track of it at all. If we consider how metric and alarming would handle this: if a pod is deleted which was originally crashing the alarm would be resolved since the object that caused the failure no longer exists. So, I'm wondering why do we need to do something more complex/different here? Also, if the user really wanted custom logic, they could always build a more complex health monitor, I think, right?

as metric and alarming doesn't have to cordon/uncordon the node, so it is easy.Also, one part of this answer is initial discussion whether we were discussing how to prevent a cycle or uncordoning and cordoning if pod is deleted and again comes back. This was the biggest bottleneck for which I had to go for tracking owner.

lalitadithya: > as metric and alarming doesn't have to cordon/uncordon the node

true, but it pages a human which imo is worse. 

> Also, one part of this answer is initial discussion whether we were discussing how to prevent a cycle or uncordoning and cordoning if pod is deleted and again comes back. This was the biggest bottleneck for which I had to go for tracking owner.

Ah, but I think that was in the context of a replicaset/deployment right? Since we dropped support for replicaset/deployment and we are focusing purely on daemonset, I'm not sure if that applies here..

deesharma24: > Ah, but I think that was in the context of a replicaset/deployment right? Since we dropped support for replicaset/deployment and we are focusing purely on daemonset, I'm not sure if that applies here.

it applies here as well because if a pod is deleted by user and comes back again and in between that window, uncordoning can happen as the pod is gone. We still don't know how to differentiate between when a pod is gone vs when it is truely gone(will not comes back).

lalitadithya: sure, agreed. But do we really care? If the user is repeated deleting pods to recover the node, they likely know what they are doing and we should just let them go ahead with that.

deesharma24: > sure, agreed. But do we really care? If the user is repeated deleting pods to recover the node, they likely know what they are doing and we should just let them go ahead with that.

We should care because this is the exact reason we have seen disprenancy in the past wrt dashboards(for cases where someone uncordoned it manually even though the fatal event is stil there with node conditions and annotations). Similar to that, in this case, it will be a 2 step for the user. 

1. First user deleted the pod. 
2. assumes new pod comes back for the same and that new pod has come to healthy.  Then,  user needs to go and remove the node condition, remove the FQ annotation, uncordon it. This is the exact thing we didn't like in the past when NVSentinel used to do that and now we are expecting the same from user. Seems not fair to me . if we can do it, we should fix it.  

In general also, if a pod crashes for some reason, in 99% percent of the case user will need to investigate why it is crashing and what's the reason so the 2 step process pain is there for the user. if we can reduce this pain, why shouldn't we.

Overall, I get that you have the concerns for the code because of the complexity but that's why we have code reviews, Tests pipelines for. In the future, if for some reason it doesn't work as expected and it is causing more harm , then we can always revisit and try to do the best we can to remediate the issue.

lalitadithya: > assumes new pod comes back for the same and that new pod has come to healthy.

I'm not sure why this would we not auto uncordon when the pod is deleted. I'm saying we uncordon as soon as the pod is deleted, we don't track what happens next. We don't have to track what happens next. 

> Overall, I get that you have the concerns for the code because of the complexity but that's why we have code reviews, Tests pipelines for. 

Sure, but also consider how newer features, tests and debugging will become more complex X months/years from now.

neerajnv: Can I suggest we add enough instrumentation to track the working and go with the implementation in the current form so that if there is any failure we have enough information to validate our assumptions and revisit the logic.

lalitadithya: > implementation in the current form

Do you mean without any code changes? Or do you mean with the changes proposed in this PR?

deesharma24: > I'm not sure why this would we not auto uncordon when the pod is deleted. I'm saying we uncordon as soon as the pod is deleted, we don't track what happens next. We don't have to track what happens next.

we can do auto uncordon when pod is deleted.  but, When we initially discussed  this, the idea was we should also be to use this in  internally. now, if we go via this route and we use this internally, then it means, We will get notified of this via alert  and we only have to remediate this(by either deleting pod/restarting node/some other issue investigation) etc. That means, whenever we will remediate it, then it is our responsibility to manually uncordon this, remove annotation, remove node condition.  That will increase our overall effort. isn't it?

neerajnv: > > implementation in the current form
> 
> Do you mean without any code changes? Or do you mean with the changes proposed in this PR?

I mean keeping the current logic but augmented with enough instrumentation: logs, metrics so that if we find our assumptions are not valid, we can change it. Lets also make sure what assumptions we have made are validated through tests.

lalitadithya: I'm still not sure where the manual steps are coming from. If we auto uncordon on pod delete, then if we reboot/restart the pod the nodes will get auto uncordoned right?

There are two possibilities here:
1. The pod gets fixed on it's own it should get uncordoned which the object monitor will do
2. The gets deleted due to reboot/manual operation, it should get uncordoned which the object monitor will do

I feel like I'm missing something here

lalitadithya: > > > implementation in the current form
> > 
> > 
> > Do you mean without any code changes? Or do you mean with the changes proposed in this PR?
> 
> I mean keeping the current logic but augmented with enough instrumentation: logs, metrics so that if we find our assumptions are not valid, we can change it. Lets also make sure what assumptions we have made are validated through tests.

Yeah, that is what I'm suggesting as well. We keep the current simple logic and re-evaluate later based on feedback. We don't have to add additional complexity at the very start. I think we have sufficient logs and metrics for this, but we can always recheck to confirm.

deesharma24: > The pod gets fixed on it's own it should get uncordoned which the object monitor will do

Correct.

> The gets deleted due to reboot/manual operation, it should get uncordoned which the object monitor will do

Correct

what about case when user only deleted the pod and ko-object-monitor uncordon and the new pod come up and it again goes into cordon state? We are okay with this cycle ? and this cycle will be applicable for our dgxc fleet as well.

lalitadithya: Considering that is the same way the alerting works, I think that is fine to start with. We can document it and if we get feedback that the current simplistic approach doesn't work then we can revisit the changes proposed in this PR.

deesharma24: if that's how we want to proceed. Then, shouldn't we keep it simple that if on the gpu node any pod is not up (irrespective of whether it is part of daemonset/replicaset/deployment) we should cordon,  and the movement pod moves or deleted, then we can just uncordon on the node. Why do we need to make it specific to daemonset? or is it like if the pod moves too much because of scheduling issues, then it cordon/uncordon cycle will be too much?

lalitadithya: Hmm... But can you give an example of a deployment/replicaset that when crashing can indicate a GPU hardware issue on the node it is running on? AFAIK the only time a GPU node becomes un-useable if when daemons on the node such as the driver stop working.

deesharma24: fair enough. For now, I am not seeing any deployment/replicaset which is running on a gpu-node and will directly be responsible for the overall gpu-health of this node.We can stick to daemonsets only.

docs/monitoring-critical-operators.md
```diff
@@ -0,0 +1,253 @@
...
+        recommendedAction: CONTACT_SUPPORT
+        errorCode:
+          - NETWORK_OPERATOR_POD_UNHEALTHY
+```
+
+## Detection Logic
+
+The policy triggers when **all** of the following conditions are true:
+
+| Condition | Check |
```
XRFXLP: Can we use http://markdowntable.com/ to format all the markdown tables so it is easier to read in plain text?

distros/kubernetes/nvsentinel/values-tilt.yaml
```diff
@@ -3797,6 +3798,40 @@ kubernetes-object-monitor:
...
+          resource.status.phase != 'Running' && 
+          resource.status.phase != 'Succeeded' &&
+          has(resource.status.startTime) &&
+          now - timestamp(resource.status.startTime) > duration('30s')
+      nodeAssociation:
+        expression: resource.spec.nodeName
+      healthEvent:
+        componentClass: Software
+        isFatal: true
+        message: "GPU Operator DaemonSet pod has not reached Running state within 30 seconds"
```
XRFXLP: Should we do some templating magic to add the ability to add pod name in the message field? thoughts?

deesharma24: it does come up in the node condition

`
 gpu-operator-pod-health      True    Wed, 04 Feb 2026 12:25:26 +0530   Wed, 04 Feb 2026 12:25:26 +0530   gpu-operator-pod-healthIsNotHealthy   ErrorCode:GPU_OPERATOR_POD_UNHEALTHY Pod:gpu-operator/test-stuck-ds-gsbc8 GPU Operator DaemonSet pod has not reached Running state within 30 seconds Recommended Action=CONTACT_SUPPORT;
`

docs/monitoring-critical-operators.md
deesharma24: update the doc. Cursor was useful in formatting :)

health-monitors/kubernetes-object-monitor/pkg/config/types.go
deesharma24: @lalitadithya updated the PR with the modified logic as discussed above. PTAL

tests/kubernetes_object_monitor_test.go
```diff
@@ -319,3 +332,534 @@ func TestKubernetesObjectMonitorWithRuleOverride(t *testing.T) {
...
+			}
+		}
+
+		return ctx
+	})
+
+	testEnv.Test(t, feature.Feature())
+}
+
+// checkPodInAnnotation checks if a pod from a DaemonSet is tracked in the node annotation
```
lalitadithya: The pattern that we try to follow is to keep the helper functions in `tests/helpers` and have only the test case code in this file. Can we follow the same here?

health-monitors/kubernetes-object-monitor/pkg/controller/reconciler.go
```diff
@@ -134,7 +145,32 @@ func (r *ResourceReconciler) cleanupDeletedResource(ctx context.Context, req ctr
...
+
+				r.matchStatesMu.Lock()
+				delete(r.matchStates, stateKey)
+				r.matchStatesMu.Unlock()
+
+				continue
+			}
+
+			resourceInfo := &config.ResourceInfo{
+				Kind:      r.gvk.Kind,
```
lalitadithya, line 166: We should use the full gvk here and not just the kind. It is possible to have same kinds in different group-version

tests/kubernetes_object_monitor_test.go
```diff
@@ -319,3 +332,534 @@ func TestKubernetesObjectMonitorWithRuleOverride(t *testing.T) {
 
 	testEnv.Test(t, feature.Feature())
 }
+
+// daemonSetTestContext holds context for DaemonSet pod health tests
+type daemonSetTestContext struct {
+	NodeName       string
+	DaemonSetName  string
+	DaemonSetName2 string // For multiple DaemonSet tests
```
lalitadithya: Can this be an array instead?

tests/kubernetes_object_monitor_test.go
```diff
@@ -319,3 +332,534 @@ func TestKubernetesObjectMonitorWithRuleOverride(t *testing.T) {
...
+		},
+		RestartPolicy: v1.RestartPolicyAlways,
+		Tolerations: []v1.Toleration{
+			{Operator: v1.TolerationOpExists},
+		},
+	}
+
+	// For failing pods, add an init container that never completes
+	// This keeps the pod in Pending (Init:0/1) state, which triggers the policy
+	// (policy checks resource.status.phase != 'Running')
```
lalitadithya: Can we check if kwok can be used to simulate daemonset crashloopbackoff so that it mirrors production more closely?

deesharma24: I have added the tests for simualating crashloopbackoff for both main container and init containers.  but did it using real time pod (with exit 1 ) in command line. That felt close to real time scenario rather than using kwok because I was using the same approach everywhere else.

tests/kubernetes_object_monitor_test.go
deesharma24: done

health-monitors/kubernetes-object-monitor/pkg/controller/reconciler.go
deesharma24: done

tests/kubernetes_object_monitor_test.go
deesharma24: moved to helper functions as well as modified existing helper functions to reduce LOC.

tests/helpers/kubernetes_object_monitor.go
```diff
@@ -76,3 +85,441 @@ func UpdateKubernetesObjectMonitorConfigMap(ctx context.Context, t *testing.T, c
...
+
+// CreateTestDaemonSetWithUniqueSelector creates a DaemonSet with optional init blocker.
+// When shouldFail=true, uses DaemonSetInitBlocking; otherwise uses DaemonSetHealthy.
+func CreateTestDaemonSetWithUniqueSelector(name, namespace, nodeName string, shouldFail bool) *appsv1.DaemonSet {
+	if shouldFail {
+		return CreateTestDaemonSet(name, namespace, nodeName, DaemonSetInitBlocking)
+	}
+
+	return CreateTestDaemonSet(name, namespace, nodeName, DaemonSetHealthy)
+}
```
lalitadithya: nit: Why do we need these single line wrappers? Why can't the caller pass in the additional arg? These are test helpers anyway right?

deesharma24: fair enough. removed it.

tests/helpers/kubernetes_object_monitor.go
```diff
@@ -76,3 +85,419 @@ func UpdateKubernetesObjectMonitorConfigMap(ctx context.Context, t *testing.T, c
...
+	return waitForCrashLoopBackOff(ctx, t, client, namespace, podNamePrefix, false)
+}
+
+// WaitForInitContainerCrashLoopBackOff waits for a pod's init container to enter
+// CrashLoopBackOff state. Returns the pod details once CrashLoopBackOff is detected.
+func WaitForInitContainerCrashLoopBackOff(
+	ctx context.Context, t *testing.T, client klient.Client, namespace, podNamePrefix string,
+) *v1.Pod {
+	return waitForCrashLoopBackOff(ctx, t, client, namespace, podNamePrefix, true)
+}
```
lalitadithya: Can we take care of these one line helper functions as well like we did with the other ones?

tests/helpers/kubernetes_object_monitor.go
```diff
@@ -76,3 +85,419 @@ func UpdateKubernetesObjectMonitorConfigMap(ctx context.Context, t *testing.T, c
...
+
+// WaitForQuarantineState polls until the node reaches the expected quarantine state or times out.
+// This is more efficient than time.Sleep as it returns as soon as the condition is met.
+func WaitForQuarantineState(
```
XRFXLP: We already have function in helpers/kube.go with signature:
```
WaitForNodesCordonState(
	ctx context.Context, t *testing.T, c klient.Client, nodeNames []string, shouldCordon bool,
)
```

Can we re-use/factor that?

tests/helpers/kubernetes_object_monitor.go
```diff
@@ -76,3 +85,419 @@ func UpdateKubernetesObjectMonitorConfigMap(ctx context.Context, t *testing.T, c
...
+		}
+
+		return node.Spec.Unschedulable == expectCordoned
+	}, timeout, WaitInterval,
+		"node %s did not reach expected cordon state %v within %v", nodeName, expectCordoned, timeout)
+}
+
+// WaitForDaemonSetPodRunning waits for a DaemonSet pod to reach Running state with all
+// containers ready. This is useful after fixing a crashing container to verify recovery.
+func WaitForDaemonSetPodRunning(
```
XRFXLP: Can we use waitForDaemonSetRollout in helpers/kube.go? Could there be erroring out pod that we don't wanna monitor and that's why we need localized check to a node?

tests/helpers/kubernetes_object_monitor.go
```diff
@@ -76,3 +85,419 @@ func UpdateKubernetesObjectMonitorConfigMap(ctx context.Context, t *testing.T, c
...
+		// Transient error - log and retry
+		t.Logf("Note: transient error checking DaemonSet deletion: %v", err)
+
+		return false
+	}, 15*time.Second, 1*time.Second, "timed out waiting for DaemonSet %s/%s to be fully deleted", namespace, name)
+}
+
+// updateDaemonSet is a helper that gets a DaemonSet, applies a modifier function,
+// and updates it.
+func updateDaemonSet(
```
XRFXLP: It looks like we're not waiting for daemonset to rollout after calling this function, can that lead to flakyness in the next tests?

docs/monitoring-critical-operators.md
```diff
@@ -0,0 +1,277 @@
...
+          has(resource.metadata.ownerReferences) &&
+          resource.metadata.ownerReferences.exists(r, r.kind == 'DaemonSet') &&
+          has(resource.spec.nodeName) && resource.spec.nodeName != "" &&
+          has(resource.status.startTime) &&
+          now - timestamp(resource.status.startTime) > duration('30m') &&
+          (
+            (resource.status.phase != 'Running' && resource.status.phase != 'Succeeded') ||
+            (
+              has(resource.status.containerStatuses) &&
+              resource.status.containerStatuses.exists(cs,
```
XRFXLP, line 75: The document above talks above monitoring even the initContainer status but this ready-to-use config misses them, should we add it here?

tests/kubernetes_object_monitor_test.go
```diff
@@ -319,3 +326,460 @@ func TestKubernetesObjectMonitorWithRuleOverride(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// Find a real worker node (non-kwok, non-control-plane)
+		// We need a real node that can actually run pods (not kwok simulated nodes)
+		nodeList := &v1.NodeList{}
+		err = client.Resources().List(ctx, nodeList)
+		require.NoError(t, err)
+
+		var testNodeName string
+		for _, node := range nodeList.Items {
```
XRFXLP: Can we use this helper function from `helpers/kube.go`?
```
// GetRealNodeName returns a real (non-KWOK) worker node name from the cluster.
// Prefers schedulable workers, falls back to unschedulable workers if needed.
func GetRealNodeName(ctx context.Context, c klient.Client) (string, error) {
```

tests/kubernetes_object_monitor_test.go
```diff
@@ -319,3 +326,460 @@ func TestKubernetesObjectMonitorWithRuleOverride(t *testing.T) {
...
+			}
+			// Found a real worker node
+			testNodeName = node.Name
+			break
+		}
+		require.NotEmpty(t, testNodeName, "no real worker node found in cluster")
+		t.Logf("Using test worker node: %s", testNodeName)
+
+		// Ensure gpu-operator namespace exists
+		err = helpers.CreateNamespace(ctx, client, gpuOperatorNamespace)
```
XRFXLP, line 322: But we will always have GPU operator namespace, right?

tests/kubernetes_object_monitor_test.go
```diff
@@ -319,3 +326,460 @@ func TestKubernetesObjectMonitorWithRuleOverride(t *testing.T) {
...
+
+	// Consolidated test: Multi-DaemonSet failure, partial recovery, full recovery
+	feature.Assess("Multi-DaemonSet failures with partial and full recovery", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		client, err := c.NewClient()
+		require.NoError(t, err)
+
+		// --- Phase 1: Create two failing DaemonSets ---
+		t.Log("=== Phase 1: Creating two failing DaemonSets ===")
+		for _, dsName := range testCtx.DaemonSetNames {
+			ds := helpers.CreateTestDaemonSet(dsName, testCtx.Namespace, testCtx.NodeName, helpers.DaemonSetInitBlocking)
```
XRFXLP: nit: Although the helper is named as CreateTestDaemonset but it doesn't actually created it. Should we move the creation logic in the helper function?

distros/kubernetes/nvsentinel/values-tilt.yaml
```diff
@@ -3797,6 +3799,55 @@ kubernetes-object-monitor:
...
+          has(resource.metadata.ownerReferences) &&
+          resource.metadata.ownerReferences.exists(r, r.kind == 'DaemonSet') &&
+          has(resource.spec.nodeName) && resource.spec.nodeName != "" &&
+          has(resource.status.startTime) &&
+          now - timestamp(resource.status.startTime) > duration('30s') &&
+          (
+            (resource.status.phase != 'Running' && resource.status.phase != 'Succeeded') ||
+            (
+              has(resource.status.containerStatuses) &&
+              resource.status.containerStatuses.exists(cs,
```
XRFXLP, line 3834: I'm not sure how the tests that check initContainer status are working as we've not written expression for init container. Is that already being captured in one of the other checks?

tests/helpers/kubernetes_object_monitor.go
```diff
@@ -76,3 +85,419 @@ func UpdateKubernetesObjectMonitorConfigMap(ctx context.Context, t *testing.T, c
...
+				break
+			}
+		}
+	}
+
+	return dsPods, nil
+}
+
+// CleanupDaemonSet deletes a DaemonSet and waits for it and its pods to be fully deleted.
+func CleanupDaemonSet(ctx context.Context, t *testing.T, client klient.Client, namespace, name string) {
```
XRFXLP: Looks like we've added a lot of functions in generic API interactions helpers in  `tests/helpers/kubernetes_object_monitor.go`, should we move them to kube.go and only keep the helpers that are specific to kubernetes_object_monitor to this file?

tests/helpers/kubernetes_object_monitor.go
deesharma24: both are different things. waitForDaemonSetRollout checks for "DaemonSet status object (aggregate counts) on all nodes whereas WaitForDaemonSetPodRunning checks for Individual Pod status on specific node. Hence, it would be simpler to use this function rather than do complex refactoring and control it via additional arguments.

docs/monitoring-critical-operators.md
deesharma24: When init container is in CrashLoopBackOff, pod phase remains Pending. The resource.status.phase != 'Running' && resource.status.phase != 'Succeeded' check catches Pending. So, it is not needed.

tests/kubernetes_object_monitor_test.go
deesharma24: hopefully yes, but I have added this just for the sake of clearity as the test should work on any namespace which we want to monitor. Tomorrow, For some reason, gpu-operator namespace is not there for whatever reason, then also the test should pass because it should be able to run on any namespace

distros/kubernetes/nvsentinel/values-tilt.yaml
deesharma24: same answer as above.

When init container is in CrashLoopBackOff, pod phase remains Pending. The resource.status.phase != 'Running' && resource.status.phase != 'Succeeded' check catches Pending. So, it is not needed.

tests/helpers/kubernetes_object_monitor.go
deesharma24: done

tests/helpers/kubernetes_object_monitor.go
deesharma24: done.

tests/helpers/kubernetes_object_monitor.go
deesharma24: to make it simpler, In the test, I have called the function helpers.WaitForDaemonSetPodRunning(ctx, t, client, testCtx.Namespace, testCtx.DaemonSetNames[0], testCtx.NodeName)
 to check this behavior after updateDaemonSet is called.

tests/kubernetes_object_monitor_test.go
deesharma24: done

tests/kubernetes_object_monitor_test.go
deesharma24: it creates the daemonset spec but yeah the apply is applying on the next line, so we can call it BuildTestDaemonSet and let the create happen in the next line. The actual  creation(build) the daemonset is happening inside the helper function only.

tests/helpers/kubernetes_object_monitor.go
deesharma24: done

tests/kubernetes_object_monitor_test.go
```diff
@@ -319,3 +320,440 @@ func TestKubernetesObjectMonitorWithRuleOverride(t *testing.T) {
...
+			_, isControlPlane := node.Labels["node-role.kubernetes.io/control-plane"]
+			if isControlPlane {
+				continue
+			}
+			// Found a real worker node
+			testNodeName = node.Name
+			break
+		}
+		require.NotEmpty(t, testNodeName, "no real worker node found in cluster")
+		t.Logf("Using test worker node: %s", testNodeName)
```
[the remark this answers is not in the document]
deesharma24: DaemonSets have special scheduling behavior in Kubernetes - they automatically tolerate node.kubernetes.io/unschedulable taint, which means they will schedule on cordoned nodes (nodes with spec.unschedulable: true). So, not needed

XRFXLP (approved):
LGTM

PR #752, opened 2026-01-27
feat: NIC health monitor design

### LINKED ISSUE #642: [Feature]: Infiniband health monitor
### Prerequisites

### Feature Summary

Hey folks, don't know if it is in scope\roadmap for NVSentiel, but would like to start the discussion about infiniband health monitoring.

Infiniband is also a very important part of HPC and it's health and performance directly influence the workloads running on the nodes.

### Problem/Use Case

We are using [gpud](https://github.com/leptonai/gpud) together with node-problem-detector and draino for health-checking, node condition setting and draining and are planning to investigate NVSentiel as a single replacement for these components.

One of the great features that gpud supports is the ability to run [infiniband checks](https://github.com/leptonai/gpud/tree/main/components/accelerator/nvidia/infiniband) and report interfaces that are either down or flap between active<>down states. This allows us to cordon the problematic nodes and evict the workloads from them until the problem is investigated and fixed. 

### Proposed Solution

Introduce a new `infiniband-health-monitor` (integrate gpud healthchecks logic?) for NVSentiel, that will implement ib ports down and flapping cheks.

### Component

New Component

KaivalyaMDabhadkar:
## Summary

[Preview the design
](https://github.com/KaivalyaMDabhadkar/NVSentinel/blob/642-nic-monitor-design/docs/nic-health-monitor/Overview.md)

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive NIC Health Monitor docs suite covering overview, end-to-end architecture, three detection areas (link state, link counters, syslog/driver errors) and centralized correlation; includes severity model, polling/threshold guidance, configuration examples, event schemas, example outputs, remediation and escalation guidance, supported hardware notes, verification steps, and operational runbooks for GPU cluster NIC monitoring.


docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+| **Log Watcher** | Detect driver/firmware failures | `/dev/kmsg` | SQLite3 (`eventstore`) |
+
+### 1.3 Key Capabilities
+
+1. **Deterministic failure detection** based on IBTA specifications and vendor documentation
+2. **Real-time kernel log monitoring** using `gpud/pkg/kmsg` for sub-second driver/firmware failure detection
+3. **Fatal counter monitoring** for counters that guarantee workload failure (e.g., `symbol_error`, `local_link_integrity_errors`, `excessive_buffer_overrun_errors`)
+4. **Link flap detection** using `gpud`'s historical scanning logic (3 cycles of persistent downtime)
+5. **Event deduplication** via `gpud/pkg/kmsg/deduper` to prevent alert storms
+6. **Persistent Event Storage** in SQLite to enable stabilization windows (Sticky Window logic)
```
lalitadithya: I'm not quite sure if we should do that. We have a different mechanism for detecting/correlating errors using the health event analyzer. I'm all for reusing existing solutions, but I think we should also make sure they fit into the design patterns we have established. Thus far, all the health monitors do not perform any aggregation, and they report raw events as is. Aggregation if required is done at a higher level.

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+| **State Monitor** | Detect hard UP/DOWN transitions and fatal counters | `sysfs` state files, counters | SQLite3 (`ibPortsStore`) |
+| **Log Watcher** | Detect driver/firmware failures | `/dev/kmsg` | SQLite3 (`eventstore`) |
+
+### 1.3 Key Capabilities
+
+1. **Deterministic failure detection** based on IBTA specifications and vendor documentation
+2. **Real-time kernel log monitoring** using `gpud/pkg/kmsg` for sub-second driver/firmware failure detection
+3. **Fatal counter monitoring** for counters that guarantee workload failure (e.g., `symbol_error`, `local_link_integrity_errors`, `excessive_buffer_overrun_errors`)
+4. **Link flap detection** using `gpud`'s historical scanning logic (3 cycles of persistent downtime)
+5. **Event deduplication** via `gpud/pkg/kmsg/deduper` to prevent alert storms
```
lalitadithya: When would we have alert storms, do we have some documentation explaining the scenarios when this would happen?

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+6. **Persistent Event Storage** in SQLite to enable stabilization windows (Sticky Window logic)
+
+### 1.4 Fatal-Only Model
+
+The monitor raises events **only for fatal conditions** that will cause workload failure:
```
lalitadithya: Since we have a `isFatal` in the health event, so I don't think we should limit the monitors to raise events only for fatal conditions.

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+
+#### 2.4.1 The `mlx5_core cmd_exec timeout`
+
+This is a severe driver-level error indicating **Driver/Firmware Interface failure**. ([Reference: RHEL mlx5_core issues](https://access.redhat.com/solutions/6955682))
+
+**Mechanism:**
+1. `mlx5_core` driver sends command to NIC firmware via mailbox
+2. Driver waits for firmware to toggle "Ownership bit" indicating completion
+3. If firmware has crashed/hung, bit never toggles
+4. Driver's watchdog expires --> logs `cmd_exec timeout` to `dmesg`
```
lalitadithya: Can you put the full log line here along with the regex we would be using for detecting them?

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+After `retry_cnt` attempts (default: 7), the NIC tears down the connection and the application receives `IBV_WC_RETRY_EXC_ERR`.
+
+**Implications:**
+- Confirms **Logical Link is broken** even if physical link is UP
+- Often indicates "Silent Drop" or **Black Hole** in the fabric
+- Local symptom of a **remote** problem
+
+**What This Monitor CAN Detect**: The `local_ack_timeout_err` and `req_transport_retries_exceeded` (native IB) hardware counters track these retry events at the NIC level. Rising counter values indicate transport-layer problems even if we can't see the application error.
+
+**Correlation**: Use with [ibdiagnet](https://docs.nvidia.com/networking/display/ibdiagnet-infiniband-fabric-diagnostic-tool-user-manual-v2-21.21.pdf) to determine if issue is local (NIC) or remote (Switch/Fabric).
```
lalitadithya: Can you add the exact commands we can run and what output we can get?

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+|------------|-----------------|--------|
+| **Soft Errors** | Probabilistic, recoverable via FEC/retries | Performance degradation, workload continues |
+| **Hard Errors** | Deterministic, exceed recovery capacity | Application failure **guaranteed** |
+
+The boundary between soft and hard errors is defined by:
+1. **Counter thresholds** that indicate recovery mechanism exhaustion
+2. **Rate of change** that exceeds retry bandwidth
+3. **Specific counter types** that indicate fundamental violation of the lossless contract
+
+#### 2.5.2 The 10⁻¹² BER Threshold
```
lalitadithya: nit: this is a bit easier to read, please update in other places as well 

```suggestion
#### 2.5.2 The 10E-12 BER Threshold
```

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+Result: QP transitions to ERROR state --> Application crashes
+```
+
+---
+
+## 3. Architecture Overview
+
+### 3.1 Design Rationale: Shared gpud Infrastructure
+
+The monitor is built upon the `gpud` architectural framework, which utilizes a **decentralized correlation model**. Instead of a monolithic "Correlation Engine," health signals are correlated directly within the monitoring component to minimize latency and complexity.
```
lalitadithya: > health signals are correlated directly within the monitoring component to minimize latency and complexity

I understand why GPUd does this, but I'm not sure if this will fit well with the NVSentinel architecture, can we have a pros/cons with both approaches in the document, please?

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+│                             │                         │                          │
+│                             │ --> Platform Connector   │                          │
+│                             └─────────────────────────┘                          │
+│                                                                                  │
+└──────────────────────────────────────────────────────────────────────────────────┘
+```
+
+### 3.6 Why Log Watcher is Essential
+
+The Log Watcher detects failures that other monitors miss:
```
lalitadithya: Would the log watcher be a part of the syslog health monitor?

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+└── /dev/kmsg --> Kernel ring buffer (real-time streaming)
+
+Pattern matching for (all FATAL):
+├── mlx5_core cmd_exec timeout    --> Firmware communication failure
+├── mlx5_core health poll failed  --> NIC health check failed
+├── mlx5_core unrecoverable       --> Hardware in error state
+├── PCIe fatal error              --> PCIe link broken
+├── module absent                 --> Transceiver removed
+├── High Temperature              --> Thermal throttling
+└── pci_power_insufficient        --> Power issue
```
lalitadithya: These seem to be very generic log pattern matches, is that expected? These feel like they can match almost anything

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+
+Deduplication:
+└── Same message within 60s window --> Suppressed (prevents alert storms)
```
lalitadithya: I don't think we need this, in our architecture we don't alert on these things, so it is perfectly valid to capture every error message, we do that for XIDs and we tend to get bursts of those.

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+
+    // Internal fields (for event generation logic, not sent to protobuf)
+    NicType            NicType            // Infiniband, Ethernet, RoCE
+    LinkLayer          string             // "InfiniBand" or "Ethernet"
+}
+
+// Entity represents an impacted hardware entity
+type Entity struct {
+    EntityType  string  // e.g., "NIC"
+    EntityValue string  // e.g., "mlx5_0", "mlx5_0_port1"
```
lalitadithya: I feel like NIC and ports are two different things, ideally, they should be different entities?

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+}
+
+type RecommendedAction int32
```
lalitadithya: If we are not making changes to this, let's not include it in the design document

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+    return nil
+}
+```
+
+**Configuration:**
+```ini
+[StateMonitoring]
+# Expected link speed in Gb/s. Links negotiating below this are FATAL.
+# Common values: 100, 200, 400 (NDR)
+TargetLinkSpeedGbps = 400
```
lalitadithya: This seems to mean we will need per cluster configuration, are there ways to detect the expected speed by looking at the hardware perhaps?

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+        NodeName:           os.Getenv("NODE_NAME"),
+        RecommendedAction:  RecommendedAction_NONE,
+        EntitiesImpacted:   []Entity{{EntityType: "NIC", EntityValue: deviceName}},
+        NicType:            Infiniband,
+        LinkLayer:          "InfiniBand",
+    }
+}
+```
+
+### 5.4 Device Discovery and Parsing
```
lalitadithya: Won't this also pick up NICs that are not configured by network operator? I thought the last time we did this, we ran into a lot of false positives because of this.

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+
+The monitor reuses the historical scanning logic from `gpud` to distinguish between persistent port drops and transient flapping.
+
+**Mechanism**:
+1. Periodic snapshots of port state are stored in the SQLite `ibPortsStore`.
+2. The scanner periodically reviews the last 10 minutes of history for each device/port.
+
+#### 5.5.1 Port Drop Detection (`findDrops`)
+An InfiniBand port is marked as "Dropped" if:
+- The state has been `DOWN` for at least 4 minutes.
```
lalitadithya: How did we arrive at the 4 minutes number?

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+The monitor identifies RoCE devices by checking if the `link_layer` file contains "Ethernet".
+
+### 5.7 NIC Vendor Detection and Mode Selection
+
+The monitor supports multiple NIC vendors. Vendor detection determines the monitoring approach:
+
+| Vendor                 | Detection                              | Counter Monitoring   | Fatal Detection                |
+|------------------------|----------------------------------------|----------------------|--------------------------------|
+| **Mellanox (IB/RoCE)** | Device name `mlx5_*` or driver symlink | Yes - fatal counters | Counters + State + Kernel Logs |
+| **AWS EFA**            | Device name `rdmap*s*`                 | **No**               | State + Kernel Logs only       |
```
lalitadithya: EFA is not in scope, let's remove it from the document. let's only focus on Mellanox

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+
+| PCI Check Result            | Severity | Action Required                         |
+|-----------------------------|----------|-----------------------------------------|
+| Device not in PCI bus       | Warning  | Reload driver (`modprobe mlx5_core`)    |
+| Config space returns `0xFF` | Fatal    | Re-seat card, check power, consider RMA |
+| PCI read error              | Fatal    | Inspect PCIe slot for physical damage   |
+
+---
+
+## 6. Ethernet Monitoring Specification
```
lalitadithya: I think this is out of scope for this design, let's focus purely on Mellanox RDMA devices

docs/nic-health-monitor.md
```diff
@@ -0,0 +1,1616 @@
...
+#                  - Uses: hw_counters/ only (no counters/ directory!)
+#                  - Devices: rdmapXXXsX naming
+#                  - Counter units: bytes (not 4-byte words)
+#
+# Field-validated:
+#   - Mellanox RoCE:  ConnectX-6, 100 Gb/s, 34 devices (OCI)
+#   - Mellanox IB:    ConnectX-6 HDR, 200 Gb/s, 9 devices (on-prem)
+#   - AWS EFA:        100 Gb/s, 32 devices (AWS p4d/p5 instances)
+#
+NICVendorMode = auto
```
lalitadithya: Let's not complicate things, let's only keep auto mode and drop the other configurations

KaivalyaMDabhadkar: Agreed. I will simplify the NICVendorMode to only support auto mode

docs/nic-health-monitor.md
KaivalyaMDabhadkar: I had added EFA in the doc just to show case how we can extend the monitor to include EFA monitoring in the future. We can choose to exclude EFA for now and add it in later, but as mentioned below, EFA monitoring is basically the same as ethernet monitoring implementation wise (both use just sysfs montoring and low watching and no counter stuff). I think it will be simpler to do the implementation in the initial development of the NIC monitor itself rather than pushing this can down the road

docs/nic-health-monitor.md
KaivalyaMDabhadkar: I think we can include it in as it requires minimal changes in the implementation (similar to EFA). The ethernet monitoring is basically just our older version of NIC health monitor which checks for state changes from sysfs plus some signals from log-watcher. Since we are already doing the sysfs monitoring for IB devices it would be simpler to implement the ethernet monitoring as well in the initial implementation itself rather than spending effort in extending the code later. What do you think?

docs/nic-health-monitor.md
KaivalyaMDabhadkar: Noted. I'll update the BER notation

docs/nic-health-monitor.md
KaivalyaMDabhadkar: Agreed. I will remove these

docs/nic-health-monitor.md
KaivalyaMDabhadkar: This is taken from gpud: https://github.com/leptonai/gpud/blob/main/components/accelerator/nvidia/infiniband/store/scan_drops.go#L14
I will clarify this further in the doc

docs/nic-health-monitor.md
KaivalyaMDabhadkar: That makes sense, good point. I will add an entity type for NIC port separately

docs/nic-health-monitor.md
KaivalyaMDabhadkar: These logs are also used by GPUd to correlate state changes (detected via sysfs) with the kmsg logs to give more context as to why that state change occurred: https://github.com/leptonai/gpud/blob/main/components/accelerator/nvidia/infiniband/kmsg_matcher.go#L14

But if we are planning to do the syslog monitoring separately in the syslog monitor rather than nic monitor, then I think doing this would be unnecessary. Need to circle back on this

docs/nic-health-monitor.md
KaivalyaMDabhadkar: Yes, good point. Missed adding some actual log examples here. Will update this

docs/nic-health-monitor.md
KaivalyaMDabhadkar: Good point, will update this in the doc

docs/nic-health-monitor/link-counter-detection.md
```diff
@@ -0,0 +1,1030 @@
...
+
+---
+
+## 5. Counter Reading and Parsing
+
+### 5.1 Mellanox Counter Reading
+
+For Mellanox devices (IB and RoCE), the monitor reads:
+1.  **Standard Counters**: `/sys/class/infiniband/<dev>/ports/1/counters/`
+    *   Fatal counters: `symbol_error`, `link_downed`, `local_link_integrity_errors`, `excessive_buffer_overrun_errors`, `req_transport_retries_exceeded` (Native IB)
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Above 120/hr it is fatal. But between 10-120/hr it is non-fatal

docs/nic-health-monitor/Overview.md
```diff
@@ -0,0 +1,286 @@
...
+> **Note**: All counter thresholds and severity levels are configurable. See [Link Counter Detection](./link-counter-detection.md#10-configuration) for customization options.
+
+### Syslog Detection (Fatal & Warning)
+
+| Source          | Conditions                                                                         | Severity    | Purpose                               |
+|-----------------|------------------------------------------------------------------------------------|-------------|---------------------------------------|
+| **Log Watcher** | `mlx5_core.*cmd_exec timeout`, `health poll failed`, `unrecoverable`, `PCIe Fatal` | **Fatal**   | Deterministic hardware/driver failure |
+| **Log Watcher** | `insufficient power`, `module absent`, `ACCESS_REG failed`                         | **Warning** | Diagnostic context for correlation    |
+
+> **Design Note**: Deterministically fatal events in logs trigger `REPLACE_VM`. Diagnostic logs remain as `Warning` and do not directly trigger remediation.
```
lalitadithya: > and do not directly trigger remediation.

meaning they will be published as non-fatal events? If yes, can we mention it that way so that it is clear?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+│  │  │  • /sys/bus/pci/devices/           │                                  │  │
+│  │  │                                    │                                  │  │
+│  │  │  CHECKS:                           │                                  │  │
+│  │  │  • InfiniBandStateCheck            │                                  │  │
+│  │  │  • EthernetStateCheck              │                                  │  │
+│  │  │                                    │                                  │  │
+│  │  │  BEHAVIOR:                         │                                  │  │
+│  │  │  • Reports RAW state events        │                                  │  │
+│  │  │  • NO aggregation                  │                                  │  │
+│  │  │  • NO local persistence            │                                  │  │
```
lalitadithya: I'm not sure if we have to mention this, I understand that the previous iteration has persistence, but now this we have removed it, to someone who is reading this the first time it looks odd as why we would call this out.

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+
+```go
+const (
+    // Port logical states
+    IBStateDown    = "1: DOWN"      // No connectivity
+    IBStateInit    = "2: INIT"      // Initializing (problematic if stuck >30s)
+    IBStateArmed   = "3: ARMED"     // Armed but not active (check SM)
+    IBStateActive  = "4: ACTIVE"    // Normal operational state
+
+    // Port physical states
```
lalitadithya, line 234: Can we add a link to where we got these states from?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+
+### 3.4 State-Based Event Generation Algorithm
+
+**Port State Evaluation Steps:**
+
+1. **Read port state** from `/sys/class/infiniband/<dev>/ports/<port>/state`
+2. **Evaluate logical state:**
+   - If `state = DOWN` → Mark as **FATAL**, message: "Port DOWN - no connectivity"
+   - If `state = INIT` → Warning, message: "Port stuck in INIT - check Subnet Manager"
+   - If `state = ARMED` → Warning, message: "Port ARMED but not ACTIVE - check fabric manager"
```
lalitadithya: Why are these warning, it looks like someone has to take some action on these right?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+   - If `state = ACTIVE` → Healthy
+
+3. **Read physical state** from `/sys/class/infiniband/<dev>/ports/<port>/phys_state`
+4. **Evaluate physical state:**
+   - If `phys_state = Disabled` → Mark as **FATAL**, message: "Physical state DISABLED"
+   - If `phys_state = Polling` → Warning, message: "Link training in progress"
```
lalitadithya: Why is this warning? In this state jobs can't run right?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+- The link "works" but at dramatically reduced throughput
+
+### 4.2 Speed Degradation Detection Diagram
+
+```
+┌─────────────────────────────────────────────────────────────────────────────────┐
+│                      LINK SPEED DEGRADATION DETECTION                           │
+├─────────────────────────────────────────────────────────────────────────────────┤
+│                                                                                  │
+│  EXPECTED STATE (H100 Cluster):                                                  │
```
lalitadithya: How do we detect the expected state? Do we need some per cluster configuration?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+│  DEGRADED STATE (Same H100 - Bad Cable):                                         │
+│  ┌─────────────────────────────────────────────────────────────────────┐        │
+│  │  NIC: mlx5_0                                                         │        │
+│  │  State: ACTIVE ✓     PhysState: LinkUp ✓     Rate: 200 Gb/s ✗       │        │
+│  │                                              └── 50% capacity loss!  │        │
+│  └─────────────────────────────────────────────────────────────────────┘        │
+│                                                                                  │
+│  ═══════════════════════════════════════════════════════════════════════════    │
+│                                                                                  │
+│  IMPACT ON NCCL AllReduce (Ring Topology):                                       │
```
lalitadithya: NCCL also has tree topology AFAIK, would the impact be the same in that too?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+| H100        | 8              | 400 Gb/s (NDR) | [DGX H100 User Guide](https://docs.nvidia.com/dgx/dgxh100-user-guide/introduction-to-dgxh100.html) |
+| H200        | 8              | 400 Gb/s (NDR) | NVIDIA Documentation                                                                               |
+| B200        | 8              | 400 Gb/s       | [DGX B200 User Guide](https://docs.nvidia.com/dgx/dgxb200-user-guide/introduction-to-dgxb200.html) |
+| GB200       | 8              | 400 Gb/s       | [GB200 NVL2](https://www.nvidia.com/en-us/data-center/gb200-nvl2/)                                 |
+
+### 4.4 Speed Degradation Detection Algorithm
+
+**GPU-to-Rate Lookup Steps:**
+
+1. **Get GPU product name** via NVML or node labels (e.g., `nvidia.com/gpu.product`)
```
lalitadithya: Using node labels would mean that the component will have to be kube aware which I don't think we want to do. Can we instead do this via the labeler and/or metadata collector?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+2. **Normalize** the product name to lowercase
+3. **Match against `gpu_port_config`** entries using longest substring match
+   - Example: "NVIDIA H100 80GB HBM3" matches "h100" → expects 8 ports at 400 Gb/s
```
lalitadithya: Does this apply to EFA/TCPXO as well? I presume we will have to exclude systems that are using such configurations from this calculation?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+│      └── device/                                                                 │
+│          └── physfn → ../0000:3b:00.0  (VF indicator - points to parent PF)     │
+│                                                                                  │
+└─────────────────────────────────────────────────────────────────────────────────┘
+```
+
+### 5.3 Vendor Detection
+
+The monitor detects Mellanox devices using the following logic:
+1. Check if device name matches `mlx5_\d+` (Mellanox).
```
lalitadithya, line 512: Can the device name be spoofed? Should we instead be matching the PCI vendor information here?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+│                  └─────────────────────────────────────────────────────────┘    │
+│                                                                                  │
+└─────────────────────────────────────────────────────────────────────────────────┘
+```
+
+> **Effect**: The Analyzer emits a new fatal event with `RecommendedAction_REPLACE_VM`. The stabilization window logic (similar to sticky XID) can be implemented as an Analyzer rule to prevent rapid re-alerting.
+
+---
+
+## 7. Device Disappearance Handling
```
lalitadithya, line 589: How would we handle this if the pod restarts or if the node is rebooted? In memory state would mean that it is lost on restarts, right?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+### 8.1 Background: Why VFs Being DOWN is Expected
+
+**SR-IOV (Single Root I/O Virtualization)** is a technology that allows a single physical NIC to appear as multiple virtual NICs. Understanding this is critical for correct alerting behavior.
+
+> **Note**: Clusters with the **NVIDIA Network Operator** installed will have SR-IOV enabled by default.
+
+**The Problem Without Understanding SR-IOV:**
+```
+Monitor starts → Sees 34 devices → 16 are DOWN → Generates 16 FATAL alerts!
+But... those 16 devices are supposed to be DOWN. False alarm storm!
```
lalitadithya, line 649: Can we add some details on why they are supposed to be down?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+|--------|-------------------|-----------------------------------------------------------------|
+| **PF** | Physical Function | The "real" NIC. Host OS controls it. Should ALWAYS be ACTIVE.   |
+| **VF** | Virtual Function  | A "virtual clone" of the PF. Created for VMs/containers to use. |
+
+### 8.3 VF Lifecycle
+
+```
+STAGE 1: System Boot
+├── PF created: mlx5_0 → ACTIVE (host uses it)
+├── VFs created: mlx5_18, mlx5_19, ... → DOWN (waiting for assignment)
```
lalitadithya, line 671: But what about in cases of baremetal environments where we don't have any virtualization layer and instead run containers directly on the host, does this still apply?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+- **State monitoring**: `state`, `phys_state` (via InfiniBand sysfs interface)
+- **Device identification**: Check `link_layer` file for "Ethernet"
+
+### 9.1 GID Table Information (RoCE Routing Diagnostics)
+
+The GID (Global Identifier) table is critical for RoCE routing. Each device exposes GIDs at:
+- `/sys/class/infiniband/<dev>/ports/<port>/gids/<index>`
+- `/sys/class/infiniband/<dev>/ports/<port>/gid_attrs/types/<index>`
+
+**GID Types discovered in production:**
```
lalitadithya: Are these the only two possible documented values?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+
+**Auto-Detection from GPU Product Name:**
+
+Rather than requiring per-cluster configuration, the monitor infers the expected link speed from the **GPU product name**. This works because GPU clusters have standardized InfiniBand configurations based on the GPU generation.
+
+**GPU-to-InfiniBand Mapping:**
+
+| GPU Product | Expected Ports | Expected Rate  | Reference                                                                                          |
+|-------------|----------------|----------------|----------------------------------------------------------------------------------------------------|
+| A100        | 1              | 200 Gb/s       | [DGX A100 User Guide](https://docs.nvidia.com/dgx/dgxa100-user-guide/introduction-to-dgxa100.html) |
```
lalitadithya: Aren't link speeds properties of the NICs, why do we have to link them back to a GPU? Can we clarify this in the design document?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+| Field             | Value                                                    |
+|-------------------|----------------------------------------------------------|
+| Agent             | `nic-health-monitor`                                     |
+| CheckName         | `InfiniBandErrorCheck`                                   |
+| ComponentClass    | `NIC`                                                    |
+| Message           | "Port mlx5_0 port 1: state DOWN - no connectivity"       |
+| IsFatal           | `true`                                                   |
+| IsHealthy         | `false`                                                  |
+| RecommendedAction | `REPLACE_VM`                                             |
+| EntitiesImpacted  | `[{EntityType: "NICPort", EntityValue: "mlx5_0_port1"}]` |
```
lalitadithya: Can we do two entities here? One for the NIC and another for the port?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+### Fatal Counters (IsFatal = true)
+
+| Counter                           | Threshold           | Recommended Action               |
+|-----------------------------------|---------------------|----------------------------------|
+| `link_downed`                     | Delta > 0 (runtime) | **RecommendedAction_REPLACE_VM** |
+| `excessive_buffer_overrun_errors` | > 0 (any)           | **RecommendedAction_REPLACE_VM** |
+| `local_link_integrity_errors`     | > 0 (any)           | **RecommendedAction_REPLACE_VM** |
+| `rnr_nak_retry_err`               | > 0 (any)           | **RecommendedAction_REPLACE_VM** |
+
+### Fatal Driver/Firmware Logs (IsFatal = true)
```
lalitadithya: This should go into the syslog design right? Can we keep the documents focussed on a single problem, please?

docs/nic-health-monitor/syslog-detection-correlation.md
```diff
@@ -0,0 +1,852 @@
...
+### 1.1 Problem Statement
+
+NIC hardware polling monitors (state checks, counter reads) can miss critical failures that occur at the **driver/firmware interface** level. A NIC can appear healthy (link UP, counters normal) while the driver is completely unable to communicate with the firmware, leading to silent workload failures.
+
+### 1.2 Scope of Syslog Detection
+
+This document covers the **Syslog Health Monitor** component for NIC driver error monitoring, which detects:
+
+- **Driver/Firmware communication failures** - `cmd_exec timeout`, firmware hangs
+- **Hardware health check failures** - `health poll failed`, unrecoverable errors
```
lalitadithya, line 37: Can we link to documentation and/or code where these log lines are written?

docs/nic-health-monitor/syslog-detection-correlation.md
```diff
@@ -0,0 +1,852 @@
...
+│  │              SYSLOG HEALTH MONITOR (Event-driven)                    │  │
+│  ├──────────────────────────────────────────────────────────────────────┤  │
+│  │                                                                      │  │
+│  │  CHECK: SysLogsNICDriverError                                        │  │
+│  │                                                                      │  │
+│  │  PATTERN MATCHING FOR (all emitted as WARNING):                      │  │
+│  │  ├── mlx5_core cmd_exec timeout       → Warning (investigate)        │  │
+│  │  ├── mlx5_core health poll failed     → Warning (investigate)        │  │
+│  │  ├── mlx5_core unrecoverable          → Warning (investigate)        │  │
+│  │  ├── PCIe Bus Error.*Fatal            → Warning (investigate)        │  │
```
lalitadithya: This seems to be too broad and can match anything. Are we sure this is scoped to NICs only?

docs/nic-health-monitor/syslog-detection-correlation.md
```diff
@@ -0,0 +1,852 @@
...
+│                                     └─────────────────────────────────────┘     │
+│                                                     │                            │
+│                                                     ▼                            │
+│                    ┌────────────────────────────────────────────────────────┐   │
+│                    │              HEALTH EVENTS ANALYZER                     │   │
+│                    │                                                         │   │
+│                    │  Rule: NICDriverErrorCorrelation                        │   │
+│                    │                                                         │   │
+│                    │  Query:                                                 │   │
+│                    │    SELECT * FROM health_events                         │   │
```
lalitadithya: Can we instead use a MongoDB query here since all the other health event analyzer queries are mongodb based?

docs/nic-health-monitor/syslog-detection-correlation.md
```diff
@@ -0,0 +1,852 @@
...
+Kernel log events are classified by their determinism of failure:
+
+| Kernel Log Event               | Severity    | Purpose                       | Recommended Action |
+|--------------------------------|-------------|-------------------------------|--------------------|
+| `mlx5_core cmd_exec timeout`   | **Fatal**   | Control plane broken          | `REPLACE_VM`       |
+| `mlx5_core health poll failed` | **Fatal**   | Firmware heartbeat lost       | `REPLACE_VM`       |
+| `mlx5_core unrecoverable`      | **Fatal**   | Hardware admission of failure | `REPLACE_VM`       |
+| `PCIe Bus Error.*Fatal`        | **Fatal**   | PCIe link broken              | `REPLACE_VM`       |
+| `NETDEV WATCHDOG`              | **Fatal**   | Data path stalled             | `REPLACE_VM`       |
+| `Detected insufficient power`  | **Warning** | Power negotiation status      | `NONE`             |
```
lalitadithya: In general, we make use of fatal v/s non-fatal as the convention, but in these designs, we are using warning which is a bit different. Can we use the same terminology so that it is less confusing?

docs/nic-health-monitor/link-counter-detection.md
```diff
@@ -0,0 +1,1030 @@
...
+```
+
+After `retry_cnt` attempts (default: 7), the NIC tears down the connection and the application receives `IBV_WC_RETRY_EXC_ERR`.
+
+**Implications:**
+- Confirms **Logical Link is broken** even if physical link is UP
+- Often indicates "Silent Drop" or **Black Hole** in the fabric
+- Local symptom of a **remote** problem
+
+**What This Monitor CAN Detect**: The `local_ack_timeout_err` and `req_transport_retries_exceeded` (native IB) hardware counters track these retry events at the NIC level. Rising counter values indicate transport-layer problems even if we can't see the application error.
```
lalitadithya, line 268: Can this also be triggered by an application? Say one of the ranks of a multi-node training crashes or hangs, would it cause this ack timeout to be incremented on the other nodes?

docs/nic-health-monitor/link-counter-detection.md
```diff
@@ -0,0 +1,1030 @@
...
+| **Port Transmit Discards** | `port_xmit_discards`              | TX discards due to congestion.                                                           | **No**  | > 100/sec                               |                                                                                                                                                   |
+
+#### 4.1.2 Extended Counters (`/sys/class/infiniband/<dev>/ports/<port>/hw_counters/`) — Non-Fatal
+
+All extended counters are **non-fatal** by default. They indicate congestion, retransmissions, or recoverable transport events. RDMA's reliable transport handles these automatically; workloads continue with potential performance impact.
+
+**Key Non-Fatal Counters** (monitor for performance degradation):
+
+| Category       | Counters                | IsFatal | Alert Threshold | Justification                                               |
```
lalitadithya: Can we link to documentation that recommend the alert thresholds?

docs/nic-health-monitor/link-counter-detection.md
```diff
@@ -0,0 +1,1030 @@
...
+|-----------------------------------|----------|-------------------------|---------|----------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------|
+| `link_downed`                     | Standard | **Delta > 0 (Runtime)** | **YES** | Logical path destruction; QP disconnect. Standard HPC apps don't support transparent QP rerouting. | [HPE ClusterStor](https://support.hpe.com/hpesc/public/docDisplay?docId=sd00001143en_us) |
+| `excessive_buffer_overrun_errors` | Standard | **> 0 (Any)**           | **YES** | Lossless guarantee violation; packet dropped immediately. HCA ingress buffer overflow.             | [IBM Redbooks](https://www.redbooks.ibm.com/redbooks/pdfs/sg247767.pdf)                  |
+| `rnr_nak_retry_err`               | Extended | **> 0 (Any)**           | **YES** | Receiver Not Ready NAK retry exhausted; connection severed. Terminal state of error handling.      | [Datadog IB](https://docs.datadoghq.com/integrations/infiniband/)                        |
```
lalitadithya: The link doesn't seem to talk about the threshold or how it will cause a fatal error, can we review the linked docs and make sure that we link to the exact section where we have documentation where it talks about the impact and fatal classification?

docs/nic-health-monitor/link-counter-detection.md
```diff
@@ -0,0 +1,1030 @@
...
+    #--------------------------------------------------------------------------
+    
+    # Physical Layer (PHY)
+    - name: symbol_error
+      path: counters/symbol_error
+      enabled: true
+      isFatal: false
+      thresholdType: velocity
+      threshold: 10.0
+      velocityUnit: per_second
```
lalitadithya:
```suggestion
      velocityUnit: second
```

docs/nic-health-monitor/Overview.md
```diff
@@ -0,0 +1,286 @@
...
+
+```
+┌────────────────────────────────────────────────────────────────────────────┐
+│                        THREE-LAYER NIC MONITORING                          │
+├────────────────────────────────────────────────────────────────────────────┤
+│                                                                            │
+│  LAYER 1: LINK STATE DETECTION                                             │
+│  ═══════════════════════════════                                           │
+│  • Polling interval: 1 second                                              │
+│  • Data source: /sys/class/infiniband/<dev>/ports/<port>/state, phys_state │
```
deesharma24, line 134: Last time when we had nic-health-monitor deployed in our prod clusters. one of the issues which we faced was whether certain interfaces(be it infiniband/ethernet) are in active use or not. if they are not in active use, then whether they stay up/down doesn't matter and I vaguely remember due to this all the nodes in some cluster(probably osmo) should nic down even though that nic interface was not getting used. how is it going to help this time? will it still be configuration driven where we need to mention the interfaces we want to monitor based cluster/csp?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+
+**GPU-to-InfiniBand Mapping:**
+
+| GPU Product | Expected Ports | Expected Rate  | Reference                                                                                          |
+|-------------|----------------|----------------|----------------------------------------------------------------------------------------------------|
+| A100        | 1              | 200 Gb/s       | [DGX A100 User Guide](https://docs.nvidia.com/dgx/dgxa100-user-guide/introduction-to-dgxa100.html) |
+| H100        | 8              | 400 Gb/s (NDR) | [DGX H100 User Guide](https://docs.nvidia.com/dgx/dgxh100-user-guide/introduction-to-dgxh100.html) |
+| H200        | 8              | 400 Gb/s (NDR) | NVIDIA Documentation                                                                               |
+| B200        | 8              | 400 Gb/s       | [DGX B200 User Guide](https://docs.nvidia.com/dgx/dgxb200-user-guide/introduction-to-dgxb200.html) |
+| GB200       | 8              | 400 Gb/s       | [GB200 NVL2](https://www.nvidia.com/en-us/data-center/gb200-nvl2/)                                 |
```
XRFXLP: GB200 system uses fifth generation nvlink whereas H100 using fourth, wouldn't there would be ~1.5x speedup?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+
+**Metric:** `/sys/class/infiniband/<dev>/ports/<port>/rate`
+
+**Auto-Detection from GPU Product Name:**
+
+Rather than requiring per-cluster configuration, the monitor infers the expected link speed from the **GPU product name**. This works because GPU clusters have standardized InfiniBand configurations based on the GPU generation.
+
+**GPU-to-InfiniBand Mapping:**
+
+| GPU Product | Expected Ports | Expected Rate  | Reference                                                                                          |
```
XRFXLP: Are these theoretical expected rate if everything E2E as per data-sheet? What if rates are like ~195 gbps for A100?

docs/nic-health-monitor/link-state-detection.md
```diff
@@ -0,0 +1,900 @@
...
+│  ┌─────────────────────────────────────────────────────────────────────┐        │
+│  │  NIC: mlx5_0                                                         │        │
+│  │  State: ACTIVE ✓     PhysState: LinkUp ✓     Rate: 200 Gb/s ✗       │        │
+│  │                                              └── 50% capacity loss!  │        │
+│  └─────────────────────────────────────────────────────────────────────┘        │
+│                                                                                  │
+│  ═══════════════════════════════════════════════════════════════════════════    │
+│                                                                                  │
+│  IMPACT ON NCCL AllReduce (Ring Topology):                                       │
+│                                                                                  │
```
XRFXLP: This diagram gives impression that communication is inside node between different GPUs i.e. link speed being nvlink speed (which is out of scope of this MR). Can we make it slightly more clear to have inter-node seperaiton?

lalitadithya (approved):

PR #759, opened 2026-01-29
fix: fault_quarantine_current_quarantined_nodes metric update

### LINKED ISSUE #835: [Bug]: UAT is flaky for syslog health monitor
### Prerequisites

### Bug Description

In UAT tests, sometimes the node is not getting uncordoned post-reboot and as a result the test are failings:
https://github.com/NVIDIA/NVSentinel/actions/runs/21715880145/job/62631555650
https://github.com/NVIDIA/NVSentinel/actions/runs/21909708266/job/63259746624
https://github.com/NVIDIA/NVSentinel/actions/runs/21940130273/attempts/1

### Component

Health Monitor

### Steps to Reproduce

Retrigger UAT enough time to get trigger the failure

### Environment

- NVSentinel version:v0.8.0
- Kubernetes version:1.32
- Deployment method:Helm

### Logs/Output

```
[2026-02-12 09:34:49] =========================================
[2026-02-12 09:34:49] Test 2: XID monitoring via syslog
[2026-02-12 09:34:49] =========================================
[2026-02-12 09:34:51] Selected GPU node: gke-nvs-d21940130273-gpu-pool-a31d5e86-3sr3 (has healthy syslog-health-monitor)
[2026-02-12 09:34:51] Original boot ID: bae9c578-9505-4276-8e03-3ee296184ae3
[2026-02-12 09:34:52] Injecting XID 119 message via logger on pod: nvidia-driver-daemonset-ccklx
[2026-02-12 09:34:54] Waiting for node condition 'SysLogsXIDError' to appear on node gke-nvs-d21940130273-gpu-pool-a31d5e86-3sr3...
[2026-02-12 09:35:00] Node condition 'SysLogsXIDError' found ✓
  Status=True Reason=SysLogsXIDErrorIsNotHealthy
[2026-02-12 09:35:01] Waiting for node gke-nvs-d21940130273-gpu-pool-a31d5e86-3sr3 to be quarantined (cordoned)...
[2026-02-12 09:35:01] Node gke-nvs-d21940130273-gpu-pool-a31d5e86-3sr3 is quarantined (cordoned) ✓
[2026-02-12 09:35:01] Waiting for node to reboot and recover...
[2026-02-12 09:35:01] Waiting for node gke-nvs-d21940130273-gpu-pool-a31d5e86-3sr3 to reboot (boot ID to change)...
[2026-02-12 09:35:01] Original boot ID: bae9c578-9505-4276-8e03-3ee296184ae3
[2026-02-12 09:35:36] Still waiting... elapsed=30s, current_boot_id='bae9c578-9505-4276-8e03-3ee296184ae3'
[2026-02-12 09:36:10] Still waiting... elapsed=60s, current_boot_id='a5ad5690-e049-4300-a5c6-44bf3cb4712c'
[2026-02-12 09:36:10] Node gke-nvs-d21940130273-gpu-pool-a31d5e86-3sr3 rebooted successfully (boot ID changed)
[2026-02-12 09:36:10]   Old: bae9c578-9505-4276-8e03-3ee296184ae3
[2026-02-12 09:36:10]   New: a5ad5690-e049-4300-a5c6-44bf3cb4712c
[2026-02-12 09:36:10] Waiting for node gke-nvs-d21940130273-gpu-pool-a31d5e86-3sr3 to be uncordoned...
[2026-02-12 09:36:45] Still waiting for uncordon... elapsed=30s, unschedulable=true
[2026-02-12 09:37:18] Still waiting for uncordon... elapsed=60s, unschedulable=true
[2026-02-12 09:37:52] Still waiting for uncordon... elapsed=90s, unschedulable=true
[2026-02-12 09:38:25] Still waiting for uncordon... elapsed=120s, unschedulable=true
[2026-02-12 09:38:59] Still waiting for uncordon... elapsed=150s, unschedulable=true
[2026-02-12 09:39:33] Still waiting for uncordon... elapsed=180s, unschedulable=true
[2026-02-12 09:40:06] Still waiting for uncordon... elapsed=210s, unschedulable=true
[2026-02-12 09:40:40] Still waiting for uncordon... elapsed=240s, unschedulable=true
[2026-02-12 09:41:13] Still waiting for uncordon... elapsed=270s, unschedulable=true
[2026-02-12 09:41:47] Still waiting for uncordon... elapsed=300s, unschedulable=true
[2026-02-12 09:42:21] Still waiting for uncordon... elapsed=330s, unschedulable=true
[2026-02-12 09:42:54] Still waiting for uncordon... elapsed=360s, unschedulable=true
[2026-02-12 09:43:28] Still waiting for uncordon... elapsed=390s, unschedulable=true
[2026-02-12 09:44:01] Still waiting for uncordon... elapsed=420s, unschedulable=true
[2026-02-12 09:44:35] Still waiting for uncordon... elapsed=450s, unschedulable=true
[2026-02-12 09:45:09] Still waiting for uncordon... elapsed=480s, unschedulable=true
[2026-02-12 09:45:42] Still waiting for uncordon... elapsed=510s, unschedulable=true
[2026-02-12 09:46:16] Still waiting for uncordon... elapsed=540s, unschedulable=true
[2026-02-12 09:46:49] Still waiting for uncordon... elapsed=570s, unschedulable=true
[2026-02-12 09:47:22] ERROR: Timeout waiting for node gke-nvs-d21940130273-gpu-pool-a31d5e86-3sr3 to be uncordoned
```

tanishagoyal2:
## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary
Recently we found following issues in prod clusters:
1.  the `fault_quarantine_current_quarantined_nodes` metric value was more than the real number of cordoned nodes. 
2. node had both annotations `quarantineHealthEvent` (with empty value) and `quarantinedNodeUncordonedManually`. Node is expected to have `quarantineHealthEvent` if node is cordoned and only `quarantinedNodeUncordonedManually` if node was manually uncordoned but we found both annotations on node.
We need to find why both type of annotations present and why `quarantineHealthEvent` annotation had empty value
3. Some nodes didn't have `quarantineHealthEvent` annotation and only `quarantinedNodeUncordonedManually` annotation present which is correct but `fault_quarantine_current_quarantined_nodes` gave more than expected value
4. Manual taint removal is not unquarantining the node and not removing the quarantine annotations. 

### For case 1
Case 1 is possible when node was manually uncordoned during the upgrade or before the fault-quarantine started running
1. Tried to reproduce it in tilt cluster
2. cordoned the node using unhealthy event
3. crashed the fault-quarantine pod
5. tried to manually uncordon the node
6. when the pod came back in running state, it checked for quarantineHealthEvent annotation which was present because we uncordoned the node when pod was crashing so it didn't detect manual uncordoning.
7. fault_quarantine_current_quarantined_nodes metric initialized with 1 value even when no node was cordoned. This is because fault-quarantine is checking for the presence of quarantineHealthEvent to count the pods cordoned by nvsentinel

### Fix
During the metric initialization, fault-quarantine should check for annotation quarantineHealthEvent precence along with unschelable state as a safety guard. If annotation is present but node is not cordoned that means we have stale annotation and we should clear stale annotations.

### For case 2
The quarantineHealthEvent annotation can have empty value when there is no unhealthy event present on that node. fault-quarantine pod calls performUncordon function to update/remove this annotation when empty. It is possible that performUncordon failed because of conflict update on the node. 

I was able to reproduce this issue in tilt cluster when all HEA tests ran. I observed that two nodes remained cordoned with stale annotations like quarantineHealthEvent with empty value.
Logs revealed that there wan a conflict in updating the node.

Screenshot 2026-02-01 at 4 52 13 PM
Screenshot 2026-02-01 at 4 51 08 PM

### Fix
Since, we were not returning the error from performUncordon function, the caller function proceed with next steps and we found stale annotation with empty value. 
We should return error in performUncordon function and catch that error
We also need to update the defaultRetries in updateNode function with more attempts and increased delays

### For case 3
1. Checked the logs of fault-quarantine pod and found that cluster was upgrading when this issue started appearing.
2. The node was cordoned because of unhealthy event and quarantineHealthEvent annotations were applied
3. cluster upgraded and PVCs were recreated
4. When upgrade was finished and fault-quarantine pod restarted, the node tried to manually uncordoned
5. During manually uncordoned operation, fault-quarantine found applied annotation and tried to update the healthEvent in db but failed to find that document in db and thus finished with error.

Screenshot 2026-01-28 at 6 22 49 PM

6. We were updating the metric fault_quarantine_current_quarantined_nodes after successfully updating the healthEvent. Since, it failed to find the event the pod finished with error and didn't update the metric 

### Fix
We should return warning instead of error in case of mongodb doc is not found in db. 

### Case 4
1. quarantined node with taint only config
2. manually untaint the node
3. fault-quarantine didn't remove the quarantineHealthEvent annotation and didn't take action on manual untaint of node

### Fix
We should listen on manually taint removal and quarantine annotation should be removed. The  fault_quarantine_current_quarantined_nodes should be updated correctly

## Summary by CodeRabbit

* **New Features**
  * Added manual untaint handling, a cleanup API for manual untaint/uncordon, new annotation keys, and a metric for manually untainted nodes.

* **Bug Fixes**
  * Improved handling of missing event documents (now logs a warning) and enhanced stale quarantine-state detection and cleanup on startup.

* **Tests**
  * Expanded end-to-end tests for quarantine lifecycle, stale-state restarts, manual uncordon/untaint, and metric verification.

* **Chores**
  * Refactored startup flow and error propagation; increased retry/backoff for node updates.


fault-quarantine/pkg/informer/node_informer.go
```diff
@@ -158,13 +158,30 @@ func (ni *NodeInformer) GetNodeCounts() (totalNodes int, quarantinedNodesMap map
 	}
 
 	quarantinedMap := make(map[string]bool, len(quarantinedObjs))
+	staleCount := 0
 
 	for _, obj := range quarantinedObjs {
 		if node, ok := obj.(*v1.Node); ok {
-			quarantinedMap[node.Name] = true
+			// confirm the node is still cordoned by checking the node's unschedulable status
```
lalitadithya: We have configuration for taints as well, so the user can configure the system to do cordon or taint or both, how are we going to handle such cases?

tanishagoyal2: This function is called during the pod start, where its trying to initialize fault_quarantine_current_quarantined_nodes so we should only count still cordoned node.
Here, are you asking to handle the case when only taints were applied?

lalitadithya:
> We also need to update the defaultRetries in updateNode function with more attempts and increased delays

Are we using the retry on conflict function while make updates? Is the conflict happening even after all the retries have completed?

fault-quarantine/pkg/informer/k8s_client.go
```diff
@@ -139,7 +139,14 @@ func (c *FaultQuarantineClient) UpdateNode(ctx context.Context, nodeName string,
 
 	defer mu.(*sync.Mutex).Unlock()
 
-	return retry.OnError(retry.DefaultBackoff, errors.IsConflict, func() error {
+	backoff := wait.Backoff{
+		Steps:    10,                    // Increased from default 5
+		Duration: 20 * time.Millisecond, // Increased from default 10ms
```
lalitadithya, line 146: Can we also add why it was increased?

fault-quarantine/pkg/reconciler/reconciler_e2e_test.go
```diff
@@ -4332,3 +4238,554 @@ func TestE2E_ConcurrentHealthyEvents_WithDelayedInformer(t *testing.T) {
...
+	labels := map[string]string{
+		statemanager.NVSentinelStateLabelKey: string(statemanager.QuarantinedLabelValue),
+	}
+
+	taints := []corev1.Taint{
+		{Key: "nvidia.com/gpu-xid-error", Value: "true", Effect: "NoSchedule"},
+	}
+
+	// Create node with stale state: has FQ annotations/taints but NOT cordoned
+	createE2ETestNode(ctx, t, nodeName, annotations, labels, taints, false) // unschedulable=false!
```
lalitadithya: Ideally, we should simulate the flow, i.e., run the reconciler once so that it cordons the nodes, then manually uncordon the node and then rerun the reconciler.

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -315,6 +328,171 @@ func (r *Reconciler) initializeQuarantineMetrics() {
...
+		slog.Error("Failed to list nodes for stale annotation cleanup", "error", err)
+		return fmt.Errorf("failed to list nodes for stale annotation cleanup: %w", err)
+	}
+
+	for _, node := range nodes {
+		_, hasCordonAnnotation := node.Annotations[common.QuarantineHealthEventIsCordonedAnnotationKey]
+
+		if hasCordonAnnotation && !node.Spec.Unschedulable {
+			if err := r.cleanupStaleCordonAnnotations(ctx, node); err != nil {
+				return err
```
lalitadithya: Please wrap the error

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -315,6 +328,171 @@ func (r *Reconciler) initializeQuarantineMetrics() {
...
+func (r *Reconciler) cleanupStaleTaints(ctx context.Context, node *corev1.Node) error {
+	// Check if taints in QuarantineHealthEventAppliedTaintsAnnotationKey are still on node
+	// If not present, that means we have stale annotations that need to be cleaned up
+	taintsApplied := []config.Taint{}
+	taintAnnotationValue := node.Annotations[common.QuarantineHealthEventAppliedTaintsAnnotationKey]
+
+	if err := json.Unmarshal([]byte(taintAnnotationValue), &taintsApplied); err != nil {
+		slog.Error("Failed to unmarshal taints", "node", node.Name, "error", err)
+		return fmt.Errorf("failed to unmarshal taints: %w", err)
+	}
```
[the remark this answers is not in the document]
tanishagoyal2: but why would QuarantineHealthEventAppliedTaintsAnnotationKey would be invalid. it has been set by FQ. And if its invalid then application should throw error

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1118,11 +1308,11 @@ func (r *Reconciler) performUncordon(
...
 	if err != nil {
 		slog.Error("Failed to prepare uncordon params for node", "node", event.NodeName, "error", err)
-		return true
+		return true, fmt.Errorf("failed to prepare uncordon params for node %s: %w", event.NodeName, err)
 	}
 
 	if len(taintsToBeRemoved) == 0 && !isUnCordon {
-		return false
+		return false, nil
 	}
```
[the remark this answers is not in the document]
tanishagoyal2: if someone is manually untainting and uncordoning the node, then we have separate flow to handle manual removal of it

fault-quarantine/pkg/reconciler/reconciler_e2e_test.go
tanishagoyal2: But how we will simulate the pod restart in unit test. We can do E2E testing with complete flow in integration test.

lalitadithya: Isn't calling the reconciler function twice the same as pod restart?

tanishagoyal2: but I want to simulate the manual uncordoning when pod is not running that means reconciler shouldn't run

lalitadithya: right, so can't we do 

```
setupE2EReconciler
uncordon
setupE2EReconciler
```

fault-quarantine/pkg/eventwatcher/event_watcher.go
```diff
@@ -270,6 +270,12 @@ func (w *EventWatcher) CancelLatestQuarantiningEvents(
 	}
 
 	if err := result.Decode(&latestEvent); err != nil {
+		if errors.Is(err, client.ErrNoDocuments) || client.IsNoDocumentsError(err) {
+			slog.Debug("No quarantining/unquarantining events found for node (during decode)", "node", nodeName)
```
XRFXLP: Why is this a debug log? We add debug logs in the operations which are frequent and unimportant usually. In this case, it is an important logs should be there irrespective of whether we've debug flag or not.

fault-quarantine/pkg/informer/k8s_client.go
```diff
@@ -139,7 +139,16 @@ func (c *FaultQuarantineClient) UpdateNode(ctx context.Context, nodeName string,
...
+	// Increased retry attempts to handle node update conflicts when multiple modules
+	// attempt concurrent updates, preventing nodes from remaining cordoned with stale annotations.
+	backoff := wait.Backoff{
+		Steps:    10,                    // Increased from default 5
+		Duration: 20 * time.Millisecond, // Increased from default 10ms
+		Factor:   2.0,
+		Jitter:   0.1,
+	}
+
+	return retry.OnError(backoff, errors.IsConflict, func() error {
```
XRFXLP: Is `errors.IsConflict` the only retry-able error we should be concerned about? what about connection refused or glitchy timeouts?

fault-quarantine/pkg/informer/k8s_client.go
```diff
@@ -475,6 +500,53 @@ func (c *FaultQuarantineClient) HandleManualUncordonCleanup(
...
+	taintsToRemove []config.Taint,
+	annotationsToRemove []string,
+	annotationsToAdd map[string]string,
+	labelsToRemove []string,
+) error {
+	updateFn := func(node *v1.Node) error {
+		// Uncordon the node if it's currently cordoned
+		if node.Spec.Unschedulable {
+			if !c.DryRunMode {
+				slog.Info("Uncordoning node as part of manual untaint cleanup", "node", nodename)
```
lalitadithya: Why do we need to uncordon the node as a part of the taint clean up?

tanishagoyal2: it is for the case when node was taint+cordon by FQM and someone manually untainted the node.

lalitadithya: But we shouldn't make that decision on behalf of the user. Maybe the user didn't want to uncordon.

tanishagoyal2: But when the node is manually uncordoned then taints are also removed from the node. How is this case diff?

lalitadithya: That is also incorrect, can you point me where in code we have that problem? I will open a ticket for this

tanishagoyal2: You can check in [HandleManualUncordon](https://github.com/tanishagoyal2/NVSentinel/blob/main/fault-quarantine/pkg/informer/k8s_client.go#L478) function

KaivalyaMDabhadkar: @lalitadithya But for the case when the Node was cordoned by FQ and the user wants to uncordon it manually, doesn't it implicitly imply that the user wants to undo all actions taken by FQ on the node (including removing any taints applied by FQ)? Would there be a case when the user decides to uncordon the node cordoned by FQ but still wants to keep the FQ applied taints?

lalitadithya: We don't that right. Maybe the presence of the taint is preventing scheduling of some validation workloads for the user, and they want to remove the taints, but they want to keep the cordon. At the end of the day, we can't make the decision on behalf of the user

fault-quarantine/pkg/informer/node_informer.go
```diff
@@ -241,9 +246,79 @@ func (ni *NodeInformer) detectAndHandleManualUncordon(oldNode, newNode *v1.Node)
...
+	return true
+}
+
+// handleUpdateNode detects and handles manual uncordon and manual untaint of quarantined nodes.
 func (ni *NodeInformer) handleUpdateNode(oldNode, newNode *v1.Node) {
-	ni.detectAndHandleManualUncordon(oldNode, newNode)
+	// Check manual uncordon first - if it triggers, it does full cleanup including taints
+	handledByUncordon := ni.detectAndHandleManualUncordon(oldNode, newNode)
+
+	// Only check manual untaint if uncordon didn't already handle it
```
KaivalyaMDabhadkar: Shouldn't we keep on retrying until `ni.detectAndHandleManualUncordon(oldNode, newNode)` succeeds? In case `handledByUncordon` returns false that would mean we failed at uncordoning the node anyways right

KaivalyaMDabhadkar: As mentioned by Lalit above, we need to removethe untaint logic from [HandleManualUncordonCleanup](https://github.com/tanishagoyal2/NVSentinel/blob/main/fault-quarantine/pkg/informer/k8s_client.go#L478). Once we do that, we would need to call the `detectAndHandleManualUntaint(oldNode, newNode)` function separately here along with `ni.detectAndHandleManualUncordon(oldNode, newNode)` to clean up the taints

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1345,28 +1446,107 @@ func (r *Reconciler) handleManualUncordon(nodeName string) error {
...
+		return fmt.Errorf("failed to clean up manually untainted node %s: %w", nodeName, err)
+	}
+
+	slog.Debug("Successfully completed K8s cleanup for manual untaint", "node", nodeName)
+
+	// Reset the quarantine gauge metric since the node is no longer quarantined
 	metrics.CurrentQuarantinedNodes.WithLabelValues(nodeName).Set(0)
-	slog.Info("Set currentQuarantinedNodes to 0 for manually uncordoned node", "node", nodeName)
+	metrics.TotalNodesManuallyUntainted.WithLabelValues(nodeName).Inc()
+	slog.Info("Set currentQuarantinedNodes to 0 after manual untaint", "node", nodeName)
```
KaivalyaMDabhadkar, line 1455: What if the Node is untainted manually but not uncordoned manually? In that case we would need to check whether the node is in a cordoned state or not as well before we update the metric. Also, should we consider having two metrics `CurrentCordonedNodes` and `CurrentTaintedNodes` rather than the ambiguous metric `CurrentQuarantinedNodes`? That way we will avoid the confusion of determining the exact state of the node since "Quarantined" can mean any of these possibilities: only Tainted, only cordoned, both Tainted and cordoned

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1345,28 +1446,107 @@ func (r *Reconciler) handleManualUncordon(nodeName string) error {
 
 	slog.Debug("Successfully completed K8s cleanup for manual uncordon", "node", nodeName)
 
+	slog.Info("Set currentQuarantinedNodes to 0 for manually uncordoned node", "node", nodeName)
+	metrics.CurrentQuarantinedNodes.WithLabelValues(nodeName).Set(0)
```
KaivalyaMDabhadkar, line 1382: Same issue here as mentioned in this comment: https://github.com/NVIDIA/NVSentinel/pull/759/changes#r2799276727

fault-quarantine/pkg/informer/node_informer.go
```diff
@@ -241,9 +246,79 @@ func (ni *NodeInformer) detectAndHandleManualUncordon(oldNode, newNode *v1.Node)
...
+	}
+
+	// If no taint was actually removed in this update, don't trigger
+	if !taintWasRemoved {
+		return false
+	}
+
+	slog.Info("Detected manual untaint of FQ-quarantined node", "node", newNode.Name)
+
+	if ni.onManualUntaint != nil {
```
KaivalyaMDabhadkar: `ni.onManualUntaint` will never be nil right?

fault-quarantine/pkg/informer/node_informer.go
```diff
@@ -241,9 +246,79 @@ func (ni *NodeInformer) detectAndHandleManualUncordon(oldNode, newNode *v1.Node)
...
+		slog.Debug("Invoking manual untaint callback", "node", newNode.Name)
+
+		if err := ni.onManualUntaint(newNode.Name); err != nil {
+			slog.Error("Manual untaint callback failed", "node", newNode.Name, "error", err)
+		} else {
+			slog.Debug("Manual untaint callback completed successfully", "node", newNode.Name)
+		}
+	} else {
+		msg := "Manual untaint callback is NOT REGISTERED - manual untaint will NOT be handled!"
```
KaivalyaMDabhadkar: We have already registered a manual untaint callback, so we can get rid of this conditional

fault-quarantine/pkg/informer/node_informer.go
```diff
@@ -241,9 +246,79 @@ func (ni *NodeInformer) detectAndHandleManualUncordon(oldNode, newNode *v1.Node)
 	return true
 }
 
-// handleUpdateNode detects and handles manual uncordon of quarantined nodes.
+// detectAndHandleManualUntaint checks if a node was manually untainted and handles it
+func (ni *NodeInformer) detectAndHandleManualUntaint(oldNode, newNode *v1.Node) bool {
+	// Check if node has FQ taint annotation
+	taintsAnnotation, hasTaintsAnnotation := newNode.Annotations[common.QuarantineHealthEventAppliedTaintsAnnotationKey]
+	if !hasTaintsAnnotation || taintsAnnotation == "" {
+		return false
```
KaivalyaMDabhadkar, line 340: I think before returning here early if `hasTaintsAnnotation` is false, we should also check if it's uncordoned/untainted as well. If the annotation is not present but the node is still cordoned or tainted, that would mean we need to uncordon and untaint it

fault-quarantine/pkg/informer/node_informer.go
```diff
@@ -241,9 +246,79 @@ func (ni *NodeInformer) detectAndHandleManualUncordon(oldNode, newNode *v1.Node)
...
+	} else {
+		msg := "Manual untaint callback is NOT REGISTERED - manual untaint will NOT be handled!"
+		slog.Warn(msg, "node", newNode.Name)
+	}
+
+	return true
+}
+
+// handleUpdateNode detects and handles manual uncordon and manual untaint of quarantined nodes.
 func (ni *NodeInformer) handleUpdateNode(oldNode, newNode *v1.Node) {
```
KaivalyaMDabhadkar, line 388: As we had discussed in the slack thread before with Ajay and Lalit, do we know for sure that if the node is cordoned/Tainted by FQ -> FQ crashes -> Node is uncordoned/Untainted manually -> FQ pod recovers, we will get the events as UPDATEs from the informer with the same oldObj and newObj rather than ADD events when the informer restarts and resyncs when the FQ pod restarts? Did we validate this?

fault-quarantine/pkg/informer/k8s_client.go
tanishagoyal2: what if the node was taint+cordon by fq but untainted manually. Should quarantineHealthEvent annotation also be removed? and should we cancel latest quarantine event also? Because if user wants to schedule the workload on node then draining and remediation should stop, right?

fault-quarantine/pkg/informer/node_informer.go
tanishagoyal2: But why would there be a case when annotation is not present but node is cordoned/tainted? If we have such case then that must be an error which needs to be handled

fault-quarantine/pkg/informer/node_informer.go
tanishagoyal2: Yes, I checked it on tilt cluster. Update informer was not getting called only addNode was called

KaivalyaMDabhadkar: But then shouldn't we have the taint/cordon cleanup logic (plus making the appropriate updates to MongoDB) in `handleAddNode` as well to take care of this restart case?

fault-quarantine/pkg/informer/node_informer.go
KaivalyaMDabhadkar: > But why would there be a case when annotation is not present but node is cordoned/tainted? If we have such case then that must be an error which needs to be handled

That's correct, I was thinking we can have the FQ applied taints cleanup logic just as a safety mechanism for the nodes which have the annotation missing (not cordon, only the taints cleanup). This would take care of potential cases where our FQ succeeded in removing the annotation from the node but failed midway before it could remove the taints. But in any case I think this is a nit and an optional change, as we are logging errors for failed taints/untaints it's fine if we skip it

fault-quarantine/pkg/informer/k8s_client.go
lalitadithya: Yes, it should stop. Any action that the user takes should result in cancelling the breakfix

fault-quarantine/pkg/reconciler/reconciler.go
tanishagoyal2: I don't think we need two metric anymore now. Because with manual action like uncordoning/untainting we are removing all quarantine annotations and cancelling healthevent. That means we are removing all applied state changes from node. So, updating the currentQuarantinedNodes to 0 will make sense because FQM is not acting on node anymore.

fault-quarantine/pkg/informer/node_informer.go
tanishagoyal2: right, I have added cleanup code in addnode function

fault-quarantine/pkg/informer/node_informer.go
tanishagoyal2: right, removed

fault-quarantine/pkg/informer/k8s_client.go
KaivalyaMDabhadkar: I think if the FQ quarantined the node due to a rulset which applied both a cordon and a taint, in this case the FQ should only remove the quarantineAnnotations and update the quarantine metric only if BOTH the cordon and the taint are removed by the user on the node. Else, we can run into cases where the user manually only removed the taint, and the FQ removes it annotation, putting the node in a state where it still has a cordon and no way of distinguishing whether that cordon was applied by a user or by FQ. What do you think?

lalitadithya: Hmm.. Under what cases would FQ need to know if it applied the cordon?

KaivalyaMDabhadkar: FQ would not necessarily need to know, but from a user's perspective if he sees a node which is cordoned in a cluster but has no FQ applied annotations on it, he can mistakenly conclude that the node was manually cordoned by a user rather than FQ. This is because if we decide to only remove the FQ annotation on the node even if the user only partially undoes the actions taken by FQ on the node, then in a way we are considering the node as "unquarantined" even though the node still has a cordon (in case we only remove a taint) or a taint (if we only remove the cordon). I think logically FQ should only consider a node as "Unquarantined" if all the actions it took on the node (taints+cordon) are undone by the user (and the onus will be on the user to undo those operations if he wants to schedule a workload on the node)

lalitadithya: I feel that is an acceptable thing because at the end of the day the cases for cordon + taint I think are fairly niche. We also have to consider the fact for a user using kubectl in many cases it will have to do commands to remove (the user can do a json patch, yes). We can document this and make sure the user knows that this is how the system works. We could also track partial unquarantined using an annotation, but that may be a bit overkill. 

Also, if we are stopping breakfix we will apply the node unquarantined manually annotation right? Wouldn't that be a sufficient indication to the user?

KaivalyaMDabhadkar (approved):

PR #879, opened 2026-02-18
fix: wait for GPUReset CRD rather than check syslog in UAT

natherz97:
## Summary
The UAT test for GPU reset is currently failing because the "/var/log/syslog" directory does not exist on GCP nodes. Since we do not have a container with journalctl available to read syslogs directly from "/var/log/journal" to detect the GPU reset in the test, we will wait for a matching GPU reset CRD to be created. This check is required to confirm that the given injected XID error resulted in a GPU reset rather than a reboot. In summary, will wait for a GPUReset CRD which:
- matches the expected node with the unhealthy GPU
- matches the GPU UUID which had the injected XID syslog error
- was created after the node was cordoned (to prevent stale GPUReset CRDs from previous test executions from being consumed)

## Testing
I locally ran the wait_for_gpu_reset function with these arguments:
```
gpu_node="10.0.6.34"
uuid="GPU-8598879c-4839-1709-231e-36a2b2844bca"
wait_for_gpu_reset "$gpu_node" "$uuid"
```
Running the function against a cluster with a matching GPUReset CRD:
```
% ./wait_for_gpu_reset.sh
[2026-02-17 16:04:06] Waiting for GPU reset for GPU-8598879c-4839-1709-231e-36a2b2844bca on 10.0.6.34 (matching GPUReset CRD)...
[2026-02-17 16:04:07] GPUReset maintenance-10.0.6.34-697bc65faecdaeca50880d66 matches GPU-8598879c-4839-1709-231e-36a2b2844bca and 10.0.6.34
...
./wait_for_gpu_reset.sh: line 56: wait_for_node_unquarantine: command not found
```
I confirmed that the test will fail if a GPUReset CRD does not exist with the expected node name or GPU UUID. Additionally, the test will fail if the there is a GPUReset CRD with the required node and GPU but was created with too old of a timestamp. 

## Type of Change
- [X] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [X] Other: UAT

## Testing
- [X] Tests pass locally
- [X]  Manual testing completed

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **Tests**
  * Switched automated GPU reset verification from syslog parsing to CRD-driven detection (matching by node and GPU UUID). Tests now provide a timestamp baseline through the wait flow, include explicit timeout/elapsed handling, skip incomplete CRD records early, and record clearer match logs to improve reliability and observability of GPU reset validation.

### DIFF
--- a/tests/uat/tests.sh
+++ b/tests/uat/tests.sh
@@ -218,33 +218,41 @@ wait_for_boot_id_change() {
 wait_for_gpu_reset() {
     local node=$1
     local uuid=$2
+    local current_ts=$3
     local timeout=${UAT_RESET_TIMEOUT:-600}
     local elapsed=0
+    local matching_crd
 
-    log "Waiting for GPU reset for $uuid on $node (GPU reset syslog message from Janitor)..."
-
-    local driver_pod
-    driver_pod=$(kubectl get pods -n gpu-operator -l app=nvidia-driver-daemonset -o jsonpath="{.items[?(@.spec.nodeName=='$node')].metadata.name}" | head -1)
-
-    if [[ -z "$driver_pod" ]]; then
-        error "No driver pod found on node $node"
-    fi
+    log "Waiting for GPU reset for $uuid on $node (matching GPUReset CRD)..."
 
     while [[ $elapsed -lt $timeout ]]; do
-        # We are ignoring log lines that include RuntimeService to prevent picking up syslog messages which result
-        # from this kubectl exec request. This is to prevent the second kubectl exec request from matching the first
-        # exec request and incorrectly determining that the GPUReset job wrote the syslog. Additionally, we
-        # have to make sure that the syslog-health-monitor itself doesn't pick up this log line and pre-maturely write
-        # a healthy event. Rather than add a similar check for the presence of RuntimeService in the syslog-health-monitor
-        # regex, we will grep the lines without the GPU UUID (which won't match the syslog-health-monitor regex) and then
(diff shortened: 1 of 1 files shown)


tests/uat/tests.sh
```diff
@@ -220,31 +220,39 @@ wait_for_gpu_reset() {
     local uuid=$2
     local timeout=${UAT_RESET_TIMEOUT:-600}
     local elapsed=0
+    local current_ts=$(date +%s)
```
lalitadithya: shouldn't this timestamp be set before the error is inject or it will be a bit race-y? For example, if all the operations happen really quickly before we reach this point then it is possible for the test to fail since it will consider the reset CR to be old?

natherz97: That's a good callout. I previously saw health event injection to CRD creation latencies of 10s-20s when testing on Kind. In the worst case, we would have 5 seconds + time to drain + create GPUReset CRD + Janitor adding StartTime. 

Let me make the fix you suggested and I'll pass the comparison as a param to wait_for_gpu_reset so we won't be dependent on setting our current timestamp before these steps complete.

lalitadithya (approved):

PR #913, opened 2026-02-26
feat: add design doc for otel tracing

tanishagoyal2:
## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

### Testing
All generated traces follows the patterns mentioned below:

1. Any error that occurred while handling the event is shown with a red dot on the span, and the error type and message are shown as errored span attributes.
2. We have tried to capture every operation done by each module like cordoning, draining, remediation, uncordoning etc.
3. All async processes and MongoDB-dependent module relationships are shown with the help of span links.
4. Each module writes its own span ID into the span_ids map in the MongoDB document after processing, so the next downstream module can read the exact upstream span to link against — creating an auditable causal chain across the pipeline.
5. We have captured all DB call, K8s API call, Cloud API call, and any external API call latency by creating a new span.
6. Span attributes include duration_ms for key operations (DB writes, K8s API calls, CSP API calls) to allow filtering slow operations via TraceQL (e.g. {db.duration_ms > 50}).
7. The trace_id is included in exported CloudEvents (event-exporter copies all health event metadata), allowing operators to link an exported event back to the full distributed trace in Grafana.
8. Log-trace correlation is enabled via a custom slog.Handler wrapper that injects trace_id and span_id from the active span into every JSON log line, so logs and traces can be correlated by trace_id.

Screenshot 2026-03-24 at 10 57 34 AM

We have tested out few scenarios to cover positive and negative cases for traces:

1. Remediation skipped because of non-actionable action: the span attribute showed the reason for skipping the remediation.

Screenshot 2026-03-24 at 10 46 08 AM

Screenshot 2026-03-24 at 10 46 19 AM

It created spans for the time when cordoning was done, the span attribute also had info of annotation and labels applied on the node

Screenshot 2026-03-24 at 10 20 48 AM

2. Node drainer's draining was cancelled because of manual uncordon of node: the cancel latest unquarantine span was generated with the reason to cancel the breakfix pipeline.

Screenshot 2026-03-24 at 10 22 30 AM

3. Breakfix pipeline was cancelled as healthy event was received: the drain_session span showed how long node-drainer waited for draining to be finished.

Screenshot 2026-03-24 at 10 23 50 AM

 We can see handle_cancelled_event span and user_pod_eviction_status span which shows when was the draining cancelled when did it update the status in DB and what was final status updated in DB.

Screenshot 2026-03-24 at 10 25 20 AM

4. Healthy event which cancelled the above breakfix pipeline: 
We can see the span for every action done when healthy event was received, new span for uncordoning the node, removing the annotation from node. Also, new spans for each k8s API call to track the latency in each API call

Screenshot 2026-03-24 at 10 29 24 AM

5. Non fatal event injected: 
Spans showed the action taken by platform-connector when non-fatal event was received, new span created for update node event as it received non-fatal event. It also tracked the event reason and type that was updated to the node.

Screenshot 2026-03-24 at 10 33 58 AM

6. Node-drainer didn't receive event as it was crashing: Node-drainer didn't receive event as it was crashing for ~4min. When we restarted the pod, it picked up the event and start processing it. The node-drainer span was correctly linked to FQ span.  

Screenshot 2026-03-24 at 10 37 42 AM

7. Cold start of node-drainer: Node-drainer was waiting for draining to be finished and then it crashed for ~4 min. When we restarted the pod, it picked the event and linked the generated span to the node-drainer's span.

Screenshot 2026-03-24 at 10 39 46 AM

The error during the draining was catch by span and attributes were added to it to show the error message.

Screenshot 2026-03-24 at 10 40 27 AM
 

8. Injected burst of event in syslog monitor: injected 3000 event in burst on the same node, the event took max ~18min to complete the processing. This delay was mainly in platform-connector k8s.fetch_and_process_health_metric as it queues the event and process them one by one.

Screenshot 2026-03-24 at 10 45 20 AM

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive OpenTelemetry tracing design: end-to-end cross-service trace continuity, canonical span naming and attributes per module, span/linking guidance for async boundaries, trace-context propagation patterns (including CR annotation patterns and gRPC interceptor propagation), datastore and service-layer tracing guidance, log–trace correlation pattern for injecting trace IDs into structured logs, and OTLP export/configuration guidance.


docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,417 @@
...
+                                               ▼
+┌──────────────────────────────────────────────────────────────────────────────-┐
+│  Grafana Tempo → Grafana Dashboards (dashboards.telemetry.dgxc.ngc.nvidia.com)│
+│  - Trace explorer, service map, latency analysis                              │
+└──────────────────────────────────────────────────────────────────────────────-┘
+```
+
+**Trace flow summary:**
+
+1. **Ingestion**: Each NVSentinel module exports spans via OTLP over gRPC to `dgxc-alloy-gateway.dgxc-alloy.svc.cluster.local:4317`. No authentication is required from NVSentinel to the gateway (in-cluster, `OTEL_EXPORTER_OTLP_INSECURE=true`).
```
KaivalyaMDabhadkar: Is alloy deployed as a daemonset or a deployment in the cluster? Will each node export traces to a single alloy deployment pod in the cluster or it's local daemonset pod?

tanishagoyal2: alloy-gateway is ds in the cluster where pods are scheduled to the monitoring nodes.

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,465 @@
...
+                                  ▼
+                         ┌──────────────────────────────────────────────────────┐
+                         │  Grafana Dashboards                                   │
+                         │  - Trace explorer, service map, latency analysis      │
+                         └──────────────────────────────────────────────────────┘
+```
+
+**Trace flow summary:**
+
+1. **Ingestion**: The OTel Collector (e.g. dgxc-alloy, Grafana Alloy, or the OpenTelemetry Collector) run separately. To connect NVSentinel to it, set `global.tracing.endpoint` to the collector's OTLP gRPC address. Helm injects this value into each module as `OTEL_EXPORTER_OTLP_ENDPOINT`, so every module exports spans via OTLP over gRPC to that collector. All NVSentinel traces therefore flow to the same collector for batching and forwarding.
```
lalitadithya:
```suggestion
1. **Ingestion**: The OTel Collector (e.g. alloy, Grafana Alloy, or the OpenTelemetry Collector) run separately. To connect NVSentinel to it, set `global.tracing.endpoint` to the collector's OTLP gRPC address. Helm injects this value into each module as `OTEL_EXPORTER_OTLP_ENDPOINT`, so every module exports spans via OTLP over gRPC to that collector. All NVSentinel traces therefore flow to the same collector for batching and forwarding.
```

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,465 @@
...
+### Fault-Quarantine
+
+**Root span:** `fault_quarantine.process_event`
+
+| What is tracked | Span / Attribute | Use case |
+|-----------------|-----------------|-------------------|
+| Cordon + taint applied | `fault_quarantine.apply_quarantine` child span; `fault_quarantine.action.cordon` (bool), `fault_quarantine.action.taint` (bool), `fault_quarantine.event.processing_status = "Quarantined"` | Did this health event lead to cordoning and tainting of the node? |
+| Unquarantine | `fault_quarantine.perform_uncordon` child span; `fault_quarantine.event.processing_status = "UnQuarantined"` | Did this health event lead to uncordoning/untainting of the node? |
+| Ruleset evaluation | `fault_quarantine.evaluate_rulesets` child span | Which rulesets were evaluated for this event? How long did evaluation take? |
+| DB status write trigger span (used as link target by node-drainer) | `fault_quarantine.db.update_status`; `span_ids.fault_quarantine` written to MongoDB | When did fault-quarantine persist status so node-drainer could pick up this event? |
```
lalitadithya: Shouldn't this be in the store client?

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,465 @@
...
+| Errors | `fault_quarantine.error.type`, `fault_quarantine.error.message` | What went wrong in fault-quarantine and why? |
+
+### Node Drainer
+
+**Root spans:** `node_drainer.enqueue_event` (change-stream entry) + `node_drainer.drain_session` (worker queue, long-lived for entire drain)
+
+| What is tracked | Span / Attribute | Use case |
+|-----------------|-----------------|-------------------|
+| Initial status set to InProgress in MongoDB | `node_drainer.set_initial_status_and_enqueue` → `node_drainer.db.update_status` child span; `node_drainer.eviction_status = "InProgress"` | Did drain start for this event? When was status set to InProgress? |
+| Final eviction status written to MongoDB | `node_drainer.update_user_pods_eviction_status` child span; `drain.status` = "Succeeded", "Failed", "Cancelled", "AlreadyDrained" | Did drain succeed, fail, get cancelled, or was the node already drained? |
```
lalitadithya: Same question as earlier

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,465 @@
...
+
+```json
+{
+  "_id": "...",
+  "trace_id": "abc123",
+  "span_ids": {
+    "platform_connector": "<span-id-of-platform_connector.db.insert>",
+    "fault_quarantine":   "<span-id-of-fault_quarantine.db.update_status>",
+    "node_drainer":       "<span-id-of-node_drainer active span>"
+  },
```
lalitadithya: This can be in the metadata field right?

docs/designs/028-OTEL-traces.md
tanishagoyal2: I didn't understand, Why should this be in store client? My goal here was to track when Fault-quarantined finished processing of event and updated status in db

docs/designs/028-OTEL-traces.md
tanishagoyal2: I had that idea as well. The reason for keeping this data separate from healthEvent is that healthEvent stores data related to node, unhealthy/healthy condition. Trace ID and Span ID is for observability purpose, for us to debug the issue in handling the event.

docs/designs/028-OTEL-traces.md
lalitadithya: yes, but if you put it in the store client any new db queries will automatically be covered in the traces

docs/designs/028-OTEL-traces.md
lalitadithya: Sure, but I think we should have it contained in the health event or health event status. Having the root span in the metadata would mean that we can then link events exported from the cloud exporter to the trace. So maybe the answer here is that we will add the root trace ID to the metadata and then rest in the status?

tanishagoyal2: Yeah, we can do that

docs/designs/028-OTEL-traces.md
tanishagoyal2: hmm, thats a good idea. In this way all the db queries will be tracked

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,465 @@
...
+
+| What is tracked | Span Attributes | Use case |
+|-----------------|----------------|-------------------|
+| Reboot signal | `janitor_provider.reboot.sent` (bool), `janitor_provider.reboot.node`, `janitor_provider.reboot.request_ref`, `janitor_provider.reboot.duration_ms` | Was a reboot request sent to the CSP? Which node? How long did the call take? |
+| Node ready check | `janitor_provider.node_ready.ready` (bool), `janitor_provider.reboot.node`, `janitor_provider.reboot.request_ref` | Did the CSP report the node as ready? For which reboot request? |
+| Terminate signal | `janitor_provider.terminate.sent` (bool), `janitor_provider.terminate.node`, `janitor_provider.terminate.request_ref` | Was a terminate request sent to the CSP? Which node? |
+| Errors | `janitor_provider.error.type` = "grpc_error"/"csp_api_error", `janitor_provider.error.message` | Did the failure come from gRPC or the CSP API? What was the error? |
+
+## Trace Context Propagation
```
lalitadithya: Can you also add some details on how this will be done with golang? Would we use context or some other mechanism?

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,465 @@
...
+
+- **Span attributes** — Key-value metadata attached to a span (e.g. `node_drainer.node_name`, `fault_remediation.cr.name`, `janitor.gpureset.processing_status`). They describe what happened during that span without needing to search logs. Attributes are used for filtering and grouping in tracing UIs (e.g. Grafana Tempo) and for understanding why a span succeeded or failed.
+
+- **Span links** — A span link connects spans that are causally related but not in a strict parent-child relationship. Links are the OpenTelemetry-standard way to model asynchronous fan-out/fan-in and queue/change-stream handoffs where work may run in parallel or outlive the trigger span.
+
+## How Can Traces Be Helpful?
+- **Module-level performance**: How long does fault-quarantine take vs node-drainer vs fault-remediation?
+- **Database query latency**: Time spent in MongoDB aggregation pipelines (health-events-analyzer)
```
lalitadithya: Can we also track latency of Kubernetes API server calls this way? What are the other external API that we invoke for which we can track the latency as well?

tanishagoyal2: Yeah, sure. We can track CSP API latency as well that we call in Janitor-provider.

docs/designs/028-OTEL-traces.md
tanishagoyal2: For spans in same service: We will pass context while creating spans for actions done in same service (similar to how we pass context  from calling func to called function). 
 ```
trace.SpanFromContext(ctx)
```
For context propagation between the services: we are storing span ID in db. through which we will connect services together using the trace links.

Comment (lalitadithya) [reply docs/designs/028-OTEL-traces.md]: sure, can you please add details about this to the document? Also, some details about what points in the code is this context used to export a trace would be helpful

Comment (tanishagoyal2) [reply docs/designs/028-OTEL-traces.md]: sure, I'll add details

File: docs/designs/028-OTEL-traces.md [hunk 8]
```diff
@@ -0,0 +1,772 @@
...
+- **Granular level performance**: Time spent in cordon, taint, drain, and CR creation operations
+- **Event lifecycle tracking**: From detection → ingestion → quarantine → drain → remediation
+- **Remediation status**: Was the last remediation succeeded or failed? 
+- **Multi-module coordination**: See which modules are actively processing the same event like fault-quarantine and health-events-analyzer pick up event at the same time
+- **Context preservation**: All relevant event metadata is attached to spans, eliminating the need to login to the cluster and to search logs in each module
+- **Concurrent processing**: Understand how multiple events are processed simultaneously
+- **Kubernetes API call latency**: Time spent in individual K8s API calls — cordon, taint, uncordon, pod eviction, CR creation, node condition updates, status patches, node lock acquisition, service teardown/restore, and job creation
+- **CSP API call latency**: Time spent in CSP provider calls — reboot signal, node-ready check, and terminate signal
+- **Circuit breaker activity**: Monitor when circuit breaker is tripped
```
lalitadithya, line 28: I don't follow: how can this tracked using traces?

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,772 @@
...
+| fault-remediation → janitor | Kubernetes CR creation / watch | CR annotations: `nvsentinel.nvidia.com/trace-id` and `nvsentinel.nvidia.com/span-id` |
+| janitor → janitor-provider | Synchronous gRPC | W3C `traceparent` header via `otelgrpc` client/server interceptors (automatic) |
+
+**MongoDB-based propagation flow:**
+
+```
+platform-connector
+  1. Creates root trace (trace_id: abc123)
+  2. Writes trace_id into healthevent.metadata["trace_id"]
+  3. Writes span_ids.platform_connector into the MongoDB document
```
lalitadithya: what if the pod is restarted before this write happens?

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,772 @@
...
+```
+
+All modules share the same trace ID (`abc123`).
+
+MongoDB document structure:
+
+```json
+{
+  "_id": "...",
+  "span_ids": {
```
lalitadithya: can we put these in the status instead to retain kube CR like semantics?

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,772 @@
...
+
+## What Will be Tracked from Each Module
+
+### Store Client Tracing
+
+Health event database operations are traced at the `DatabaseClient` and `HealthEventStore` layers in `store-client/`. Every module that reads or writes a health event automatically gets a child span with latency, operation metadata, and error tracking — no per-module instrumentation needed. Only DB operations related to health events in the breakfix pipeline are traced.
+
+The table below lists every health-event DB operation that will be traced, the store method involved, which module(s) call it, and the span attributes recorded. Modules use either the `DatabaseClient` or `HealthEventStore` interfaces in `store-client/` — tracing is added once in those layers and automatically covers every caller.
+
+| Span name | Method | Called by | Attributes | Use case |
```
lalitadithya, line 420: how does the store client get the trace IDs?

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,772 @@
...
+
+1. Create `commons/pkg/tracing/http_transport.go` — Add a new `conditionalHTTPTracingRT` struct implementing `http.RoundTripper`. It wraps any existing transport and, on each `RoundTrip`, checks whether `req.Context()` carries a valid parent span. If no active trace exists, the request passes through unchanged. If an active trace exists, it creates a child span named `http.client.<method>`, executes the request, records span attributes and errors, then ends the span. Export `NewConditionalHTTPTracingRoundTripper(next http.RoundTripper) http.RoundTripper` for use by the audit logger.
+
+2. Update `commons/pkg/auditlogger/roundtripper.go` — In `NewAuditingRoundTripper`, wrap the delegate transport with `tracing.NewConditionalHTTPTracingRoundTripper(delegate)` before assigning it. This ensures every module that already uses the auditing wrapper automatically gets conditional HTTP tracing with no additional wiring.
```
lalitadithya, line 450: what if we don't want audit logs, but want tracing?

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,772 @@
...
+| Ruleset evaluation | `fault_quarantine.evaluate_rulesets` child span | Which rulesets were evaluated for this event? How long did evaluation take? |
+| DB status write (used as link target by node-drainer) | `span_ids.fault_quarantine` written to MongoDB; DB latency tracked automatically by store client span (`db.health_event.update_quarantine_status`) | When did fault-quarantine persist status so node-drainer could pick up this event? How long did the DB write take? |
+| Cancellation of quarantining events on manual uncordon/untaint | `fault_quarantine.cancel_latest_quarantining_events` root span (linked to `fault_quarantine.db.update_status` of the cancelled event) | Were quarantining events cancelled for this event due to manual uncordon/untaint? |
+| Errors | `fault_quarantine.error.type`, `fault_quarantine.error.message` | What went wrong in fault-quarantine and why? |
```
lalitadithya, line 532: aren't error already written to logs? if something fails, shouldn't the span for that be marked as failed, or do we need a separate span for this?

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,772 @@
...
+| Drain scope (partial vs full) | `node_drainer.drain.scope` = "partial" or "full"; `node_drainer.partial_drain.entity_type`, `node_drainer.partial_drain.entity_value` — set on `drain_session` at session end | Was this a partial or full drain? What was the partial-drain target (entity type/value)? |
+| Pods targeted at phase entry | `node_drainer.immediate_eviction_pods`, `node_drainer.allow_completion_pods`, `node_drainer.delete_after_timeout_pods` — set on `drain_session` at session end | Which pods were in each phase (immediate eviction, allow completion, delete after timeout)? |
+| Phase durations (wall-clock) | `node_drainer.immediate_eviction_duration_s`, `node_drainer.allow_completion_duration_s`, `node_drainer.delete_after_timeout_duration_s` — set on `drain_session` at session end | How long did each drain phase take? |
+| Force-deleted pods | `node_drainer.force_deleted_pods` (comma-separated), `node_drainer.pods_force_deleted_count` — set on `drain_session` | Which pods were force-deleted and how many? |
```
lalitadithya: won't this cause cardinality explosion if we put pod names here? Or does this not apply?

docs/designs/028-OTEL-traces.md
```diff
@@ -0,0 +1,772 @@
...
+| GPU targets (UUIDs and PCI bus IDs) | `janitor.gpureset.gpu_uuids` (comma-separated), `janitor.gpureset.pci_bus_ids` (comma-separated) | Which GPU(s) are being reset (UUIDs and PCI bus IDs)? |
+| Node lock | `janitor.node_lock.acquired` (bool), `janitor.node_lock.node`, `janitor.k8s.node_lock.duration_ms` (float) | Was the node lock acquired? Which node? How long did the K8s API call take? |
+| Service teardown / restore | `janitor.services.teardown.success` (bool), `janitor.services.restore.success` (bool), `janitor.k8s.services.teardown.duration_ms` (float), `janitor.k8s.services.restore.duration_ms` (float) | Did service teardown and restore succeed? How long did each K8s API call take? |
+| Reset job | `janitor.reset_job.created` (bool), `janitor.reset_job.name`, `janitor.reset_job.completed` (bool), `janitor.reset_job.failed` (bool), `janitor.k8s.reset_job.create.duration_ms` (float) | Was a reset job created? Did it complete or fail? How long did the K8s job creation API call take? |
```
lalitadithya, line 611: what about traces for the GPU reset job itself? Is that possible?

lalitadithya (commented):
in the PR description can you also please add all the use cases you tested out as a part of your POC and the outcomes you saw?

docs/designs/028-OTEL-traces.md
tanishagoyal2: This will be done by adding one span if circuit breaker is tripped during event processing.
This will tell us that the event was not processed because circuit breaker was tripped. And in the dashboard we can filter out the events for which the span attribute `fault_quarantine.circuit_breaker.tripped` is true

docs/designs/028-OTEL-traces.md
tanishagoyal2: If platform-connector pod restart, then isn't it expected that it will retry to push the event in db? This should already be handled in platform-connector.
The platform-connector is adding span_ids to the health-event object and then its inserting the whole event to db. It is not updating the already inserted event.

I think I need to update the above written statement. It should be 
`Writes span_ids into healthevent object`

docs/designs/028-OTEL-traces.md
lalitadithya: but if the circuit breaker is tripped then no new health events are read from mongodb change stream

docs/designs/028-OTEL-traces.md
tanishagoyal2: Store client gets the trace ID from the context. It will check if current context already has valid trace going on or not. If yes, then it will add span for db query otherwise it will skip it.
As we only want to track db calls done for health-event related operation that's why checking for valid trace is best option as it will avoid noice of traces for general db queries done like resumeToken find operation, maintenance event insertion call etc.

docs/designs/028-OTEL-traces.md
tanishagoyal2: Yes, if something fails the span will show the error with red dot. I'll not be adding new span for error. The errors will be tracked in the already going on trace as span attributes.

docs/designs/028-OTEL-traces.md
tanishagoyal2: It should not create any issue, as force-deleted pods will be shown as span attribute when draining session has been finished so, multiple spans will not be created for each deletion call.
But if I think now, which pods were affected is not an useful info for us. We can avoid adding this detail in span.

docs/designs/028-OTEL-traces.md
tanishagoyal2: right, but for the last event received event, it can show the span for circuit breaker

docs/designs/028-OTEL-traces.md
tanishagoyal2: Yeah, sure we can do that

docs/designs/028-OTEL-traces.md
tanishagoyal2: I checked the code, we are not adding NewAuditingRoundTripper conditionally. it is independent from auditLogging flag.

docs/designs/028-OTEL-traces.md
tanishagoyal2: Do you mean to track the reset_job status? or to track every action done to perform the reset

tanishagoyal2: If later, then we are already planning to track every action done for reset job like to track when job status was updated completed and when restoreServices was called

lalitadithya: latter - but what about the operations done in the job?

tanishagoyal2: I checked on it, the span creation using context (as we are following for all modules) won't be possible as gpu_reset script is written in bash not in GO. 
One possible way to track the span for it is to manually create the span and to send the generated span to otel-collector's endpoint. This manual work could cause many issues, as we would have to manually create the child-parent relationship, we have to generate the span ID and duration sec and have to send the generated span to the configured endpoint while making sure we are correctly following the span format.

If its really necessary to track the script operation then we can still try with mentioned option, otherwise we can skip it as manual actions are error-prone and don't align with the pattern we have for rest of the modules.

lalitadithya: I see, okay, then let's document it as a known limittation

docs/designs/028-OTEL-traces.md
lalitadithya: okay, can we confirm this works as a part of your tests?

tanishagoyal2: We will get span for the HTTP call made to update the circuit breaker to TRIPPED state

Screenshot 2026-03-24 at 1 09 01 PM

And when circuit-breaker is closed, FQ processed the event and new span were generated within the same trace

Screenshot 2026-03-24 at 1 08 27 PM

docs/designs/028-OTEL-traces.md
tanishagoyal2: added details in doc

docs/designs/028-OTEL-traces.md
tanishagoyal2: updated doc and remove this `node_drainer.force_deleted_pods` point

docs/designs/028-OTEL-traces.md
tanishagoyal2: Updated doc to add span IDs in healtheventstatus

lalitadithya (approved):

Issue #790, opened 2026-02-04
[Bug]: Fault remediation module level E2E tests are flaky


XRFXLP:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Bug Description

Every now and then FR module level reconciler tests fail: 
https://github.com/NVIDIA/NVSentinel/actions/runs/21671705374/job/62481010622?pr=772
https://github.com/NVIDIA/NVSentinel/actions/runs/21522366586/job/62016794718?pr=752

Even though MR touches nothing of FR

### Component

Fault Management

### Steps to Reproduce

Trigger pipeline enough times

### Environment

- NVSentinel version: v0.8.0
- Kubernetes version:
- Deployment method:

lalitadithya:
@tanishagoyal2 , is this the same issue that you were debugging with the HEA or a different issue?

PR #915, opened 2026-02-27
fix: resolve race condition in annotation read-modify-write (#790)

Gyan172004:
## Summary

The flaky e2e test failures in #790 were caused by a race condition in how we update node annotations.

In `checkExistingCRStatus`, we were calling `RemoveGroupFromState` inside a for loop  so each iteration would GET the node, delete one group, and WRITE it back. But if two of these run close together, the second one reads the old state (before the first write), and when it writes back, it overwrites what the first one did. So some group removals just get lost.

This also explains why it was flaky as Go randomizes map iteration order, so depending on timing and ordering, sometimes the overwrites happen and sometimes they don't.

**What I changed:**
- Added a `RemoveGroupsFromState` method that removes all groups in one GET-modify-WRITE instead of one at a time in a loop
- Wrapped `UpdateRemediationState`, `ClearRemediationState`, and `RemoveGroupsFromState` with `retry.RetryOnConflict` so if there's a conflict it just retries with fresh state
- Updated `checkExistingCRStatus` to collect groups first and then call the batch method once
- Also fixed error handling `checkExistingCRStatus` as errors were being silently logged instead of returned
- Added concurrency tests to reproduce and verify the fix

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added batch removal of multiple remediation groups in a single operation, exposed via the public API.

* **Bug Fixes**
  * Made remediation-state updates retry-safe and atomic to prevent conflicts under concurrent updates; reconciler now performs batched removals and surfaces errors for failed batches.

* **Tests**
  * Added concurrency tests validating safe concurrent group removals and mixed update/remove scenarios.


fault-remediation/pkg/annotation/annotation_test.go
```diff
@@ -219,3 +221,99 @@ func TestRemoveGroupFromState(t *testing.T) {
...
+	client := fake.NewClientBuilder().WithObjects(node).Build()
+	annotationManager := NodeAnnotationManager{client: client}
+
+	var wg sync.WaitGroup
+	for i := 0; i < 5; i++ {
+		wg.Add(1)
+		go func(idx int) {
+			defer wg.Done()
+			_ = annotationManager.RemoveGroupFromState(context.TODO(), nodeName, fmt.Sprintf("group-%d", idx))
+		}(i)
```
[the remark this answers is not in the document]
Gyan172004: these were written specifically for race condition tests using a fake k8s client so i think it's not gonna throw random errors. if something actually breaks the state assertions will catch it anyway please analyze is it worth changing?

fault-remediation/pkg/annotation/annotation_interface.go
```diff
@@ -32,6 +32,7 @@ type NodeAnnotationManagerInterface interface {
 	UpdateRemediationState(ctx context.Context, nodeName string, group string, crName string, actionName string) error
 	ClearRemediationState(ctx context.Context, nodeName string) error
 	RemoveGroupFromState(ctx context.Context, nodeName string, group string) error
```
KaivalyaMDabhadkar: Is this method now being used anywhere if we have added `RemoveGroupsFromState`? Can we get rid of it in that case?

fault-remediation/pkg/annotation/annotation.go
```diff
@@ -83,105 +84,143 @@ func (m *NodeAnnotationManager) GetRemediationState(
...
+			if err = m.client.Update(ctx, updatedNode); err != nil {
+				return err
+			}
+
+			slog.Info("Cleared remediation state annotation for node", "node", nodeName)
+
+			return nil
+		}
+
+		// Marshal to JSON
```
KaivalyaMDabhadkar: nit: can we get rid of obvious comments like these?

fault-remediation/pkg/annotation/annotation_interface.go
Gyan172004: It is currently being used at multiple places in tests, I will replace them with the new function there and remove this single group removal function.

fault-remediation/pkg/annotation/annotation.go
Gyan172004: Done

fault-remediation/pkg/annotation/annotation_interface.go
Gyan172004: Done

KaivalyaMDabhadkar (approved):

KaivalyaMDabhadkar (changes requested):
@Gyan172004 it seems there are some linter errors, can you address them?

Gyan172004:
@KaivalyaMDabhadkar @XRFXLP Could you please trigger the CI tests again when you get some time?
Thanks in advance.

KaivalyaMDabhadkar (approved):

Issue #954, opened 2026-03-03
[Bug]: fault-remediation resume token name comes as watcher-factory which is not correct.


deesharma24:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Bug Description

[Bug]: fault-remediation resume token name comes as watcher-factory which is not correct. the resume token name should be fault-remediation only.

### Component

Fault Management

### Steps to Reproduce

1. login to mongodb pod.
2. exec into the healts-events database and see  all resume token.
3. fault-remediation is represented as watcher-factory which is not right

### Environment

- NVSentinel version: 0.7.1
- Kubernetes version:
- Deployment method: helm chart

### Logs/Output

```
  {"time":"2026-02-26T11:55:55.346401359Z","level":"ERROR","msg":"Failed to watch change stream","module":"fault-remediation","version":"v0.7.1","client":"watcher-factory","error":"(ChangeStreamFatalError) Executor error during getMore :: caused by :: cannot resume stream; the
  resume token was not found. {_data: \"8269A0042C000000732B...\"}"}
```

PR #971, opened 2026-03-07
fix: store resume token under correct clientName in fault-remediation (Issue #954)

Gyan172004:
## Summary
Solves Issue #954 
`MongoWatcherFactory` had `clientName` hardcoded as `"watcher-factory"` so resume token was always stored under wrong name. Although postgres factory in the same code was already doing it right, had its own `extractClientName` helper reading from config.

Other components are fine since they pass `clientName` directly to the legacy adapter, only `fault-remediation` goes through the new `WatcherFactory` path so it was the only one affected.

Quick fix would be copy what postgres was doing into mongo as well. But same logic in two places felt off so moved it to `ClientName()` on `WatcherConfig` instead, deleted the postgres helper, both factories just call `config.ClientName()` now. Open to change if  i missed to understand this as good decision

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Refactor**
  * Change stream watchers now use configurable client names instead of hard-coded defaults for consistent identification across providers.
  * Watcher creation now requires a client name and will fail if it is not provided.

* **Chores**
  * Removed an unused helper and streamlined client-name handling across watcher implementations.
  * Initial configuration sets a default client name for the remediation service.

### DIFF
--- a/fault-remediation/pkg/initializer/init.go
+++ b/fault-remediation/pkg/initializer/init.go
@@ -165,6 +165,9 @@ func initDatastoreAndWatcher(
 	watcherConfig := watcher.WatcherConfig{
 		Pipeline:       pipeline,
 		CollectionName: "HealthEvents",
+		Options: map[string]interface{}{
+			"ClientName": "fault-remediation",
+		},
 	}
 
 	watcherInstance, err := watcher.CreateChangeStreamWatcher(ctx, ds, watcherConfig)
--- a/fault-remediation/pkg/initializer/init.go
+++ b/fault-remediation/pkg/initializer/init.go
@@ -165,9 +165,7 @@ func initDatastoreAndWatcher(
 	watcherConfig := watcher.WatcherConfig{
 		Pipeline:       pipeline,
 		CollectionName: "HealthEvents",
-		Options: map[string]interface{}{
-			"ClientName": "fault-remediation",
-		},
+		ClientName:     "fault-remediation",
 	}
 
 	watcherInstance, err := watcher.CreateChangeStreamWatcher(ctx, ds, watcherConfig)
--- a/store-client/pkg/datastore/providers/mongodb/watcher_factory.go
+++ b/store-client/pkg/datastore/providers/mongodb/watcher_factory.go
@@ -50,10 +50,10 @@ func (f *MongoWatcherFactory) CreateChangeStreamWatcher(
 	}
 
 	// Create the change stream watcher using the existing factory
-	watcher, err := mongoStore.CreateChangeStreamWatcher(ctx, "watcher-factory", mongoPipeline)
+	watcher, err := mongoStore.CreateChangeStreamWatcher(ctx, config.ClientName(), mongoPipeline)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create MongoDB change stream watcher: %w", err)
-	}
+	} 
 
 	return watcher, nil
 }
--- a/store-client/pkg/datastore/providers/mongodb/watcher_factory.go
+++ b/store-client/pkg/datastore/providers/mongodb/watcher_factory.go
@@ -53,7 +53,7 @@ func (f *MongoWatcherFactory) CreateChangeStreamWatcher(
 	watcher, err := mongoStore.CreateChangeStreamWatcher(ctx, config.ClientName(), mongoPipeline)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create MongoDB change stream watcher: %w", err)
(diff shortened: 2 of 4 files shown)


store-client/pkg/watcher/factory.go
```diff
@@ -37,6 +37,16 @@ type WatcherConfig struct {
...
 }
 
+func (c WatcherConfig) ClientName() string {
+	if c.Options != nil {
+		if name, ok := c.Options["ClientName"].(string); ok && name != "" {
+			return name
+		}
+	}
+
+	return "watcher-factory"
```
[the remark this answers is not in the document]
Gyan172004: Added the warn log.
For the typed field : same pattern was already in postgres with its local extractClientName helper, just moved it to WatcherConfig to avoid duplication. Only one caller right now so feels like overkill?

fault-remediation/pkg/initializer/init.go
```diff
@@ -165,6 +165,9 @@ func initDatastoreAndWatcher(
 	watcherConfig := watcher.WatcherConfig{
 		Pipeline:       pipeline,
 		CollectionName: "HealthEvents",
+		Options: map[string]interface{}{
```
KaivalyaMDabhadkar: Can't we just pass the `ClientName` as a typed field instead? That will be more convenient:
```
type WatcherConfig struct {
    Pipeline       datastore.Pipeline `json:"pipeline,omitempty"`
    CollectionName string             `json:"collectionName"`
    ClientName     string             `json:"clientName"`
    ResumeToken    []byte             `json:"resumeToken,omitempty"`
    Options        map[string]interface{} `json:"options,omitempty"`
}
```

We can then directly initialize it like this:

```
watcherConfig := watcher.WatcherConfig{
    Pipeline:       pipeline,
    CollectionName: "HealthEvents",
    ClientName:     "fault-remediation",
}
```

store-client/pkg/datastore/providers/mongodb/watcher_factory.go
```diff
@@ -50,7 +50,7 @@ func (f *MongoWatcherFactory) CreateChangeStreamWatcher(
 	}
 
 	// Create the change stream watcher using the existing factory
-	watcher, err := mongoStore.CreateChangeStreamWatcher(ctx, "watcher-factory", mongoPipeline)
+	watcher, err := mongoStore.CreateChangeStreamWatcher(ctx, config.ClientName(), mongoPipeline)
```
KaivalyaMDabhadkar: With the previously suggested implementation, we would need to have a check for an empty clientname as well:
```
func CreateChangeStreamWatcher(
    ctx context.Context, ds datastore.DataStore, config WatcherConfig,
) (datastore.ChangeStreamWatcher, error) {
    if config.ClientName == "" {
        return nil, fmt.Errorf("ClientName is required in WatcherConfig")
    }
    // ...
}
```

fault-remediation/pkg/initializer/init.go
Gyan172004: Yes, this looks better. The earlier PostgreSQL implementation was passing ClientName via Options, so I kept the same approach initially. But having it as a struct field is cleaner.

Gyan172004: Done

store-client/pkg/datastore/providers/mongodb/watcher_factory.go
Gyan172004: Done

KaivalyaMDabhadkar:
@Gyan172004 Just to confirm, Did you also check manually that the resume token for the FR module is being set properly now in mongodb with the new changes?

Gyan172004:
> @Gyan172004 Just to confirm, Did you also check manually that the resume token for the FR module is being set properly now in mongodb with the new changes?

Yes @KaivalyaMDabhadkar , I had already verified it manually in MongoDB, FR was using `clientName: "fault-remediation"` and there were no entries with `watcher factory`

KaivalyaMDabhadkar (approved):

PR #1005, opened 2026-03-10
feat: gRPC TLS + SA token auth for CSP provider connection

jtschelling:
## Summary
- Add TLS support for the janitor → janitor-provider gRPC connection with configurable CA bundle loading
- Add client-side projected SA token interceptor that reads the token on every call (handles K8s auto-rotation) and attaches it as Bearer metadata
- Add server-side TokenReview interceptor in janitor-provider that validates incoming tokens via K8s TokenReview API with audience matching
- Add server-side TLS to janitor-provider gRPC server with cert-manager Certificate and Helm chart wiring
- Wire Helm chart values for both janitor (TLS config, projected SA token volume) and janitor-provider (TLS cert, auth config, RBAC for tokenreviews)
- Enable TLS + auth in the Tilt dev environment and add an end-to-end integration test

### Client side (janitor)
- `pkg/client/grpc_tls.go` — `NewCSPProviderDialOptions()` builds TLS credentials from CA path
- `pkg/client/grpc_auth.go` — `TokenInterceptor()` re-reads projected SA token file on each gRPC call
- `pkg/config/` — Added `CSPProviderCAPath`, `CSPProviderInsecure`, `CSPProviderTokenPath` with cascading defaults
- Both controllers updated to use new dial options + conditional token interceptor
- Helm chart: CA secret volume mount, projected SA token volume with audience binding

### Server side (janitor-provider)
- `main.go` — TLS server credentials from `TLS_CERT_PATH`/`TLS_KEY_PATH` env vars; rejects partial TLS config and auth-without-TLS (fail closed)
- `pkg/auth/token_review.go` — `TokenReviewInterceptor()` extracts Bearer token from gRPC metadata, validates via K8s TokenReview API
- Audience values are trimmed to handle comma-separated lists with whitespace
- `charts/janitor-provider/templates/certificate.yaml` — cert-manager Certificate for provider gRPC TLS
- `charts/janitor-provider/templates/deployment.yaml` — TLS cert volume mount + env vars
- ClusterRole updated with `authentication.k8s.io/tokenreviews` create permission

### Dev environment (Tilt)
- `values-tilt.yaml` — Enable TLS + auth for both janitor and janitor-provider
- `tilt/Tiltfile` — Add provider cert to cert-manager objects, wait for cert before starting services

### Tests
- Unit tests for TLS dial options, client token interceptor, server TokenReview interceptor
- `tests/helpers/kube.go` — Added `GetCRCondition()` and `WaitForCRByName()` helpers
- `tests/janitor_test.go` — `TestJanitorRebootWithTLSAuth` integration test validates the full TLS + auth chain end-to-end (janitor → TLS → provider → kind CSP → reboot complete)

## Test plan
- [x] All existing unit tests pass
- [x] New unit tests pass (TLS, auth interceptors)
- [x] Lint passes
- [x] `devenv env start nvsentinel` — Tilt deploys successfully with TLS + auth
- [x] Provider logs: `"gRPC TLS enabled"` + `"gRPC TokenReview auth enabled"`
- [x] Janitor logs: `"CSP provider gRPC auth enabled"` + `insecure: false`
- [x] Integration test: `go test -tags=arm64_group -run TestJanitorRebootWithTLSAuth -v` — PASS
- [x] RebootNode CR completes with `SignalSent=True` and `NodeReady=True`

Refs: JSC-55


janitor-provider/main.go
```diff
@@ -145,7 +147,19 @@ func run() error {
...
 	}
 
-	svr := grpc.NewServer()
+	var serverOpts []grpc.ServerOption
+
+	if audiences := os.Getenv("AUTH_AUDIENCES"); audiences != "" {
+		auds := strings.Split(audiences, ",")
+		serverOpts = append(serverOpts,
+			grpc.UnaryInterceptor(
+				auth.TokenReviewInterceptor(k8sClient, auds)))
```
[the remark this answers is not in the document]
jtschelling: Fixed in e675634c — audiences are now trimmed with `strings.TrimSpace` and empty entries are filtered out.

janitor-provider/main.go
```diff
@@ -145,7 +149,29 @@ func run() error {
...
+
+		slog.Info("gRPC TokenReview auth enabled",
+			"audiences", auds)
+	}
```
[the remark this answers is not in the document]
jtschelling: Fixed in e675634c — the server now rejects partial TLS config (only one of cert/key set) and refuses to start auth without TLS enabled. Both cases return a clear error and prevent startup.

tests/janitor_test.go
```diff
@@ -181,6 +181,72 @@ func TestJanitorWebhookRejectsNonExistentNode(t *testing.T) {
...
+	feature.Assess("RebootNode completes successfully over TLS with auth", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		client, err := c.NewClient()
+		require.NoError(t, err, "failed to create kubernetes client")
+
+		crName := fmt.Sprintf("reboot-tls-%s", selectedNodeName)
+		_, err = helpers.CreateRebootNodeCR(ctx, client, selectedNodeName, crName)
+		require.NoError(t, err, "RebootNode CR creation should succeed")
+
+		completedCR := helpers.WaitForCR(ctx, t, client, selectedNodeName, helpers.RebootNodeGVK)
+		require.NotNil(t, completedCR, "RebootNode should complete")
```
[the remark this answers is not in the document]
jtschelling: Fixed in e675634c — added `WaitForCRByName` helper that matches on `metadata.name` instead of `spec.nodeName`, and the test now uses it to wait on the exact CR it created.

tests/janitor_test.go
```diff
@@ -181,6 +181,72 @@ func TestJanitorWebhookRejectsNonExistentNode(t *testing.T) {
...
+
+		// Verify SignalSent condition is True (proves the gRPC call over TLS+auth succeeded)
+		signalSent := helpers.GetCRCondition(completedCR, "SignalSent")
+		require.NotNil(t, signalSent, "SignalSent condition should exist")
+		assert.Equal(t, "True", signalSent["status"], "SignalSent should be True")
+
+		// Verify NodeReady condition exists (reboot completed)
+		nodeReady := helpers.GetCRCondition(completedCR, "NodeReady")
+		require.NotNil(t, nodeReady, "NodeReady condition should exist")
```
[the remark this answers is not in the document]
jtschelling: Fixed in e675634c — now asserts `NodeReady` status is `"True"`, matching the `SignalSent` check.

distros/kubernetes/nvsentinel/charts/janitor/templates/deployment.yaml
```diff
@@ -126,6 +136,24 @@ spec:
...
         {{- end }}
+        {{- if and .Values.config.cspProvider .Values.config.cspProvider.tls .Values.config.cspProvider.tls.enabled }}
+        - name: csp-ca
+          secret:
+            secretName: {{ .Values.config.cspProvider.tls.caSecretName }}
+            items:
+              - key: ca.crt
+                path: ca.crt
+            defaultMode: 420
+        {{- end }}
```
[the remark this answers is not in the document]
jtschelling: Acknowledged — the CA comes from a cert-manager Certificate with 90-day duration and 30-day renewBefore, so the window for stale trust is limited. The janitor also restarts on any Helm upgrade. Adding a checksum annotation is a nice-to-have but I'll defer it to avoid scope creep in this PR.

janitor-provider/pkg/auth/token_review.go
```diff
@@ -0,0 +1,130 @@
...
+
+	result, err := k8sClient.AuthenticationV1().TokenReviews().Create(
+		ctx, review, metav1.CreateOptions{},
+	)
+	if err != nil {
+		slog.Error("TokenReview API call failed", "error", err)
+
+		return status.Errorf(codes.Internal,
+			"token validation failed: %v", err)
+	}
```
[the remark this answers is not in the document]
jtschelling: Good catch — mapping transient apiserver errors to `codes.Unavailable` would let the janitor retry. However, this changes error semantics in the auth interceptor and deserves its own focused PR with tests for the various failure modes. Tracking as a follow-up.

janitor-provider/pkg/auth/token_review.go
```diff
@@ -0,0 +1,130 @@
...
+		return status.Error(codes.Unauthenticated,
+			fmt.Sprintf("token not authenticated: %s",
+				result.Status.Error))
+	}
+
+	slog.Info("Request authenticated",
+		"user", result.Status.User.Username,
+		"audiences", result.Status.Audiences)
+
+	return nil
```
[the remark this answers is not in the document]
jtschelling: Audience-scoped projected SA tokens already limit the blast radius — only tokens minted with the `nvsentinel-csp-provider` audience are accepted, and that audience is only projected onto the janitor's service account. Adding an explicit SA identity allowlist is a worthwhile hardening measure but it's additive security, not a bug in the current implementation. Will track as a follow-up.

janitor/pkg/config/config.go
```diff
@@ -63,6 +66,12 @@ type RebootNodeControllerConfig struct {
 	Exclusions []metav1.LabelSelector
 	// CSPProviderHost is the host of the CSP provider
 	CSPProviderHost string
+	// CSPProviderCAPath is the path to the CA bundle PEM file for TLS verification
+	CSPProviderCAPath string
+	// CSPProviderInsecure skips TLS when true (for local development)
+	CSPProviderInsecure bool
+	// CSPProviderTokenPath is the path to the SA token file for gRPC auth
+	CSPProviderTokenPath string
```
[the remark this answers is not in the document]
jtschelling: Valid concern, but this touches the config inheritance logic across all controllers (RebootNode, TerminateNode, GPUReset) and is a significant refactor. The current per-controller override pattern works correctly for this PR's use case since we're enabling TLS globally, not selectively per-controller. Will address the tri-state `*bool` refactor in a separate PR.

janitor/pkg/config/default.go
```diff
@@ -105,6 +105,32 @@ func applyCSPProviderHostDefaults(config *Config) {
...
+		config.RebootNode.CSPProviderInsecure = config.Global.CSPProviderInsecure
+	}
+
+	if len(config.TerminateNode.CSPProviderCAPath) == 0 {
+		config.TerminateNode.CSPProviderCAPath = config.Global.CSPProviderCAPath
+	}
+
+	if !config.TerminateNode.CSPProviderInsecure {
+		config.TerminateNode.CSPProviderInsecure = config.Global.CSPProviderInsecure
+	}
```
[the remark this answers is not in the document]
jtschelling: Same as the above comment — the `*bool` tri-state refactor for `CSPProviderInsecure` will be addressed together with the config override changes in a follow-up PR.

janitor-provider/main.go
```diff
@@ -145,7 +149,29 @@ func run() error {
...
-	svr := grpc.NewServer()
+	var serverOpts []grpc.ServerOption
+
+	if certPath, keyPath := os.Getenv("TLS_CERT_PATH"), os.Getenv("TLS_KEY_PATH"); certPath != "" && keyPath != "" {
+		cert, err := tls.LoadX509KeyPair(certPath, keyPath)
+		if err != nil {
+			return fmt.Errorf("failed to load TLS key pair: %w", err)
+		}
+		tlsCfg := &tls.Config{Certificates: []tls.Certificate{cert}, MinVersion: tls.VersionTLS12}
+		serverOpts = append(serverOpts, grpc.Creds(credentials.NewTLS(tlsCfg)))
```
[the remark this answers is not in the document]
jtschelling: Good point for production — cert-manager rotates at the 60-day mark (renewBefore: 30 days of the 90-day duration), and the kubelet updates the mounted secret automatically. Adding `tls.Config` with `GetCertificate` for live reload is a worthwhile enhancement but it's not blocking since the cert has a long lifetime and pod restarts (deploys, scaling) will pick up new certs. Will track as a follow-up.

tilt/Tiltfile
```diff
@@ -238,14 +239,20 @@ local_resource(
...
 k8s_resource(
     'janitor',
-    resource_deps=['wait-for-janitor-cert'],
+    resource_deps=['wait-for-janitor-cert', 'wait-for-janitor-provider-cert'],
 )
 
 k8s_resource(
     'janitor-provider',
-    resource_deps=['janitor'],
+    resource_deps=['janitor', 'wait-for-janitor-provider-cert'],
```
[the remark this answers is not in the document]
jtschelling: The Tiltfile is only used in the devenv (local development) where TLS is always enabled via `values-tilt.yaml`. Adding conditional Starlark logic for a configuration that never occurs in this context would add complexity for no practical benefit. If someone disables TLS in a custom values file, the `kubectl wait` timeout message will clearly indicate the certificate doesn't exist.

distros/kubernetes/nvsentinel/charts/janitor-provider/values.yaml
```diff
@@ -69,6 +69,26 @@ service:
 
 extraEnv: []
 
+# gRPC TLS configuration for the provider server
+tls:
+  # Enable TLS for the gRPC server
+  enabled: false
```
lalitadithya: it should be enabled by default; customers can choose to turn it off if required.

janitor-provider/pkg/auth/token_review.go
```diff
@@ -0,0 +1,130 @@
...
+) grpc.UnaryServerInterceptor {
+	return func(
+		ctx context.Context,
+		req any,
+		info *grpc.UnaryServerInfo,
+		handler grpc.UnaryHandler,
+	) (any, error) {
+		token, err := extractBearerToken(ctx)
+		if err != nil {
+			return nil, err
```
lalitadithya: we should wrap errors

janitor-provider/main.go
```diff
@@ -145,7 +149,46 @@ func run() error {
...
+		slog.Info("gRPC TLS enabled", "certPath", certPath)
+	}
+
+	if audiences := os.Getenv("AUTH_AUDIENCES"); audiences != "" {
+		if !tlsEnabled {
+			return fmt.Errorf("AUTH_AUDIENCES requires TLS to be enabled (set TLS_CERT_PATH and TLS_KEY_PATH)")
```
lalitadithya: why so? say a customer is using a service mesh that already does TLS, maybe they just want auth?

janitor-provider/main.go
```diff
@@ -145,7 +149,46 @@ func run() error {
...
+	var serverOpts []grpc.ServerOption
+
+	certPath, keyPath := os.Getenv("TLS_CERT_PATH"), os.Getenv("TLS_KEY_PATH")
+	tlsEnabled := certPath != "" && keyPath != ""
+
+	if certPath != "" != (keyPath != "") {
+		return fmt.Errorf("both TLS_CERT_PATH and TLS_KEY_PATH must be set, got cert=%q key=%q", certPath, keyPath)
+	}
+
+	if tlsEnabled {
```
lalitadithya, line 161: I'm also thinking if we can just do mTLS, maybe that is option we should think about and document in the design document.

distros/kubernetes/nvsentinel/charts/janitor/templates/deployment.yaml
```diff
@@ -126,6 +136,24 @@ spec:
...
+        {{- if and .Values.config.cspProvider .Values.config.cspProvider.auth .Values.config.cspProvider.auth.enabled }}
+        - name: csp-provider-token
```
lalitadithya, line 149: will this token have all the permissions of the service account that the pod is attached to? If yes, can we list out what permissions it has?

distros/kubernetes/nvsentinel/charts/janitor-provider/templates/deployment.yaml
```diff
@@ -150,23 +150,48 @@ spec:
...
               readOnly: true
+            {{- end }}
           {{- end }}
-      {{- if and (eq (.Values.csp.provider | default "kind") "nebius") .Values.csp.nebius.serviceAccountKeySecret }}
+      {{- if or .Values.tls.enabled (and (eq (.Values.csp.provider | default "kind") "nebius") .Values.csp.nebius.serviceAccountKeySecret) }}
       volumes:
+        {{- if .Values.tls.enabled }}
+        - name: tls-cert
+          secret:
+            secretName: {{ include "provider.fullname" . }}-grpc-cert
```
natherz97, line 186: Can we be consistent with how janitor-provider and janitor generate the secret name each needs for TLS? Looks like we're using 2 approaches:
```
secretName: {{ include "provider.fullname" . }}-grpc-cert
secretName: {{ .Values.config.cspProvider.tls.caSecretName }}
```

distros/kubernetes/nvsentinel/charts/janitor/templates/configmap.yaml
```diff
@@ -50,7 +50,16 @@ data:
         {{- end }}
       {{- end }}
       cspProviderHost: {{ .Values.config.cspProviderHost }}
```
natherz97, line 52: Nit: should this a property within the cspProvider struct?

distros/kubernetes/nvsentinel/charts/janitor/templates/configmap.yaml
```diff
@@ -50,7 +50,16 @@ data:
...
+      cspProviderCAPath: /etc/nvsentinel/janitor/csp-ca/ca.crt
+      {{- end }}
+      {{- if and .Values.config.cspProvider .Values.config.cspProvider.tls .Values.config.cspProvider.tls.insecure }}
+      cspProviderInsecure: true
+      {{- end }}
+      {{- if and .Values.config.cspProvider .Values.config.cspProvider.auth .Values.config.cspProvider.auth.enabled }}
+      cspProviderTokenPath: /var/run/secrets/nvsentinel/csp-provider/token
+      {{- end }}
+
     rebootNodeController:
```
natherz97, line 63: Should we allow a cspProviderHost override for individual controllers? Right now it looks like we allow setting a custom cspProviderHost which is different from the global setting but none of the new properties that are being introduced.

janitor-provider/main.go
```diff
@@ -145,7 +149,46 @@ func run() error {
 		return fmt.Errorf("failed to create csp client: %w", err)
 	}
 
-	svr := grpc.NewServer()
+	var serverOpts []grpc.ServerOption
```
natherz97, line 152: What's our plan for rotating the 3 TLS artifacts? It looks like we're only reloading tls.crt/tls.key in janitor-provider and ca.crt in janitor during component start-up right?

distros/kubernetes/nvsentinel/charts/janitor/templates/deployment.yaml
natherz97: Since we're using a different audience, trying to use this token will result in a 401 error from the apiserver so it will fail authN and not be authenticated as the Janitor SA.

distros/kubernetes/nvsentinel/charts/janitor/templates/configmap.yaml
jtschelling: i think this is a good idea but will follow up in another pr

jtschelling: if you think this is a hard requirement im down to discuss further

distros/kubernetes/nvsentinel/charts/janitor/templates/configmap.yaml
jtschelling: yes, but i think maybe that changes if we include your below comment on per controller settings?

janitor-provider/main.go
jtschelling: adding that to the design document

janitor-provider/main.go
jtschelling: adding cert rotation to the design doc, we can discuss more there

distros/kubernetes/nvsentinel/charts/janitor-provider/values.yaml
jtschelling: Done — TLS and auth are now enabled by default in both charts. Customers can set `tls.enabled: false` and `auth.enabled: false` to opt out.

janitor-provider/pkg/auth/token_review.go
jtschelling: Done — wrapped both `extractBearerToken` and `validateToken` errors with context in the interceptor.

janitor-provider/main.go
jtschelling: Good point — removed the TLS requirement for auth. Auth can now be enabled independently for environments where a service mesh handles transport security.

janitor-provider/main.go
jtschelling: mTLS is covered as an option in the design doc (PR #1008, ADR-029). It's listed under "Other Options" with trade-offs. The short version: mTLS via cert-manager client certs is viable but adds certificate lifecycle management on both sides. The current approach (server TLS + SA token auth) reuses Kubernetes-native primitives for the auth layer, keeping the cert management simpler. That said, the design doc presents it as an option rather than a rejected alternative — open to discussion.

distros/kubernetes/nvsentinel/charts/janitor-provider/templates/deployment.yaml
jtschelling: Fixed — the janitor's `caSecretName` now defaults to `"janitor-provider-grpc-cert"` in values.yaml, matching the provider's derived cert secret name (`{{ include "provider.fullname" . }}-grpc-cert`). Both charts reference the same secret by default without manual configuration.

jtschelling: they're different helm charts so im not sure we'd get to a perfect solution. theres an argument that we should just combine the helm charts, i'd be in support of that

natherz97 (approved):

Issue #1078, opened 2026-03-27
[Bug]: DCGM Handle Leak in gpu-health-monitor Causes GpuDcgmConnectivityFailure


suket22:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Bug Description

**Component:** `NVSentinel/health-monitors/gpu-health-monitor`
  **File:** `gpu_health_monitor/dcgm_watcher/dcgm.py`
  **Version:** `nvsentinel-gpu-health-monitor:1.1` (we run a private fork)
  **Severity:** Medium — causes recurring false `GpuDcgmConnectivityFailure` node conditions and alerts

  ---

  ### Summary

The `DCGMWatcher` leaks TCP connections to the DCGM host engine during periods of transient DCGM connectivity failures. Each failed initialization cycle opens a connection that is not properly closed by `pydcgm.DcgmHandle.Shutdown()`. Over time, leaked connections accumulate until DCGM's max client connection limit is reached, at which point all subsequent handle creations fail — triggering `GpuDcgmConnectivityFailure` health events and associated alerts to fire.

  ---

  ### Observed Symptoms

  - Node condition `GpuDcgmConnectivityFailure = True` with message: `ErrorCode:DCGM_CONNECTIVITY_ERROR Failed to connect to DCGM for health check`
  - Repeating log pattern in the `gpu-health-monitor` pod once max connections is hit:

  ```
  INFO  Successfully created DCGM handle
  INFO  supported gpus are [0, 1, 2, 3, 4, 5, 6, 7]
  INFO  supported switches are [0, 1, 2, 3]
  ERROR Error getting DCGM handle: Max limit reached for the object
  ERROR DCGM connectivity failure detected, sending GpuDcgmConnectivityFailure health event
  ```

  ---

### Root Cause

**Location:** `_cleanup_dcgm_resources()` lines 273–288, specifically `dcgm_handle.Shutdown()`

The leak is triggered by **transient DCGM connectivity failures** during `_initialize_dcgm_monitoring()`. The logs show a clear buildup pattern prior to hitting Max limit:

  ```
  11:51:18 ERROR [dcgm.py:342] Error getting DCGM handle: Timeout
  11:51:48 INFO  [dcgm.py:278] Successfully created DCGM handle
  11:52:48 ERROR [dcgm.py:342] Error getting DCGM handle: Timeout
  11:53:19 INFO  [dcgm.py:278] Successfully created DCGM handle
[...]

PR #1089, opened 2026-03-31
fix: prevent DCGM handle/connection leak on connectivity failure (#1078)

deesharma24:
## Summary

Bug => https://github.com/NVIDIA/NVSentinel/issues/1078

When DCGM becomes transiently unresponsive or not reachable , _cleanup_dcgm_resources() had Delete() and Shutdown() in a single try block. Delete() throws when the connection is broken, causing Shutdown() to be skipped entirely. Each retry cycle leaks one TCP connection + one server-side DCGM group. After 64 leaks (DCGM_MAX_NUM_GROUPS), DCGM returns Max limit reached, triggering permanent GpuDcgmConnectivityFailure alerts until the DCGM pod is manually recycled.

As part of the fix, Split the try blocks so Shutdown() always runs regardless of Delete() outcome, and added group rollback on partial initialization failure.

  1. Split try blocks in _cleanup_dcgm_resources() — Delete() and Shutdown() now have independent error handling. Delete() failing no longer prevents Shutdown() from running and closing the TCP socket.
  2. Added init rollback in _initialize_dcgm_monitoring() — If any step fails after group creation (e.g., health.Set(), GetGpuIds()), the group is deleted before the exception propagates, preventing server-side group leaks from partial initialization.
  3.  Added 3 unit tests — test_cleanup_dcgm_resources[delete_throws] (core bug), test_cleanup_dcgm_resources[none_group] (partial init path), and test_init_monitoring_rolls_back_group_on_failure (init rollback).

 Testing:- Tested on dev cluster tg220000-dgxc-runai-gcp-cmh-dev0 with custom image builds across three scenarios:

  - Service port change (4 cycles, 5555→5556→5555): Client maintained exactly 1 DCGM connection and 9 FDs throughout all cycles with zero accumulation. Logs confirmed Delete() failing independently while Shutdown() ran successfully each time.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU health monitoring initialization to roll back partially created resources on failure.
  * Ensured cleanup always attempts to release all GPU-related resources so shutdown proceeds even if group deletion fails.

* **Tests**
  * Added tests covering initialization rollback and various cleanup failure scenarios to prevent resource leaks.

### DIFF
--- a/health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
+++ b/health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
@@ -273,7 +273,8 @@ class DCGMWatcher:
             log.info(f"dcgm gpu_id are {gpu_ids}")
 
             return dcgm_group, gpu_ids, gpu_serials
-        except Exception:
+        except Exception as e:
+            log.warning(f"DCGM monitoring initialization failed, rolling back group: {e}")
             try:
                 dcgm_group.Delete()
             except Exception as del_err:
--- a/health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
+++ b/health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
@@ -259,33 +259,47 @@ class DCGMWatcher:
 
         Returns:
             A tuple of (dcgm_group, gpu_ids, gpu_serials)
+
+        If any step after group creation fails the group is deleted before the
+        exception propagates so that it does not leak on the DCGM server.
         """
         dcgm_group = self._create_dcgm_group_with_all_entities(dcgm_handle)
-        with metrics.dcgm_api_latency.labels("group_health_set").time():
-            dcgm_group.health.Set(dcgm_structs.DCGM_HEALTH_WATCH_ALL)
+        try:
+            with metrics.dcgm_api_latency.labels("group_health_set").time():
+                dcgm_group.health.Set(dcgm_structs.DCGM_HEALTH_WATCH_ALL)
 
-        gpu_ids = dcgm_group.GetGpuIds()
-        gpu_serials = self._get_gpu_serial_numbers(dcgm_handle)
-        log.info(f"dcgm gpu_id are {gpu_ids}")
+            gpu_ids = dcgm_group.GetGpuIds()
+            gpu_serials = self._get_gpu_serial_numbers(dcgm_handle)
+            log.info(f"dcgm gpu_id are {gpu_ids}")
 
-        return dcgm_group, gpu_ids, gpu_serials
+            return dcgm_group, gpu_ids, gpu_serials
+        except Exception:
+            try:
+                dcgm_group.Delete()
+            except Exception as del_err:
+                log.warning(f"Failed to delete DCGM group during init rollback: {del_err}")
+                metrics.dcgm_api_failures.labels("init_group_rollback").inc()
+            raise
 
     def _cleanup_dcgm_resources(
         self,
         dcgm_group: pydcgm.DcgmGroup,
         dcgm_handle: pydcgm.DcgmHandle,
     ):
-        """Clean up DCGM resources safely."""
-        try:
-            if dcgm_group:
+        """Clean up DCGM resources safely.
+
+        Group deletion and handle shutdown are in separate try blocks so that
+        a failure in Delete() does not prevent Shutdown() from running.
+        """
+        if dcgm_group:
+            try:
                 dcgm_group.Delete()
-                dcgm_group = None
-            if dcgm_handle:
-                # Clean up the handle
-                dcgm_handle.Shutdown()
-                del dcgm_handle
-        except Exception as e:
-            log.error(f"Error cleaning up DCGM handle: {e}")
+            except Exception as e:
+                log.warning(f"Error deleting DCGM group (will still shut down handle): {e}")
+                metrics.dcgm_api_failures.labels("group_delete_error").inc()
+
+        if dcgm_handle:
+            dcgm_handle.Shutdown()
 
     def start(self, fields_to_monitor: list[str], exit: Event) -> None:
         dcgm_handle = None
--- a/health-monitors/gpu-health-monitor/gpu_health_monitor/tests/test_dcgm_watcher/test_dcgm.py
+++ b/health-monitors/gpu-health-monitor/gpu_health_monitor/tests/test_dcgm_watcher/test_dcgm.py
@@ -17,6 +17,7 @@ from unittest.mock import MagicMock, patch
 import dcgm_structs, dcgm_errors, dcgm_fields, dcgm_field_helpers
 from threading import Event, Thread
 from ctypes import pointer
+import pytest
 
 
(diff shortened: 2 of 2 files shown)


health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
```diff
@@ -259,33 +259,47 @@ def _initialize_dcgm_monitoring(self, dcgm_handle: pydcgm.DcgmHandle) -> tuple:
...
-                dcgm_handle.Shutdown()
-                del dcgm_handle
-        except Exception as e:
-            log.error(f"Error cleaning up DCGM handle: {e}")
+            except Exception as e:
+                log.warning(f"Error deleting DCGM group (will still shut down handle): {e}")
+                metrics.dcgm_api_failures.labels("group_delete_error").inc()
+
+        if dcgm_handle:
+            dcgm_handle.Shutdown()
```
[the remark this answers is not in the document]
deesharma24: ⏺ Shutdown() is intentionally not wrapped in try-except because it cannot raise an exception by design:

  1. dcgmDisconnect() always returns DCGM_ST_OK — every code path in the C implementation returns OK, with explicit comments: "Returning OK here to prevent errors from being logged from the python framework"
  dcgmlib/src/DcgmApi.cpp#L6203-L6231
  2. Shutdown() already catches AttributeError internally for the edge case where modules are unloaded during multi-threaded teardown:
  testing/python3/DcgmHandle.py#L162-L167
  3. Shutdown() is a safe no-op if already called — it checks _handleCreated first and returns immediately if the handle was already shut down:
  testing/python3/DcgmHandle.py#L151-L170

  Adding a try-except around Shutdown() would be dead code that misleads future readers into thinking failure is expected. Updated the docstring to clarify why Delete() needs protection but Shutdown() does not.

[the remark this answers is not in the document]
deesharma24: Should you resolve it or I should resolve it?

lalitadithya, line 303: previously we had this in the try block, but with the new changes it looks like we have moved it out of the try block so this would mean that we will end up crashing when the shutdown errors, is that expected?

deesharma24: Shutdown() is intentionally not in a try block because it cannot raise an exception. This is the shutdown code https://github.com/NVIDIA/DCGM/blob/e137795bc755d14fabc0e44c91178a364952cb43/testing/python3/DcgmHandle.py#L151 . The old code had it in a try block only because it shared one with
  Delete() — not because Shutdown() needed protection(Shutdown() catches AttributeError internally). Moving it out makes the intent clearer: Delete() can fail (and we handle it), Shutdown() can't.

lalitadithya: > Shutdown() is intentionally not in a try block because it cannot raise an exception.

why? what if there is a nil pointer how is that prevented? I understand that `AttributeError ` is skipped, what about others?

deesharma24: we are just calling the Shutdown api with dcgm_handle. We are proactively checking whether dcgm_handle is nil or not. I Am not clear on why it needs to be part of try/catch block as it can't raise an exception. So, what other thing we can do?

Also, what do you mean by **what about others?**

lalitadithya: My question is: what if the shutdown raises an exception other than `AttributeError `  for which they have a try catch internally, what do we want to do? If the pod restarts are the server-side state (if any) is cleaned up correctly?

deesharma24: The only place else from where exception can come is from here https://github.com/NVIDIA/DCGM/blob/e137795bc755d14fabc0e44c91178a364952cb43/testing/python3/DcgmHandle.py#L159

But it will never go inside this if loop because fromDecorator will be false https://github.com/NVIDIA/DCGM/blob/e137795bc755d14fabc0e44c91178a364952cb43/testing/python3/DcgmHandle.py#L64-L68 when we initially create the dcgm handle from here https://github.com/NVIDIA/NVSentinel/blob/main/health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py#L249.

 the pod restarts are the server-side state (if any) is cleaned up correctly?
 That is the best case if the dcgm pod restarts. When the pod restarts:
  1. Client side(gpu-health-monitor ): Process dies → OS closes ALL fds → TCP RST sent to server
  2. Server side(dcgm) : Receives RST → RemoveConnectionByBev() → RemoveAllGroupsForConnection() → all groups freed

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
```diff
@@ -259,33 +259,47 @@ def _initialize_dcgm_monitoring(self, dcgm_handle: pydcgm.DcgmHandle) -> tuple:
...
-        gpu_ids = dcgm_group.GetGpuIds()
-        gpu_serials = self._get_gpu_serial_numbers(dcgm_handle)
-        log.info(f"dcgm gpu_id are {gpu_ids}")
+            gpu_ids = dcgm_group.GetGpuIds()
+            gpu_serials = self._get_gpu_serial_numbers(dcgm_handle)
+            log.info(f"dcgm gpu_id are {gpu_ids}")
 
-        return dcgm_group, gpu_ids, gpu_serials
+            return dcgm_group, gpu_ids, gpu_serials
+        except Exception:
```
lalitadithya: can we also log the original exception here?

deesharma24: added.

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
```diff
@@ -259,33 +259,48 @@ def _initialize_dcgm_monitoring(self, dcgm_handle: pydcgm.DcgmHandle) -> tuple:
 
         Returns:
             A tuple of (dcgm_group, gpu_ids, gpu_serials)
+
+        If any step after group creation fails the group is deleted before the
+        exception propagates so that it does not leak on the DCGM server.
```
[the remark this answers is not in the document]
deesharma24: the phrasing is correct.

lalitadithya (approved):

PR #1095, opened 2026-04-02
metadata-collector: make kubelet host configurable via KUBELET_HOST env var

### LINKED ISSUE #1096: metadata-collector: kubelet connection fails when kubelet binds to node IP instead of 0.0.0.0
## Description

The metadata-collector hardcodes the kubelet endpoint as `localhost:10250`. This fails on clusters where kubelet binds to the node's primary IP instead of `0.0.0.0`, causing all metadata-collector pods to crash-loop:

```
dial tcp 127.0.0.1:10250: connect: connection refused
```

The code assumes that because the pod runs with `hostNetwork: true`, localhost will reach the kubelet. This is not the case when kubelet is configured with `address: <node-ip>` rather than `address: 0.0.0.0`. This is common on bare-metal Kubernetes clusters.

## Impact

All metadata-collector DaemonSet pods crash-loop on affected clusters. GPU metadata is collected successfully, but the pod device mapper phase fails on every reconcile attempt.

## Proposed Fix

PR #1095 makes the kubelet host configurable via a `KUBELET_HOST` environment variable (defaults to `localhost` for backward compatibility). The Helm chart injects this via the Kubernetes Downward API (`status.hostIP`).

yysindi:
Fixes #1096

## Problem

The metadata-collector hardcodes the kubelet endpoint as `localhost:10250`. This fails on clusters where kubelet binds to the node's primary IP instead of `0.0.0.0`, causing all metadata-collector pods to crash-loop:

```
dial tcp 127.0.0.1:10250: connect: connection refused
```

This is common on bare-metal Kubernetes clusters where the kubelet config sets `address` to the node IP.

## Fix

- Read kubelet host from the `KUBELET_HOST` environment variable, falling back to `localhost` for backward compatibility
- Inject `KUBELET_HOST` in the Helm chart daemonset template via the Kubernetes Downward API (`status.hostIP`), which provides the correct node IP on every node automatically
- Use `net.JoinHostPort` for correct IPv6 address bracketing

## Testing

- Table-driven tests covering default (localhost), IPv4, and IPv6 hosts
- All existing tests pass unchanged

## Summary by CodeRabbit

* **New Features**
  * Helm chart adds a configurable kubelet host override for the metadata collector, allowing non-default kubelet endpoints (node primary IPs, custom hosts). Defaults to localhost when not set; supports IPv4 and IPv6.
* **Tests**
  * Added unit tests validating host-resolution and request URI construction for empty, IPv4, and IPv6 host cases.

### DIFF
--- a/distros/kubernetes/nvsentinel/charts/metadata-collector/templates/daemonset.yaml
+++ b/distros/kubernetes/nvsentinel/charts/metadata-collector/templates/daemonset.yaml
@@ -58,6 +58,10 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: spec.nodeName
+            {{- if .Values.kubeletHost }}
+            - name: KUBELET_HOST
+              {{- toYaml .Values.kubeletHost | nindent 14 }}
+            {{- end }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
--- a/distros/kubernetes/nvsentinel/charts/metadata-collector/values.yaml
+++ b/distros/kubernetes/nvsentinel/charts/metadata-collector/values.yaml
@@ -29,3 +29,19 @@ resources:
 
 # Runtime class name for the pod. If empty or not set, the field will be omitted.
 runtimeClassName: "nvidia"
+
+# Override the kubelet host used by the pod device mapper. When unset (default),
+# the metadata-collector connects to localhost. Set this to a valid Kubernetes
+# env var spec to override. On clusters where kubelet binds to the node's
+# primary IP instead of 0.0.0.0, use the Downward API:
+#
+#   kubeletHost:
+#     valueFrom:
+#       fieldRef:
+#         fieldPath: status.hostIP
+#
+# A static value also works:
+#
+#   kubeletHost:
+#     value: "10.0.0.1"
+kubeletHost: {}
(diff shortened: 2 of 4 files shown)


distros/kubernetes/nvsentinel/charts/metadata-collector/templates/daemonset.yaml
```diff
@@ -58,6 +58,11 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: spec.nodeName
+            - name: KUBELET_HOST
```
lalitadithya, line 62: I believe this will inject the env var in all environments, so we will not fallback to `localhost` right? Should we wrap this by a helm flag/value?

yysindi: the behavior should be equivalent to localhost on clusters where kubelet binds to 0.0.0.0 but agreed, probably cleaner to make it opt in so it defaults to localhost

yysindi: updated the PR to gate it behind a Helm value. kubeletHost defaults to empty (no env var injected, falls back to localhost). Users who need it set:

```
  kubeletHost:
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
```

lalitadithya (approved):

PR #1113, opened 2026-04-05
feat: add gRPC sink connector to platform-connectors

shaq918:
### Platform-connectors
- Add gRPC sink connector that forwards HealthEvent protos to an external gRPC server
- Reuses existing PlatformConnector gRPC service (HealthEventOccurredV1)
- Own ring buffer with independent retry/backoff (same pattern as Store and K8s connectors)
- Prometheus metrics: send counter, duration histogram, retry counter
- Configurable via Helm values (enable flag, target address, max retries)
- Disabled by default

### Helm
- Add `grpcSinkConnector` section to platform-connectors values
- Add config keys to ConfigMap template

## Test plan

- [x] Unit tests: 7 tests covering send success/failure, retry, max retries drop, shutdown
- [x] `go vet ./...` passes
- [x] Helm lint passes
- [x] E2E: deploy with `grpcSinkConnector.enabled=true` and a test gRPC server, verify events arrive

## Summary by CodeRabbit

* **New Features**
  * Added a configurable gRPC sink connector to forward health metrics to external services (disabled by default).
  * New configuration options to enable the connector, set target endpoint, and control max retry attempts.
  * Added Prometheus metrics for send counts, durations, and retry totals.

* **Tests**
  * Added unit tests covering forwarding, retry behavior, shutdown/drain semantics, and safe close.

* **Documentation**
  * Added design ADR describing connector behavior and configuration.


distros/kubernetes/nvsentinel/templates/configmap.yaml
```diff
@@ -29,6 +29,9 @@ data:
       "StoreConnectorMaxRetries": {{ .Values.platformConnector.mongodbStore.maxRetries }},
       "enableMongoDBStorePlatformConnector": "{{ or .Values.global.mongodbStore.enabled (and .Values.global.datastore (eq (default "" .Values.global.datastore.provider) "mongodb")) }}",
       "enablePostgresDBStorePlatformConnector": {{ if and .Values.global.datastore .Values.global.datastore.provider }}{{ eq .Values.global.datastore.provider "postgresql" | quote }}{{ else }}"false"{{ end }}
+      ,"enableGRPCSinkConnector": "{{ .Values.platformConnector.grpcSinkConnector.enabled }}"
```
lalitadithya, line 32: Can you please some details on how auth would work in this case?

lalitadithya (commented):
This is great PR! Thank you for putting this together! 

I think it would be good to start with a simple design in https://github.com/NVIDIA/NVSentinel/tree/main/docs/designs so that we understand the scope and use cases we are targeting for this feature. What do you think?

distros/kubernetes/nvsentinel/templates/configmap.yaml
shaq918: Hi Lalit!! thanks for looking into this so prompty!! The gRPC sink uses insecure credentials because the target is a component within the cluster's internal network, not exposed externally. This seems to fit the existing trust model and the NVSentinal orverarching architecture (plug in to a single cluster), almost all internal gRPC like health monitors, health-events-analyzer, metadata-collector is insecure (except: the janitor → janitor-provider).   They all use `insecure.NewCredentials()`.

docs/designs/033-grpc-sink-connector.md
```diff
@@ -0,0 +1,114 @@
...
+   HealthEvent proto (17 fields) into a condition type, boolean status, and a
+   truncated message string. Entity-level detail (GPU UUID, GPC/TPC/SM,
+   register values), the full error code array, recommended action enum,
+   processing strategy, and metadata are lost or truncated at the 4KB
+   Kubernetes message limit.
+
+2. **Query the database** — requires access to the NVSentinel MongoDB or
+   PostgreSQL instance, coupling the external system to NVSentinel's internal
+   storage layer.
+
```
lalitadithya, line 34: We also have the event exporter which can be used to send information outside, but that also reads from MongoDB. I think the ask here is for a solution that can export events externally without having to maintain state in cluster?

docs/designs/033-grpc-sink-connector.md
```diff
@@ -0,0 +1,114 @@
...
+within this window, the send is treated as a failure and retried via the ring
+buffer's exponential backoff.
+
+## Security
+
+The gRPC sink target is expected to be a component within the same cluster,
+outside of NVSentinel but still within the cluster's internal network. The
+connector uses insecure credentials (`insecure.NewCredentials()`), consistent
+with all other internal NVSentinel gRPC connections (health monitors,
+health-events-analyzer, metadata-collector).
```
lalitadithya: The key different here with the health monitors is that they use UDS sockets, so it is not over the network, so unless the host itself is compromised we won't see fake data. In the case of sending data in cluster to a different cluster, wouldn't it be possible for anyone running cluster to also send fake events to that service? Do we want to recommend using network policies or something like that here? I think it is fairly simple to also include the kube token like we do in Janitor - Janitor provider so that we provide some options to the tenants, if they don't want to use the auth that is fine, but as good citizens we will provide them with all the required information, what do you think?

docs/designs/033-grpc-sink-connector.md
shaq918: Yup on the money! Did look into the exporter but sadly wouldnt work for our usecase as you mentioned -- we would need persistence.

lalitadithya: that makes sense, let's just call it out in the document for posterity

docs/designs/033-grpc-sink-connector.md
shaq918: Agree can add auth optionality -- kubernetes SA token like the janitor path. Our use case is from in-cluster to in-cluster.

docs/designs/033-grpc-sink-connector.md
shaq918: updated the doc.

lalitadithya (approved):

shaq918:
Thanks for the quick turnaround on this @lalitadithya !! do we have a rough estimate on when you'll be creating a new release with a proper tag? v(1.1.1) or v(1.2.0) I see a lot of activity on main.

PR #1119, opened 2026-04-07
docs: Add design doc for deduplication layer in syslog health monitor

### LINKED ISSUE #1120: [Feature]: Deduplicate XID/SXID in syslog health monitor
### Prerequisites

### Feature Summary

Add de-duplication logic in syslog health monitor that avoid the monitor to send a new health event if the same XID message has already been sent in the same remediation cycle.

### Problem/Use Case

In some of the cluster,s we had over ~800K entries of XID 95 on a problematic node but we had only 29 distinct entries. Treating each of these duplicated entries only causes extra load on the modules down the line and achieve nothing that would aid in smoother/better remediation process.

### Proposed Solution

Design PR: https://github.com/NVIDIA/NVSentinel/pull/1119

### Component

Health Monitor

XRFXLP:
## Summary

Markdown preview: [Platform connector — Event Deduplication Until Remediation](https://github.com/XRFXLP/NVSentinel/blob/deduplication-design-doc/docs/designs/039-health-event-deduplication.md)

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Health-event deduplication now suppresses duplicate events in the processing pipeline.
  * Deduplication uses a configurable burst-window TTL and canonical matching to reliably detect duplicates.
  * Healthy events clear cached unhealthy counterparts to avoid stale cancellations.
  * Pipeline includes a non-mutating filter stage to drop events when needed.
  * Per-check opt-out via configuration and a Prometheus metric to monitor suppressed events.
  * Deduplication state is cleared on service restarts and expires by TTL.


docs/designs/033-syslog-health-monitor-event-deduplication.md
```diff
@@ -0,0 +1,380 @@
...
+    CheckHealthy -->|No| Normalize["Strip kernel timestamp<br/>from event.Message"]
+    Normalize --> DedupCheck{"Normalized message<br/>already in seen set?"}
+    DedupCheck -->|Yes| Suppress["Suppress - do not send<br/>(log at Info level)"]
+    DedupCheck -->|No| Mark["Add to seen set"]
+    Mark --> Send
+```
+
+### What counts as "the same message"
+
+The dedup key is the **exact message string** with the kernel timestamp prefix stripped. Two messages that differ in any field — PCI address, XID code, pid, channel, process name — are treated as distinct and are **not** deduplicated. Only truly identical repeated error lines are suppressed.
```
lalitadithya: how does this play with the health event analyzer rules where it checks for multiple XID 13s/31s? Can we document if there will be any impact to those rules?

docs/designs/033-syslog-health-monitor-event-deduplication.md
```diff
@@ -0,0 +1,380 @@
...
+
+### What clears the dedup
```
lalitadithya: Can this be a leaky bucket kind of approach so we will still continue to see independent bursts if they occur at two different times before remediation?

docs/designs/033-syslog-health-monitor-event-deduplication.md
```diff
@@ -0,0 +1,380 @@
...
+
+
+## Rationale
+
+- **Monitor-level dedup, not handler-level**: placing dedup in `handleSingleLine` avoids modifying the `Handler` interface or each handler implementation. The `Handler.ProcessLine` contract stays unchanged — it returns events, and the monitor decides whether to send them.
+- **Exact message matching (minus timestamp)**: this is the simplest correct key. Any semantic difference in the message (different PCI, different pid, different XID code) is a different error and should not be suppressed.
+- **State file persistence**: the monitor is a long-running daemon, so in-memory tracking covers cross-poll dedup. Persisting to the state file additionally covers pod restarts without needing to re-report errors that downstream components have already processed.
+- **No TTL/expiry**: dedup is scoped to "until remediation". Healthy events and reboots are the remediation signals. A TTL would introduce a tuning parameter with no universally correct value and risk either premature re-reporting or unbounded suppression.
```
lalitadithya: I would say that having without a TTL we have unbounded suppression since we don't get feedback into the different bursts of XIDs that can happen prior to remediation

docs/designs/033-syslog-health-monitor-event-deduplication.md
XRFXLP: Why do we need a leaky bucket? If the original event has already triggered breakfix pipeline then why do we need to follow up with the same error?

lalitadithya: also what if the XIDs are in patterns XID 1 -> XID-2 -> XID-1 -> XID-2 -> XID-1, should we dedup or no?

lalitadithya: There is also the use case outside of breakfix like in the case of the exporter where we want to publish the health events for analytics purposes outside of the cluster. Now if we supress in the vast majority of the cases, I'm worried we'd end up losing data on bursts and potentially might delay our response to a bug

XRFXLP: As per the current design, we will only have the first `XID 1 -> XID-2`, rest of the chain would be de-duplicated.

XRFXLP: In theory, we can monitor the suppression metric to get an idea of the burst that we're getting, as prioritizing analytics over resilience of breakfix system just for getting an idea of size of the burst doesn't make sense. Adding any dedup window duration doesn't make sense because we've seen the burst of size of 10+ hours.


Also are we assuming that large burst indicates an error of higher severity here?

docs/designs/033-syslog-health-monitor-event-deduplication.md
XRFXLP: We will have impact on the rules that assume repetition of XIDs with same impact entities. So it will be RepatedXIDonSameGPU and RepeatedXIDonSameGPCAndTPC, will add this into a doc.

docs/designs/033-syslog-health-monitor-event-deduplication.md
lalitadithya: > In theory, we can monitor the suppression metric

Sure, but that won't have all the information as we have with the health events. 

> as prioritizing analytics over resilience of breakfix system just for getting an idea of size of the burst doesn't make sense

I'm not saying prioritize analytics over the breakfix system, I'm saying we provide all the parts of the system with as much data as they can handle. I'm thinking about this in terms of throughput. What is the maximum possible throughput that the system can handle which we can saturate. 

> Adding any dedup window duration doesn't make sense because we've seen the burst of size of 10+ hours.

I think the idea here is that we'd want to see them as multiple bursts 

> Also are we assuming that large burst indicates an error of higher severity here?

to a certain extent, yes.

XRFXLP: I guess it would be better to optimize the system throughout first (if we can). One very obvious I can think of is using event ID rather than full event ID in node drainer queue that would decrease the memory usage as a function of number of events consequently increasing the max throughput. Checking that possibility, meanwhile keeping this MR as draft for now as dedup window straightaway looks like arbitrary.

lalitadithya: I agree, let's find opportunities for optimization first. One that @KaivalyaMDabhadkar is already working on is the cold start and there could be other optimizations that we can do as well.

neerajnv: +1 to prioritising system stability over analytics. If we need a record of all health events regardless of duplication, we should think of an alternate mechanism such as exporting all the events via logs to Kratos and it does not need to go through MongoDB either.

neerajnv: does order matter here? I mean are you using a pattern match or a search in unordered set for deduplication?

docs/designs/033-syslog-health-monitor-event-deduplication.md
```diff
@@ -0,0 +1,380 @@
...
+```
+
+**State file changes** in [`pkg/syslog-monitor/types.go`](health-monitors/syslog-health-monitor/pkg/syslog-monitor/types.go):
+
+```go
+type syslogMonitorState struct {
+    Version          int                     `json:"version"`
+    BootID           string                  `json:"boot_id"`
+    CheckLastCursors map[string]string       `json:"check_last_cursors"`
+    SeenMessages     map[string][]string     `json:"seen_messages,omitempty"`
```
neerajnv: ignore the comment. can't get rid of it.

docs/designs/033-syslog-health-monitor-event-deduplication.md
```diff
@@ -0,0 +1,380 @@
...
+        tracker.Clear()
+        slog.Info("Cleared entire dedup tracker on healthy event (no PCI entity)",
+            "check", checkName)
+    }
+}
+```
+
+### 5. Boot ID change handling
+
+In [`handleBootIDChange`](health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor.go), clear all dedup trackers alongside cursors:
```
neerajnv: does GPU reset also resets the `seen` messages?

XRFXLP: yes, but only XIDs for that specific GPU.

docs/designs/033-syslog-health-monitor-event-deduplication.md
KaivalyaMDabhadkar: I think it would also make sense to reduce the frequency of re-emissions rather than complete deduplication. Between the spectrum of complete deduplication (proposed approach) and no deduplication ( current implementation) we have a case where we do re-emit some duplicates but not as frequently, but just enough to give us an idea of different "bursts" in time as @lalitadithya has suggested. This will also provide the HEA with more information regarding the frequency of the XIDs compared to the proposed approach where the HEA loses all information regarding duplicates in time and their frequency. A simple way to accomplish this would be to re-emit the same error after an exponentially increasing number of suppressed occurrences, for eg. emit the 1st, then suppress until the 10th occurence, then suppress until the 20th, 40th, 80th and so on (or until an upper limit which we set is reached). This means that our total duplicates emitted would only grow logarithmically with the total frequency of ocurrences, which is still a huge amount of deduplication and which is enough for the HEA to detect repetition patterns and for rules like RepeatedXIDonSameGPU to fire. Thoughts about this?

docs/designs/039-syslog-health-monitor-event-deduplication.md
```diff
@@ -0,0 +1,268 @@
...
+
+`SysLogsXIDError` and `SysLogsSXIDError` only. `SysLogsGPUFallenOff` already runs its own XID-windowed correlation and is excluded.
+
+### Assumption: single dominant XID per burst
+
+Production XID bursts on a given GPU are dominated by a single XID code (e.g., XID 31, 45, 95 producing ~100k repeated events). Distinct XID codes arriving back-to-back within the same burst window are rarely observed in field data, except XID 162 <-> XID 163 pair. This justifies tracking only one `(currentXidCode, seenMessages)` slot per `(node, GPU_UUID)`: a different XID arriving on the same GPU is treated as a state transition that supersedes the prior slot.
+
+## Decision
+
+Per `(node, GPU_UUID)` slot, hold one current XID code and a TTL'd set of seen normalized messages. Suppress an unhealthy event when its normalized message is already in the slot for the same XID code and the entry is within `BurstWindow`. Reset the slot on XID change, healthy event, or reboot.
```
lalitadithya: Do we also need to consider additional metadata such as GPC, TPC and NVLink4 registers for slot?

docs/designs/039-syslog-health-monitor-event-deduplication.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Whole-check clearing on healthy events
+Clear every slot for the check whenever any healthy event arrives, instead of clearing per-GPU.
+
+**Rejected:** when GPU-A is reset, GPU-B's suppressed errors would be re-reported.
+
+### Dedup only in downstream components
+Rely on `deduplicateMessagesByIdentity` in the platform connector instead of suppressing at source.
+
+**Rejected:** that operates on node-condition annotations, not the gRPC event stream. Redundant events still consume bandwidth, storage, and inflate counts.
```
lalitadithya: What about adding a gRPC middleware to handle this? That will ensure the dedup logic will work for all future health monitors if required.

docs/designs/039-syslog-health-monitor-event-deduplication.md
XRFXLP: Not following, do you mean slot reset on the basis of this metadata or we're being too strict in deduplication just on the basis of XID number?

lalitadithya: the later, I think dedup based on just the XID is too strict and it will likely result in lost events. We will need to consider the additional metadata as well

docs/designs/039-syslog-health-monitor-event-deduplication.md
XRFXLP: gRPC middleware would be too coarse to handle this since this is quite domain specific de-duplication, for instance in case of syslog HM it is (GPU UUID, XID, normalized message). But in case of GPU health monitor, it might be very different. Even being aggressive and abstracting (GPU_UUID, ErrorCode) away, the normalized message still remains a domain-specific dedup key (for syslog we strip the kernel timestamp but other monitors would have their own noise fields to mask)

lalitadithya: Say we extract everything that really mattered to a health event into entities impacted and we have a clear error code. In the case syslog health monitor, it would have all the GPC, TPC, register etc as entities impacted and GPU health monitor would do the same. Would a generic enough baseline for dedup be just the combination of entities impacted, node and error_code with a TTL? 

If we are relying on the message then, yes, it becomes domain specific.

docs/designs/039-syslog-health-monitor-event-deduplication.md
XRFXLP: Ah, no. We're not doing deduping just on basis XID and GPU_UUID, it is including message field as well.

> Per `(node, GPU_UUID)` slot, hold one current XID code and a TTL'd set of seen normalized messages

So if XID 1 comes with two messages message 1 followed by message 2 (with content being different, ignore timestamp for now) then both message 1 and message 2 will be published.

Also we've:
> Suppress an unhealthy event when its normalized message is already in the slot for the same XID code

normalized message = kernel log line after masking timestamp. That means we're not losing any information and all the metadata available in that log line would be sent (if not already sent)

docs/designs/039-syslog-health-monitor-event-deduplication.md
XRFXLP: yeah, that can work. There is nothing extra in the message field that the other modules rely on. 

Are you suggesting this to move to platform connector like a transformer or somewhere before that?

docs/designs/039-health-event-deduplication.md
```diff
@@ -0,0 +1,241 @@
...
+
+### What clears the dedup
+
+| Signal                          | Cleared                                             |
+|---------------------------------|-----------------------------------------------------|
+| TTL elapses on a seen entry     | That entry only — next identical event re-emits     |
+| System reboot (boot ID change)  | All entries                                         |
+| Platform-connector restart      | Nothing — entries restored from state file          |
+
+There is no explicit "healthy event clears the corresponding unhealthy key" hook. A healthy event has its own dedup key (`IsHealthy=true` separates it from any unhealthy event with the same `(node, entities, ErrorCode)` shape) and is itself deduplicated against repeated healthy emissions; the corresponding unhealthy entry just TTL-expires.
```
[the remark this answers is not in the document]
XRFXLP: addressed

docs/designs/039-syslog-health-monitor-event-deduplication.md
XRFXLP: I've made changes in that direction, PTAL.

docs/designs/039-health-event-deduplication.md
```diff
@@ -0,0 +1,249 @@
...
+
+type Tracker struct {
+    mu     sync.RWMutex
+    seen   map[uint64]time.Time   // canonical-key hash → first-seen
+    ttl    time.Duration
+    now    func() time.Time       // injectable for tests
+}
+
+// Snapshot is the on-disk representation of a single seen entry.
+type Snapshot struct {
```
KaivalyaMDabhadkar: Can we include `CheckName` in the key too? Because for example NIC events currently don’t set `ErrorCode` and thus different checknames can collide in the map (eg. InfiniBandStateCheck and InfiniBandDegradationCheck for the same NIC port)

docs/designs/039-health-event-deduplication.md
```diff
@@ -0,0 +1,249 @@
...
+
+func NewTracker(ttl time.Duration) *Tracker
+func NewTrackerFromSnapshot(ttl time.Duration, snap []Snapshot) *Tracker  // drops already-expired entries
+
+// Key extracts the canonical key from an event.
+func Key(event *pb.HealthEvent) uint64
+
+// IsDuplicate is true iff the event's key is already in the tracker and within ttl.
+// Side effect: evicts TTL-expired entries it scans.
+func (t *Tracker) IsDuplicate(event *pb.HealthEvent) bool
```
KaivalyaMDabhadkar, line 81: Can you also call out that we would need to periodically sweep expired entries from the map? else entries whose keys don't occur again after TTL for a long time will remain in memory and can cause memory leakage

docs/designs/039-health-event-deduplication.md
```diff
@@ -0,0 +1,249 @@
...
+type Deduplicator struct {
+    tracker *dedup.Tracker
+    skip    map[string]bool   // checks excluded from dedup
+}
+
+func (d *Deduplicator) Filter(ctx context.Context, event *pb.HealthEvent) (bool, error) {
+    if d.skip[event.CheckName] {
+        return true, nil
+    }
+    if d.tracker.IsDuplicate(event) {
```
KaivalyaMDabhadkar, line 139: a potential ordering issue: the pseudocode checks IsDuplicate(event) before calling ClearUnhealthyCounterpart(event), so duplicate healthy events return early without clearing the matching unhealthy key.

for example within the same TTL window and the same (node, entities, ErrorCode) tuple:

1. unhealthy event arrives, so (A, IsHealthy=false) is marked.
2. healthy event arrives, so (A, IsHealthy=true) is marked and (A, IsHealthy=false) is cleared.
3. the same unhealthy condition occurs again, so (A, IsHealthy=false) is marked again.
4. another healthy event arrives. Since (A, IsHealthy=true) from 2 is still within TTL, this healthy event is treated as a duplicate and returns before clearing (A, IsHealthy=false) from 3.

A later unhealthy event can now be suppressed against the stale unhealthy key, even though there was a healthy signal at 4.
Should healthy events clear the unhealthy counterpart before the duplicate-return path, or at least still call ClearUnhealthyCounterpart when event.IsHealthy && IsDuplicate(event)?

XRFXLP: Good catch, ordering is fixed.

docs/designs/039-health-event-deduplication.md
XRFXLP: Added the logic for it.

docs/designs/039-health-event-deduplication.md
XRFXLP: Right, my sample space was only Syslog HM and GPU HM.

docs/designs/039-syslog-health-monitor-event-deduplication.md
lalitadithya: > Are you suggesting this to move to platform connector like a transformer or somewhere before that?

There was thinking of using https://grpc.io/docs/guides/interceptors/

docs/designs/039-health-event-deduplication.md
```diff
@@ -0,0 +1,263 @@
...
+- The hash function is stable across processes (e.g., FNV-1a over the canonical encoding) so state files round-trip cleanly.
+
+### What clears the dedup
+
+| Signal                                                          | Cleared                                                                                              |
+|-----------------------------------------------------------------|------------------------------------------------------------------------------------------------------|
+| Healthy event passes the filter for `(node, check, entities, ErrorCode)` | The corresponding unhealthy entry (same tuple, `IsHealthy=false`) — fresh recurrence emits as new   |
+| TTL elapses on a seen entry                                     | That entry only — next identical event re-emits                                                      |
+| System reboot (boot ID change)                                  | All entries                                                                                          |
```
lalitadithya: what is the use case for boot ID based tracking? Since this is at the platform connector layer and the cache is cleared on restart emitting an additional event shouldn't be too big of a deal right?

docs/designs/039-syslog-health-monitor-event-deduplication.md
XRFXLP: I see.  

From the looks of it, it needs to be done before the transformers pipelines and we can't do that since override transformer have ability to mutate some of the fields that are part of de-duplication key so the ordering of (transformer, dedup) is not swappable.

docs/designs/039-health-event-deduplication.md
XRFXLP: Yeah, you're right. I've removed the entire state tracking with in-memory store.

lalitadithya (approved):

PR #1121, opened 2026-04-07
Optimize node drainer memory footprint with lazy access to DB

### LINKED ISSUE #1133: [Optimization]: Reduce workqueue memory footprint in node-drainer by lazy-fetching events from DB
### Prerequisites

### Feature Summary

Node drainer memory footprint can be optimized ~10x if we store only the reference (ID) instead of full event to workqueue memory

### Problem/Use Case

The node-drainer workqueue stores the full `datastore.Event` document (a `map[string]interface{}` deserialized from the DB) in each `NodeEvent` queue item. Each real-world health event document is ~13 KB. Under burst conditions (e.g., a GPU repeatedly generating the same XID error every 30 seconds), the queue accumulates thousands of items, each holding a full document copy. At 50k queued events, the node-drainer pod consumes ~700 Mi of heap — enough to OOM on a typical 512 Mi pod memory limit.

The Kubernetes `TypedRateLimitingQueue` holds all queued items in memory (pending + dirty + processing sets). Since each `NodeEvent` includes a `*datastore.Event` pointer to a unique heap allocation, none of these documents are GC'd until the single-threaded worker processes and dequeues them.

### Proposed Solution

Replace the full event storage with a lazy-fetch pattern:

- At enqueue time, store only the `DocumentID` (native DB identifier — `ObjectID` for MongoDB, `UUID` for PostgreSQL) in the `NodeEvent` struct. Drop the Event `*datastore.Event` field.

- At processing time, the worker fetches the full event from the DB via `FindDocument` using the stored `DocumentID` before passing it to the reconciler.

- Inject `_id` into the fetched event if the DB backend doesn't include it in the document column (PostgreSQL stores id as a separate column, not inside the JSONB).

### Component

Fault Management

XRFXLP:
## Summary


  Previously, `NodeEvent` stored the full `*datastore.Event` object in the workqueue — meaning every                                                                                                                                                    
  event queued for processing held the complete health event payload (including all metadata, entities,                                                                                                                                                 
  and DB references) in memory for its entire lifetime in the queue.                                                                                                                                                                                    
                                                                                                                                                                                                                                                        
  This change stores only the document ID (`DocumentID interface{}`) in the queue instead of the full                                                                                                                                                   
  event. The worker fetches the full event from MongoDB lazily via `FindDocument` only when it is                                                                                                                                                       
  actually processing the item, and the object is GC'd immediately after. The public                                                                                                                                                                    
  `EnqueueEventGeneric` interface is unchanged — callers are unaffected.                                                                                                                                                                                
                                                                                                                                                                                                                                                        
  **Side effect:** deduplication is now semantically correct. Previously, two enqueues of the same                                                                                                                                                      
  logical event (same `_id`) bypassed workqueue deduplication because each created a distinct pointer.                                                                                                                                                  
  With value-based `DocumentID`, the workqueue correctly marks a re-enqueued in-flight event as dirty                                                                                                                                                   
  rather than creating a duplicate queue entry. Other than that there is no change in node drainer functionality                                                                                                                                                                                              
                                                                                                                                                                                                                                                        
  ## Memory Benchmark (injected XIDs in 10k batches in a 25 node cluster)                                                                                                                                                               
                                                                                                                                                                                                                                                        
  | Events | Old (Mi) | New (Mi) | Savings (Mi) | Reduction |                                                                                                                                                                                           
  |--------|----------|----------|--------------|-----------|
  | 0 (baseline) | 44 | 59 | — | — |                                                                                                                                                                                                                    
  | 10k | 192 | 67 | 125 | 65% |
  | 20k | ~310 | 76 | ~234 | ~75% |                                                                                                                                                                                                                     
  | 30k | ~440 | 80 | ~360 | ~82% |                                                                                                                                                                                                                     
  | 40k | ~570 | 90 | ~480 | ~84% |
  | 50k | ~700 | 94 | ~606 | ~87% |                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                        
  At 50k queued events the new image uses ~6.5% of what the old one would have consumed                                                                                      

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [x] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Refactor**
  * Queue items now carry only document IDs; workers fetch full event data on demand.
  * Deduplication is based on document ID instead of in-memory event payloads.
  * Workers retry on lookup failures with rate limiting and drop items if datastore handles are missing.
  * Cold-start enqueue skips events when a document ID cannot be extracted.

* **Tests**
  * Unit and integration tests updated to seed documents and validate ID-based lookup, deduplication, and worker fetch behavior.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/node-drainer/main.go
+++ b/node-drainer/main.go
@@ -298,6 +298,8 @@ func handleColdStart(ctx context.Context, components *initializer.Components) er
 
 	slog.Info("Found events to re-process", "count", len(healthEvents))
 
+	dbAdapter := &dataStoreAdapter{DatabaseClient: components.DatabaseClient}
+
 	// Re-process each event
 	for _, he := range healthEvents {
 		// Use the RawEvent from the database query which includes _id
@@ -325,9 +327,6 @@ func handleColdStart(ctx context.Context, components *initializer.Components) er
 			slog.Error("Node name is empty in cold start event")
 			continue
 		}
-
-		// Create adapter to bridge interface differences
-		dbAdapter := &dataStoreAdapter{DatabaseClient: components.DatabaseClient}
 
 		documentID, err := utils.ExtractDocumentIDNative(event)
 		if err != nil {
--- a/node-drainer/main.go
+++ b/node-drainer/main.go
@@ -35,6 +35,7 @@ import (
 	"github.com/nvidia/nvsentinel/node-drainer/pkg/initializer"
 	"github.com/nvidia/nvsentinel/store-client/pkg/client"
 	"github.com/nvidia/nvsentinel/store-client/pkg/query"
+	"github.com/nvidia/nvsentinel/store-client/pkg/utils"
 )
 
 var (
@@ -328,7 +329,15 @@ func handleColdStart(ctx context.Context, components *initializer.Components) er
 		// Create adapter to bridge interface differences
 		dbAdapter := &dataStoreAdapter{DatabaseClient: components.DatabaseClient}
 
-		if err := components.QueueManager.EnqueueEventGeneric(ctx, nodeName, event, dbAdapter, healthStore); err != nil {
+		documentID, err := utils.ExtractDocumentIDNative(event)
+		if err != nil {
+			slog.Error("Failed to extract document ID from cold start event", "error", err)
+			continue
+		}
+
+		err = components.QueueManager.EnqueueEventGeneric(
+			ctx, nodeName, event, dbAdapter, healthStore, documentID)
+		if err != nil {
 			slog.Error("Failed to enqueue cold start event", "error", err, "nodeName", nodeName)
 		} else {
 			slog.Info("Re-queued event from cold start", "nodeName", nodeName)
--- a/node-drainer/pkg/queue/worker.go
+++ b/node-drainer/pkg/queue/worker.go
@@ -72,6 +72,7 @@ func (m *eventQueueManager) processNextWorkItem(ctx context.Context) bool {
 			"node", nodeEvent.NodeName, "eventID", nodeEvent.EventID)
 		m.queue.Forget(nodeEvent)
 		metrics.QueueDepth.Set(float64(m.queue.Len()))
+
 		return true
 	}
 
@@ -81,11 +82,11 @@ func (m *eventQueueManager) processNextWorkItem(ctx context.Context) bool {
 			"node", nodeEvent.NodeName, "eventID", nodeEvent.EventID, "error", fetchErr)
 		m.queue.AddRateLimited(nodeEvent)
 		metrics.QueueDepth.Set(float64(m.queue.Len()))
+
 		return true
 	}
 
 	err := m.processEventGeneric(ctx, event, nodeEvent.Database, nodeEvent.HealthEventStore, nodeEvent.NodeName)
-
 	if err != nil {
 		slog.Warn("Error processing event for node (will retry)",
 			"node", nodeEvent.NodeName,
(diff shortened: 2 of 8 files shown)


node-drainer/main.go
```diff
@@ -328,7 +329,15 @@ func handleColdStart(ctx context.Context, components *initializer.Components) er
...
 
-		if err := components.QueueManager.EnqueueEventGeneric(ctx, nodeName, event, dbAdapter, healthStore); err != nil {
+		documentID, err := utils.ExtractDocumentIDNative(event)
+		if err != nil {
+			slog.Error("Failed to extract document ID from cold start event", "error", err)
+			continue
+		}
+
+		err = components.QueueManager.EnqueueEventGeneric(
+			ctx, nodeName, event, dbAdapter, healthStore, documentID)
```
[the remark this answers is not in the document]
XRFXLP: should be handled

node-drainer/pkg/queue/worker.go
```diff
@@ -46,14 +71,26 @@ func (m *eventQueueManager) processNextWorkItem(ctx context.Context) bool {
...
+	if nodeEvent.Database == nil || nodeEvent.HealthEventStore == nil {
+		slog.Error("NodeEvent missing database or health event store, dropping",
+			"node", nodeEvent.NodeName, "eventID", nodeEvent.EventID)
+		m.queue.Forget(nodeEvent)
+		metrics.QueueDepth.Set(float64(m.queue.Len()))
+
+		return true
+	}
+
+	event, fetchErr := fetchEventFromDB(ctx, nodeEvent.DocumentID, nodeEvent.Database)
```
lalitadithya, line 83: this is trading memory for additional mongodb api calls, do we have an idea of how many more api calls this will result in?

XRFXLP: For enqueue phase, it is the same number of operation with old vs new (we set status to `InProgress`). For process phase, if the number of operation earlier was N then with the new code it would N + 1 (for `FindDocument` call).

Now for realistic number:
```
rate(mongodb_mongod_op_latencies_latency_total{type="read"}[5m]) / rate(mongodb_mongod_op_latencies_ops_total{type="read"}[5m]) / 1000
```

is average over 0.4 - 0.6 ms, if we take cap for 1ms then for 50k event the extra-overhead would be 50 seconds spread across all the events.

XRFXLP: that 50k count will decrease after we have dedup at syslog HM level

lalitadithya: I see and this call is done every time the object is requeued right? So, if a node is stuck in drain for a long time, we will make mongodb api calls for every requeue? I'm trying to understand the diff b/w read calls per second before and after this change

XRFXLP: Right. After hitting exponential backoff ceiling it would be 1 extra read call every 2 minutes for long drains.

lalitadithya (approved):

PR #1124, opened 2026-04-07
feat: support component-based recommended actions

alexscjundev:
Signed-off-by: Alex Jun <aljun@nvidia.com>

## Summary

Support `ComponentRemediationActions` in fault remediation, which maps from component type and recommended action to created CR.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Component-class-specific remediation: define per-component remediation actions that override global actions with automatic fallback.

* **Documentation**
  * Docs updated to describe component-scoped configuration, resolution precedence, template variables, and new templates for component-specific workflows.

* **Behavior**
  * Remediation tracking now records component class to improve deduplication and ensure correct action/template selection.

* **Tests**
  * Added/updated tests and e2e coverage for component-scoped resolution and template loading.


fault-remediation/pkg/remediation/remediation.go
```diff
@@ -70,6 +70,7 @@ type FaultRemediationClient struct {
 	statusChecker     *crstatus.CRStatusChecker
 }
 
+// nolint:cyclop
```
XRFXLP, line 73: Can we modularize the function so we have to not add nolint for cyclop?

Looks at the diff below:
```golang
	for actionName, maintenanceResource := range remediationConfig.RemediationActions {
		if err := loadTemplateForAction(templates, templateMountPath, actionName, maintenanceResource); err != nil {
			return nil, err
		}
	}
```

is repeated in the line 96-100, perhaps we can create a function for this that validate for map?

docs/configuration/fault-remediation.md
```diff
@@ -52,18 +52,46 @@ Defines the Custom Resource that will be created to trigger remediation actions.
...
-        templateFileName: "gpureset-template.yaml"
-        equivalenceGroup: "reset"
-        supersedingEquivalenceGroups: ["restart"]
-        impactedEntityScope: "GPU_UUID"
+        completeConditionType: "NodeReady"
+        templateFileName: "rebootnode-template.yaml"
+        equivalenceGroup: "restart"
+
+    componentActions:
+      "GPU":
```
XRFXLP, line 65: Maybe I'm missing some context here: What forbids an user to add `Node` here and do node level remediation flowing through this new pathaway?

fault-remediation/pkg/config/config.go
```diff
@@ -186,6 +233,114 @@ func validateOneSupersedingGroup(actionName, group string, resource MaintenanceR
...
+		if resource, exists := c.RemediationActions[actionName]; exists {
+			return resource, true
+		}
+
+		return MaintenanceResource{}, false
+	}
+
+	for _, componentActions := range c.ComponentRemediationActions {
+		if _, ok := componentActions[actionName]; ok {
+			return MaintenanceResource{}, false
```
KaivalyaMDabhadkar, line 333: This returns false as soon when it finds the action in any component's overrides, even if only one component has it. This means an older annotation (with the previous version of annotation) for COMPONENT_RESET is treated as ambiguous even if only GPU (and no other component) has an override. So when nvsentinel is upgraded in a cluster, this means the existing annotation entry gets treated as stale and cleaned up, potentially causing a new CR to be created even though the in-progress one is still valid. I think a better alternative would be to treat it as unambiguous if exactly one component class defines the action, and resolve to that action. What do you think?

KaivalyaMDabhadkar:
In `fault-remediation/pkg/reconciler/reconciler_e2e_test.go`, the e2e tests compile with the new fucntion signature but don't actually test the new feature we have added. All health events in the e2e tests have an empty ComponentClass, and the config only has shared RemediationActions, no ComponentRemediationActions. So every test still is still going through the old fallback path.

It would be good to add one e2e test case that uses the component-aware path end to end. Something like: a GPU health event (ComponentClass: "GPU") with a GPU-specific override in the config, verify it creates a GPUReset CR (not the shared RebootNode), verify the annotation records ComponentClass: "GPU", then send a second identical event and verify it correctly deduplicates against the stored GPU-specific CR. This would catch any issues in the e2e flow with this new code path

PR #1146, opened 2026-04-10
feat: add node-drainer otel changes

tanishagoyal2:
## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Testing
Tested traces integration changes in tilt and dev cluster. We have tested various negative and positive cases and output of exported traces are mentioned below

1. Node recovery: Injected an fatal error and node went in draining state. The healthy event later on injected to simulate the node recovery.

node recovery during drain

The span is created for status update along with final status that has been updated in DB

node-recovery along with status and cancel received

2. Healthy event injected to cancel the breakfix pipeline:

Screenshot 2026-04-11 at 5 08 27 PM

3. Mixed mode of eviction: the node went in draining with pods present in each mode of eviction (Immediate, Allow completion and delete after timeout). The span was created for each mode to show the start and end of each mode.

mixed-modes

mixed-modes 2

4. Node drainer crashed while draining the node: the node drainer was crashing for 5min, when it came back in running state it picked up the event and continued the draining with spans added to the old trace. We can see two `node_drainer.phase.allow_completion.start` span both at the gap of 5min. 

Screenshot 2026-04-11 at 5 28 38 PM

 
5. Draining failed as no supported entities were present for partial drain

drain-failed

drain-failed 2

6. Node is already in draining state: the node drainer updated status to AlreadyDrained

already drained

already drained 1

7. Partial drain in node-drainer: Spans were created with drain scope, entity type and entity value info

partial-drain 

## Summary by CodeRabbit

* **New Features**
  * Added OpenTelemetry distributed tracing across node-drainer with per-operation spans, context-aware logs, and persisted drain-session trace info for cross-service correlation.
  * Exposed configurable OTLP exporter settings in Kubernetes deployments.

* **Bug Fixes**
  * Corrected a log message spelling.

* **Chores**
  * Promoted OpenTelemetry dependencies to v1.43.0 and added a new service identifier constant for tracing.


node-drainer/pkg/queue/worker.go
```diff
@@ -82,22 +85,62 @@ func (m *eventQueueManager) processNextWorkItem(ctx context.Context) bool {
...
+
+	processCtx := ctx
+
+	if session.DrainSessionSpan == nil {
+		if traceID != "" {
+			var spanCtx context.Context
+
+			spanCtx, session.DrainSessionSpan = tracing.StartSpanWithLinkFromTraceContext(
+				ctx, traceID, parentSpanID, "node_drainer.drain_session")
+			processCtx = spanCtx
```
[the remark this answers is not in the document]
tanishagoyal2: the event is always expected to have fault-quarantine span ID

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -367,6 +492,72 @@ func (r *Reconciler) executeAction(ctx context.Context, action *evaluator.DrainA
...
+		return
+	}
+
+	r.setDrainScope(ds, action)
+
+	traceID := tracing.TraceIDFromMetadata(healthEvent.HealthEvent.GetMetadata())
+	parentSpanID := tracing.ParentSpanID(
+		healthEvent.HealthEventStatus.SpanIds, tracing.ServiceFaultQuarantine)
+
+	r.endPhaseMarkers(ctx, ds, action.Action, traceID, parentSpanID)
```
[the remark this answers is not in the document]
tanishagoyal2: we can be sure that node-drainer will receive the event when FQ has processed it

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -229,24 +273,48 @@ func (r *Reconciler) setInitialStatusAndEnqueue(ctx context.Context, document ma
...
+	if err := r.queueManager.EnqueueEventGeneric(
+		ctx, nodeName, datastore.Event(document), r.databaseClient, r.healthEventStore, documentID,
+	); err != nil {
+		tracing.RecordError(span, err)
+		span.SetAttributes(
+			attribute.String("node_drainer.error.type", "enqueue_error"),
+			attribute.String("node_drainer.error.message", err.Error()),
+		)
+
+		return err
```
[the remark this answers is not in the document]
tanishagoyal2: why do we need context wrapping with error. Can you please point me to the line where its mentioned?

node-drainer/pkg/evaluator/evaluator.go
```diff
@@ -408,21 +431,32 @@ func (e *NodeDrainEvaluator) evaluateCustomDrain(ctx context.Context, healthEven
...
+	span := tracing.SpanFromContext(ctx)
 	if !isComplete {
-		slog.Debug("Drain CR in progress",
+		slog.DebugContext(ctx, "Drain CR in progress",
 			"node", nodeName,
 			"crName", crName)
 
+		span.SetAttributes(
+			attribute.String("node_drainer.custom_cr.name", crName),
+			attribute.String("node_drainer.custom_cr.status", "in_progress"),
```
lalitadithya: do we need the status? won't that be evident from the trace not completing?

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -213,6 +250,13 @@ func (r *Reconciler) handleEventCancellation(
 // setInitialStatusAndEnqueue sets the initial status to InProgress and enqueues the event
 func (r *Reconciler) setInitialStatusAndEnqueue(ctx context.Context, document map[string]any,
 	documentID any, nodeName string) error {
+	ctx, span := tracing.StartSpan(ctx, "node_drainer.set_initial_status_and_enqueue")
+	defer span.End()
+
+	span.SetAttributes(
+		attribute.String("node_drainer.user_pods_eviction_status.status", string(model.StatusInProgress)),
```
lalitadithya: Same question as earlier, aren't the status evident from the span itself?

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -367,6 +492,72 @@ func (r *Reconciler) executeAction(ctx context.Context, action *evaluator.DrainA
...
+func (r *Reconciler) setDrainScope(ds *queue.DrainSession, action *evaluator.DrainActionResult) {
+	if ds.ScopeSet {
+		return
+	}
+
+	if action.PartialDrainEntity != nil {
+		ds.DrainSessionSpan.SetAttributes(
+			attribute.String("node_drainer.drain.scope", "partial"),
+			attribute.String("node_drainer.partial_drain.entity_type", action.PartialDrainEntity.EntityType),
+			attribute.String("node_drainer.partial_drain.entity_value", action.PartialDrainEntity.EntityValue),
```
lalitadithya: aren't these already in the health event, do we want these in the trace too?

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -418,10 +617,33 @@ func (r *Reconciler) executeSkip(ctx context.Context,
...
+
+	if ds := queue.DrainSessionFromContext(ctx); ds != nil && !ds.ImmediateEvictionStarted {
+		traceID := tracing.TraceIDFromMetadata(healthEvent.HealthEvent.GetMetadata())
+		parentSpanID := tracing.ParentSpanID(
+			healthEvent.HealthEventStatus.SpanIds, tracing.ServiceFaultQuarantine)
+
+		if traceID != "" {
+			_, startSpan := tracing.StartSpanWithLinkFromTraceContext(
+				ctx, traceID, parentSpanID, "node_drainer.phase.immediate_eviction.start")
+			startSpan.End()
```
lalitadithya: I'm a little confused by this, shouldn't the span continue till the `immediate_eviction` is completed? why do we want short-lived spans like this?

node-drainer/pkg/reconciler/reconciler.go
```diff
@@ -907,25 +1250,33 @@ func (r *Reconciler) deleteCustomDrainCRIfEnabled(ctx context.Context, nodeName
...
 			"node", nodeName,
 			"crName", crName,
 			"error", err)
+		tracing.RecordError(span, err)
 	} else {
-		slog.Info("Deleted custom drain CR",
+		slog.InfoContext(ctx, "Deleted custom drain CR",
 			"node", nodeName,
 			"crName", crName)
+		span.SetAttributes(attribute.Bool("node_drainer.custom_cr.deleted", true))
```
lalitadithya: do we need this attribute, isn't it understood due to the execution flow?

node-drainer/pkg/reconciler/reconciler.go
tanishagoyal2: I added this intentionally to avoid noise of spans in trace. We check the pods in each mode in every reconciliation which runs at 2min of interval and pods can run for days. So, to avoid adding new spans for every reconcile I am adding short lived span.
We are adding mode.start as soon as we enter in that mode of eviction for the first time and mode.end when we are moving to next eviction mode (immediate mode to deleteaftertimeout mode)

lalitadithya: Can't we just keep start one span and leave it open till completed? 

We do need to trace every get api call made during that duration as well irrespective of how ever long the reconcile cycle lasts

tanishagoyal2: We can leave it open but if node-drainer crashed then we won't get old created spans as long running span never finish during crash.

lalitadithya: can't we save the id to mongodb like we do with the parent ones?

tanishagoyal2: if we go with current approach then we will have all spans traced. For example shown in the below screenshot where we have two allowcompletion.start because node-drainer was crashing for ~5 min

Screenshot 2026-04-13 at 8 56 41 PM

node-drainer/pkg/reconciler/reconciler.go
tanishagoyal2: right, I have removed it

node-drainer/pkg/evaluator/evaluator.go
tanishagoyal2: right, that will be clear from span name itself. The span `node_drainer.set_initial_status_and_enqueue` indicates that event is marked in-progress and has been requeued.

node-drainer/pkg/reconciler/reconciler.go
tanishagoyal2: Agree, we will have http.delete call to indicate the deletion of cr. I have removed storing this attribute from code

node-drainer/pkg/reconciler/reconciler.go
tanishagoyal2: As we discussed, I have removed the short running spans for each mode of eviction. Now the draining duration  is captured through drain_session span. Tested it with tilt cluster where node-drainer was crashing for ~3min. 

Screenshot 2026-04-14 at 1 19 27 PM

lalitadithya (approved):

PR #1198, opened 2026-04-16
feat: device platform connector

pteranodan:
## Summary

> [!WARNING]
> **Do not merge until Device API repo has been open-sourced**

adding Device API platform connector
- unified health event utilities
- K8s platform connector
  - replaced event processing utilities with unified ones 

## Type of Change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [x] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Optional device platform connector to monitor device health and update GPU status (disabled by default; enable via platformConnector.deviceConnector.enabled)
  * Chart/DaemonSet support to mount device API socket and set runtime group/supplemental groups when the connector is enabled

* **Improvements**
  * Consolidated health-event utilities for sorting, timestamps, status/reason/message generation and deterministic truncation
  * Centralized temporary-error detection used in retries and logging
  * Node/event processing now emits Warning events consistently and aggregates errors

* **Tests**
  * Extensive unit and integration tests covering health-event utilities, device connector processing, and event/condition behavior

* **Bug Fixes**
  * Restored missing newline in chart values file


platform-connectors/pkg/connectors/device/device_connector.go
```diff
@@ -0,0 +1,356 @@
...
+				events := healthEvents.GetEvents()
+				count := len(events)
+
+				logFields := []any{
+					"connector", connectorName,
+					"error", err,
+					"event_count", count,
+					"version", healthEvents.GetVersion(),
+				}
+				if count > 0 {
```
lalitadithya: I'm not sure if we'd need this, we can just log the complete health event array right?

platform-connectors/pkg/connectors/device/device_connector.go
```diff
@@ -0,0 +1,356 @@
...
+type statusPatch struct {
+	Status struct {
+		Conditions []metav1.Condition `json:"conditions"`
+		// TODO(pteranodan): Change *string once GPUStatus.RecommendedAction is *string.
+		RecommendedAction string `json:"recommendedAction,omitempty"`
+	} `json:"status"`
+}
+
+// nolint:cyclop,gocognit // Pipeline for sorting, deduplicating, and mapping events
+func (r *DeviceConnector) processGPUEvents(ctx context.Context, name string, events []*pb.HealthEvent) error {
```
lalitadithya: the logic here seems fairly similar to what we have in the kubernetes connector, can we reuse some of the existing functions?

platform-connectors/pkg/connectors/device/device_connector.go
pteranodan: I went with the window logic to keep logs concise and searchable. Would you prefer I just log the raw array, or perhaps just a count/version to keep it minimal?

lalitadithya: I would just do the counts as info and the array as debug

platform-connectors/pkg/connectors/device/device_connector.go
pteranodan: done.

commons/pkg/eventutil/health_event.go
```diff
@@ -0,0 +1,214 @@
...
+	if len(e) == 0 {
+		return e
+	}
+
+	s := slices.Clone(e)
+	if len(s) == 1 {
+		return s
+	}
+
+	slices.SortFunc(s, compareEvents)
```
XRFXLP, line 39: Do we need to special handling for len(e) == 0 and len(e) == 1 or this function already handles those case gracefully?

platform-connectors/main.go
```diff
@@ -296,51 +297,79 @@ func initializeGRPCSinkConnector(
...
+		if err != nil {
+			return nil, nil, nil, nil, fmt.Errorf("failed to initialize device platform connector: %w", err)
 		}
 	}
 
-	return k8sRingBuffer, storeConnector, grpcSinkConnector, nil
+	return k8sRingBuffer, storeConnector, grpcSinkConnector, deviceConnector, nil
 }
 
+// nolint:cyclop // cleanupResources naturally grows with the number of connectors
```
XRFXLP, line 363: should we collect the connectors into a struct or something similar?

platform-connectors/pkg/connectors/device/device_connector.go
```diff
@@ -0,0 +1,306 @@
...
+// from the NVIDIA_DEVICE_API_PATH environment variable.
+func InitializeConnector(
+	ctx context.Context,
+	ringBuffer *ringbuffer.RingBuffer,
+	stopCh <-chan struct{}) (*DeviceConnector, error) {
+	target := os.Getenv("NVIDIA_DEVICE_API_PATH")
+
+	cs, err := deviceset.NewForTarget(target)
+	if err != nil {
+		return nil, err
```
XRFXLP, line 71: Can we wrap the error here and perhaps add logging possible as well? Looks like that is missing in few more places.

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -465,25 +426,28 @@ func (r *K8sConnector) fetchHealthEventMessage(healthEvent *protos.HealthEvent)
...
 	if healthEvent.Message != "" {
 		// Replace semicolons with dots in the message to prevent delimiter collision
 		sanitizedMessage := strings.ReplaceAll(healthEvent.Message, ";", ".")
-		message += fmt.Sprintf("%s ", sanitizedMessage)
+		parts = append(parts, sanitizedMessage)
 	}
 
-	message += fmt.Sprintf("Recommended Action=%s;", healthEvent.RecommendedAction.String())
+	// Only include if there is an actionable recommendation
+	if rec := eventutil.GetRecommendedAction(healthEvent); rec != nil {
```
XRFXLP: Do we need to make the appropriate change in fault-quarantine module CEL config to filter out any health event that doesn't have recommended action?
Otherwise this might introduce a bug where we don't have any node condition but the nodes are still cordoned because a buggy health monitor sent `isFatal: true` and `recommendedAction: None`

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -507,8 +475,17 @@ func filterProcessableEvents(ctx context.Context, healthEvents *protos.HealthEve
...
 // createK8sEvent creates a Kubernetes event from a health event.
-func (r *K8sConnector) createK8sEvent(ctx context.Context, healthEvent *protos.HealthEvent) *corev1.Event {
-	ts := safeTimestamp(ctx, healthEvent.GeneratedTimestamp)
+func (r *K8sConnector) createK8sEvent(healthEvent *protos.HealthEvent) *corev1.Event {
+	ts := eventutil.GetTime(healthEvent, time.Now())
+
+	status := eventutil.GetConditionStatus(healthEvent)
+
+	var eventType string
+	if status == metav1.ConditionFalse {
```
XRFXLP: We've:
```golang
	for _, healthEvent := range processableEvents {
		if !healthEvent.IsHealthy && !healthEvent.IsFatal {
			start := time.Now()
			err := r.writeNodeEvent(ctx, r.createK8sEvent(healthEvent), healthEvent.NodeName)
```

and if the event is the healthy the condition would be false, so this branch won't be ever called, right?

platform-connectors/pkg/connectors/device/device_connector.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	latestByCheck := make(map[string]*pb.HealthEvent)
+
+	var latest *pb.HealthEvent
+
+	sortedEvents := eventutil.SortByAge(events)
+	for _, event := range sortedEvents {
+		checkName := event.GetCheckName()
+		if checkName == "" {
```
XRFXLP: We've already done filtering on basis of processing strategy and null values of entites impact in the caller(`processHealthEvents`) of this function, can we do same check name there as well? Maybe by that we can just have a function to check for necessary attributes and skip otherwise.

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
deesharma24: why do we need to do this?.For a s single node condition with multiple messages some showing recommedation_Action and some don't creates confusion and breaks the consistency in the messaging format. So, I think we should keep it. it is causing no harm.

distros/kubernetes/nvsentinel/templates/daemonset.yaml
```diff
@@ -46,6 +46,9 @@ spec:
       securityContext:
         fsGroup: 65532
       {{- end }}
+      {{- if .Values.platformConnector.deviceConnector.enabled }}
+        supplementalGroups: [5000]
```
deesharma24, line 52: what is it used for ?

deesharma24: Also, I am not sure. how do you make sure if the receiver is not ready to accept the health events which have already created til now. how are you preserving them till the consumer is not ready or if UDS is not created  by the receiver.

platform-connectors/pkg/connectors/device/device_connector.go
```diff
@@ -0,0 +1,306 @@
...
+)
+
+const connectorName = "device"
+
+// DeviceConnector synchronizes health events from a ring buffer to the NVIDIA Device API.
+type DeviceConnector struct {
+	clientset  deviceset.Interface
+	ringBuffer *ringbuffer.RingBuffer
+	stopCh     <-chan struct{}
+	ctx        context.Context
```
KaivalyaMDabhadkar: We should avoid storing contexts in structs: https://go.dev/blog/context-and-structs#storing-context-in-structs-leads-to-confusion
Consider passing this as a param instead

platform-connectors/pkg/connectors/device/device_connector.go
```diff
@@ -0,0 +1,306 @@
...
+	patch := struct {
+		Status v1alpha1.GPUStatus `json:"status"`
+	}{Status: status}
+
+	patchBytes, err := json.Marshal(patch)
+	if err != nil {
+		return fmt.Errorf("failed to marshal status patch for GPU %q: %w", name, err)
+	}
+
+	_, err = r.clientset.DeviceV1alpha1().GPUs().Patch(ctx,
```
KaivalyaMDabhadkar: Can we wrap this in a `retry.RetryOnConflict` as well so as to be consistent with the pattern we are following for other update k8s API calls?

platform-connectors/pkg/connectors/device/device_connector_test.go
```diff
@@ -0,0 +1,774 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/nvidia/device-api/api/device/v1alpha1"
+	devicev1alpha1 "github.com/nvidia/device-api/api/device/v1alpha1"
```
KaivalyaMDabhadkar: nit: duplicate import

platform-connectors/main.go
```diff
@@ -296,51 +297,79 @@ func initializeGRPCSinkConnector(
 	return connector, nil
 }
 
+func initializeDeviceConnector(
+	ctx context.Context,
+	config map[string]interface{},
```
KaivalyaMDabhadkar: nit: we are not using config in the function body anywhere

platform-connectors/pkg/connectors/device/device_connector.go
pteranodan: Agreed. It looks like the [k8s connector](https://github.com/pteranodan/NVSentinel/blob/main/platform-connectors/pkg/connectors/kubernetes/k8s_connector.go#L49) set a bit of a bad precedent here. I’ll update this PR to pass it as a param, and we should probably track a separate issue to refactor the other connectors to match.

commons/pkg/eventutil/health_event.go
pteranodan: It handles those cases gracefully through its nil and equality checks.

platform-connectors/pkg/connectors/device/device_connector.go
pteranodan: The error wrapping here is redundant and unnecessary. The underlying `deviceset.NewForTarget` already returns a descriptive error (e.g., 'error parsing target URI...'), and the top-level caller also wraps the result with 'failed to initialize device platform connector'.

Not wrapping here prevents the error message from stuttering while still maintaining a clean trace for the caller.

platform-connectors/pkg/connectors/device/device_connector.go
pteranodan: Makes sense. I'll move the check name validation into `processHealthEvents` that we only group and process valid events.

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
pteranodan: I'll update to always append a recommended action.

pteranodan: I'll update to always append a recommended action.

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
pteranodan: You're right! I'll simplify the logic to always use `EventTypeWarning` here

platform-connectors/main.go
pteranodan: Removed.

platform-connectors/main.go
pteranodan: I agree that a struct or a registry would clean this up. However, since the goal of this PR is just to add the device connector, can we handle that refactor in a separate follow-up?

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
pteranodan: any issues with removing the trailing semi-colon?

distros/kubernetes/nvsentinel/templates/daemonset.yaml
pteranodan: Required because the NVIDIA Device API Unix domain socket (UDS) is restricted to that specific group ID.

pteranodan: Added a comment in the chart.

platform-connectors/pkg/connectors/device/device_connector.go
pteranodan: Done. Migrated `isTemporaryError` and its helpers from the Kubernetes connector to a shared `commons/pkg/errutil` package.

platform-connectors/pkg/connectors/device/device_connector_test.go
pteranodan: removed

commons/pkg/errutil/errors.go
```diff
@@ -0,0 +1,180 @@
...
+		IsNetworkError(err) ||
+		IsSyscallError(err) ||
+		IsStringBasedError(err) ||
+		errors.Is(err, io.EOF) ||
+		strings.Contains(err.Error(), "EOF")
+}
+
+// IsContextError checks if the error is a context-related error that should be retried
+func IsContextError(err error) bool {
+	return errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled)
```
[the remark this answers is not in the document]
pteranodan: Thoughts @deesharma24? This code was moved without change from `platform-connectors/pkg/connectors/kubernetes/process_node_events.go`

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
pteranodan: Decided to add the trailing semi-colon back. keep with the status quo.

commons/pkg/errutil/errors.go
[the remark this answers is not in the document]
deesharma24: @pteranodan we should remove this `errors.Is(err, context.Canceled)` . this function is only called with retryable scenarios only.

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -465,25 +426,26 @@ func (r *K8sConnector) fetchHealthEventMessage(healthEvent *protos.HealthEvent)
...
-		message += fmt.Sprintf("%s ", sanitizedMessage)
+		parts = append(parts, sanitizedMessage)
 	}
 
-	message += fmt.Sprintf("Recommended Action=%s;", healthEvent.RecommendedAction.String())
+	rec := healthEvent.GetRecommendedAction()
+	parts = append(parts, fmt.Sprintf("%s%s", recommendedActionMarker, rec))
 
-	return message
+	return strings.Join(parts, " ")
```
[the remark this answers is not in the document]
pteranodan: Adding trailing semi-colon back

commons/pkg/errutil/errors.go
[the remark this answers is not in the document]
pteranodan: done

platform-connectors/pkg/connectors/device/device_connector_test.go
```diff
@@ -0,0 +1,774 @@
...
+		},
+	}
+
+	err := connector.processGPUEvents(context.Background(), "gpu-0", events)
+	assert.NoError(t, err)
+
+	for _, action := range fakeCS.Actions() {
+		if p, ok := action.(clienttesting.PatchAction); ok {
+			var patch statusPatch
+			json.Unmarshal(p.GetPatch(), &patch)
```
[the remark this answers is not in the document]
pteranodan: done

platform-connectors/pkg/connectors/device/device_connector.go
```diff
@@ -0,0 +1,306 @@
...
+		name,
+		types.StrategicMergePatchType,
+		patchBytes,
+		metav1.PatchOptions{},
+		"status",
+	)
```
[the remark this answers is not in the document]
pteranodan: Custom type (not CRD). `types.StrategicMergePatchType` is supported, including patches to the `/status` subresource.

lalitadithya:
closing as stale, we can always reopen later :)

PR #1229, opened 2026-04-24
feat: add state detection logic for the NIC health monitor

### LINKED ISSUE #1211: [Feature]: Add NIC state detection logic + Metadata collector changes
### Prerequisites

### Feature Summary

Implement logic for NIC state detection + the changes in metadata collector as outlined in this design doc: https://github.com/KaivalyaMDabhadkar/NVSentinel/blob/642-nic-monitor-design/docs/nic-health-monitor/link-state-detection.md

### Problem/Use Case

Implement logic for NIC state detection + the changes in metadata collector as outlined in this design doc: https://github.com/KaivalyaMDabhadkar/NVSentinel/blob/642-nic-monitor-design/docs/nic-health-monitor/link-state-detection.md

### Proposed Solution

Refer to the design doc: Implement logic for NIC state detection + the changes in metadata collector as outlined in this design doc: https://github.com/KaivalyaMDabhadkar/NVSentinel/blob/642-nic-monitor-design/docs/nic-health-monitor/link-state-detection.md

### Component

Health Monitor

KaivalyaMDabhadkar:
## Summary

Implements link state detection for InfiniBand and Ethernet/RoCE NICs (Ref: [Design preview](https://github.com/KaivalyaMDabhadkar/NVSentinel/blob/642-nic-monitor-design/docs/nic-health-monitor/link-state-detection.md)). The monitor polls sysfs for port state changes and emits health events to the platform connector when ports go DOWN, recover, or disappear:

- **State detection checks**: `InfiniBandStateCheck` and `EthernetStateCheck` monitor logical state (ACTIVE/DOWN), physical state (LinkUp/Disabled), and device disappearance from sysfs
- **NIC classification**: Automatically classifies NICs as Compute, Storage, or Management based on `nvidia-smi topo -m` topology data from the metadata collector. Management NICs (default route) are excluded from monitoring
- **Persistent state**: Port states survive pod restarts via a hostPath-backed JSON file. Boot-ID comparison detects host reboots and re-emits healthy baselines to clear stale platform conditions
- **First-poll suppression**: Avoids false FATAL events for expected-down ports on first poll using card homogeneity analysis
- **VF exclusion**: SR-IOV Virtual Functions (detected via `device/physfn` symlink) are automatically skipped
- **Helm chart**: DaemonSet with configurable checks, polling interval, sysfs paths, and NIC exclusion regex

### Testing

**Real clusters**: Metadata collector + NIC monitor validated end-to-end on Azure AKS (8x A100-SXM4-80GB), OCI (4x L40S), on-prem DGXC (L40S, GB200), and DGXC H100 - covering topology parsing, NIC classification, VF detection, and state check behavior across different GPU/NIC configurations.

**Local E2E (Tilt + KIND)**: 10 scenarios validated with fake sysfs injection using real A100 topology — healthy baseline, DOWN/recovery for IB and Ethernet, device disappearance, persistence across restart, boot-ID change, VF exclusion, multi-port fault, multiple cycles, and message format.

Automated E2E tests will be committed in a follow-up PR.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * New NIC health monitor service/image: InfiniBand and Ethernet port discovery, state-change detection, persisted state, metrics, and reliable event delivery with retries.
  * Topology classifier now treats default-route exclusion with higher precedence during role classification.

* **Refactor**
  * Startup/config streamlined (TOML), clearer startup errors and wiring for checks and metrics.

* **Documentation**
  * Clarified default-route exclusion behavior.

* **Tests**
  * Added comprehensive unit tests for topology, state persistence, and checks.

* **Chores**
  * Build/image and local dev build updates added.


distros/kubernetes/nvsentinel/charts/nic-health-monitor/templates/_helpers.tpl
```diff
@@ -0,0 +1,191 @@
...
+{{- define "nic-health-monitor.name" -}}
+{{- .Chart.Name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+*/}}
+{{- define "nic-health-monitor.fullname" -}}
+{{- "nic-health-monitor" | trunc 63 | trimSuffix "-" }}
+{{- end }}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: This matches the pattern used by ALL other NVSentinel subchart helpers (syslog, gpu, etc.). Changing it just for NIC would be inconsistent.

health-monitors/nic-health-monitor/pkg/checks/state/ethstate.go
```diff
@@ -0,0 +1,542 @@
...
+	firstPoll bool, expectedCards map[string]struct{}, portCard map[string]string,
+) *pb.HealthEvent {
+	if pi.snap.State != checks.IBStateDown {
+		slog.Debug("RoCE port in non-ACTIVE state, ignoring",
+			"device", pi.snap.Device, "port", pi.snap.Port,
+			"state", pi.snap.State, "physState", pi.snap.PhysicalState,
+		)
+
+		return nil
+	}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: ACTIVE + Disabled is not a real sysfs state combination. The kernel sets logical state to DOWN when phys_state is Disabled. This is a theoretical concern that doesn't occur in practice.

health-monitors/nic-health-monitor/Tiltfile
```diff
@@ -12,8 +12,10 @@
...
-docker_build(
-    "ghcr.io/nvidia/nvsentinel/nic-health-monitor",
-    context="../..",
-    dockerfile="./Dockerfile",
+# Build and deploy
+custom_build(
+    'ghcr.io/nvidia/nvsentinel/nic-health-monitor',
+    '../../scripts/ko-tilt-build.sh . $EXPECTED_REF',
+    deps=['./', '../../commons', '../../data-models'],
+    skips_local_docker=True
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Same pattern as all other modules

distros/kubernetes/nvsentinel/values.yaml
```diff
@@ -176,6 +176,8 @@ global:
     enabled: false
   k8sdatastoreCrds:
     enabled: false
+  nicHealthMonitor:
+    enabled: false
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Not necessart

health-monitors/nic-health-monitor/pkg/checks/state/ethstate.go
```diff
@@ -0,0 +1,542 @@
...
+
+		for i := range dev.Ports {
+			p := dev.Ports[i]
+			if !discovery.IsEthernetPort(&p) {
+				continue
+			}
+
+			c.recordPort(st, dev, card, p)
+		}
+	}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Has been resolved now

distros/kubernetes/nvsentinel/charts/nic-health-monitor/templates/_helpers.tpl
```diff
@@ -0,0 +1,191 @@
...
+Selector labels
+*/}}
+{{- define "nic-health-monitor.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "nic-health-monitor.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+DaemonSet template that can be customized for kata or regular mode.
+Usage: include "nic-health-monitor.daemonset" (dict "root" . "kataMode" true)
```
lalitadithya: I don't quite follow what does NIC monitor have to do with kata mode, can you please explain?

distros/kubernetes/nvsentinel/charts/nic-health-monitor/templates/configmap.yaml
```diff
@@ -0,0 +1,34 @@
...
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "nic-health-monitor.fullname" . }}-config
+  labels:
+    {{- include "nic-health-monitor.labels" . | nindent 4 }}
+data:
+  config.yaml: |
```
lalitadithya: other modules use toml, can we do the same?

health-monitors/nic-health-monitor/pkg/checks/state/ethstate.go
lalitadithya, line 1: A lot of this seems to be a copy of IB, can we refactor to avoid code duplication?

health-monitors/nic-health-monitor/pkg/topology/topology.go
```diff
@@ -0,0 +1,602 @@
...
+	}
+
+	_, ok := c.gpuNUMASet[node]
+
+	return ok
+}
+
+// blueFieldHCATypes lists HCA type prefixes for BlueField DPUs.
+// DPU NICs are managed by SmartNIC firmware and excluded from monitoring.
+var blueFieldHCATypes = []string{
```
lalitadithya: Can we move all the consts to a separate file, I saw that we also had consts for other things like `Ethernet`

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
```diff
@@ -0,0 +1,605 @@
...
+func (c *InfiniBandStateCheck) runCardHomogeneityCheck(
+	cardActive, cardTotal map[string]int,
+	cardRole map[string]topology.Role,
+) []*pb.HealthEvent {
+	anomalies := c.classifier.CheckCardHomogeneity(cardActive, cardTotal, cardRole)
+
+	var events []*pb.HealthEvent
+
+	for card, a := range anomalies {
+		if _, already := c.previousAnomalies[card]; already {
```
lalitadithya: is we run this on only on first poll, won't the `previousAnomalies` be empty by definition?

health-monitors/nic-health-monitor/pkg/discovery/discovery.go
```diff
@@ -0,0 +1,255 @@
...
+
+// DiscoverDevices enumerates all IB/RoCE devices from sysfs, parsing each
+// device's metadata and ports. SR-IOV VFs are tagged via IBDevice.IsVF so
+// callers can filter them out; they are never elided from the returned
+// slice so operators can still observe their presence.
+//
+// The exclusionRegex argument is a comma-separated list of regexes that
+// filter device *names* (e.g., "^veth.*,^lo$"). Devices matching any
+// pattern are skipped entirely.
+func DiscoverDevices(reader sysfs.Reader, exclusionRegex string) ([]IBDevice, error) {
```
lalitadithya: Can we filter out VFs and devices that we know we don't want to monitor here so that we don't have to do these checks as part of the health check?

distros/kubernetes/nvsentinel/charts/nic-health-monitor/templates/_helpers.tpl
KaivalyaMDabhadkar: Right, kata containers will not expose the sysfs files of the host to us, so we will not have a kata mode for NIC health monitor at all. Fixed it

distros/kubernetes/nvsentinel/charts/nic-health-monitor/templates/configmap.yaml
KaivalyaMDabhadkar: Done

health-monitors/nic-health-monitor/pkg/checks/state/ethstate.go
KaivalyaMDabhadkar: Refactored to reduce duplicated code

health-monitors/nic-health-monitor/pkg/topology/topology.go
KaivalyaMDabhadkar: Done

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
KaivalyaMDabhadkar: Good catch, It was just a safety check which was later made redundant. Removed it

health-monitors/nic-health-monitor/pkg/discovery/discovery.go
KaivalyaMDabhadkar: Good idea, done.

health-monitors/nic-health-monitor/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,50 @@
...
+		Help:    "Duration of each poll cycle in seconds",
+		Buckets: prometheus.DefBuckets,
+	}, []string{"node", "category"})
+
+	// StateCheckErrors counts per-device state-check error events (port
+	// DOWN, device disappeared, etc.).
+	StateCheckErrors = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "nic_health_monitor_state_check_errors_total",
+		Help: "Total number of state check error events",
+	}, []string{"node", "check", "device", "port"})
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: the cardinality is bounded by physical NICs *  ports per node so it's not truly unbounded

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
```diff
@@ -0,0 +1,405 @@
...
+		return nil, fmt.Errorf("device discovery failed: %w", err)
+	}
+
+	metrics.DevicesDiscovered.WithLabelValues(c.nodeName).Set(float64(len(devices)))
+
+	firstPoll := c.previousDevices == nil
+	baselineRun := firstPoll && c.emitHealthyBaselines
+	st := newIBPollState()
+
+	c.collectDevicesAndPorts(devices, st)
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: In practice, NicExclusionRegex is set at deploy time via the ConfigMap and doesn't change without a pod restart. On restart, statefile.Load() discards the persisted state if the schema version has changed, and seedFromPersistedState() rebuilds previousDevices from the current PortStates so a config change + restart won't produce a phantom "device disappeared" event.

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
```diff
@@ -0,0 +1,405 @@
...
+		"prevState", prev.State, "prevPhysState", prev.PhysicalState,
+		"state", current.State, "physState", current.PhysicalState,
+		"wasHealthy", wasHealthy, "isHealthy", isHealthy,
+	)
+
+	if isHealthy {
+		return c.healthyBaselineEvent(current)
+	}
+
+	return c.unhealthyPortEvent(current)
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: This is intentional. The monitor only reports boundary crossings between the healthy and unhealthy buckets (ACTIVE/LinkUp vs everything else). Intermediate unhealthy->unhealthy transitions like INIT/Polling -> DOWN/Disabled are suppressed because:

1. INIT/Polling is a transient state that resolves within seconds during link negotiation ,  emitting events for it would flood the platform with noise.
2. The downstream remediation system acts on fatal vs non-fatal, not on sub-states within the unhealthy bucket. A severity-class comparison would add complexity without changing the remediation outcome.
3. If a port transitions ACTIVE -> INIT/Polling -> DOWN/Disabled, the ACTIVE -> INIT/Polling crossing already emits the fatal event. The subsequent INIT -> DOWN is a continuation of the same failure, not a new escalation.

health-monitors/nic-health-monitor/pkg/statefile/statefile.go
```diff
@@ -0,0 +1,395 @@
...
+		seen[v.Device] = true
+	}
+
+	devices := make([]string, 0, len(seen))
+	for d := range seen {
+		devices = append(devices, d)
+	}
+
+	sort.Strings(devices)
+	m.state.KnownDevices = devices
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: This is the intended behavior . The previous implementation merged knownDevices with the existing set, which caused stale devices (e.g. VFs that were once visible but later removed) to persist indefinitely in the state file. On every poll, detectDeviceDisappearance() would compare previousDevices against seenDevices, find the stale entry missing, and emit a repeated "device disappeared" FATAL event , flooding logs and the platform.

health-monitors/nic-health-monitor/pkg/topology/topology.go
```diff
@@ -0,0 +1,557 @@
...
+
+	if checkDPU && c.isBlueFieldDPU(device) {
+		return RoleManagement
+	}
+
+	if c.isInfiniBandDevice(device) {
+		return RoleCompute
+	}
+
+	return RoleStorage
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: This is intentional design and is documented in the design doc. Please check. The InfiniBand link-layer promotion handles platforms (like GB200) where nvidia-smi topo -m shows NODE or SYS relationships for NICs that are actually used for GPU-to-GPU RDMA traffic. On these platforms, the topology matrix doesn't expose fine-grained NIC-GPU PCIe relationships, so NODE/SYS is the best we get.

The heuristic: if a NIC is on a GPU NUMA socket and uses InfiniBand link layer, it's almost certainly a compute NIC carrying GPU traffic, regardless of what the topology matrix says. Ethernet NICs on the same path correctly fall through to RoleStorage).

health-monitors/nic-health-monitor/main.go
```diff
@@ -87,49 +191,273 @@ func run() error {
...
+	}
+
+	slog.Warn("Failed to load config file, using defaults", "error", err, "path", path)
+
+	slog.Info("Using default configuration with CLI flags")
+
+	return &config.Config{
+		SysClassInfinibandPath: "/nvsentinel/sys/class/infiniband",
+		SysClassNetPath:        "/nvsentinel/sys/class/net",
+	}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: The fallback is intentional and documented in the function's comment

health-monitors/nic-health-monitor/pkg/topology/topology.go
```diff
@@ -0,0 +1,557 @@
...
+		"storage", storage,
+		"management", management,
+	)
+}
+
+// classify encapsulates the compute/storage/management decision. It is
+// separated from RoleOf so tests can drive it directly without priming
+// the cache.
+func (c *Classifier) classify(device string) Role {
+	if c.defaultRouteDevice != "" && device == c.defaultRouteDevice {
```
XRFXLP, line 211: This appear to be inconsistent with the doc where we have things other way round:
```
    if any GPU has PIX or PXB:
        return Compute
    # Step 2: Default route exclusion (runs before topology classification)
    # Reads /proc/net/route to find the default route interface, maps it
    # to an IB device via sysfs, and excludes it as Management.
    if device == default_route_device:
        return Management
```

health-monitors/nic-health-monitor/pkg/checks/state/ethstate.go
```diff
@@ -0,0 +1,358 @@
...
+		return nil
+	}
+
+	if firstPoll {
+		card := portCard[pi.key]
+		if _, expected := expectedCards[card]; expected {
+			slog.Info("Suppressing first-poll fatal for expected-down RoCE port",
+				"device", pi.snap.Device, "port", pi.snap.Port, "card", card)
+
+			return nil
```
XRFXLP: This appears to be different than the first poll suppression trail that we've in case of IB: where we send the unhealthy non-fatal event and bump up the metrics. Is this intentional?

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
```diff
@@ -0,0 +1,400 @@
...
+// CheckCardHomogeneity) converts "stuck in a non-fatal intermediate
+// state" into a fatal event by observing that the card has fewer
+// active ports than its peers.
+func (c *InfiniBandStateCheck) unhealthyPortEvent(snap portSnapshot) *pb.HealthEvent {
+	isFatal := snap.State == checks.IBStateDown
+
+	switch snap.PhysicalState {
+	case checks.IBPhysDisabled:
+		isFatal = true
+	case checks.IBPhysLinkErrorRecovery, checks.IBPhysPolling:
```
XRFXLP, line 383: In multiples places in the doc: https://github.com/NVIDIA/NVSentinel/blob/main/docs/nic-health-monitor/link-state-detection.md?plain=1#L257, https://github.com/NVIDIA/NVSentinel/blob/main/docs/nic-health-monitor/link-state-detection.md?plain=1#L1166
physical state `LinkErrorDiscovery` is treated as fatal but in the code it is otherwise. Can we fix the inconsistency?

health-monitors/nic-health-monitor/pkg/checks/state/base.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	devices := make([]string, 0, len(currentDevices))
+	for d := range currentDevices {
+		devices = append(devices, d)
+	}
+
+	b.state.UpdatePortStates(snapshots, devices, linkLayer)
+
+	if err := b.state.Save(); err != nil {
```
XRFXLP, line 257: Do we need to run this even if state are unchanged?

health-monitors/nic-health-monitor/pkg/checks/state/base.go
```diff
@@ -0,0 +1,265 @@
...
+		events = append(events, checks.NewHealthEvent(
+			b.nodeName, b.strategy.checkName(),
+			b.strategy.formatDeviceDisappearance(device),
+			checks.DeviceEntities(device),
+			true, false, pb.RecommendedAction_REPLACE_VM, b.processingStrategy,
+		))
+
+		// Remove so we don't re-emit on the next poll if
+		// previousDevices hasn't been replaced yet.
+		delete(b.previousDevices, device)
```
XRFXLP: Do we need this?

Since we're overwriting this anyway in the next cycle down the line: https://github.com/NVIDIA/NVSentinel/blob/pull-request/1229/health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go#L153?

Same question for deleting entry in previous ports.

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
```diff
@@ -0,0 +1,400 @@
...
+}
+
+// Run executes a single poll cycle and returns the resulting events.
+func (c *InfiniBandStateCheck) Run() ([]*pb.HealthEvent, error) {
+	result, err := discovery.DiscoverDevices(c.reader, c.cfg.NicExclusionRegex)
+	if err != nil {
+		return nil, fmt.Errorf("device discovery failed: %w", err)
+	}
+
+	metrics.DevicesDiscovered.WithLabelValues(c.nodeName).Set(float64(len(result.Devices)))
```
XRFXLP: We're calling the same metrics with same labels both in Eth state check and IB state check, shuold we add a label in the metric?

Same question for:
```
	result, err := discovery.DiscoverDevices(c.reader, c.cfg.NicExclusionRegex)
	if err != nil {
		return nil, fmt.Errorf("device discovery failed: %w", err)
	}
```

looks like called twice for each check.

health-monitors/nic-health-monitor/pkg/discovery/discovery.go
```diff
@@ -0,0 +1,267 @@
...
+
+	for _, pat := range strings.Split(commaSeparated, ",") {
+		pat = strings.TrimSpace(pat)
+		if pat == "" {
+			continue
+		}
+
+		re, err := regexp.Compile(pat)
+		if err != nil {
+			slog.Warn("Invalid regex, skipping", "pattern", pat, "error", err)
```
XRFXLP, line 242: should this be validated in LoadConfig rather than silently logged?

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
```diff
@@ -0,0 +1,400 @@
...
+
+	evt.IsFatal = false
+	evt.RecommendedAction = pb.RecommendedAction_NONE
+}
+
+// logDiscoverySummaryIfChanged emits a one-line summary whenever the
+// discovered set of devices/ports changes size. On the first poll
+// previousDevices is nil (len 0), so the size always differs and the
+// summary is logged unconditionally.
+func (c *InfiniBandStateCheck) logDiscoverySummaryIfChanged(st *ibPollState) {
```
XRFXLP, line 289: should we add similar function in ethstate.go?

health-monitors/nic-health-monitor/pkg/topology/topology.go
KaivalyaMDabhadkar: Good catch, the doc pseudocode had the default route check after PIX/PXB, but the code runs it first. The code ordering is correct: if a NIC carries the host's default route, it should be excluded as Management nic regardless of PCIe topology. This matters on platforms like GB200 where a management NIC could theoretically share a PCIe switch with a GPU. Updated the doc to match the code.

health-monitors/nic-health-monitor/pkg/checks/state/ethstate.go
KaivalyaMDabhadkar: Ethernet path now emits a non-fatal event with RecommendedAction_NONE, matching the IB behavior

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
KaivalyaMDabhadkar: The code is correct, LinkErrorRecovery is non-fatal transient state. Fixed the docs

health-monitors/nic-health-monitor/pkg/checks/state/base.go
KaivalyaMDabhadkar: Updated the code and UpdatePortStates now compares old vs new state entry-by-entry and returns bool indicating whether anything changed. persistState in base.go skips the Save() call entirely when we return false

health-monitors/nic-health-monitor/pkg/checks/state/base.go
KaivalyaMDabhadkar: Removed both deletes (previousDevices and previousPorts).

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
KaivalyaMDabhadkar: good point, added a check label to DevicesDiscovered so the IB and Ethernet checks emit distinct metric series

health-monitors/nic-health-monitor/pkg/discovery/discovery.go
KaivalyaMDabhadkar: Fixed

health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go
KaivalyaMDabhadkar: Added

XRFXLP (approved):

PR #1240, opened 2026-04-30
feat: fallback to a reboot if a GPU reset fails

natherz97:
## Summary

This PR adds logic to fallback to a reboot if a GPU reset fails.

**1. Create a RESTART_VM event when GPU resets fail:** currently, a healthy event is emitted by the syslog-health-monitor when it detects that a GPU was successfully reset by consuming a syslog event written by the GPU reset job. Now, we will also start emitting an unhealthy event with a RESTART_VM recommended action by the syslog-health-monitor when it detects that a GPU failed to be reset. The resulting reboot will clear both events for the initial XID needing a reset and subsequent failed reset event needing a reboot and the node will be uncordoned.
**2. Add a new writeSyslogEvent option to the Janitor config:** this gpuResetController config property defaults to true if not explicitly set. If true, the GPU reset job will write syslog events for successful and failed resets that will be consumed by the syslog-health-monitor. If false, the GPU reset will not write syslog events which will prevent automatic uncordoning for successful resets and reboot fallbacks for failed resets. Opting out of the syslog writing behavior might be desirable for GPUResets triggered outside of NVSentinel so that debugging failed resets can occur without reboots being triggered.
```
  apiVersion: janitor.dgxc.nvidia.com/v1alpha1                                                                                                                                                                                                                                                                    
  kind: GPUReset                                                                                                                                                                                                                                                                                                  
  metadata:                                                                                                                                                                                                                                                                                                       
    name: maintenance-my-node-abc123                                                                                                                                                                                                                                                                              
    annotations:                    
      nvsentinel.nvidia.com/trace-id: "some-trace-id"                                                                                                                                                                                                                                                             
      nvsentinel.nvidia.com/span-id: "some-span-id"                                                                                                                                                                                                                                                               
  spec:                                                                                                                                                                                                                                                                                                           
    nodeName: my-node                                                                                                                                                                                                                                                                                             
    selector:                                                                                                                                                                                                                                                                                                     
      uuids:                                                                                                                                                                                                                                                                                                      
        - GPU-455d8f70-2051-db6c-0430-ffc457bff834
    writeSyslogEvent: false 
```

## Overview for the reboot fallback

A COMPONENT_RESET remediation will result in the following log line being emitted to syslog and consumed by the syslog-health-monitor (regardless of which health-monitor created the original COMPONENT_RESET unhealthy event):

```
GPU reset executed: GPU-455d8f70-2051-db6c-0430-ffc457bff834, success: <true/false>
```

1. Successful resets: will result in a healthy event that can clear XID errors from the SysLogsXIDError check for matching impacted entities (GPU UUID and PCI ID). Note that if a different health-monitor emitted the original event, it would need to also check syslog or implement a different detection mechanism for GPU resets (for example the gpu-health-monitor relies on resets fixing the underlying DCGM watch or by having the nvidia-dcgm pod restarted as part of the GPU reset workflow). 
4. Failed resets: will result in a new unhealthy event from the SysLogsXIDError check with the same impacted entities (GPU UUID and PCI ID) as the original health event and a RESTART_VM recommended action. Note that this flow will be triggered regardless of which health-monitor emitted the original COMPONENT_RESET event. This serves as a fallback where we will reboot a node if a GPU reset fails. This will result in the node being cordoned due to 2 events which are the original XID event and this subsequent GPU reset failed event. The syslog-health-monitor has logic to clear all unhealthy events for each of its checks in response to a reboot by sending a healthy event with empty impacted entities. As a result, we should expect that the node will be uncordoned after the reboot completes.

Example event:
```
	{
	  createdAt: ISODate('2026-04-30T10:46:50.263Z'),
	  healthevent: {
	    agent: 'syslog-health-monitor',
	    componentclass: 'GPU',
	    checkname: 'SysLogsXIDError',
	    isfatal: true,
	    ishealthy: false,
	    message: ‘GPU reset failed, proceeding with a node reboot',
	    recommendedaction: RESTART_VM,
	    errorcode: [
	      'GPU_RESET_FAILURE'
	    ],
	    entitiesimpacted: [
	      {
	        entitytype: 'PCI',
	        entityvalue: '000b:00:00'
	      },
	      {
	        entitytype: 'GPU_UUID',
	        entityvalue: 'GPU-123’
	      }
	    ],
	    nodename: ‘node-123’,
	  }
	}
```
Notes:

- A follow-up unhealthy event is required to trigger a full drain in node-drainer because the original COMPONENT_RESET event would've done a partial drain only against pods using the GPU needing reset.
- If a burst of XIDs occur with both COMPONENT_RESET and RESTART_VM recommended actions and the GPU reset fails, this logic is not necessary because a reboot would already be triggered. We will rely on the existing fault-remediation de-duplication logic to only process one of the reboots.
- This logic is meant as a fallback when there are one of more XIDs with COMPONENT_RESET recommended actions which all result in failed GPU resets. When this logic is triggered, the following steps should be followed:
   - Fix the underlying cause for the GPU reset failure.
   - If the reset failure is isolated to a given XID, override the recommended action from COMPONENT_RESET to RESTART_VM.
   - If the GPU reset failure is not unique to a specific XID error, disable the GPU reset feature to always reboot.

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ X] Health Monitors
- [X] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * GPU reset syslog entries now include a success/failure flag and can be disabled (default: enabled).
  * GPU reset job exposes a configurable option to control syslog emission.

* **Behavior Changes / Health**
  * Health monitor distinguishes successful vs failed GPU resets and emits appropriate health events and recommended actions.

* **Tests**
  * Expanded unit, integration, and UAT coverage for success/failure flows and boot-ID verification.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/distros/kubernetes/nvsentinel/charts/janitor/crds/janitor.dgxc.nvidia.com_gpuresets.yaml
+++ b/distros/kubernetes/nvsentinel/charts/janitor/crds/janitor.dgxc.nvidia.com_gpuresets.yaml
@@ -28,11 +28,11 @@ spec:
   scope: Cluster
   versions:
   - additionalPrinterColumns:
-    - description: The target node for the GPU reset
+    - description: Target node for GPU reset
       jsonPath: .spec.nodeName
       name: Node
       type: string
-    - description: The current status of the reset
+    - description: Current status of reset
       jsonPath: .status.phase
       name: Status
       type: string
--- a/distros/kubernetes/nvsentinel/charts/janitor/templates/configmap.yaml
+++ b/distros/kubernetes/nvsentinel/charts/janitor/templates/configmap.yaml
@@ -85,6 +85,7 @@ data:
       {{- end }}
       {{- if .Values.config.controllers.gpuReset.resetJob }}
       resetJob:
+        writeSysLogEvent: "{{ .Values.config.controllers.gpuReset.resetJob.writeSysLogEvent }}"
         runtimeClassName: "{{ .Values.config.controllers.gpuReset.resetJob.runtimeClassName }}"
         imageConfig:
           image: "{{ .Values.config.controllers.gpuReset.resetJob.image.repository }}:{{ .Values.config.controllers.gpuReset.resetJob.image.tag | default ((.Values.global).image).tag | default .Chart.AppVersion }}"
--- a/distros/kubernetes/nvsentinel/charts/janitor/values.yaml
+++ b/distros/kubernetes/nvsentinel/charts/janitor/values.yaml
@@ -165,6 +165,7 @@ config:
       serviceManager:
         name: "gpu-operator"
       resetJob:
+        writeSysLogEvent: true
         runtimeClassName: "nvidia"
         # imagePullSecrets can be specified under image.imagePullSecrets, otherwise we will
         # fall back to global.imagePullSecrets.
--- a/distros/kubernetes/nvsentinel/values-tilt.yaml
+++ b/distros/kubernetes/nvsentinel/values-tilt.yaml
@@ -3777,6 +3777,7 @@ janitor:
               enabledValue: "true"
               disabledValue: "false"
         resetJob:
+          writeSysLogEvent: false
           runtimeClassName: ""
           image:
             repository: "public.ecr.aws/docker/library/alpine"
--- a/gpu-reset/gpu_reset.sh
+++ b/gpu-reset/gpu_reset.sh
@@ -163,10 +163,13 @@ DURATION=$(printf "%.3f" "$DURATION_RAW")
 
 if [ "$FINAL_EXIT_STATUS" -ne 0 ]; then
   log "FAILED: GPU reset workflow failed in ${DURATION}s."
+  SYSLOG_SUCCESS="false"
 else
   log "SUCCESS: GPU reset workflow completed in ${DURATION}s."
+  SYSLOG_SUCCESS="true"
+fi
 
-  # Write "GPU reset executed" message to notify syslog-health-monitor of GPU reset success
+if [ "${WRITE_SYSLOG_EVENT:-true}" = "true" ]; then
   ORIGINAL_IFS=$IFS
   IFS=','
   for UUID in $TARGET_UUIDS; do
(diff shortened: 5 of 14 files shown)


janitor/api/v1alpha1/gpureset_types.go
```diff
@@ -114,6 +114,12 @@ type GPUResetSpec struct {
 	// +kubebuilder:validation:Optional
 	// +optional
 	Selector *GPUSelector `json:"selector,omitempty"`
+
+	// WriteSyslogEvent controls whether the GPU reset job writes a syslog entry
+	// with the result of the reset operation.
+	// +kubebuilder:default:=true
+	// +optional
+	WriteSyslogEvent *bool `json:"writeSyslogEvent,omitempty"`
```
pteranodan: This feels like a leaky abstraction. Should logging behavior be part of the API spec, or would it be better handled as a configuration for the controller/job itself?

natherz97: This is a good callout. If we're fine with requiring that Janitor is updated in order to change whether GPUResets result in a reboot or not, then we can remove this new WriteSyslogEvent field from the CRD and add it to the Janitor config. If an operator wants to change the behavior, they'll need to follow these steps:

1. Update the Janitor configmap with the new WriteSyslogEvent value
2. Initiate a rolling update of the Janitor deployment
3. Trigger manual GPUResets
4. Update the Janitor configmap back to the old WriteSyslogEvent value
5. Initiate a rolling update of the Janitor deployment

I think this is a reasonable tradeoff since manual GPUResets will likely only be required for development work and the fallback to reboot would be desirable for NVSentinel and manual oncall actions in most situations.

pteranodan: Agreed.

neerajnv:
LGTM!

lalitadithya (approved):

PR #1253, opened 2026-05-06
docs: Add health event based cancellation logic design doc

### LINKED ISSUE #1252: [Feature]: Add ability to cancel XIDs based on other XIDs
### Prerequisites

### Feature Summary

Enhance NVSentinel to uncordon based on individual XIDs as in this case since XID163 occured followed by XID 162 so issue is gone.

### Problem/Use Case

One of new XID pair that fits this pattern is XID 162 and XID 163 which mean PSHC_REENGAGED and PSHC_DISENGAGED respectively, with recommended action for XID 163 of reset gpu if PSHC need to be enabled, but we need to cancel that reset GPU pipeline if we already got XID 162 (i.e. PSHC has been re-engaged)

Ref: https://docs.nvidia.com/deploy/xid-errors/analyzing-xid-catalog.html

### Proposed Solution

TBD

### Component

Health Monitor

XRFXLP:
## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added a design document describing health monitor cancellation rules: per-monitor cancellation behavior, synthetic "healthy" event emission, TOML configuration schema examples, notes on resolver and deployment integration, and observability counters for tracking cancellations.

### DIFF
--- a/docs/designs/038-health-monitor-cancellation-rules.md
+++ b/docs/designs/038-health-monitor-cancellation-rules.md
@@ -7,7 +7,7 @@ A fatal `HealthEvent` quarantines a node and is released only by:
 1. **Manual uncordon** — operator runs `kubectl uncordon`; fault-quarantine detects the transition and calls `CancelLatestQuarantiningEvents`. See [`cancelling-breakfix.md`](../cancelling-breakfix.md).
 2. **Reboot / device reset** — a monitor (typically syslog) observes the recovery and emits a healthy `HealthEvent` for the same `CheckName`, which flips `node.status.conditions` to `False` and clears the fault-quarantine annotation map.
 
-There is no way to express that a *different* observation clears a prior fault. An operator who knows that XID 162 implies recovery from XID 163 cannot configure NVSentinel to act on that knowledge — they must wait for a device-specific recovery signal that may never fire, or uncordon manually.
+There is no way to express that a *different* error code clears a prior fault. An operator who knows that XID 162 implies recovery from XID 163 cannot configure NVSentinel to act on that knowledge — they must wait for a device-specific recovery signal that may never fire, or uncordon manually.
 
 This ADR adds a third path: **per-monitor cancellation rules** that declare "observation A clears observation B" pairs in configuration owned by the monitor that produces both observations.
 
--- a/docs/designs/038-health-monitor-cancellation-rules.md
+++ b/docs/designs/038-health-monitor-cancellation-rules.md
@@ -22,7 +22,7 @@ The natural owner of cancellation pairs is therefore the monitor that produces t
 
 ## Decision
 
-When a monitor's parser observes an event matching a configured rule, the monitor emits one or more synthetic healthy `HealthEvent`s in the same gRPC batch. Those events flow through the unchanged platform-connector pipeline and are folded by every downstream consumer (K8s connector, store, device connector, fault-quarantine) exactly as if produced by a real recovery observation. Two downstream clearers that today ignore `ErrorCode` are extended to honour it, so a cancellation event clears only the targeted code on the targeted entity rather than every code on that entity (see *Error-code precision in downstream clearers* in Implementation).
+On a configured rule match, the monitor appends synthetic healthy `HealthEvent`s to the same gRPC batch alongside the original; the original is never mutated. Each synthetic event flows through the unchanged platform-connector pipeline and is folded by every downstream consumer exactly like a real recovery event. Two downstream clearers are extended to honour `ErrorCode` so a cancellation clears only its targeted code (see *Error-code precision in downstream clearers*).
 
 Initial schema (`syslog-health-monitor`):
 
@@ -147,6 +147,7 @@ One new counter: `syslog_health_monitor_cancellations_emitted_total{check, sourc
 ### Positive
 
 - Adds a declarative resolution path without any change to platform-connectors or fault-quarantine.
+- **Additive, never mutating.** The original observation is left intact, so the upstream signal stays auditable and its own semantics (metrics, downstream rules keyed off the source code) are unaffected. A single source code can fan out to multiple cancellations, which mutation could not express.
 - Synthetic events carry additional metadata (`Metadata["nvsentinel.io/cancel-source-error-code"]`) for correlation between the trigger and its cancellation.
 - The existing hardcoded GPU-reset cancel in `createHealthEventGPUResetEvent` becomes a candidate for later migration onto this same mechanism.
 
--- a/docs/designs/038-health-monitor-cancellation-rules.md
+++ b/docs/designs/038-health-monitor-cancellation-rules.md
@@ -1,0 +1,174 @@
+# ADR-038: Health Monitors — Health Event Cancellation Rules
+
+## Context
+
+A fatal `HealthEvent` quarantines a node and is released only by:
+
+1. **Manual uncordon** — operator runs `kubectl uncordon`; fault-quarantine detects the transition and calls `CancelLatestQuarantiningEvents`. See [`cancelling-breakfix.md`](../cancelling-breakfix.md).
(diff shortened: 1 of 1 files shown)


docs/designs/038-health-monitor-cancellation-rules.md
```diff
@@ -0,0 +1,175 @@
...
+| `IsFatal`                                                                         | `false`                                                             |
+| `RecommendedAction`                                                               | `NONE`                                                              |
+| `ErrorCode`                                                                       | the matched entry from `cancelErrorCodes`, as a single-element list |
+| `Message`                                                                         | `Cancelled by <CheckName> error code <onErrorCode>`                 |
+| `Metadata["nvsentinel.io/cancel-source-error-code"]`                              | the rule's `onErrorCode`                                            |
+| `GeneratedTimestamp`                                                              | `time.Now()`                                                        |
+
+
+The source event is unchanged; synthetic events are appended to the same `*pb.HealthEvents` batch.
```
lalitadithya, line 57: Where is the source event stored? How is it presisted in case of container reboots/upgrades?

docs/designs/038-health-monitor-cancellation-rules.md
```diff
@@ -0,0 +1,175 @@
...
+### Observability
+
+One new counter: `syslog_health_monitor_cancellations_emitted_total{check, source_error_code, target_error_code}`, incremented per synthetic event. Downstream layers see synthetic events as ordinary healthy events and account for them through existing metrics.
+
+## Consequences
+
+### Positive
+
+- Adds a declarative resolution path without any change to platform-connectors or fault-quarantine.
+- Synthetic events are persisted with `Metadata["nvsentinel.io/cancel-source-error-code"]`, providing a join from the trigger to its cancellation.
```
lalitadithya: persisted where? Is this an additional metadata that is added to the health event?

docs/designs/038-health-monitor-cancellation-rules.md
```diff
@@ -0,0 +1,175 @@
...
+- The existing hardcoded GPU-reset cancel in `createHealthEventGPUResetEvent` becomes a candidate for later migration onto this same mechanism.
+
+### Negative
+
+- **Entity-level granularity.** The K8s connector's `removeImpactedEntitiesMessages` and the fault-quarantine annotation map both clear by `EntitiesImpacted`, ignoring `ErrorCode`. A synthetic healthy event for one XID will also clear other XIDs on the same entity until they re-fire. Tightening this to error-code precision requires changes to both clearers and is out of scope.
```
lalitadithya: Why would this be out of scope? This would cause real errors to be masked right?

docs/designs/038-health-monitor-cancellation-rules.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Alternative C — Promote each XID code to its own `CheckName`
+
+Emit `CheckName = "GpuXid163"`, `"GpuXid162"`, etc., and use the existing per-check recovery path.
+
+**Rejected** because every distinct XID would become its own `node.status.conditions[Type]`, fragmenting the consolidated `SysLogsXIDError` view and breaking every existing fault-quarantine rule and runbook that references it.
+
+### Alternative D — Hardcoded XID pairs in `XIDHandler`
+
+**Rejected** because pairs vary by driver version, hardware family, and operator policy.
```
lalitadithya, line 181: Even with the current approach pair differences by driver version or hardware family can't be handled right?

docs/designs/038-health-monitor-cancellation-rules.md
XRFXLP: we don't have to store the source event, it is the incoming event.

docs/designs/038-health-monitor-cancellation-rules.md
XRFXLP: We can have conditional templates on basis of hardware family and driver version as it is only the configs, right?

docs/designs/038-health-monitor-cancellation-rules.md
XRFXLP: yes, additional metadata added to health event.

docs/designs/038-health-monitor-cancellation-rules.md
lalitadithya: The table above says `copied from the source event`, in this context what is `copied` and from where?

docs/designs/038-health-monitor-cancellation-rules.md
lalitadithya: okay, can we update the language to something like `Additional metadata included` or something like that?

docs/designs/038-health-monitor-cancellation-rules.md
lalitadithya: If one were just helm install that wouldn't be possible right? That would only be possible if we have another layer of templating above the helm install. Is this additional layer what you are referring to here?

XRFXLP: even if we do helm install then we can change the configuration, it is not that we've to do the code changes. templating is just one easy to way store the config before deployment.

docs/designs/038-health-monitor-cancellation-rules.md
XRFXLP: ``Agent`, `ComponentClass`, `CheckName`, `Version`, `NodeName`, `EntitiesImpacted`` are copied from source event. For concreteness: source event can be XID 162 and synthetic event emitted would be XID 163.

docs/designs/038-health-monitor-cancellation-rules.md
lalitadithya: right, yeah

docs/designs/038-health-monitor-cancellation-rules.md
lalitadithya: right, so in this case 163 can be emitted at T0 and 162 at T1. So, what we are saying here is that at T1 the copy will be from 162 into a new event. My next question is why we can't just mutate 162 event, why do we need a synthetic copy?

XRFXLP: If we just mutate XID 162, that original event would be lost and we would lose all the observability around it like node events.

docs/designs/038-health-monitor-cancellation-rules.md
XRFXLP: changed

docs/designs/038-health-monitor-cancellation-rules.md
XRFXLP: removed this line, added changes required in FQ and PC

docs/designs/038-health-monitor-cancellation-rules.md
lalitadithya: I see, that is a good point. Can we add this reason into the document? 

If I were to restate it, the reason we are doing this because the driver sort of overloads error code to signify both health and error in this case. So, we need to publish the fact that an "error" occurred along with our interpretation that this signifies a healthy state. Is this what we are saying?

lalitadithya: maybe we already have it in `There is no way to express that a *different* observation clears a prior fault. `, maybe we can say that a different `error code` instead of observation

XRFXLP: done

lalitadithya (approved):

Issue #1276, opened 2026-05-11
[Feature/Enhancement]: Coordinate node ownership with external remediation systems


jtschelling:
### Prerequisites

- [x] I searched existing issues

### Feature Summary

Provide a way for NVSentinel and an external remediation system to safely transfer ownership of a node between them, so that destructive actions (cordon, drain, reboot, terminate, replace, etc.) cannot race or duplicate work between the two systems.

### Problem/Use Case

NVSentinel is currently the sole owner of node-level destructive actions in the clusters where it runs. Its pipeline (health-monitor → fault-quarantine → node-drainer → fault-remediation) assumes nothing else is concurrently acting on the same node, and there is no in-band mechanism for an outside system to declare "I need to take this node" or for NVSentinel to declare "I am done with this node, you may have it."

In practice, multiple operational scenarios require an external system to perform actions NVSentinel cannot perform itself:

- **Faults whose remediation lives outside the cluster** — Long-running workflows, CSP-side hardware repair, manual operator intervention, physical maintenance. NVSentinel can detect, quarantine, and drain, but cannot complete the fix.
- **Faults NVSentinel cannot observe** — CSP-scheduled maintenance windows, out-of-band monitoring signals, operator-driven decommissioning. An external system detects the issue and needs NVSentinel to prepare the node (cordon + drain) before the external work begins.

Without a coordination mechanism:

- Both systems can try to act on the same node at the same time.
- An external system has no reliable way to know NVSentinel has finished draining and is no longer touching the node.
- NVSentinel has no reliable way to know an external system is currently mid-remediation and that it should not start its own remediation flow on the same node.
- Returning the node to normal operation (clearing quarantine taints, uncordoning, re-enabling scheduling) requires NVSentinel's pipeline, but the external system has no way to signal "I'm done; please resume."

Today this is worked around through side channels — operator coordination, ad-hoc taints, separate runbooks. That works at small scale but does not scale to multiple integrated systems or unattended automation.

### Goals

A coordination mechanism between NVSentinel and external systems should:

1. **Define ownership unambiguously.** At any point in time, a given node is owned by NVSentinel *or* by exactly one external system — never both, never neither.
2. **Provide a single entry point** into external ownership. Any external system that needs to take a node says so in one well-defined way.
3. **Provide a single exit point** out of NVSentinel ownership and back. Any external system that has finished its work returns the node in the same well-defined way.
[...]

PR #1282, opened 2026-05-13
wip: add ADR-040 — External Remediations (EF / ERR) coordination

jtschelling:
## Summary

Adds **ADR-040: External Remediations (EF / ERR)**, defining the API and reconciler design for coordinating node ownership between NVSentinel and external remediation systems (CSP-side repair, RMA workflows, hardware swaps, operator-driven maintenance).

Two new CRDs in the existing `nvsentinel.nvidia.com/v1alpha1` API group:

- **`ExternalRemediationRequest` (ERR)** — exit door from NVSentinel ownership. Created by `fault-remediation` (when triage hits the `CUSTOM` action from ADR-036) or by the `ExternalFault` reconciler. Applies a release taint to the node carrying its own name in the value; external system reads it, does work, writes a terminal completion condition back.
- **`ExternalFault` (EF)** — entry door for external systems declaring "I need to take this node." Triggers the standard pipeline (quarantine → drain → fault-remediation), which produces the ERR.

Both spec types are the existing `datamodels.HealthEvent` (reused by reference in the proto — no schema duplication). Coordination state lives in `status.conditions`, following the `Unknown`/`True`/`False` terminal-state convention already established by `RebootNode`/`GPUReset`/`TerminateNode`.

CRDs are generated proto-driven via the existing `protoc-gen-crd` plugin (same plugin already used for `HealthEventResource`). Reconcilers + a small `metav1.Condition` adapter live alongside the existing maintenance reconcilers in `janitor/`. A validating admission webhook (modelled on the existing `janitor_webhook.go`) rejects malformed EFs at admission.

## Design highlights

- **Single coordination surface** — one entry point (EF), one exit point (ERR). The same protocol works for automated orchestrators and human operators.
- **`HealthEvent`-driven** — same proto already produced by every health-monitor and consumed by every downstream stage.
- **No timeouts, no NVSentinel assumptions about external behaviour** — `ExternalRemediationComplete` can stay `Unknown` indefinitely; observability surfaces the wait.
- **Asymmetric handling of `ExternalRemediationComplete=False`** — for NVSentinel-detected ERRs, taint comes off and existing health-monitors re-detect persistent faults. For EF-owned ERRs, the EF reconciler does NOT emit a healthy event (NVSentinel can't verify the external fault is gone); fault-quarantine taint stays on, the EF goes terminal, and the external system can create a new EF as the explicit retry path.
- **Reuses `fault-remediation`'s existing equivalence-group + `latestFaultRemediationState` machinery** for "single active ERR per node" — no bespoke dedup code.
- **Release taint value carries the owning ERR name** as the operator's discovery path (no extra Node labels or annotations).
- **Cross-kind supersedence** between ERR and `RebootNode`/`GPUReset` declared via existing `supersedingEquivalenceGroups` config — no new code.

Refs: #1276

## Summary by CodeRabbit

* **Documentation**
  * Added ADR describing an external remediations protocol for coordinated node handoff between systems, including message/condition semantics, reconciliation behaviors, admission validation and safety checks, integration with existing remediation flows, operator discovery and RBAC guidance, observability considerations (metrics, events, tracing, alerts), failure modes, alternatives, and detailed example workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NVSentinel/pull/1282?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


docs/designs/040-external-remediations.md
```diff
@@ -0,0 +1,735 @@
...
+
+Both CRD specs are a `datamodels.HealthEvent` (the same proto already emitted by every health-monitor and consumed by every downstream stage — see `data-models/protobufs/health_event.proto`). All coordination state lives in `status.conditions`. There is exactly **one** entry point into external ownership (EF) and exactly **one** exit point (ERR); every external system speaks the same protocol.
+
+EF creation always results in ERR creation by routing through the normal pipeline with `recommendedAction = CUSTOM` (per [ADR-036](036-custom-remediation-actions.md)). The EF reconciler does not create the ERR directly — that keeps a single, well-tested code path responsible for materialising release requests.
+
+## Implementation
+
+### Module layout
+
+The CRD types are defined in proto and generated via the existing `protoc-gen-crd` plugin (the same plugin that generates `HealthEventResource` today — see `data-models/protobufs/health_event.proto` line 121). The reconcilers live alongside the existing maintenance reconcilers in `janitor/`.
```
lalitadithya, line 35: I don't think this is strictly required. The reason why we have protos for the health api is because it goes over the UDS. In this case I think these are pure Kubernetes APIs that we can apply using go types / CRDs without the addition proto overhead

docs/designs/040-external-remediations.md
```diff
@@ -0,0 +1,735 @@
...
+
+With `FaultCleared=False` set, the EF's three conditions are all terminal. The webhook's "no active EF for this node" dedup check sees terminal-EFs as "not active" and allows a new EF to be created for the same node — this is the retry path. The external system creates a fresh EF (typically with the same `HealthEvent.id` or an incremented one); a new health event flows, a new ERR is produced (with a new release taint value), and the cycle restarts. The old failed EF/ERR pair can be deleted by the external system or an operator at any time (cascade-deletes via the EF ownerRef on the ERR); they do not block forward progress.
+
+During the brief window between the old release taint being removed and the new ERR's taint being applied, the node still carries fault-quarantine's taint — workloads remain off. The "released to external" signal is briefly absent, which is correct: for that window, NVSentinel formally owns the node, the external system has acknowledged failure, no one is actively doing anything.
```
lalitadithya, line 310: wouldn't this be a single patch operation, would there be a brief window?

docs/designs/040-external-remediations.md
```diff
@@ -0,0 +1,735 @@
...
+
+### ERR reconciler
+
+**Watches:** `ExternalRemediationRequest` (primary); `Node` (secondary, to detect taint drift).
+
+**Reconcile loop:**
+
+1. On first reconcile, call `SetInitialConditions`: write `NVSentinelOwnershipReleased=Unknown (Initializing)` and `ExternalRemediationComplete=Unknown (AwaitingExternalSystem)`.
+2. If `NVSentinelOwnershipReleased` is `Unknown`:
+   - Apply the configured release taint to `spec.nodeName` (idempotent).
```
lalitadithya, line 332: We will also have to tear down the health monitors, right? Else we will end up racing with the external remediation system. Also tearing down the health monitors will also mean that we don't have to do additional checks in other modules to work with the taint

docs/designs/040-external-remediations.md
```diff
@@ -0,0 +1,735 @@
...
+  .items[]
+  | select(.spec.taints[]?.key == "nvsentinel.nvidia.com/external-remediation")
+  | [.metadata.name,
+     (.spec.taints[] | select(.key=="nvsentinel.nvidia.com/external-remediation") | .value)]
+  | @tsv'
+```
+
+The trade-off vs. duplicating the ERR name to a Node label is explicit: filtering nodes via `kubectl get nodes -l …` is unavailable; operators use the jq query above. Given how rarely this enumeration is needed in practice, the cost of maintaining a parallel Node label was judged not worth it.
+
+### EF reconciler
```
lalitadithya, line 388: This would be in the CSP health monitor, I presume?

docs/designs/040-external-remediations.md
```diff
@@ -0,0 +1,735 @@
...
+
+**Watches:** `ExternalFault` (primary); `ExternalRemediationRequest` keyed by `ownerReferences[].uid` (secondary, to detect ERR creation).
+
+**Reconcile loop:**
+
+1. On first reconcile, call `SetInitialConditions`: write `FaultReported=Unknown (Initializing)`, `ExternalRemediationComplete=Unknown`, `FaultCleared=Unknown (FaultActive)`.
+2. **Observe-then-act for `FaultReported`:**
+   - List ERRs in the configured ERR namespace whose `ownerReferences` point to this EF (cheap via the ownerRef indexer).
```
lalitadithya, line 396: I'm not sure if we need to list the EERs here, can't we track if the event has already been remediated using just status conditions?

docs/designs/040-external-remediations.md
lalitadithya, line 1: this is a fairly long ADR, can we break this down into two parts (possibly PRs) one for ERR and one more for EF?

neerajnv:
some questions

1. can there be multiple ERR for the same entity, for example a node? 
2. what if there are two GPU and a NVLink on a node found unhealthy. will there be multiple ERR?
3. I read in the doc that there is no TTL or expectation from the duration of state transition after an entity is handed over to the external system. can we introduce a limit on the duration later on as we gain a better understanding of external remediation system?

docs/designs/040-external-remediations.md
jtschelling: Yes it got quite lengthy, I'll figure out a way to split it up today after going over your reviews. I kept attempting to shorten it but ended removing information that I thought was necessary for the whole thing to work

docs/designs/040-external-remediations.md
jtschelling: should the taint effect be `NoExecute`? that would evict on taint apply, i don't think the reconciler would have to do anything extra in that case. ill call it out better in the doc that the nvsentinel daemonsets need to be evicted from the node

lalitadithya: not really, the nvsentinel daemonsets will have to tolerate all the taints by default since users can configure any taint they like in the quarantine configuration, which is why we will need a label-based tear down so irrespective of what taints the users may apply the daemonsets continues to run while nvsentinel owns the node

jtschelling:
@neerajnv 

> some questions
> 
> 1. can there be multiple ERR for the same entity, for example a node?
> 2. what if there are two GPU and a NVLink on a node found unhealthy. will there be multiple ERR?
> 3. I read in the doc that there is no TTL or expectation from the duration of state transition after an entity is handed over to the external system. can we introduce a limit on the duration later on as we gain a better understanding of external remediation system?

1. No. Fault-remediation blocks this via its equivalence groups config. There is also blocking via the validation webhook if an external system attempts to make an additional EF that would generate an ERR for the node.
2. You're saying new faults being discovered while the node is already owned by the external system? In that case no, whichever one was created first will block the creation of the other. The first fault that ended up creating an ERR will also have caused the NVSentinel daemonsets to stop running, so there wouldn't be detection of the later fault. Lalit had an additional comment to this effect so I need to clarify in the document how the eviction of the NVS DS's works.
3. Yes, that statement was to make this change overall easier. We can introduce later as a configuration value somewhere in the system.

docs/designs/040-external-remediations.md
jtschelling: when i started working on the details here i see 3 options:

1. define the CRD in the proto defs
2. use `runtime.RawExtension` to dynamically allocate the health event object to the EF/ERR object spec
3. duplicate the health event schema in a go struct

I went with 1 mainly because it will be regenerated automatically on health event schema changes, and we get apiserver validation by default. With 2, we don't validate any fields on the EF/ERR objects with the apiserver, we have to do typing validation with an admission webhook. However with 2 health event schema updates don't matter. For 3, we get strict validation but also any health event schema changes need to propagate to the go structs, which I'm not a fan of for maintenance reasons. I think it would be easy to miss.

docs/designs/040-external-remediations.md
jtschelling: the EF reconciler? not sure what you mean here, can you expand?

docs/designs/040-external-remediations.md
jtschelling: should i enumerate these options in the doc and decide which one to go with there? definitely looking for opinions on this part specifically

docs/designs/040-external-remediations.md
lalitadithya: yeah, we already have a CSP health monitor which watches for CSP events, I'm thinking external faults is similar to a CSP notification

docs/designs/040-external-remediations.md
lalitadithya: don't we already have the health event as a gotype? Can't we use that?

jtschelling: yes, but some of the fields are problematic for controller-gen. in particular are RecommendedAction and GeneratedTimestamp, both would need some sort of handling by the janitor pkg before controller-gen would read them correctly. so you end up having to inspect health event schemas and mapping new changes back to the EF and ERR definitions on every change anyways. 

is there any particular reason you're against using protoc-gen-crd? fwiw i'd rather have all the CRDs generated via that (not that its in scope here)

docs/designs/040-external-remediations.md
jtschelling: i dont have a preference. the code changes for adding this in csp-health-monitor would be much more significant, and it also requires deployment manifest changes in the helm chart if we put it there. i'll add the option to the document and we can decide then.

docs/designs/040-external-remediations.md
jtschelling: i'm not sure what you mean, in order for the EF reconciler to find any ERR's that result from the EF we would have to index on that field, no? we're not tracking if the event was remediated, only that the fault generated by the EF resulted in an ERR being created

docs/designs/040-external-remediations.md
jtschelling: i'm going back and forth on this (evicting the various health-monitor pods). i'm not really understanding technically how we could ensure that the health-monitors are evicted from the node. we can't rely on taints since tolerations are user-configurable. there's also the issue that not all health-monitors are daemonsets, csp and k8s monitors are both deployments

> which is why we will need a label-based tear down

what do you mean here? something like gpu-operator has for its components? that would mean introducing required nodeAffinities on the ds manifests for the health-monitors. tmk this works in gpu-operator because the operator owns the component deployments, which we dont have in nvsentinel

docs/designs/040-external-remediations.md
jtschelling: no, because the external system has determined that it can no longer continue remediation, and also determined that it could not clear the fault. in this case we have to "finish" the ERR, which means untainting the node, and it also means that we cannot resolve the EF - because the external system could not remediate it. so in this interim period we have the node owned by nvsentinel (no taint) but we can't return it to schedulable.

docs/designs/040-external-remediations.md
```diff
@@ -0,0 +1,774 @@
...
+
+`fault-remediation` already uses `GetEffectiveActionName(he)` (per ADR-036) to resolve the action name for routing. To produce ERRs:
+
+- **TOML config** — declare the external-remediation action's `MaintenanceResource` with `apiGroup: "nvsentinel.nvidia.com"` and `kind: "ExternalRemediationRequest"`. No schema extension is needed; this reuses the same `apiGroup` + `kind` knobs that already route the reboot action to `janitor.dgxc.nvidia.com` / `RebootNode`.
+- **`fault-remediation/pkg/remediation/remediation.go`** — `CreateMaintenanceResource` already builds the maintenance CR from `apiGroup` + `kind` via the dynamic client. The branch for `ExternalRemediationRequest` mostly falls out of the existing template-render path; only the ERR-specific fields (ownerRef wiring, labels) need explicit handling.
+
+ERR construction details:
+
+- `metadata.name` is a deterministic hash of `(nodeName, healthEvent.id)` so repeated reconciles of the same event do not create duplicate ERRs. The name **also** becomes the value of the release taint applied by the ERR reconciler — so it MUST be a valid Kubernetes taint value (≤ 63 chars, alphanum + `.`, `-`, `_`).
```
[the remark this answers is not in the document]
jtschelling: hand-waving is fine here

docs/designs/040-external-remediations.md
jtschelling: i'm having difficulty coming up with a split here. i thought about doing a singular EF and another ERR adr but that didnt end up making much sense because most of EF is dependent on an ERR implementation existing. maybe an ERR adr on its own, then follow with an EF adr? i can also remove the webhook validation part and move that to its on adr. any other ideas?

maybe a "definitional" ADR with the spec on the APIs then the technical implementation in another?

lalitadithya: I think we can do just the ERR first, get that reviewed, merged and then we can do a EF ADR

docs/designs/040-external-remediations.md
```diff
@@ -0,0 +1,777 @@
...
+
+External breakfix systems — automated orchestrators driving CSP-side repair, long-running workflows, hardware swaps, or human operators performing manual maintenance — also have authority to perform destructive actions on the same nodes. Today these systems coordinate with NVSentinel ad-hoc, through side channels and operator knowledge.
+
+A node must be owned by **exactly one** system at any point in time. Without a formal protocol, NVSentinel and an external system can both decide to act on the same node and race — duplicating remediation work, fighting over taints, or interleaving operations that corrupt cluster state.
+
+The two crossing points needed are:
+
+1. **NVSentinel-detected fault that an external system must fix.** NVSentinel's triage maps a known failure mode to a remediation that NVSentinel cannot perform itself (e.g. CSP-side repair, long-running workflows). NVSentinel needs to signal "I no longer own this node" and stop attempting remediations.
```
lalitadithya, line 13: nit:

```suggestion
1. **NVSentinel-detected fault that an external system must fix.** NVSentinel's triage maps a known failure mode to a remediation that NVSentinel cannot perform itself (e.g. CSP-side repair, long-running workflows). NVSentinel needs to signal "I no longer own this node" and stop attempting remediations and health checking.
```

docs/designs/040-external-remediations.md
lalitadithya: > yes, but some of the fields are problematic for controller-gen. in particular are RecommendedAction and GeneratedTimestamp, both would need some sort of handling by the janitor pkg before controller-gen would read them correctly. so you end up having to inspect health event schemas and mapping new changes back to the EF and ERR definitions on every change anyways.

I see, okay

> is there any particular reason you're against using protoc-gen-crd?

I'm not against it; it is just that CRDs are typically not generated this way is all.

docs/designs/040-external-remediations.md
lalitadithya: > what do you mean here? something like gpu-operator has for its components?

yes, we can have a label something like `nvsentinel.nvidia.com/monitoring.active: true` much like the `gpu.present` label and then we can have the daemon sets key off of that. You don't have to do nodeAffinity, we can just add this as an additional label to the selector set. The health monitors that are not daemonsets will have to be handled separately where they ignore all nodes that are labeled with the above. 

>  tmk this works in gpu-operator because the operator owns the component deployments, which we dont have in nvsentinel

right, yes. We will have to document this pattern for health monitors that are owned by the community. I agree there is not really a way to enforce this with the label approach. 

The other option would be for the platform connector to automatically drop events from the health monitors based on node labels and then replay healthy events once the ownership is transferred back, but I feel that is going to get really complex.

docs/designs/040-external-remediations.md
lalitadithya: In the case of EF shouldn't the entity that said there was a fault also be the one confirms the remediation of the fault? So, I was thinking there would be a status condition where the external entity will patch the resolution when it completes the remediation and detects that the fault has truly been resolved.

lalitadithya:
superseeded by #1316

PR #1283, opened 2026-05-14
feat: add health event deduplication in platform connector

### LINKED ISSUE #1120: [Feature]: Deduplicate XID/SXID in syslog health monitor

XRFXLP:
## Summary

Add TTL-based health event deduplication in the platform connector, following ADR-039:
https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/039-health-event-deduplication.md

The dedup filter runs once inside the platform connector pipeline, after transformers and before fan-out to the enabled connectors. Duplicate events are suppressed before they reach either the Kubernetes connector or the store connector.

```mermaid
flowchart TD
    HM[Health Monitor] -->|HealthEvent over UDS/gRPC| RX

    subgraph PC[Platform Connector]
        RX[gRPC Receiver]
        TF[Transformers]
        DD{Dedup Filter}
        FANOUT[Fan-out to enabled connectors]
        K8S[K8s Connector]
        STORE[Store Connector]

        RX --> TF
        TF --> DD
        DD -->|new event or skipped check| FANOUT
        DD -.->|duplicate event suppressed| END(( ))
        FANOUT --> K8S
        FANOUT --> STORE
    end

    K8S -->|node conditions / annotations| API[Kubernetes API]
    STORE -->|persist health events| MONGO[(MongoDB)]
```

Dedup filter behavior:

```mermaid
flowchart TD
    IN[HealthEvent]
    SKIP{skipChecks<br/>contains checkName?}
    HEALTHY{isHealthy?}
    CLEAR[Clear matching<br/>unhealthy key]
    KEY[Build dedup key]
    SEEN{key seen within<br/>suppressionWindow?}
    MARK[Store key with<br/>firstSeen timestamp]
    KEEP[Allow event]
    DROP[Suppress duplicate]
    TICK[cleanupInterval<br/>ticker]
    EVICT[Evict expired keys]

    IN --> SKIP
    SKIP -->|yes| KEEP
    SKIP -->|no| HEALTHY
    HEALTHY -->|yes| CLEAR --> KEEP
    HEALTHY -->|no| KEY --> SEEN
    SEEN -->|yes| DROP
    SEEN -->|no| MARK --> KEEP
    TICK -.-> EVICT
    EVICT -.-> SEEN
```

## Implementation Notes

The dedup key is built from the stable parts of a health event: node name, check name, impacted entities, error codes, processing strategy, and health state. The message is intentionally not part of the key, since repeated syslog events often differ only by timestamp, PID, or message text while representing the same underlying fault.

The tracker is in-memory and TTL-based. `suppressionWindow` controls how long a repeated key is suppressed. `cleanupInterval` controls how often expired keys are removed from the tracker.

Healthy events are always allowed through. When a healthy event arrives, the filter also clears the matching unhealthy dedup key so a future unhealthy event for that same fault is not incorrectly suppressed.

`skipChecks` lets us avoid deduplicating checks where repeated events are meaningful to downstream components. In the default/full values, `SysLogsGPUFallenOff` and `SysLogsNICDriverError` are skipped so dedup remains focused on XID/SXID-style syslog spam.

One compatibility note: if `health-events-analyzer` is enabled in a full-stack configuration, dedup should be disabled for that environment unless the analyzer workflow has been explicitly validated with dedup enabled. The analyzer intentionally reasons over repeated unhealthy events, and global deduplication can hide the repeated signals it needs.

## Testing

Added an E2E test for platform connector deduplication.

The test sends 26 unhealthy syslog XID health events:

- 20 XID 95 events:
  - 10 events for the same `(node, check, entities, errorCode)` key with message variations.
  - 10 events across 5 other PCI/GPU entity pairs, each pair sent twice with entity order reversed on the duplicate.
- 6 XID 119 events:
  - 4 events for one PCI/GPU entity key with message variations.
  - 2 events for another PCI/GPU entity key with entity order reversed on the duplicate.

After deduplication, fault-quarantine should show 8 unique active faults in `quarantineHealthEvent`:

- 6 XID 95 entries
- 2 XID 119 entries

That means 18 duplicate events are suppressed:

- 14 suppressed XID 95 events
- 4 suppressed XID 119 events

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added health-event deduplication via an automatic **Deduplicator** pipeline stage, configurable with suppression window, cleanup interval, and a per-check allowlist.
  * Introduced the **`STORE_AND_ANALYSE`** processing strategy across event handling and related CRD schemas.
* **Bug Fixes**
  * Improved pipeline shutdown and updated processing so `STORE_AND_ANALYSE` is correctly included/excluded where appropriate.
* **Documentation**
  * Updated platform-connector pipeline and deduplication docs with configuration and dedup-key semantics.
* **Tests**
  * Added unit and end-to-end coverage for deduplication behavior and pipeline mutation/batching.


XRFXLP:
this cannot work

distros/kubernetes/nvsentinel/values-full.yaml
```diff
@@ -292,6 +292,16 @@ platformConnector:
     # Burst: Maximum burst size above QPS
     burst: 10
 
+  # Health event deduplication filter
+  # Suppresses repeated events with the same node, check, impacted entities, error code, and health state.
+  dedup:
+    enabled: true
```
lalitadithya: can we start this off as default disabled?

lalitadithya (commented):
can we clearly document somewhere which of the HEA rules would this impact and what the impact would be?

lalitadithya:
what are you thoughts on having the duplicate events still be persisted, but marked as skipped? would that help the HEA case?

lalitadithya:
@XRFXLP , can you also look at https://github.com/NVIDIA/NVSentinel/issues/1447 to see if this would help with the power throllting case?

XRFXLP:
> what are you thoughts on having the duplicate events still be persisted, but marked as skipped? would that help the HEA case?

This can work, if we go by this then we don't really need filter pipeline and suppression would just be another transformer, so simpler implementation wise.

> can you also look at https://github.com/NVIDIA/NVSentinel/issues/1447 to see if this would help with the power throllting case?

Problem with this is that is an toggle behaviour and with the current implementation healthy version of an event kicks out the stored unhealthy counter part => duplication won't work and if we go with the STORE_ONLY mechanism then it would not solve the storage problem of flappy events.

lalitadithya:
> This can work, if we go by this then we don't really need filter pipeline and suppression would just be another transformer, so simpler implementation wise.

Won't we still need all the logic you have for the dedup? The outcome is changing, instead of drop it is mutated with an attribute to skip. This would allow the HEA to function. Or am I missing something?

XRFXLP:
> Won't we still need all the logic you have for the dedup? The outcome is changing, instead of drop it is mutated with an attribute to skip. This would allow the HEA to function. Or am I missing something?

yeah, we would still need all the logic + It would be just simpler, implementation-wise.

platform-connectors/pkg/transformers/dedup/transformer.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil
+	}
+
+	if d.tracker.checkAndMark(event) {
+		dedupStoreOnlyCounter.WithLabelValues(
+			event.GetCheckName(),
+			event.GetNodeName(),
+			errCodeLabel(event),
+		).Inc()
+		event.ProcessingStrategy = pb.ProcessingStrategy_STORE_ONLY
```
lalitadithya: should we also add an additional metadata to the event so that we can distinguish between events that originated with `ProcessingStrategy_STORE_ONLY` and ones that we marked as `ProcessingStrategy_STORE_ONLY`. I think this will also help with the HEA since now we can only include store only events that have this metadata and not store only which would preserve the existing behaviour as well.

XRFXLP: how about adding another value in processing strategy? One for store only, another for analysis and third for remediation?

lalitadithya: that works as well, we can maybe call it `DUPLICATE_ANALYSIS_ONLY` or something similar?

XRFXLP: using `STORE_AND_ANALYSE`.

docs/designs/025-processing-strategy-for-health-checks.md
```diff
@@ -37,7 +37,7 @@ When a health monitor wants observability-only behavior, it publishes health eve
 - Node quarantine (fault-quarantine skips - no taint, labels, or annotations)
 - Node draining (node-drainer module won't receive them - filtered by pipeline)
 - Remediation CR creation (fault-remediation module won't receive them - filtered by pipeline)
-- Pattern analysis (health-events-analyzer ignores incoming `processingStrategy=STORE_ONLY` events for pattern detection and does not consider them while running the query)
+- Direct remediation side effects from the source event. Health-events-analyzer may still use `STORE_ONLY` source events as analysis input; the analyzer output event's own `processingStrategy` controls whether an inferred pattern can modify cluster state.
```
lalitadithya: this change is no longer accurate right? we now have a new enum for this, can you please check?

docs/designs/025-processing-strategy-for-health-checks.md
```diff
@@ -421,7 +419,7 @@ func ToCloudEvent(event *pb.HealthEvent, metadata map[string]string) (*CloudEven
 
 We need new methods in mongodb and postgres pipeline builder:
 1. `BuildProcessableHealthEventInsertsPipeline` which filters `processingStrategy=EXECUTE_REMEDIATION` inserted events for fault-quarantine module
-2. `BuildProcessableNonFatalUnhealthyInsertsPipeline` which filters `processingStrategy=EXECUTE_REMEDIATION`, non-fatal and unhealthy inserted events for health-events-analyzer module.
+2. `BuildProcessableNonFatalUnhealthyInsertsPipeline` which filters non-fatal and unhealthy inserted events for health-events-analyzer module. It intentionally includes `STORE_ONLY` source events as analyzer input; the analyzer output event's own processing strategy controls downstream side effects.
```
lalitadithya: need to update here as well

docs/designs/039-health-event-deduplication.md
```diff
@@ -14,48 +14,50 @@ These reach fault-quarantine and node-drainer as distinct events: redundant drai
...
 flowchart TD
     Recv["HealthEventOccurredV1"] --> Pipe["Pipeline.Process(event)"]
     Pipe --> MA["MetadataAugmentor<br/>(transformer)"]
     MA --> OT["OverrideTransformer<br/>(transformer)"]
-    OT --> Dedup["DeduplicationFilter<br/>(filter)"]
+    OT --> Dedup["Deduplicator<br/>(transformer)"]
     Dedup --> KeyExtract["key = (NodeName,<br/>CheckName,<br/>canonical(EntitiesImpacted),<br/>canonical(ErrorCode),<br/>IsHealthy)"]
     KeyExtract --> Seen{"key in seen set<br/>and within TTL?"}
-    Seen -->|Yes| Drop["Drop event<br/>(log Info, increment counter)"]
+    Seen -->|Yes| StoreOnly["Mark STORE_ONLY<br/>(log Info, increment counter)"]
```
lalitadithya: need to update here as well

tests/platform_connector_dedup_test.go
lalitadithya, line 1: Can we add a small test showing that store_and_analyze doesn't result in cordon/remediation as a defensive test?

XRFXLP: added in fault_quarantine_test.go

docs/designs/039-health-event-deduplication.md
XRFXLP: done

docs/designs/025-processing-strategy-for-health-checks.md
XRFXLP: done

docs/designs/025-processing-strategy-for-health-checks.md
XRFXLP: should be fixed now

lalitadithya (approved):

PR #1304, opened 2026-05-18
fix: fix node condition cleanup for truncated entity messages

### LINKED ISSUE #1299: [Bug]: Platform connector doesn't remove node condition after healthy event due to truncation of node condition message
### Prerequisites

### Bug Description

`platform-connectors` can fail to clear a node condition after receiving healthy health events when the condition message has been compacted/truncated.

The affected flow is for healthy events with a non-empty `entitiesImpacted` list. The connector tries to remove only matching entity messages from the existing node condition, but matching depends on finding the full `entityType:entityValue` token inside the condition message. If the node condition message was truncated earlier, the entity token may no longer be present in full, so the stale message remains and the condition stays `True`.

### Component

Core Service

### Steps to Reproduce

1. Send enough unhealthy event to platform connector like this to truncate the entities impacted like:
```
PodStuckAfterDeletion True ... PodStuckAfterDeletionIsNotHealthy
ErrorCode:POD_STUCK_AFTER_DELETION v1/Pod:prod/61f345d08c9a432a-134... Recommended Action=RESTART_VM
```
2. Send healthy event for all of the entities impacted
3. Node condition will not be be removed because the healthy event entities impacted doesn't  match to any of the truncated entities impacted

### Environment

- NVSentinel version: v1.3.0
- Kubernetes version:
- Deployment method:

### Logs/Output

A node condition remained unhealthy:
```text
PodStuckAfterDeletion True ... PodStuckAfterDeletionIsNotHealthy
ErrorCode:POD_STUCK_AFTER_DELETION v1/Pod:prod/61f345d08c9a432a-134... Recommended Action=RESTART_VM
```

Even though:

```
{
  "checkName": "PodStuckAfterDeletion",
  "isFatal": true,
  "isHealthy": true,
  "errorCode": ["POD_STUCK_AFTER_DELETION"],
  "entitiesImpacted": [
    {
      "entityType": "v1/Pod",
      "entityValue": "prod/61f345d08c9a432a-134a464884734f90"
    }
  ],
  "nodeName": "node-name-0291",
  "processingStrategy": 1
...
}
```

KaivalyaMDabhadkar:
## Problem

When multiple health events exceed `MaxNodeConditionMessageLength` (1024 bytes), the platform connector compacts messages in two stages:

- **Tier 1 (per-message compaction):** Each individual message's content before `Recommended Action=` is cut to `CompactedHealthEventMsgLen` (72 bytes), appending `...` as a truncation marker.
- **Tier 2 (byte-level hard limit):** After all messages are compacted, if the combined `;`-joined string still exceeds `MaxNodeConditionMessageLength` (1024 bytes), the last message entry is cut at an arbitrary byte boundary to fit the limit.

For entity types with long values like `v1/Pod:prod/61f345d08c9a432a-134a464884734f90`, the Tier 1 byte cut lands inside the entity value:

```
ErrorCode:POD_STUCK_AFTER_DELETION v1/Pod:prod/61f345d08c9a432a-134... Recommended Action=RESTART_VM
```

When a healthy event arrives to clear that entity, the cleanup code searches for the full token `v1/Pod:prod/61f345d08c9a432a-134a464884734f90 ` via exact substring match. Since only the truncated form exists in the stored condition, the match fails and the condition remains stuck as unhealthy indefinitely.

### Forward fix -- identity-aware compaction

`compactMessageField` is rewritten to parse the structured identity prefix (ErrorCode + entity tokens) and only truncate the trailing diagnostic free-text. Identity tokens are never byte-truncated.

**Before (old behavior):**

```
Input:    "ErrorCode:POD_STUCK_AFTER_DELETION v1/Pod:prod/61f345d08c9a432a-134a464884734f90 Pod stuck after deletion timeout exceeded Recommended Action=RESTART_VM"
                                                                                          ^--- blind cut at byte 72
Output:   "ErrorCode:POD_STUCK_AFTER_DELETION v1/Pod:prod/61f345d08c9a432a-134... Recommended Action=RESTART_VM"
```

Entity name is damaged. Future healthy events cannot clear this condition.

**After (new behavior):**

```
Input:    "ErrorCode:POD_STUCK_AFTER_DELETION v1/Pod:prod/61f345d08c9a432a-134a464884734f90 Pod stuck after deletion timeout exceeded Recommended Action=RESTART_VM"

Identity: "ErrorCode:POD_STUCK_AFTER_DELETION v1/Pod:prod/61f345d08c9a432a-134a464884734f90"  (preserved in full)
Diagnostic: "Pod stuck after deletion timeout exceeded"  (truncated)

Output:   "ErrorCode:POD_STUCK_AFTER_DELETION v1/Pod:prod/61f345d08c9a432a-134a464884734f90 ... Recommended Action=RESTART_VM"
```

Entity name intact. Only the human-readable diagnostic is removed. `MaxNodeConditionMessageLength` is still enforced by Tier 2.

### Backward fix -- truncation-tolerant entity matching

Nodes that already have truncated conditions stored (from before this fix, or from Tier 2 byte-level truncation) need to be clearable. The new `entityMatchesMessage` helper falls back to prefix matching when exact match fails, but only when there is evidence of truncation.

**Example -- Tier 1 truncation (entity ends with `...`):**

```
Stored condition: "ErrorCode:POD_STUCK v1/Pod:prod/61f345d08c9a432a-134... Recommended Action=RESTART_VM"
Healthy event:    entity = v1/Pod:prod/61f345d08c9a432a-134a464884734f90

Step 1: Exact match for "v1/Pod:prod/61f345d08c9a432a-134a464884734f90 " → NOT FOUND
Step 2: Tokenize prefix, find "v1/Pod:prod/61f345d08c9a432a-134..."
        - Ends with "..." → evidence of truncation
        - Strip "...", candidate = "v1/Pod:prod/61f345d08c9a432a-134"
        - Full entity starts with candidate? YES → MATCH
Result: Condition is cleared.
```

**Example -- Tier 2 truncation (entity is last token, no `Recommended Action=`):**

```
Stored condition: "ErrorCode:POD_STUCK v1/Pod:prod/61f345d08c9a432a-134"
Healthy event:    entity = v1/Pod:prod/61f345d08c9a432a-134a464884734f90

Step 1: Exact match for "v1/Pod:prod/61f345d08c9a432a-134a464884734f90 " → NOT FOUND
Step 2: Tokenize prefix, find "v1/Pod:prod/61f345d08c9a432a-134"
        - No "Recommended Action=" in message AND token is last → evidence of Tier 2 truncation
        - candidate = "v1/Pod:prod/61f345d08c9a432a-134"
        - Full entity starts with candidate? YES → MATCH
Result: Condition is cleared.
```

**Example -- false positive prevention (complete shorter entity):**

```
Stored condition: "ErrorCode:48 v1/Pod:prod/app-a error Recommended Action=DRAIN"
Healthy event:    entity = v1/Pod:prod/app-abcdef

Step 1: Exact match for "v1/Pod:prod/app-abcdef " → NOT FOUND
Step 2: Tokenize prefix, find "v1/Pod:prod/app-a"
        - Does NOT end with "..."
        - Message HAS "Recommended Action=" so last-token rule does not apply
        - isTruncated = false → skip prefix matching
Result: No match. Condition for app-a is NOT incorrectly cleared.
```

## Testing

- **Unit tests (`make test` in `platform-connectors/`):** 245 tests pass including all envtest/kubebuilder integration tests. New test functions: `TestIsStructuredEntityToken`, `TestEntityMatchesMessage`, and truncation cases added to `TestRemoveImpactedEntitiesMessagesScoped` and `TestCompactMessageField`.
- **Lint (`golangci-lint`):** Passes for both `platform-connectors/` and `tests/`.
- **Tilt/e2e regression test:** Seeds a pre-truncated node condition directly via the Kubernetes status subresource, sends a scoped healthy event, and verifies clearance. Verified on a Kind cluster:
  - With the fix stashed (old code running): condition stays stuck as `ConditionTrue`, test times out and fails.
  - With the fix unstashed (new code running): condition clears to `ConditionFalse` with "No Health Failures" in ~5 seconds, test passes.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

### DIFF
--- a/tests/platform_connector_node_condition_test.go
+++ b/tests/platform_connector_node_condition_test.go
@@ -1,0 +1,129 @@
+//go:build amd64_group
+// +build amd64_group
+
+// Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"context"
+	"testing"
+
+	"tests/helpers"
+
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+
+	pb "github.com/nvidia/nvsentinel/data-models/pkg/protos"
+)
+
+const (
+	truncatedEntityRegressionCheckName = "PodStuckAfterDeletionRegression"
+	truncatedEntityRegressionErrorCode = "POD_STUCK_AFTER_DELETION"
+	truncatedEntityRegressionType      = "v1/Pod"
+	truncatedEntityRegressionValue     = "prod/61f345d08c9a432a-134a464884734f90"
+)
+
+// TestPlatformConnectorClearsCompactedLongEntityCondition verifies that platform-connector
+// can clear node conditions written by the old compaction behavior. The test seeds a
+// node condition directly with a truncated entity token:
+//
+//	v1/Pod:prod/61f345d08c9a432a-134...
+//
+// Then it sends a healthy event with the full entity value. Without truncation-aware
+// matching, the connector searches for the full token, does not find it in the
+// stored condition message, and leaves the condition stuck as unhealthy.
+func TestPlatformConnectorClearsCompactedLongEntityCondition(t *testing.T) {
+	feature := features.New("Platform Connector - Clears Truncated Long Entity Condition").
+		WithLabel("suite", "platform-connector").
+		WithLabel("component", "node-condition")
+
+	var nodeName string
+
+	feature.Setup(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		client, err := c.NewClient()
+		require.NoError(t, err, "failed to create kubernetes client")
+
+		nodeName = helpers.AcquireNodeFromPool(ctx, t, client, helpers.DefaultExpiry)
+		sendTruncatedEntityClearEvent(ctx, t, nodeName)
+
+		return ctx
+	})
+
+	feature.Assess("Healthy event clears a pre-existing condition whose entity was truncated", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		client, err := c.NewClient()
+		require.NoError(t, err, "failed to create kubernetes client")
+
+		helpers.UpsertNodeCondition(ctx, t, client, nodeName, v1.NodeCondition{
+			Type:    v1.NodeConditionType(truncatedEntityRegressionCheckName),
+			Status:  v1.ConditionTrue,
+			Reason:  truncatedEntityRegressionCheckName + "IsNotHealthy",
+			Message: "ErrorCode:POD_STUCK_AFTER_DELETION v1/Pod:prod/61f345d08c9a432a-134... Recommended Action=NONE;",
+		})
+
+		healthyEvent := helpers.NewHealthEvent(nodeName).
+			WithAgent("platform-connector-regression").
+			WithComponentClass("Pod").
+			WithCheckName(truncatedEntityRegressionCheckName).
+			WithFatal(false).
+			WithHealthy(true).
+			WithErrorCode(truncatedEntityRegressionErrorCode).
+			WithEntitiesImpacted([]helpers.EntityImpacted{{
+				EntityType:  truncatedEntityRegressionType,
+				EntityValue: truncatedEntityRegressionValue,
+			}}).
+			WithMessage("pod recovered").
+			WithRecommendedAction(int(pb.RecommendedAction_NONE))
+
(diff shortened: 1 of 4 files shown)


tests/platform_connector_node_condition_test.go
```diff
@@ -0,0 +1,129 @@
...
+		sendTruncatedEntityClearEvent(ctx, t, nodeName)
+
+		return ctx
+	})
+
+	feature.Assess("Healthy event clears a pre-existing condition whose entity was truncated", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		client, err := c.NewClient()
+		require.NoError(t, err, "failed to create kubernetes client")
+
+		helpers.UpsertNodeCondition(ctx, t, client, nodeName, v1.NodeCondition{
```
lalitadithya, line 71: Can we instead send the health events via `helpers.NewHealthEvent` like the other tests?

KaivalyaMDabhadkar: So actually we need to insert the node condition directly through k8s api because we need to simulate the old broken truncated state that the old code would have produced. If we send a health event through the normal path, the current platform connector (which already has the fix) would produce the correct non-truncated condition, and the test would pass trivially

lalitadithya: ah, can we add that as a comment?

lalitadithya: nvm, it's there on the top

XRFXLP (approved):

PR #1310, opened 2026-05-18
feat: add drainGPUPods configuration flag to restrict pod eviction to GPU workloads

### LINKED ISSUE #1264: [Feature]: Add option to only drain GPU pods in node-drainer
### Prerequisites

### Feature Summary

Add a Helm variable that allows the node-drainer to only drain pods in the selected namespaces which are requesting GPUs. This will prevent the node-drainer from being blocked on CPU workloads if there is a non-daemonset CPU pod running in one of the specified userNamespaces needing drained.

### Problem/Use Case

Pod selection for draining by the node-drainer is currently namespaced-based where an operator can configure systemNamespaces where pods in these namespaces are skipped for draining or userNamespaces where pods in these namespaces must be drained. If a pod is in a namespace requiring draining but belongs to a daemonset, is in a terminal state, is not ready, or is stuck terminating, we will ignore that pod ([ref](https://github.com/NVIDIA/NVSentinel/blob/main/node-drainer/pkg/informers/informers.go#L326)).

While namespace selection support regexes, an operator may want to target all pods with a specific property regardless of the namespace (in addition to the hard-coded properties referenced above). A useful feature would be to allow an operator to specify a field selector that allows only draining all pods with a specific property in any namespace (or in conjunction with the namespaces they're already selecting). If this too general of a feature, a setting that would require only pods requesting GPUs to be drained (or potentially pods from statefulsets) would satisfy most use-cases.

### Proposed Solution

Add a Helm variable called drainGPUPods which defaults to false to allow operators to restrict their draining to only pods with GPU requests.

### Component

Node drainer

coderuhaan2004:
## Summary

Implements Issue #1264 by adding a new `drainGPUPods` Helm configuration flag that allows the node-drainer to selectively drain only pods requesting GPU resources. This prevents the node-drainer from being blocked on CPU-only workloads during GPU fault remediation, enabling faster recovery while preserving availability of critical infrastructure services.

When `drainGPUPods: true`, the node-drainer filters pod eviction to target only workloads with GPU resource requests (nvidia.com/gpu or nvidia.com/pgpu), leaving CPU-only pods (logging agents, monitoring sidecars) running on the node.

### Changes Made

- **Helm Configuration**: Added `drainGPUPods: false` to node-drainer values with comprehensive documentation
- **Configuration Layer**: Added `DrainGPUPods bool` field to TomlConfig struct with proper TOML tag
- **Filtering Logic**: Implemented `filterPodsWithGPURequests()` in informers that:
  - Detects GPU resource requests in pod containers and init containers
  - Checks for both `nvidia.com/gpu` and `nvidia.com/pgpu` resource names
- **E2E Tests**: 
  - Added E2E testing within reconciler_integration_test.go for GPU pod filtering
- **Documentation**: Updated node-drainer module documentation with feature explanation

## Type of Change
- [x] ✨ New feature
- [ ] 🐛 Bug fix
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Fault Management
- [x] Documentation/CI
- [ ] Core Services
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: Node Drainer

## Testing
- [x] Tests pass locally
- [x] No breaking changes

## Checklist
- [x] Self-review completed
- [x] Documentation updated
- [x] Ready for review

## Related Issue
Closes #1264

## Summary by CodeRabbit

* **New Features**
  * Added a `drainGPUPods` option to enable GPU-only pod draining during node remediation; when enabled, only GPU-requesting pods are eligible for eviction. Defaults to `false`.

* **Documentation**
  * Updated docs to describe GPU-only draining behavior, eligibility rules, and differences vs. full-node drain scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NVSentinel/pull/1310?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


node-drainer/pkg/informers/informers.go
```diff
@@ -189,6 +191,8 @@ func (i *Informers) FindEvictablePodsInNamespaceAndNode(namespace, nodeName stri
 
 	pods = i.filterEvictablePods(pods)
 
+	pods = i.filterPodsWithGPURequests(pods)
+
```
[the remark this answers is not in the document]
coderuhaan2004: Resolved.

XRFXLP:
Should we use the same pattern that we've in `filterPodsUsingEntity`?

```
		deviceAnnotationJSON, podHasDeviceAnnotation := pod.Annotations[model.PodDeviceAnnotationName]
		// While we can't detect a stale annotation for devices, we can detect pods that are requesting GPUs that do
		// not have the device annotation at all which would indicate an issue with the metadata-collector.
		isPodRequestingDevices := areContainersRequestingDevice(pod.Spec.Containers, resourceNames) ||
			areContainersRequestingDevice(pod.Spec.InitContainers, resourceNames)

		if isPodRequestingDevices && !podHasDeviceAnnotation {
			return nil, fmt.Errorf("pod %s is requesting devices but is missing device annotation", pod.Name)
		}
```

Relying on metadata collector as a single source of truth would be more general, like for DRA based allocations.

coderuhaan2004:
> Should we use the same pattern that we've in `filterPodsUsingEntity`?
> 
> ```
> 		deviceAnnotationJSON, podHasDeviceAnnotation := pod.Annotations[model.PodDeviceAnnotationName]
> 		// While we can't detect a stale annotation for devices, we can detect pods that are requesting GPUs that do
> 		// not have the device annotation at all which would indicate an issue with the metadata-collector.
> 		isPodRequestingDevices := areContainersRequestingDevice(pod.Spec.Containers, resourceNames) ||
> 			areContainersRequestingDevice(pod.Spec.InitContainers, resourceNames)
> 
> 		if isPodRequestingDevices && !podHasDeviceAnnotation {
> 			return nil, fmt.Errorf("pod %s is requesting devices but is missing device annotation", pod.Name)
> 		}
> ```
> 
> Relying on metadata collector as a single source of truth would be more general, like for DRA based allocations.

Yes, you're right. Relying on resource limits directly isn't flexible enough. I have updated the implementation to use device annotations as the single source of truth, which the metadata-collector adds to pods that are assigned GPU 
devices.

docs/node-drainer.md
```diff
@@ -108,4 +110,7 @@ Multiple timeout mechanisms prevent stuck drains:
 Automatically resumes drain operations after restarts - queries datastore for in-progress drains and continues from where it left off.
 
 ### Partial Drain Functionality
-For GPU faults that can be remediated with a GPU reset, the Node Drainer will only drain pods which are leveraging the unhealthy GPU. For GPU faults that require a node reboot, all pods on the given node in the configured namespaces will be drained.
\ No newline at end of file
+For GPU faults that can be remediated with a GPU reset, the Node Drainer will only drain pods which are leveraging the unhealthy GPU. For GPU faults that require a node reboot, all pods on the given node in the configured namespaces will be drained.
+
+### GPU-Only Draining
+When `drainGPUPods: true` is set, the Node Drainer filters pod eviction to only target workloads that request GPU resources. The feature detects GPU resources by scanning pod containers and init containers for `nvidia.com/gpu` or `nvidia.com/pgpu` in resource limits. Pods without GPU requests are not evicted, preserving availability of critical infrastructure services like logging agents and monitoring sidecars during GPU maintenance. Default is `false`.
```
XRFXLP: > The feature detects GPU resources by scanning pod containers and init containers for `nvidia.com/gpu` or `nvidia.com/pgpu` in resource limits

Can we update this? Or better include the details in the docs/configuration/node-drainer.md?

docs/node-drainer.md
```diff
@@ -49,6 +49,7 @@ node-drainer:
   systemNamespaces: "^(nvsentinel|kube-system|gpu-operator)$"  # Namespaces to skip
   deleteAfterTimeoutMinutes: 60      # Force delete after this timeout
   notReadyTimeoutMinutes: 5          # Timeout for stuck pods
+  drainGPUPods: false                # Only drain GPU workloads
```
XRFXLP: Can we add this field here as well: https://github.com/NVIDIA/NVSentinel/blob/main/docs/configuration/node-drainer.md ?

node-drainer/pkg/informers/informers_test.go
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
XRFXLP: Can we add tests in `reconciler_integration_test.go` instead of here? As that looks like the existing pattern we've been following and that would increase the overall E2E-iness of the tests if the existing pattern of tests in that file is followed.

XRFXLP (changes requested):

node-drainer/pkg/initializer/init.go
```diff
@@ -98,7 +98,9 @@ func InitializeAll(ctx context.Context, params InitializationParams) (*Component
 		return nil, fmt.Errorf("failed to initialize dynamic client and mapper: %w", err)
 	}
 
-	informersInstance, err := initializeInformers(clientSet, &configs.tomlCfg.NotReadyTimeoutMinutes, params.DryRun)
+	informersInstance, err := initializeInformers(
+		clientSet, &configs.tomlCfg.NotReadyTimeoutMinutes, &configs.tomlCfg.DrainGPUPods, params.DryRun,
```
natherz97: Nit: we're reading the value as a primitive and we don't need to differentiate not-set from true/false, can you pass a non-ptr value to simplify the check we make before calling filterPodsWithGPURequests?

node-drainer/pkg/informers/informers_test.go
natherz97: +1 there's some tests for the partial draining functionality under TestReconciler_ProcessEvent that can be adapted to cover the new filterPodsWithGPURequests functionality.

node-drainer/pkg/informers/informers.go
```diff
@@ -189,6 +191,10 @@ func (i *Informers) FindEvictablePodsInNamespaceAndNode(namespace, nodeName stri
 
 	pods = i.filterEvictablePods(pods)
 
+	if i.drainGPUPods != nil && *i.drainGPUPods {
+		pods = i.filterPodsWithGPURequests(pods)
```
natherz97, line 195: This looks reasonable as long as we only support partial drains targeting pods using GPU devices. I am realizing that a partial drain should likely ignore the namespace filter passed by the user completely since we have to capture all GPU workload pods regardless of whether the user specifies we should drain in that namespace (or we should be explicit that running GPU pods in other namespaces will cause GPU resets to fail). This wasn't required when we only supported reboots since the remediation action would still succeed if we didn't drain all GPU pods. I wouldn't worry about that in this PR but I will open another issue for tracking.

docs/node-drainer.md
coderuhaan2004: Updated docs and shifted the details to docs/configuration/node-drainer.md.

docs/node-drainer.md
coderuhaan2004: Done.

node-drainer/pkg/informers/informers_test.go
coderuhaan2004: Sure, added test in `TestReconciler_ProcessEvent`

node-drainer/pkg/initializer/init.go
coderuhaan2004: Done.

XRFXLP:
LGTM!

XRFXLP (approved):

natherz97 (approved):

PR #1316, opened 2026-05-20
docs: add ADR-040 for ExternalRemediationRequest (ERR)

jtschelling:
## Summary

Introduces a new CRD, **`ExternalRemediationRequest` (ERR)**, in the existing `nvsentinel.nvidia.com/v1alpha1` API group, plus the reconciler that drives it and the `fault-remediation` integration that creates it. ERR is the formal mechanism for NVSentinel to release a node to an external party — CSP-side support workflows, RMA flows, vendor maintenance, individual operators — when the appropriate remediation lives outside NVSentinel's control plane.

Today this coordination is ad-hoc (operators manually uncordon, fight over taints, use side channels). ERR makes it in-band: `fault-remediation` produces an ERR when triage maps a fault to a `CUSTOM` action configured to produce one, the ERR reconciler applies a release taint to the target node, the external system reads the ERR and performs its remediation, then writes a terminal `ExternalRemediationComplete` condition back. The reconciler removes the taint, and existing health-monitors and `fault-quarantine` take over from there.

## What the ADR covers

- The `ExternalRemediationRequest` CRD, defined in proto (`data-models/protobufs/external_remediation.proto`) and generated via the existing `protoc-gen-crd` plugin — same automation already used for `HealthEventResource`. The `HealthEvent` spec is reused by reference, so HealthEvent schema changes propagate to the ERR CRD automatically.
- Condition state machine: `NVSentinelOwnershipReleased` and `ExternalRemediationComplete`, with `Unknown`/`True`/`False` terminal semantics matching `RebootNode`/`GPUReset`/`TerminateNode`.
- The ERR reconciler — taint apply/remove, idempotency, the single-active-ERR-per-node invariant (which falls out of `fault-remediation`'s existing equivalence-group + `latestFaultRemediationState` skip logic, no bespoke dedup code required).
- The release-taint mechanism: the ERR's metadata name is encoded in the taint value, giving operators a direct discovery path from a tainted node (`kubectl describe node …` → see the taint value → `kubectl get err <name>`). No new Node labels or annotations.
- `fault-remediation` integration: TOML config pointing `apiGroup` + `kind` at the new CRD, a `StatusChecker` for `ExternalRemediationRequest`, and ERR construction details.
- A Mermaid sequence diagram for the end-to-end flow.
- RBAC for the ERR reconciler and for external systems patching ERR status conditions.
- Observability — metrics, events, tracing, plus the controller-runtime freebies.
- Failure modes (external system never acknowledges, external system reports failure, node deleted mid-flight, multiple distinct faults on one node, etc.) and operations notes.

## What's intentionally not in scope

- **Mechanisms for external systems to inject faults into NVSentinel.** This ADR is the exit door only. A complementary entry-door surface — letting external systems declare faults directly to NVSentinel — is captured as a non-goal here and may be the subject of a future ADR. It will produce `CUSTOM`-action health events that flow through the standard pipeline, which is exactly what this ERR design already responds to; no changes to ERR are required to support it.
- **New remediation actions.** Builds on the existing `CUSTOM` action from ADR-036.
- **Changes to in-cluster remediation CRs** (`GPUReset`, `RebootNode`, `TerminateNode`) — those continue to be created directly by whoever needs them, including external systems if their workflow involves in-cluster operations.
- **Standalone-ERR garbage collection** — deferred to a follow-up if it becomes operationally necessary.

## What's not implemented yet

This PR is **docs only**. Implementation lands in follow-up PRs tracking the proto/CRD generation, the reconciler, the fault-remediation integration, and the cross-component release-taint guard in `fault-quarantine` and `node-drainer`.

Refs: #1276

## Summary by CodeRabbit

* **Documentation**
  * Added ADR-040 defining the External Remediation Request (ERR) API as the in-band handoff for external remediation systems.
  * Documents request lifecycle and reconciliation behavior, taint-based ownership transfer and release semantics, single-active-request constraints per node, operator discovery, required RBAC for both sides, and observability expectations (metrics, events, and trace/span propagation).

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NVSentinel/pull/1316?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,423 @@
...
+
+**Idempotency:** taint application uses a server-side patch. The ERR reconciler verifies both key AND value match its own ERR name before treating the taint as "its own" — protecting against drift from a previous, no-longer-existing ERR.
+
+**Node deletion:** if `spec.nodeName` no longer exists at reconcile time, the reconciler logs and treats the taint operation as complete. The ERR object remains until the external system writes a terminal value to `ExternalRemediationComplete`; if the external system never acknowledges, the ERR persists with `ExternalRemediationComplete=Unknown` (visible via the open-ERR gauge — see Observability).
+
```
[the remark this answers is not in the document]
jtschelling: i think you're wrong here, we still need to send the notification to the external system that nvsentinel doesnt own the node (ofc it doesnt, the node doesnt exist anymore)

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,423 @@
...
+When a field is added to `HealthEvent` in `health_event.proto`, the ERR CRD schema updates automatically on the next `make generate`; no Go struct edits required. This is the same automation already used for `HealthEventResource`.
+
+### API
+
+Defined in `data-models/protobufs/external_remediation.proto`:
+
+```protobuf
+import "google/protobuf/timestamp.proto";
+import "health_event.proto";
+import "github.com/yandex/protoc-gen-crd/library/go/k8s/protoc_gen_crd/proto/crd.proto";
```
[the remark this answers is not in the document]
jtschelling: i ran an experiment with claude and you seem to be wrong. output from claude:

▎ I ran the generator end-to-end to verify. Using the pinned protoc-gen-crd v1.1.0 from .versions.yaml, with a proto file matching the ADR's design (HealthEvent spec = 1;), the generated CRD has the HealthEvent schema fully
  ▎ inlined under spec — every scalar, enum, array, map, and nested message type expanded recursively as inline properties. The full output is 151 lines and grep -c x-kubernetes-preserve-unknown-fields returns 0.
  ▎
  ▎ The K8s "no $ref" constraint you cited is real, but the generator handles it by inline expansion at generation time, not by marking fields opaque. Marking opaque is the fallback for recursive (self-referential) types;
  ▎ HealthEvent and its transitive types are not recursive, so they expand cleanly.
  ▎
  ▎ This matches the empirical behaviour visible in the already-committed health_event.crd.yaml (distros/kubernetes/nvsentinel/charts/k8s-datastore/crds/), which is what HealthEventResource produces today using the same generator —
  ▎ also zero x-kubernetes-preserve-unknown-fields.

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,460 @@
...
+
+Both `True` and `False` on `ExternalRemediationComplete` are meaningful, terminal outcomes from the external system's perspective. The ERR reconciler treats them symmetrically: remove the release taint, log the outcome, and leave the rest of the system to handle whatever follows. Specifically:
+
+- On `True`: the external system has reported success. NVSentinel's health-monitors will, on their next observation, naturally either emit a healthy event (clearing fault-quarantine state and returning the node to service) or re-detect the same fault (re-triggering the pipeline and producing a fresh ERR).
+- On `False`: the external system has reported "I tried, couldn't fix it." The same downstream path applies — health-monitors decide. If the underlying fault is still present, a new ERR will be produced via the standard pipeline. If the fault is gone (the external system gave up but the issue resolved separately), the node returns to service naturally.
```
lalitadithya: I don't think the same downstream path should apply here, I think in this case there needs to be a human in the loop here, so we should have a metric that we can alert on saying "hey we tried some automation, it failed and it gave up, come help fix"

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,460 @@
...
+### ERR reconciler
+
+**Watches:** `ExternalRemediationRequest` (primary); `Node` (secondary, to detect taint drift).
+
+**Reconcile loop:**
+
+1. On first reconcile, call `SetInitialConditions`: write `NVSentinelOwnershipReleased=Unknown (Initializing)` and `ExternalRemediationComplete=Unknown (AwaitingExternalSystem)`.
+2. If `NVSentinelOwnershipReleased` is `Unknown`:
+   - Apply the configured release taint to `spec.nodeName` (idempotent).
+   - Set the `nvsentinel.nvidia.com/monitoring.active=false` label on `spec.nodeName` to tear down NVSentinel's own health monitors for the duration of external remediation (see "Health-monitor teardown" below).
```
lalitadithya: both of these should be done in a single patch api call right?

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,460 @@
...
+   - On success: set `NVSentinelOwnershipReleased=True` with `reason=ReleaseTaintApplied`.
```
lalitadithya, line 223: should this wait for the health monitors to be removed from the node before the conditions are set?

docs/designs/040-external-remediation-request.md
jtschelling: yeah that would be fine

docs/designs/040-external-remediation-request.md
jtschelling: yeah that'd be fine

docs/designs/040-external-remediation-request.md
jtschelling: i think the human is the external system that an ERR would expect to fulfill the request. if we want a human operator to do something then we still have to assume that nvsentinel releases ownership of the node. the coordination between the automated external system and the human operator is out of scope of nvsentinel IMO, that should be on the end user.

for example, nvsentinel finds fault X and creates an ERR. Automated external system fails to remediate, turns health monitors back on. nvsentinel still sees fault X and creates an ERR. Automated external system needs to have historical tracking and would notify the human operator, and when the human operator gets the notification the node is prepared for them to work on already

lalitadithya: sure, but my question is if the external system already knows if it couldn't fix it and set the status to false, why would it come back to NVSentinel shouldn't it as a part of its workflow to escalate to a human?

docs/designs/040-external-remediation-request.md
jtschelling: lol i responded too quick, i actually think its a better design to set the label and forget about monitoring them. we've ran into cases with nodes that are stuck and i dont want a non-terminating pod to block on the ERR continuing. also we'd have to maintain some sort of knowledge of which pods need to be removed which is additional overhead. i'd rather implement the reconciler as described now and if there's operational issues around this segment we can revisit later

docs/designs/040-external-remediation-request.md
jtschelling: what you described there makes sense and should probably be the external system's behavior. however on nvsentinel's side i dont think we can make assumptions of what the external system is going to do and need to keep behavior consistent: if an ERR is in a terminal state, nvsentinel owns the node and should continue normal operation

> in this case there needs to be a human in the loop here, so we should have a metric that we can alert on saying "hey we tried some automation, it failed and it gave up, come help fix"

just to be more specific, the ERR should be the coordination point between a human operator and nvsentinel. in the doc there is a `err_total` metric tracking failures that could be used for alerting if the automated external system doesn't contact a human

docs/designs/040-external-remediation-request.md
lalitadithya: > nodes that are stuck and i dont want a non-terminating pod to block on the ERR continuing

fair, we can always have a timeout for when pods don't exit nodes right? 

> also we'd have to maintain some sort of knowledge of which pods need to be removed which is additional overhead

We can always take a label selector or namespace selector as input right? We have to filter out daemonset pods on the node and then select only ones that match. 

My worry is that when the pods don't terminate in time for whatever reason and we end up racing the remediation system if it starts before the node is fully cleaned up.

docs/designs/040-external-remediation-request.md
lalitadithya: > if an ERR is in a terminal state, nvsentinel owns the node and should continue normal operation

I don't think so. I think the hand-off contract should be that nvsentinel takes back the node only if the external system says it has successfully remediated the node. If the external system says it has failed, nvsentinel should not enable monitoring on it. Having a metric and alert from the NVSentinel side is fine, but I'd rather not have the node go back into service because it passes health checking. For example, what if the ERR failed at the final stages of remediation and the node is in an unsafe state and shouldn't be used, in such cases we shouldn't return the node to service.

jtschelling: how do we return it to service if the ERR is in a terminal state and the health monitors are not running?

lalitadithya: two case:
1. terminal with status=false: we don't return it to service
2. terminal with status=true: enable health monitors and return to service if they pass

jtschelling: when we're in state 1 how do we get nvsentinel to return the node to service? thats what i dont understand, it seems like it would be a lot of manual undoing of actions nvsentinel took

lalitadithya: I was imagining the flow would be something like:
1. external system sets status=false
2. someone gets paged
3. engineer looks though all the steps there were done to find what the error was and manually fixes the error 
4. user manually updates the CR with status=true 
5. our reconciler sees 4 and then it will enable health monitors, untaint etc

lalitadithya: 4 doesn't have to be manual, the external system could have some button somewhere where it will then programmatically set the status, but that is orthogonal.

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,468 @@
...
+
+The ERR reconciler manages a single Node label, `nvsentinel.nvidia.com/monitoring.active`, alongside the release taint:
+
+- **Value `"false"`** — monitors should be off for this node. Set by the ERR reconciler atomically with the release taint apply (step 2 of the reconcile loop).
+- **Absence (or any other value)** — monitors run normally. The default state; restored by the ERR reconciler when the release taint is removed (step 3).
+
+Monitors consume the label in two different shapes depending on whether they run per-node or cluster-wide:
+
+**DaemonSet monitors** (gpu-health-monitor, syslog-health-monitor, nic-health-monitor) carry a `nodeAffinity` clause requiring the label to NOT be `"false"`:
```
natherz97: Are we sure that the changing the nodeAffinity rule will result in the daemonset being deleted from the node? I know that's the case for nodeSelectors but in general nodeAffinity isn't re-evaluated for running pods unless this is something handled by the daemonset-controller (which is hinted from the "IgnoredDuringExecution" part of the name). I think you would need to leverage the nodeSelector label the daemonset uses. 

What if we instead of the nvsentinel.nvidia.com/monitoring.active label, we just leveraged the existing ManagedByNVSentinel=false label?  We could pair this with our existing node-labeler which is currently the source of truth for determining which health-monitor should run on each node. I think we could simplify this this flow by:
1. Add the existing ManagedByNVSentinel=false label rather than adding nvsentinel.nvidia.com/monitoring.active.
2. This prevents us from needing to update cluster-scope monitors to reference the label since fault-quarantine will ignore these health events.
3. The node-labeler can remove the daemonset labels from any nodes which have the ManagedByNVSentinel=false label added to tear down the existing health-monitors. 

It seems like ManagedByNVSentinel=false in general should result in the health-monitor tear down even outside of this use-case.

docs/designs/040-external-remediation-request.md
natherz97: We have the invariant that the node will remain cordoned until the original health-monitor recovers. So we shouldn't need to update our alerting behavior whether we choose either behavior:
- Only return to service is the node is healthy from external remediation + our health-monitor
- Only return to service after our health monitor recovers (regardless of what external remediation did or if it succeeded).

I think we should make sure that either automatic or manual uncordons clears the taint + labels added by external remediation if they don't end up updating the CR status to true.

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,468 @@
...
+  creationTimestamp: "2026-05-11T21:16:02Z"
+  generation: 1
+  labels:
+    nvsentinel.nvidia.com/component-class: gpu
+    nvsentinel.nvidia.com/node: node-01.example-cluster.internal
+    nvsentinel.nvidia.com/recommended-action: external-remediation
+  name: gpu0-xid79-node-01
+  namespace: nvsentinel
+spec:
+  agent: gpu-health-monitor
```
natherz97: Should we have a higher-level field named healthEvent in the spec? I think it would be cleaner to nest all the healthEvent properties under this higher level field, particularly if we add additional properties to the spec that aren't from the healthEvent. We would lose visibility into what fields we're setting manually vs. what's in the healthEvent.

docs/designs/040-external-remediation-request.md
lalitadithya: there is a bit of history behind that label `ManagedByNVSentinel`. The reason why we introduced that label was that we wanted a way where we could just turn off cordon+drain+remediate on nodes that we were debugging while leaving the health monitors running. The use case that drove that label was that in some cases we found that some user operations cause errors, so we wanted a way to stop break/fix while we root cause the flow using node conditions to know when it errored and when it cleared. If we are changing the behaviour of the existing label, let's call it out in the doc and what is the alternative solution that users can leverage for this use case.

jtschelling: the `k8saas.nvidia.com/ManagedByNVSentinel` is only a gate on FQ, and is a configuration value for dropping health events. it doesnt do anything with scheduling/placement, so all the health monitors would still be running on the node. i'm wary of using this label as a pivotal component of the descheduling flow because 1. its not anywhere in code, only end-user configurable behavior and 2. the namespace isnt correct

nodeAffinity works for descheduling daemonsets, just ran a quick test. its also in the docs here https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#updating-a-daemonset . the field naming doesnt really match the behavior though, i agree with you there.

i went with nodeAffinity because i didnt want to have a bootstrap step labeling all the nodes running nvsentinel to have a specific label that i would use nodeSelector on, since the label would have to exist before the DS was able to schedule pods there. using node-labeler would remove the bootstrap problem because it can automatically add a label to match the DS nodeSelector on node join and then ERR reconciler will be able to flip that back and forth. i think it'd look something like:

- labeler stamps nvsentinel.dgxc.nvidia.com/deploy.healthMonitors=true on monitored nodes by default (matching labeler's existing label domain). Idempotent — only writes when the label is absent, so it doesn't fight the ERR reconciler's writes.
- Monitor DaemonSets carry nodeSelector: nvsentinel.dgxc.nvidia.com/deploy.healthMonitors: "true".
- ERR reconciler flips the label "true" → "false" alongside the release-taint apply (still a single PATCH), and back to "true" on terminal completion.
- Cluster-scope monitors still need a code-level check on the label since they don't run on the node but they need that for any descheduling anyways.

to sum it up i think there's three options right now for descheduling:

1. use nodeAffinities on the DS specs and toggle with ERR reconciler
2. update node-labeler to include a new label `nvsentinel.dgxc.nvidia.com/deploy.healthMonitors` on new nodes, use nodeSelectors on the DS specs and toggle with ERR reconciler
3. make a new controller to manage this portion of the design

docs/designs/040-external-remediation-request.md
jtschelling: i think its fine how it is now in the document. i'm not sure what else you would need in the API that you couldnt put as additional information in the object metadata or health event metadata. @lalitadithya ?

(to be clear im also fine with changing it to spec.healthEvent or whatever, this isnt a sticking point for me)

docs/designs/040-external-remediation-request.md
jtschelling: oh that would work. i was stuck thinking the object was immutable after reaching a terminal state. that would be fine

docs/designs/040-external-remediation-request.md
natherz97: Thanks for checking the daemonset behavior. So it sounds like you're proposing: node-labeler add the nvsentinel.dgxc.nvidia.com/deploy.healthMonitors label to nodes + ERR reconciler toggles its value directly.

The alternative would be: node-labeler add the nvsentinel.dgxc.nvidia.com/deploy.healthMonitors label to nodes + and removes it when the ManagedByNVSentinel label is false and ERR reconciler only adds/removes the label.

I'm in favor of being consistent with the existing logic and using node-labeler so I think either approach is fine since the existing labeling behavior is already based on pod + node observations which result in labels being added/removed from nodes. For comparing 1 vs. 2, do we want to maintain the logic for all cluster-wide health-monitors needing knowledge of this label compared to deferring to fault-quarantine gate?

docs/designs/040-external-remediation-request.md
natherz97: This is a two-door decision, we can always update the ERR reconciler later. What JT is saying is consistent with our existing Janitor remediations where we'll uncordon the node even if Janitor says it failed the action. 

Maybe to start we keep the node tainted since it does seem like we should be aware of external-remediation failures and we can revisit that decision if we see cases where NVSentinel would've uncordoned successfully even if Maestro reported a failure.

lalitadithya: yeah, it is a two-door decision, certainly. But there is a bit of nuance here because in the Janitor we know that it is not going to take action after it fails and we don't have taints signifying ownership. In the case of external remediation I think we need to be a bit more defensive so maybe we start with being defensive and then we can loosen up over time if required. 

Then we also have to consider the angle of alerting where most of the SLI/SLO alerts are based on cordoned nodes (AFAIK) and not necessarily based on taints.

docs/designs/040-external-remediation-request.md
lalitadithya: @natherz97 , did you have a use case in mind for this? 

It doesn't hurt to make it extensible certainly. I'm fine with nesting it to keep the door open for additional fields we may need in future it will save us doing conversation.

natherz97: The main one that came to mind was if we're asking the external system to do an action for us vs. if we're deferring to the external system to investigate + remediate. I think we could solve that without nesting depending on the contract we have by mapping both CUSTOM -> ExternalRemediation and RESTART_VM -> ExternalRemediation and then the external system would need to be smart enough to know that the recommended action will tell it what to do (CUSTOM means troubleshoot and try to remediate where RESTART_VM means just do the reboot for us and pass back). 

It might be cleaner to embed this context in the spec directly like shouldReboot, shouldTerminate, shouldTriage rather than having to derive this from the HealthEvent.

docs/designs/040-external-remediation-request.md
natherz97: From our previous discussions, there's not a situation where the node would be uncordoned but the taint is still present right? My understanding is we have the invariant that an uncordon can't occur until NVSentinel is re-enabled at the same time the taint is removed.

jtschelling: correct the ERR reconciler is not involved in the uncordoning behavior. if the fault still exists when the health-monitors are scheduled on the node then the node will not be uncordoned

lalitadithya: > From our previous discussions, there's not a situation where the node would be uncordoned but the taint is still present right?

correct, yes. But what confused me is that in this thread we also said the followings:

> Maybe to start we keep the node tainted since it does seem like we should be aware of external-remediation failures

My main point in this thread here is that if the external system says the fault is not remediated, we should consider that a terminal state and not move forward. The key different between this and janitor is that in the case of Janitor we control the system so we know at what points would failure be marked and if it is safe, but the same can't be said for an external system.

docs/designs/040-external-remediation-request.md
```diff
@@ -218,82 +227,88 @@ All conditions are append-or-update in place via the `SetCondition` helper, whic
...
+`node-labeler` (an existing NVSentinel `Deployment`) is extended to manage this label as part of its existing infrastructure-detection role. Behaviour:
 
-Monitors consume the label in two different shapes depending on whether they run per-node or cluster-wide:
+- Watches Node objects via its existing informer.
+- On observing a Node without `monitoring.active` set, evaluates a configurable **eligibility expression** against the Node. If eligible, stamps the label to `"true"`. The expression lives in `node-labeler`'s Helm values as a CEL expression, mirroring the pattern `fault-quarantine` already uses to gate its own rule matching (see `charts/fault-quarantine/values.yaml`). The default expression excludes nodes carrying `k8saas.nvidia.com/ManagedByNVSentinel=false`, matching `fault-quarantine`'s existing default so operators using that opt-out continue to see the same behaviour. The label itself is referenced **only in chart values**, never in `node-labeler`'s Go code — changing the eligibility rule (including removing the `ManagedByNVSentinel` check entirely) is a chart-values change, not a code change.
```
lalitadithya: what would be the behaviour if no expression is specified?

Specifying the `k8saas.nvidia.com/ManagedByNVSentinel` would change the current behaviour where only cordon+drain is stopped but with this monitoring would also be stopped.

docs/designs/040-external-remediation-request.md
jtschelling: okay updating the document to say that if an ERR fails then we should not re-turn on health monitors

docs/designs/040-external-remediation-request.md
jtschelling: move a lot of this logic to node-labeler

docs/designs/040-external-remediation-request.md
jtschelling: change to nvsentinel.dgxc.nvidia.com/managed but keep that label as the main one

docs/designs/040-external-remediation-request.md
jtschelling: updated. this line is the important one now:

> Ownership is transferred back to NVSentinel by exactly two events: (1) the external system writing `ExternalRemediationComplete=True` on the ERR, or (2) an operator running `kubectl delete err <name>` to invoke the finalizer-driven cleanup path. `ExternalRemediationComplete=False` is **not** a transfer event — it signals failure on the external system's side but leaves ownership with them until one of the two transfer events occurs.

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,503 @@
...
+2. **A status checker for `ExternalRemediationRequest`** plugging into the existing `StatusChecker` interface: returns `ShouldSkip=true` while the ERR is still claiming the node — `ExternalRemediationComplete=Unknown` (in-flight) **or** `ExternalRemediationComplete=False` (failed-but-still-released, see *Condition state machine*). Returns `ShouldSkip=false` only when `ExternalRemediationComplete=True`, at which point the entry is pruned from the `latestFaultRemediationState` annotation as usual. If the operator deletes the ERR via the finalizer escape hatch, the ERR object disappears entirely and the corresponding annotation entry is treated as missing on the next reconcile — same effect.
```
natherz97, line 355: Do we want to only allow 1 ERR per time the node is cordoned? We discussed that if the external system does not fix the underlying issue, we would trigger a new unhealthy event when the health-monitors are re-enabled which would likely create a new ERR. Would it be preferable to skip new ERR creation regardless of its status and we'd only allow a new ERR if the node is uncordoned and cordoned again?

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,503 @@
...
+
+`metadata-collector` is included in the list above for completeness: it gates on `driver.installed` AND `nvidia.com/gpu.present`. When `driver.installed` is withheld, the AND fails and `metadata-collector` is evicted too. This is the desired behaviour during external remediation — NVSentinel's collection should also stop on a released node.
+
+#### Cluster-scope monitors: code-level emission gating
+
+Cluster-scope monitors (`csp-health-monitor`, `kubernetes-object-monitor`, `slurm-drain-monitor`) run as `Deployment`s, not DaemonSets, and target nodes by name from outside the node. They cannot be evicted from a node because they do not run on it. Instead, each cluster-scope monitor reads the target node's `managed` label from its Kubernetes informer cache and skips emission when `managed="false"`. Any other state (label absent, label set to `"true"`, or any other value) means the cluster-scope monitor emits as usual.
```
natherz97, line 307: Do we not want to rely on fault-quarantine dropping events from the cluster-wide monitors from the managed="false" node label?

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,503 @@
...
+
+**Node deletion:** if `spec.healthEvent.nodeName` no longer exists at reconcile time, the reconciler logs and treats taint and label operations as no-ops. The ERR object remains until the external system writes `ExternalRemediationComplete=True` or until an operator deletes it. If the external system never acknowledges, the ERR persists with `ExternalRemediationComplete=Unknown` (visible via the open-ERR gauge — see Observability); the operator escape hatch below is how that state is reclaimed.
+
+**Operator-driven release.** Because `ExternalRemediationComplete=False` (or persistent `Unknown`) intentionally leaves the node released — NVSentinel does not infer that the external system has given up — there must be an explicit operator-facing knob to force the node back into NVSentinel ownership. The mechanism is a finalizer (`nvsentinel.nvidia.com/external-remediation-cleanup`) added to every ERR by the reconciler on first reconcile. The operator-facing workflow:
```
natherz97, line 253: I think the desired flow would be to delete the ERR, restore health-monitors, send a healthy event, and fault-quarantine uncordons the node. If someone deletes the ERR and the node stays cordoned, they can investigate the node exactly as they do now, but what happens if someone manually uncordons the node when an ERR is still active? The node would stay as tainted until they delete the ERR. Do you think that's the best operator experience? 

I don't think we want to to enforce that you can't uncordon a node with an active taint on the server side but maybe our cordon + uncordon scripts could enforce this where we block an uncordon if there's an active ERR?

docs/designs/040-external-remediation-request.md
```diff
@@ -0,0 +1,503 @@
...
+
+**Single active ERR per node.** Because the `(key, effect)` taint tuple is unique on a Node, only one ERR can have its release taint applied at a time. This invariant is enforced by `fault-remediation`'s existing equivalence-group machinery (see the *`fault-remediation` integration* section): ERRs declare their own equivalence group and a status checker, and the existing "skip CR creation when a CR for a matching group is in progress" logic prevents a second concurrent ERR from ever being created. No new dedup code is required in the ERR reconciler.
```
natherz97, line 247: If two ERRs are created concurrently, it seems like there would be a race condition where we could try to apply two different taints via a patch from the ERRs against the same node at the same time. If we used an update instead of a patch or a patch that also sets the resource version, you could detect that a conflicting operation occurred: https://kubernetes.io/docs/reference/using-api/api-concepts/#patch-and-apply. I don't think we need to implement this since we have the gate from fault-remediation dropping duplicates but if we ever have multiple systems creating ERRs I think we'll need to add this conditional update logic or add the node-level locking we have in Janitor (since with CSP APIs we had to add our own locking).

docs/designs/040-external-remediation-request.md
jtschelling: i think we need the check at the health monitor. if we have edge based monitoring on any of the checks we would need to requeue sending the health event if the node was not managed at the time of the event. if the event was sent through and dropped at FQ (which should happen anyways because of the equivalence groups) then the health monitor wouldnt know to re-send it

docs/designs/040-external-remediation-request.md
jtschelling: doesnt that fault still need to be remediated by _something_? the ERR isnt prescriptive on what does the remediation, only that nvsentinel cannot do it internally. the finalizer here is for preventing locking up the system on a non-terminal state ERR deletion. if we didnt have the finalizer undoing its operations we'd have the taint/health-monitors dead and would have to manual undo the ERR reconciler actions

alexander-demicev (commented):
Thanks a lot for writing it the ADR. I have a couple of questions:
- When external remediator writes `ExternalRemediationComplete=True` will fault-remediation set `healtheventstatus.faultremediated.value=true`? This is crucial for `health-events-analyzer` rules to count attempted remediations.
- ERR sets `nvsentinel.dgxc.nvidia.com/managed=false` , the remediator fires an action requiring human intervention and potentially can take some time. In this case, can we leave monitors running on the node so person looking into the issue can have observability over broken node?

docs/designs/040-external-remediation-request.md
natherz97: We're making the call to ignore health signals sent from the daemonset health-monitors by not running them, should we be consistent with that behavior and drop/not send any health events from the cluster-wide monitors? A health event that fires when the external remediation is active may not be valid and we're already dropping events from the core monitors. Maybe we can choose to opt-in specific monitors? For example the CSP health monitor should probably re-send the event, but the K8s object monitor should likely not re-send.

docs/designs/040-external-remediation-request.md
natherz97: To clarify, I wasn't referring to the finalizer, I was just saying that we might only want to create one ERR per node when it is cordoned to prevent us from continually retrying and recreating ERRs if the original ERR successfully completes but the original health monitor does not and re-fires the same unhealthy event which originally led to the ERR. 

For RebootNodes, we create a new CRD if the status is success or failed and skip if it is in-progress. For ERR, we could always skip creating a new CRD if the existing status is success, failed, or in-progress.

docs/designs/040-external-remediation-request.md
jtschelling: i agree this is a confusing situation for operators. there are multiple options for handling this (auto-triggered ERR cleanup on uncordon through some nvsentinel component being one) that we will handle in follow up designs

docs/designs/040-external-remediation-request.md
jtschelling: yep this should be easy to do in the implementation, will include it

natherz97 (approved):

jtschelling:
@alexander-demicev 

> When external remediator writes ExternalRemediationComplete=True will fault-remediation set healtheventstatus.faultremediated.value=true? This is crucial for health-events-analyzer rules to count attempted remediations.

that gets set at remediation creation time by fault-remediation, so the ERR reconciler doesnt have to do anythign to get this benefit

> ERR sets nvsentinel.dgxc.nvidia.com/managed=false , the remediator fires an action requiring human intervention and potentially can take some time. In this case, can we leave monitors running on the node so person looking into the issue can have observability over broken node?

i think we're trying to avoid this with this implementation, it kind of relies on the health-monitors not running because it can interfere with the remediation process. **but**, there's nothing stopping someone from removing the label, if the label is removed while an ERR is active the health-monitors will start to schedule again and publish health events. ERR reconciler does not watch that label, only sets it at the beginning of the reconciliation

PR #1323, opened 2026-05-21
docs: ADR for priority queue in node-drainer

### LINKED ISSUE #1322: [Feature]: Improve node-drainer performance by coalescing InProgress health events
### Prerequisites

### Feature Summary

Coalesce overlapping built-in node-drainer events for the same drain scope so that only one active event performs the drain while later covered events wait and eventually resolve as AlreadyDrained

### Problem/Use Case

Today, built-in drains skip redundant work only after a previous drain has completed. If another health event arrives for the same node and same impacted entity while the first built-in drain is still `InProgress`, node-drainer does not treat the active drain as covering the new event.

Because drain events are retried through the workqueue, overlapping events can interleave across queue turns and repeatedly run the same drain evaluation, pod scans, eviction attempts, completion checks, node-event updates, and retries. This is especially noisy for bursts of events that represent the same physical fault, such as repeated partial-drain events for the same `GPU_UUID`.

This does not generally create unsafe parallel eviction because node-drainer currently has a single worker and Kubernetes eviction is mostly idempotent, but it is sub-optimal: it adds queue churn, Kubernetes API calls, logs/events/metrics noise, and can slow overall convergence to terminal drain status.

### Proposed Solution

TBD

### Component

Fault Management

XRFXLP:
## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added ADR describing scope-aware coalescing for overlapping node-drain events so built-in drains coordinate rather than duplicate work.
  * Clarified "completed coverage wins" semantics and an owner/follower ownership model where only the owner performs work.
  * Documented two-phase evaluation, cold-start/recovery behavior, observable broken states, and required metrics/logging for overlap waits and recovery.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NVSentinel/pull/1323?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


docs/designs/040-node-drainer-overlapping-built-in-drains.md
```diff
@@ -0,0 +1,259 @@
+# ADR-040: Node Drainer — Coalesce Overlapping Built-in Drains
+
+## Context
+
+Node-drainer processes quarantined health events and executes either:
+
+- the built-in drain flow, which evicts or waits for Kubernetes pods directly, or
+- the custom-drain flow, which delegates drain work to a configured CR.
+
+This ADR is scoped only to built-in drains. Custom drain already has its own active-drain handling through drain CRs and is intentionally out of scope.
```
lalitadithya: would a simpler solution here be to move all the existing in-tree drain plugins to CRs? Is that an alternative we should consider?

docs/designs/040-node-drainer-overlapping-built-in-drains.md
```diff
@@ -0,0 +1,259 @@
...
+
+A full-node drain covers all later built-in drains for the node. A partial drain covers only later partial drains for the same impacted entity. A partial drain must not block a later full-node drain.
+
+Completed coverage wins before active ownership is considered. If any covered prior event has `userpodsevictionstatus.status == Succeeded`, the current event is marked `AlreadyDrained`, even if the current event has stale follower metadata or another covered event is still `InProgress`.
+
+## Ownership Resolution
+
+Ownership resolution is the gate before built-in drain work. Only an event that has resolved as owner reaches the built-in drain action evaluator.
+
+Node-drainer persists ownership state under `healtheventstatus.userpodsevictionstatus.metadata`:
```
lalitadithya: Can this be done using Kubernetes lease objects https://kubernetes.io/docs/concepts/architecture/leases/? What you are describing in this document is sort of a leader election, which is also what the lease object was designed to solve. We can use heartbeats to ensure that there is always at least one thread acting on the drain

docs/designs/040-node-drainer-overlapping-built-in-drains.md
lalitadithya, line 1: can we add some alternative options that we have considered?

docs/designs/040-node-drainer-overlapping-built-in-drains.md
XRFXLP: added that in alternatives considered.

docs/designs/040-node-drainer-overlapping-built-in-drains.md
XRFXLP: added that in alternatives considered.

docs/designs/040-node-drainer-overlapping-built-in-drains.md
XRFXLP: Done, PTAL.

Common downside in all them is they would add extra source of truths.

docs/designs/040-node-drainer-overlapping-built-in-drains.md
```diff
@@ -252,6 +252,24 @@ Broken states are operational signals. They should be visible and bounded rather
...
+
+### Use Kubernetes Lease objects
+
+Node-drainer could also represent drain ownership with Kubernetes `Lease` objects keyed by node and drain scope. `Lease` is a valid general coordination primitive, but using it here would move drain ownership into the Kubernetes API while drain status remains in the health-event datastore. That adds another state plane, RBAC and object lifecycle concerns, and cleanup requirements for every drain scope. Keeping ownership in the health-event datastore avoids that split and keeps ownership visible with the events that fault-quarantine, node-drainer, and remediation already use.
```
lalitadithya: I agree that it adds additional RBAC and cleanup, but wouldn't it be easier to use a solution that has already been built and tested rather than building our own copy? Plus, it will also make the implementation a lot simpler. 

WRT to cleanup, it could be something as simple as delete all on startup.

lalitadithya: When you say additional source of truth do you mean the answer to "who is the drain leader?" if that is the case then where else do we have this information stored that will cause duplication?

XRFXLP: > but wouldn't it be easier to use a solution that has already been built and tested rather than building our own copy? Plus, it will also make the implementation a lot simpler.

If this were only full-node drains, I agree that Kubernetes Lease would fit very well. The tricky part is that node-drainer has drain-scope semantics: full drains cover partial drains, partial drains only cover the same impacted entity, and unrelated partial drains should not block each other. A Lease would give us the storage/claim primitive, but node-drainer would still need to build all of the scope mapping and coverage logic around it. At that point we are mostly using Lease as storage, not getting much semantic help from it.



> When you say additional source of truth do you mean the answer to "who is the drain leader?"

By “additional source of truth,” I don’t mean that the drain leader would be stored twice. I mean the drain lifecycle would be split across two systems: the DB would still store drain status (`InProgress`, `Succeeded`, `AlreadyDrained`, etc.), while Kubernetes Lease would store who is currently driving the drain. That split means node-drainer has to keep those two pieces reconciled during retries, cold start, cancellation, and cleanup.

docs/designs/040-node-drainer-priority-queue.md
```diff
@@ -0,0 +1,161 @@
...
+The priority is node-level: events for nodes not yet marked `draining` are more urgent than events for nodes already marked `draining`. This is intentionally narrower than scope-aware coalescing; it optimizes drain start latency without introducing owner/follower event lifecycle state.
+
+## Priority Model
+
+Node-drainer already receives one queue item per event. The priority model changes the order in which ready queue items are processed; it does not change the event lifecycle.
+
+The queue tracks a small in-memory set of nodes that node-drainer has successfully marked `draining`.
+
+```text
+if node is not in draining set:
```
lalitadithya: would an optimization here be to also check that the node is not already present in the high priority queue?

XRFXLP: Addressed

lalitadithya: I don't see the change, can you please check?

lalitadithya (approved):

PR #598, opened 2025-12-17
feat: add github action to clean up old untagged images

KaivalyaMDabhadkar:
## Summary
Over time, container registries accumulate orphaned/untagged images and outdated versions, consuming storage and making it harder to manage packages. This workflow enables automated cleanup based on configurable retention policies.
This PR adds a github actions workflow to automatically clean up old container images from GitHub Container Registry (ghcr.io) using [snok/container-retention-policy](https://github.com/snok/container-retention-policy).

## Features

- Manual trigger via workflow_dispatch for controlled execution
- Dry-run mode currently (default) to preview deletions before applying
- Configurable retention policies

## Testing

Validated in a test repository with tagged and untagged images - successfully identified and deleted target images while preserving intended versions.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [x] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [] Other: ____________

## Testing
- [x] Tests pass locally
- [x] Manual testing completed
- [ ] No breaking changes (or documented)

## Checklist
- [x] Self-review completed
- [ ] Documentation updated (if needed)
- [x] Ready for review

## Summary by CodeRabbit

* **Chores**
  * Added automated cleanup of untagged container images with daily scheduling and a manual trigger.
  * Configurable retention: dry-run option, age cutoff (default 30 days), and keeps the 5 most recent untagged images.
  * Targets untagged images only, uses update timestamp for age, 30-minute job timeout, and applies across many components.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1339, opened 2026-05-27
fix: clean up main branch container images

### LINKED ISSUE #171: [Feature]: Clean up old images published form main in container registry
### Prerequisites

### Feature Summary

Each PR merge on main publishes images with `main-sha`. 

### Problem/Use Case

Bloating the container registry. Clean these up using age

### Proposed Solution

https://github.com/marketplace/actions/ghcr-io-cleanup-action

### Component

Multiple Components

valdo766hi:
Closes #171.

## Summary

This completes the cleanup work started in #598.

The existing workflow only removes untagged GHCR images. Images published from `main` are tagged as `main-<shortsha>`, so they were not covered by the untagged cleanup policy.

## Changes

- keep the existing untagged image cleanup
- add cleanup for old `main-*` tagged images
- keep the latest 10 matching main images
- preserve release tags by only matching `main-*`
- keep manual runs dry-run safe by default
- allow scheduled runs to perform real cleanup
- add currently published packages missing from the cleanup matrix

## Validation

```bash
ruby -e 'require "yaml"; YAML.load_file(".github/workflows/cleanup-untagged-images.yml"); puts "valid yaml"'
grep -nE 'delete-tags: main-\*|keep-n-tagged: 10|older-than:|validate: true' .github/workflows/cleanup-untagged-images.yml
grep -c "dry-run: \${{ github.event_name == 'workflow_dispatch' && github.event.inputs.dry_run || 'false' }}" .github/workflows/cleanup-untagged-images.yml
git diff --check
```

## Summary by CodeRabbit

* **Chores**
  * Renamed and clarified the container image cleanup workflow and its manual-run inputs.
  * Reformatted cutoff input to a human-readable "30 days" and standardized retention defaults.
  * Expanded the list of image packages covered and improved ordering.
  * Adjusted behavior for dry-run between scheduled and manual runs.
  * Added automated removal of older build tags, preserving the 10 most recent and enabling validation.

### DIFF
--- a/.github/workflows/cleanup-untagged-images.yml
+++ b/.github/workflows/cleanup-untagged-images.yml
@@ -114,7 +114,7 @@ jobs:
           timestamp-to-use: updated_at
 
       - name: Delete old main branch images for ${{ matrix.package }}
-        uses: dataaxiom/ghcr-cleanup-action@v1.2.1
+        uses: dataaxiom/ghcr-cleanup-action@f092b48ba3b604b2a83690dc4b2bbb3392e1045f  # v1.2.1
         with:
           owner: NVIDIA
           package: ${{ matrix.package }}
--- a/.github/workflows/cleanup-untagged-images.yml
+++ b/.github/workflows/cleanup-untagged-images.yml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Cleanup Untagged Container Images
+name: Cleanup Container Images
 
 on:
   # Run daily at midnight UTC
@@ -30,9 +30,9 @@ on:
           - 'true'
           - 'false'
       cut_off:
-        description: 'Delete images older than this (e.g., 1h, 1d, 7d, 30d)'
+        description: 'Delete images older than this (e.g., 1h, 1d, 7d, 30 days)'
         required: true
-        default: '30d'
+        default: '30 days'
         type: string
 
 concurrency:
@@ -44,8 +44,8 @@ permissions:
   contents: read
 
 jobs:
-  cleanup-untagged-images:
-    name: Cleanup Untagged Images
+  cleanup-container-images:
+    name: Cleanup Container Images
     runs-on: linux-amd64-cpu32
     if: github.repository == 'NVIDIA/NVSentinel'
     timeout-minutes: 30
@@ -56,8 +56,10 @@ jobs:
         package:
           - nvsentinel
           - nvsentinel/platform-connectors
+          - nvsentinel/plugins-slinky-drainer
           - nvsentinel-platform-connectors
           - nvsentinel/gpu-health-monitor
+          - nvsentinel/gpu-reset
           - nvsentinel/syslog-health-monitor
           - nvsentinel/csp-health-monitor
           - nvsentinel-gpu-health-monitor
@@ -79,6 +81,7 @@ jobs:
           - nvsentinel/log-collector
           - nvsentinel/file-server-cleanup
           - nvsentinel/janitor
+          - nvsentinel/janitor-provider
           - nvsentinel/health-events-analyzer
           - nvsentinel/maintenance-notifier
           - nvsentinel/event-exporter
@@ -92,6 +95,7 @@ jobs:
           - nvsentinel/preflight-dcgm-diag
           - nvsentinel/nic-health-monitor
           - nvsentinel/preflight-nccl-allreduce
+          - nvsentinel/preflight-nccl-loopback
 
     steps:
       - name: Delete untagged images for ${{ matrix.package }}
@@ -100,12 +104,23 @@ jobs:
           account: NVIDIA
           token: ${{ secrets.GITHUB_TOKEN }}
           image-names: ${{ matrix.package }}
-          cut-off: ${{ github.event.inputs.cut_off || '30d' }}
+          cut-off: ${{ github.event.inputs.cut_off || '30 days' }}
           # Keep at least 5 most recent untagged versions as safety buffer
           keep-n-most-recent: 5
-          dry-run: ${{ github.event.inputs.dry_run || 'true' }}
+          dry-run: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.dry_run || 'false' }}
           # Only target untagged images - preserve all tagged releases
           tag-selection: untagged
           # Use updated_at timestamp for age calculation
           timestamp-to-use: updated_at
+
+      - name: Delete old main branch images for ${{ matrix.package }}
+        uses: dataaxiom/ghcr-cleanup-action@v1.2.1
+        with:
+          owner: NVIDIA
+          package: ${{ matrix.package }}
+          delete-tags: main-*
+          older-than: ${{ github.event.inputs.cut_off || '30 days' }}
+          keep-n-tagged: 10
+          dry-run: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.dry_run || 'false' }}
+          validate: true
 
--- a/.github/workflows/cleanup-untagged-images.yml
+++ b/.github/workflows/cleanup-untagged-images.yml
@@ -99,7 +99,7 @@ jobs:
 
     steps:
       - name: Delete untagged images for ${{ matrix.package }}
-        uses: snok/container-retention-policy@3b0972b2276b171b212f8c4efbca59ebba26eceb  # v3.0.1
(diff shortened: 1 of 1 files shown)


valdo766hi:
please take a look sir @KaivalyaMDabhadkar

KaivalyaMDabhadkar:
Thanks for the PR @valdo766hi !
Requesting some changes before we enable scheduled real deletion here:

This PR changes scheduled runs from dry-run to real deletion, because schedule events now resolve `dry-run` to `false`. That is fine for the new `main-*` cleanup goal, but it also makes the existing untagged cleanup step delete for real.

This step is still using `snok/container-retention-policy@v3.0.1`. NVSentinel publishes multi-arch images (`linux/amd64,linux/arm64`), and `snok`’s docs state that GHCR represents multi-arch child manifests as untagged package versions; deleting those children can break the tagged parent image. `snok` documents automatic multi-arch child protection starting in `v3.1.0`, while this workflow is pinned to `v3.0.1`.

Can you please:
1. Run this workflow manually with `dry_run: true` and share the output before we enable scheduled deletion.
2. Since this PR also changes scheduled runs from dry-run to real deletion, can we make the untagged cleanup multi-arch safe before merge, either by upgrading to a pinned `snok` `v3.1.0` SHA or by moving untagged cleanup to `dataaxiom/ghcr-cleanup-action` as well?

valdo766hi:
Thanks for the review @KaivalyaMDabhadkar.

I updated the existing untagged cleanup to use `snok/container-retention-policy` v3.1.0 pinned by full commit SHA. 

For the dry-run output, running this from my fork would not be useful because the job is guarded with `github.repository == 'NVIDIA/NVSentinel'`, so it would skip outside the upstream repo.

Could a maintainer/vetter trigger or approve a manual `workflow_dispatch` run in `NVIDIA/NVSentinel` with `dry_run=true`?

lalitadithya:
@KaivalyaMDabhadkar , can you please help here?

KaivalyaMDabhadkar:
Hi @valdo766hi, apologies for the long silence here, this fell off my radar.

I re-reviewed the latest changes and the `snok/container-retention-policy` v3.1.0 bump (SHA-pinned) resolves the multi-arch concern from my earlier review, v3.1.0 automatically detects and retains child manifests of tagged multi-arch images, so the untagged cleanup is now safe to run. The `dataaxiom/ghcr-cleanup-action` usage also checks out: `keep-n-tagged` combined with `delete-tags: main-*` restricts retention to the `main-*` subset, so release tags are untouched, and the pinned SHA matches v1.2.1.

On the dry run: you're right, I'll push this branch to a temporary branch in `NVIDIA/NVSentinel` and trigger a `workflow_dispatch` with `dry_run=true` myself, then share the run output here. If the logs look clean (only untagged versions and old `main-*` tags staged for deletion, no release tags), I'll approve and merge.

No further action currently needed from your side, thanks for the contribution and for your patience!

KaivalyaMDabhadkar:
Hi @valdo766hi,

**About the `snok` v3.1.0 bump:** every nightly run of this workflow on `main` is currently failing because v3.0.1 rejects GitHub’s new-format `GITHUB_TOKEN` ([[example failing run](https://github.com/NVIDIA/NVSentinel/actions/runs/30317091734)](https://github.com/NVIDIA/NVSentinel/actions/runs/30317091734)). Your SHA-pinned v3.1.0 bump fixes that and correctly protects multi-arch child manifests.

**The `dataaxiom` action needs to be allowlisted.** The validation run failed at startup because `dataaxiom/ghcr-cleanup-action` is not currently allowed in the NVIDIA organization ([[run](https://github.com/NVIDIA/NVSentinel/actions/runs/30339605627)](https://github.com/NVIDIA/NVSentinel/actions/runs/30339605627)).

I’m going to request that this exact SHA be added to the Actions allowlist:

```text
dataaxiom/ghcr-cleanup-action@f092b48ba3b604b2a83690dc4b2bbb3392e1045f
```

After reviewing the alternatives, I prefer keeping the purpose-built action rather than replacing it with an inline `actions/github-script` implementation. It handles multi-arch images and referrers explicitly and includes throttling and retries for GitHub API rate limits—which is important because the dry run identified over 10,000 stale `main-*` versions.

No change is currently needed from your side for the `dataaxiom` step. Once it is allowlisted, I’ll rerun the workflow with `dry_run=true` and verify that only eligible `main-*` tags are selected.

**One change is still needed:** `nvsentinel/plugins-slinky-drainer` does not exist in GHCR; the published package is `nvsentinel/slinky-drainer`. Please update the matrix entry to:

```yaml
          - nvsentinel/slinky-drainer
```

Everything else reviewed clean. Thanks for the contribution and your patience!

KaivalyaMDabhadkar:
Hi @valdo766hi, closing the loop on the `dataaxiom` question, sorry for the churn on this one. I know my last comment said I would rather keep the purpose-built action, but after allowlisting it and testing it properly, I have to walk that back beacause testing changed my mind, so let me share exactly what I found.

**Round 1, allowlisted, ran as-is:** the policy gate cleared, but the run failed on all 23 of the current-generation `nvsentinel/*` packages with `Request failed with status code 429` ([run 30426162779](https://github.com/NVIDIA/NVSentinel/actions/runs/30426162779)). The earlier claim about the action retries on rate limits turned out to be wrong. It fails the step on the first 429, and our matrix (16 concurrent jobs sharing the repo `GITHUB_TOKEN`, with packages holding 4-6k versions each) trips GitHub's secondary limits pretty reliably.

**Round 2, throttled with `max-parallel: 3`:** that did fix the rate limiting ([run 30431628191](https://github.com/NVIDIA/NVSentinel/actions/runs/30431628191),  41/42 green in about 26 minutes). But when I did the log verification, I found something more concerning: the action staged four `nvsentinel/nic-health-monitor` versions that carry **release tags** (`v0.8.0`, `v0.9.0`, `v1.2.0`, `v1.3.0` which are the releases we promoted by re-tagging main builds, now older than the cutoff). I read through the action's source at the pinned SHA to understand what would happen in a real run: when `keep-n-tagged` is set, deletions go through a phase that removes whole versions by digest, without checking for other tags. There *is* an untag -> reload -> recompute sequence designed to protect these versions, but it relies on read-after-write consistency between the registry and the Packages API and dry-run mode can't exercise that path at all, so we'd have no way to validate it before real deletions start running nightly right next to our release images. Since promoting releases by re-tagging main builds is normal practice in this repo (slinky-drainer's `v1.16.0` is another example), I'm not comfortable taking that risk.

So two asks, and apologies for moving the goalposts after saying no change was needed on this step:

**1. Could you replace the `dataaxiom` step with an `actions/github-script` step that does the same cleanup via the Packages API?** `actions/github-script` is GitHub-created (so no allowlist concerns) and its octokit client retries on rate limits out of the box. Please pin it to a full commit SHA like the other actions in the workflow. The step should:

- List all versions of the package (`${{ matrix.package }}`) using the org package-versions API, paginated, with the workflow's `GITHUB_TOKEN`.
- Select only versions where **every** tag starts with `main-` as that way a version carrying a release tag (or any other tag) alongside a `main-*` alias can never be touched, which is the safety property the dataaxiom step couldn't give us.
- From that set, keep anything newer than the `cut_off` input (same shared input as the snok step; treating the leading number as days is fine), sort the expired ones newest-first, keep the 10 most recent as a safety buffer, and delete the rest via the delete-package-version-for-org endpoint.
- Honor the same dry-run expression as the snok step: when dry-run is on, only log what would be deleted (id, updated date, and tags per version) instead of calling the delete endpoint.

I've already validated exactly this logic across the full matrix at full parallelism ([run 30342041763](https://github.com/NVIDIA/NVSentinel/actions/runs/30342041763)): 41/42 green, 10,548 stale `main-*` versions staged, zero release tags anywhere, and no `max-parallel` needed since octokit absorbed the same rate-limit conditions. Happy to share the exact step I tested if that saves you time.

**2. The slinky fix from my last comment still applies:** `nvsentinel/plugins-slinky-drainer` -> `nvsentinel/slinky-drainer` (the package doesn't exist under the current name, which is the one red job in every validation run).

One optional extra if you're up for it: the same step can also clean up orphaned attestation referrers (versions whose only tags are `sha256-<digest>` pointing at images that no longer exist) but that's a nice-to-have we can also handle in a follow-up, so don't let it hold up the two changes above.

Once you've pushed both changes, I'll re-run the dry run against your branch, post the log here, and approve. Thanks so much for your patience, this PR ended up being even more valuable than it first looked, since the snok bump also fixes our nightly cleanup, which has been silently failing.

PR #1376, opened 2026-06-09
feat(janitor): add ExternalRemediationRequest CRD foundation

jtschelling:
## Summary

First in a series of changes that introduces the ExternalRemediationRequest (ERR) CRD — a coordination surface that lets NVSentinel hand off node ownership to an external breakfix system. This PR ships the **API shape only** so apiserver accepts ERR objects but nothing observes them yet. The reconciler, fault-remediation producer, and node-labeler / monitor gating land in follow-up PRs.

## What's included

- **Proto + CRD generation** — new `data-models/protobufs/external_remediation.proto` and Makefile target that generates the CRD YAML into the janitor Helm chart (the ERR controller will ship with janitor).
- **Scheme registration** — separate `SchemeBuilder` for the `nvsentinel.nvidia.com` API group registered in `janitor/main.go`.
- **protojson MarshalJSON/UnmarshalJSON** — Spec and Status carry proto well-known types (notably `Timestamp` on `Condition.lastTransitionTime`). Without custom JSON serialization the default `encoding/json` marshals these as `{seconds, nanos}` objects and the apiserver rejects status PATCHes with HTTP 422. An envtest spec verifies the full apiserver round-trip.
- **RBAC** — janitor ClusterRole adds `get/list/watch/update/patch` on `externalremediationrequests` plus the `status` and `finalizers` subresources, ready for the reconciler.
- **`commons/pkg/managed` package** — centralises the `nvsentinel.dgxc.nvidia.com/managed` Node label (the opt-out signal an external system flips when it takes a node), plus ERR identity constants (`ERRApiGroup`, `ERRKind`, `ReleaseTaintKey`, `IsERRResource`). One source of truth for the strings every downstream code path needs to import.

## What's NOT included (lands in follow-ups)

- Controller reconciling ERR objects (apply release taint, observe completion, finalize cleanup)
- fault-remediation creating ERRs for the `EXTERNAL_REMEDIATION` action
- node-labeler and cluster-scope monitors backing off when `managed=false`
- Observability (metrics, events, OTEL tracing)

## Test Plan

- [x] `make test` passes in `commons/` and `janitor/`
- [x] envtest spec in `janitor/api/v1alpha1/external_remediation_envtest_test.go` exercises the JSON round-trip with a `Condition.LastTransitionTime` set
- [x] Hand-applied ERR YAML accepted by the apiserver on a local kind cluster
- [x] Janitor manager binary boots cleanly with the new scheme registration

## Summary by CodeRabbit

* **New Features**
  * Added ExternalRemediationRequest CRD, API types, and protobuf schema for external remediation workflows
  * Custom JSON marshal/unmarshal with RFC3339 timestamps and omission of nil spec/status
  * Cluster-wide "managed" node label with explicit opt-out behavior
  * Extended RBAC to allow ExternalRemediationRequest operations (including status and finalizers)

* **Chores**
  * Build updates to generate the new CRD and protobuf artifacts; updated module requirements

* **Tests**
  * Added unit and envtest coverage for JSON round-trips, deep-copy/scheme registration, and node opt-out behavior


commons/pkg/managed/external_remediation.go
lalitadithya, line 1: nit: can we all this file `external-remediation-request`? err is typically used for errors in golang

commons/pkg/managed/err.go
```diff
@@ -0,0 +1,47 @@
...
+//     release taint during reconciliation.
+//   - fault-remediation/crstatus: detects ERR resources to apply the asymmetric
+//     completion semantics in StatusChecker.
+//   - fault-remediation/remediation: constructs ERR CRs from the
+//     EXTERNAL_REMEDIATION action template.
+//
+// Keeping the strings in one place prevents drift between the apply side and
+// the observe side — a typo would silently break the dispatcher's predicate.
+const (
+	ERRApiGroup = "nvsentinel.nvidia.com"
```
lalitadithya: maybe this is a preference but can we use the full names err seems to be to close to error

data-models/protobufs/external_remediation.proto
```diff
@@ -0,0 +1,99 @@
...
+// design including the asymmetric ExternalRemediationComplete=True vs False
+// handling, the cleanup finalizer, the release taint, and the
+// nvsentinel.dgxc.nvidia.com/managed label coordination mechanism.
+message ExternalRemediationRequest {
+  option (protoc_gen_crd.k8s_crd) = {
+    api_group: "nvsentinel.nvidia.com",
+    kind: "ExternalRemediationRequest",
+    plural: "externalremediationrequests",
+    singular: "externalremediationrequest",
+    short_names: ["err"],
```
lalitadithya: would it be possible to get a different short name here, to me err sounds too similar to error.

janitor/api/v1alpha1/external_remediation_register.go
```diff
@@ -0,0 +1,173 @@
...
+type ExternalRemediationRequest struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec and Status are pointer fields because the underlying proto types
+	// embed protoimpl.MessageState (which contains a sync.Mutex). Copying a
+	// proto message by value would trip go vet's copylocks analyzer. Pointer
+	// fields also match the CRD schema's `nullable: true` for spec and status
+	// produced by protoc-gen-crd.
+	Spec   *protos.ExternalRemediationRequestSpec   `json:"spec,omitempty"`
```
natherz97, line 84: An alternative would be to embed the ExternalRemediationRequest proto Go struct directly in this v1alpha1.ExternalRemediationRequest struct (like what we're already doing for metav1.TypeMeta  + metav1.ObjectMeta. I think either is fine if we only need to extract 2 fields but it might prevent some confusion that clients should not use protos.ExternalRemediationRequest directly.

commons/pkg/managed/managed.go
```diff
@@ -0,0 +1,92 @@
...
+		if apierrors.IsNotFound(err) {
+			// Stale or empty cache for this node — fail open. No log spam: this
+			// is expected during informer warmup and after node deletion.
+			return false
+		}
+
+		slog.ErrorContext(ctx, "managed-label lookup failed; falling back to managed (fail-open)",
+			"node", nodeName, "error", err)
+
+		return false
```
natherz97: Should we return an error so that the client can decide the appropriate action? It seems like the node-labeler should decide to not modify node labels if it cannot determine the node state since if we failed open when a node was under ERR, we'd re-deploy the health-monitors with an active ERR.

commons/pkg/managed/external_remediation.go
```diff
@@ -0,0 +1,47 @@
...
+	ExtRRApiGroup = "nvsentinel.dgxc.nvidia.com"
+	ExtRRVersion  = "v1"
+	ExtRRKind     = "ExternalRemediationRequest"
+
+	// ReleaseTaintKey is the taint the ExtRR reconciler applies during its
+	// apply path to release NVSentinel's ownership of the node. The taint
+	// value carries the ExtRR CR's name so the cleanup path can find and
+	// remove only its own taint, even if other controllers apply taints to
+	// the same node.
+	ReleaseTaintKey = "nvsentinel.dgxc.nvidia.com/external-remediation"
```
natherz97: Nit: only the ERR controller will need the taint name right, is there a reason that it's in commons vs. Janitor package?

commons/pkg/managed/managed.go
jtschelling: yep good catch, implemented

commons/pkg/managed/external_remediation.go
jtschelling: removed, we dont need the taint imported anywhere. the other fields will be used by janitor and fault-remediation module

janitor/api/v1alpha1/external_remediation_register.go
jtschelling: you mean like

```go
type ExternalRemediationRequest struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
    protos.ExternalRemediationRequest
```

? personally i think the way it is in the PR right now is a bit cleaner, only because the client usage looks more like other existing client-go objects

```go
import (
    nvsentinelv1 "github.com/nvidia/nvsentinel/janitor/api/v1alpha1"
    "github.com/nvidia/nvsentinel/data-models/pkg/protos"
)

extrr := &nvsentinelv1.ExternalRemediationRequest{
    ObjectMeta: metav1.ObjectMeta{Name: "extrr-node-5-abc123"},
    Spec: &protos.ExternalRemediationRequestSpec{
        HealthEvent: &protos.HealthEvent{
            NodeName: "node-5",
            Id:       "abc123",
        },
    },
}
client.Create(ctx, extrr)
```

jtschelling: i might not be understanding your comment fully tho if you expand a bit more

natherz97 (approved):

PR #1378, opened 2026-06-10
docs: add ADR for tracking hardware device-count

XRFXLP:
## Tl;dr

Make changes in labeler to digest the following config:
```yaml
labeler:
  expectedInventory:
    enabled: true
    classes:
      - name: gpu
        labels:
          current: nvsentinel.dgxc.nvidia.com/gpu.count.current
          expected: nvsentinel.dgxc.nvidia.com/gpu.count.expected
        groupingLabels:
          - node.kubernetes.io/instance-type
          - karpenter.sh/nodepool
          - nvidia.com/gpu.product
          - nvidia.com/gpu.sharing-strategy
        currentExpression: |
          int(node.metadata.labels['nvidia.com/gpu.count'])
      - name: nic.roce
        labels:
          current: nvsentinel.dgxc.nvidia.com/nic.roce.count.current
          expected: nvsentinel.dgxc.nvidia.com/nic.roce.count.expected
        groupingLabels:
          - node.kubernetes.io/instance-type
          - karpenter.sh/nodepool
        currentExpression: |
          sum(resourceSlices
            .filter(rs,
              has(rs.spec.driver) &&
              rs.spec.driver == 'dra.networking.k8s.aws' &&
              has(rs.spec.devices)
            )
            .map(rs, rs.spec.devices
              .filter(d,
                has(d.attributes) &&
                'dra.vpc.amazonaws.com/deviceType' in d.attributes &&
                has(d.attributes['dra.vpc.amazonaws.com/deviceType'].string) &&
                d.attributes['dra.vpc.amazonaws.com/deviceType'].string == 'roce'
              )
              .size()
            ))
```

So it can label a node for a resource current and expected value, irrespective of the whether is NIC or GPU or it is advertized by device-plugin model or DRA model. 

This information can be consumed by kubernetes object monitor that can track the history of the node and decide whether to reboot or replace the node.

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added ADR describing normalization of device inventory into node labels carrying both current and expected counts, partitioned by hardware-class grouping, with rules for label validity, evaluation failure behavior, and expected-count learning/relearn semantics.
  * Documents monitoring behavior that raises health events when current < expected, and adds observability guidance for counts and label update outcomes.


docs/designs/043-expected-device-inventory-labels.md
```diff
@@ -0,0 +1,294 @@
...
+- a CEL expression that returns the current count for the node.
+
+The CEL environment is intentionally small and read-only. `labeler` builds the input context and the expression only counts from that context:
+
+- `node`: the Kubernetes `Node` object being reconciled;
+- `resourceSlices`: all raw `ResourceSlice` objects associated with that node.
+
+Inventory class expressions can use either legacy node labels or DRA inventory:
+
+- device-plugin deployments can read labels such as `nvidia.com/gpu.count` and `nvidia.com/gpu.product` from `node`;
```
lalitadithya: I don't follow this, why would the device plugin read those labels?

docs/designs/043-expected-device-count-labels.md
```diff
@@ -0,0 +1,294 @@
...
+
+- an admin removing the expected-count label from a node or hardware class;
+- an explicit labeler command or configuration flag to relearn a class;
+- a future configuration object that declares expected inventory per class.
+
+Cold start behavior should be conservative. If no expected label exists and only one node exists in a class, `labeler` may initialize expected from current, but that provides no protection against a device that was already missing before NVSentinel started. The ADR accepts this limitation for auto-learned baselines and leaves admin-provided expected inventory as a future hardening path.
+
+## Consumer Policies
+
+Kubernetes Object Monitor consumes the normalized labels with Node policies. The replace policy uses the existing fault-quarantine `quarantineHealthEvent` annotation as evidence that the restart event was already processed, instead of depending on node conditions written by the platform connector.
```
lalitadithya: would it be possible to use the nvsentinel state label here instead?

docs/designs/043-expected-device-count-labels.md
```diff
@@ -0,0 +1,294 @@
...
+          int(resource.metadata.labels['nvsentinel.dgxc.nvidia.com/gpu.count.current']) <
+          int(resource.metadata.labels['nvsentinel.dgxc.nvidia.com/gpu.count.expected'])
+      healthEvent:
+        componentClass: Node
+        isFatal: true
+        message: "GPU inventory is below expected count"
+        recommendedAction: RESTART_VM
+        errorCode:
+          - LESS_THAN_TOTAL_GPUS_RESTART
+    - name: LessThanTotalGPUsReplace
```
lalitadithya: Is this intended to be an escalation of `LessThanTotalGPUsRestart`? If so can we use the node condition that it would triggered?

If that is not the case and it is relying on say slow DCGM startup then it is possible that we can have a extra reboot inbetween right?

docs/designs/043-expected-device-count-labels.md
```diff
@@ -0,0 +1,294 @@
...
+
+Inventory class expressions can use either legacy node labels or DRA inventory:
+
+- device-plugin deployments can read labels such as `nvidia.com/gpu.count` and `nvidia.com/gpu.product` from `node`;
+- DRA deployments can count matching `ResourceSlice.spec.devices[]` entries from `resourceSlices`;
+- DRA device semantics are driver-specific, so the expression is responsible for matching the right driver and device attributes for that inventory class.
+
+The expression must return an integer. If it errors or returns a non-integer, `labeler` skips the current-count update for that class and records the evaluation failure. The first implementation will support standard CEL list macros such as `filter`, `map`, and `size`, and will register a `sum(list<int>) -> int` helper so DRA counts can be expressed across multiple `ResourceSlice` objects.
+
+Example shape:
```
lalitadithya, line 45: Can you add an example of how an admin could hardcode this value if they so desire using the `expectedInventory`?

docs/designs/043-expected-device-inventory-labels.md
```diff
@@ -0,0 +1,294 @@
+# ADR-043: Labeler - Expected Device Inventory Labels
```
lalitadithya: I think we should call this device count labels. inventory is a rather broad team and it can often be confused to mean many other things

docs/designs/043-expected-device-count-labels.md
```diff
@@ -0,0 +1,294 @@
...
+
+```yaml
+labeler:
+  expectedInventory:
+    enabled: true
+    classes:
+      - name: gpu
```
lalitadithya, line 52: can we also have an option to turn off policies at an individual class level so that we can easily contain a expensive cel expression for example?

docs/designs/043-expected-device-count-labels.md
```diff
@@ -0,0 +1,294 @@
...
+          - karpenter.sh/nodepool
+          - nvidia.com/gpu.product
+          - nvidia.com/gpu.sharing-strategy
+        currentExpression: |
+          int(node.metadata.labels['nvidia.com/gpu.count'])
+      - name: nic.roce
+        labels:
+          current: nvsentinel.dgxc.nvidia.com/nic.roce.count.current
+          expected: nvsentinel.dgxc.nvidia.com/nic.roce.count.expected
+        groupingLabels:
```
lalitadithya, line 73: what if there are no grouping labels available? For example, some legacy environments may not have the `node.kubernetes.io/instance-type` label.

docs/designs/043-expected-device-count-labels.md
XRFXLP: Do you mean remediation-succeeded? That can be ambiguous as that can also mean reboot is stuck.

docs/designs/043-expected-device-count-labels.md
XRFXLP: We can tie this escalation to node conditions as well but the problem is node condition are like leaf level information since k8s-connector can be be enabled/disabled.

We have grace time of 1 hour after node becomes not ready to catch the slow DCGM startup.

docs/designs/043-expected-device-count-labels.md
XRFXLP: It will break if somehow we have heterogeneous cluster yet no way to figure out it from the labels.

docs/designs/043-expected-device-count-labels.md
XRFXLP: done

docs/designs/043-expected-device-count-labels.md
XRFXLP: added an example

docs/designs/043-expected-device-inventory-labels.md
XRFXLP: re-worded

docs/designs/043-expected-device-inventory-labels.md
XRFXLP: done

docs/designs/043-expected-device-count-labels.md
lalitadithya: yeah, but the same problem exists with `quarantineHealthEvent` too right? we have to check for that as well as the ready transition time like you are doing

docs/designs/043-expected-device-count-labels.md
lalitadithya: ah, good point on the k8s-connector enabled/disabled. 

maybe should we check for cordon state, i.e., wait for the node to be uncordoned and ready?

docs/designs/043-expected-device-count-labels.md
lalitadithya: of course, but in the case of some our of on-prem cluster we have no labels but they are heterogenous in the sense that there are only GPU and CPU nodes in there.

docs/designs/043-expected-device-count-labels.md
```diff
@@ -93,16 +101,17 @@ nvsentinel.dgxc.nvidia.com/nic.roce.count.expected
 ```
 
 - `current` is the count visible in the selected inventory source.
-- `expected` is the stable baseline for the node's hardware class.
-- `expected` may rise automatically when a higher count is observed in the same class, but it must not fall automatically when `current` drops.
+- `expected` is either an operator-provided override for the matching hardware class or the learned baseline for that class.
+- Learned `expected` may rise automatically when a higher count is observed in the same class, but it must not fall automatically when `current` drops.
+- Disabled classes are not compiled, evaluated, or written to node labels. This allows expensive or environment-specific CEL expressions to be contained without disabling the entire expected-inventory feature.
```
lalitadithya: nit:

> expected-inventory

device count. please update in other places if required

docs/designs/043-expected-device-count-labels.md
XRFXLP: IIRC we always have labels in those cases in terms of node-role or nodeGroup.

docs/designs/043-expected-device-count-labels.md
XRFXLP: Right. I've removed this section entirely, we can always enable escalation based on node condition for the environment where we've k8s-connector enabled.

docs/designs/043-expected-device-count-labels.md
XRFXLP: > maybe should we check for cordon state, i.e., wait for the node to be uncordoned and ready?

No, that would be problematic as well as k8s object monitor will never send healthy even unless expected_device_count == actual_device_count which won't happen in unhealthy case.

I've removed this escalation point as there is no simple way to do it.

docs/designs/043-expected-device-count-labels.md
XRFXLP: done

docs/designs/043-expected-device-count-labels.md
```diff
@@ -0,0 +1,274 @@
+# ADR-043: Labeler - Expected Device Count Labels
+
+## Context
+
+NVSentinel needs a way to detect missing GPUs and NICs even when the kernel or driver does not emit a useful XID, SXID, or fallen-off-bus syslog line. A common failure mode is that a node reports fewer devices than its platform is expected to have; if no error log is produced, Kubernetes Object Monitor needs a normalized Kubernetes signal to consume.
+
+Today the expected device count is implicit. Operators may know that a class of nodes should have eight GPUs or eight RoCE interfaces, but NVSentinel does not persist that expectation in a normalized form that Kubernetes Object Monitor policies can consume.
```
deesharma24: is this LessThanTotalNic problem specific to OCI is the reason we are mentioning nics.roce everywhere in this doc?

deesharma24: I see you have mentioned AWS in different part of the problem. So, I am assuming this problem is there in Every CSP.right? if yes,then why we are calling it as ROCE?

XRFXLP: Everything is config driven and that was just an example that I picked.

XRFXLP: replaced roce with NICs wherever applicable.

lalitadithya (approved):

PR #1380, opened 2026-06-10
docs(design): ADR-050 — system-services-monitor scope

dmvevents:
## Summary

Design doc only (PR 1 of 5 from the #891 split, requested by @XRFXLP): `docs/designs/050-system-services-monitor-scope.md` — ADR-050, the scope contract the implementation (#1382), CI (#1381), Helm chart (#1385), and demo (#1383) PRs build against. The ADR resolves the two architectural questions raised in #891 review: (1) PCIe / NVLink / clock-throttling signals stay in `gpu-health-monitor` via the existing `pydcgm` path (not reimplemented); (2) the monitor (renamed from fabric-manager-monitor to system-services-monitor) is scoped to non-DCGM host-service health — FM systemd state, responsiveness, fabric-state-stuck detection, and GPU service lifecycle — with the signal-ownership table and event taxonomy (FM_DOWN / FM_UNRESPONSIVE / FM_FLAPPING) in the doc.

## Type of Change
- [x] 📚 Documentation

## Component(s) Affected
- [x] Documentation/CI

## Testing
- [x] Tests pass locally (docs-only change; no code paths)
- [x] Manual testing completed (cross-references verified against `docs/designs/` numbering at head — 050 is the next free number)
- [x] No breaking changes (or documented)

## Checklist
- [x] Self-review completed
- [x] Documentation updated (if needed)
- [x] Ready for review

## Series (from the #891 split)
This PR (ADR-050) · #1382 (implementation + tests) · #1381 (CI matrices) · #1385 (Helm chart) · #1383 (demo). Open scope discussion: the DCGM-4.5.2 coverage question from @deesharma24 is answered in the comments below, including an offer to drop the now-DCGM-duplicated `fabric_state_check.py` from #1382.

## Summary by CodeRabbit

* **Documentation**
  * Documented node-local monitoring through NPD custom plugins and opt-in Kubernetes Operator Monitoring policies.
  * Defined monitoring for Fabric Manager liveness, flapping, installation presence, and configurable GPU-support services.
  * Added host-service and DaemonSet deployment requirements for flap-state storage, including permissions, atomic updates, and invalid-state handling.
  * Clarified bounded probe behavior, initial-status skipping, permanent-condition defaults, unknown-state handling, restart tracking, and recovery behavior.

PR #1381, opened 2026-06-10
ci: register system-services-monitor in build/lint/publish matrices

dmvevents:
## Summary

CI wiring for system-services-monitor (PR 3 of 5 from the #891 split; scope in ADR-050, #1380): a `container-build-test.yml` matrix row, a `lint-test.yml` registration, and a minimal `health-monitors/system-services-monitor/Makefile` so the implementation PR's (#1382) first CI run hits a real matrix entry instead of failing on a missing module.

## Type of Change
- [x] 🔨 Build/CI

## Component(s) Affected
- [x] Documentation/CI

## Testing
- [x] Tests pass locally
- [x] Manual testing completed (workflow YAML validated; Makefile targets runnable)
- [x] No breaking changes (or documented)

## Checklist
- [x] Self-review completed
- [x] Documentation updated (if needed)
- [x] Ready for review

## Open sequencing question for @XRFXLP
This PR's diff is also carried inside #1382 (it needs the matrix entries to go green). Two clean ways to land the pair — your call, since the split was your request: (a) land this PR first and I rebase #1382 to drop the duplicated files, or (b) close this PR and let #1382 carry its own CI wiring. Either is a small change on my side; I'll follow whichever you prefer.

## Series (from the #891 split)
#1380 (ADR-050, docs) · #1382 (implementation + tests) · this PR (CI matrices) · #1385 (Helm chart) · #1383 (demo).

## Summary by CodeRabbit

* **Chores**
  * Extended automated validation to include the system services health monitoring component, including lint checks and test coverage reporting.
  * Added automated Docker image build and publishing support for the component.
  * Added build tooling with commands for validation, image creation, publishing, and usage guidance.

PR #1382, opened 2026-06-10
docs(npd): add GPU system-service checks — plugin scripts, reference configs, and tutorial

dmvevents:
### What does this PR do?

Ships the deliverables of [ADR-050](https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/050-gpu-system-services-npd-checks.md) (merged in #1380), replacing this PR's previous in-tree monitor with the NPD/KOM path the ADR settled on:

- **`docs/tutorials/assets/npd-gpu-services/`** — four NPD custom-plugin scripts (Fabric Manager liveness with consecutive-probe debounce and transitional-state handling; FM crash-loop detection with `systemctl reset-failed` disambiguation; FM unit presence for required fleets; parameterized GPU-service liveness) and one single-condition `CustomPluginMonitor` configuration per check.
- **`docs/tutorials/monitoring-gpu-system-services.md`** — operator tutorial in the `integrating-node-problem-detector.md` format: host-service and DaemonSet installation (including the `/var/run/nvsentinel/npd` per-check state mount), KOM policy configuration, a safe validation ladder, recovery-semantics caveats, troubleshooting, and the one-shot AI prompt appendix.
- **ADR rename** to `050-gpu-system-services-npd-checks.md` per review request on #1380 (the doc no longer describes an in-tree monitor; no inbound references existed).
- **Withdraws** the Python monitor module, its CI matrix entry, and its root protos registration.

The opt-in KOM values overlay (`values-npd-gpu-services.yaml`) ships with the companion chart change in #1385; merging that first keeps the tutorial's file references live from day one.

### Semantics mapping (reviewed behavior → new home)

| Reviewed in the withdrawn module | Now lives in |
| --- | --- |
| fm-presence tri-state (`required`/`auto`/`disabled`) | Configuration presence per the ADR's Configuration matrix (presence config only on required, host-systemd-managed fleets) |
| `NRestarts` reset-vs-restart gating (`ExecMainStartTimestamp`) | `check_fm_flapping.sh` (validated: pure flush records nothing; flush-masked restart records one) |
| Per-service instance identity (cache key) | One condition type + one monitor configuration per service |
| UNKNOWN ≠ DOWN (probe failure) | Scripts hold last confirmed state through probe failures; internal 8s probe bound under the 12s rule timeout |
| `FabricManagerNotInstalled` → `CONTACT_SUPPORT` override | `NPDFabricManagerNotInstalled` KOM policy (in #1385) |

### Testing

- `shellcheck -S warning` clean on all four scripts; `jq` validation on all four configurations.
- Behavioral matrix exercised against live systemd (and a stubbed `systemctl` for failure paths): debounce ladder (fires on the 3rd consecutive non-running probe; transitional states never fire), hold-fault through probe failures with recovery only on a confirming probe, healthy-hold TTL, `reset-failed` disambiguation, corrupt-state re-baseline.
- NPD itself is never deployed by NVSentinel (ADR-053), so live NPD execution is validated per the tutorial's walkthrough rather than repo CI.

## Summary by CodeRabbit

- **New Features**
  - Added monitoring checks for NVIDIA Fabric Manager presence, liveness, and crash-loop flapping.
  - Added liveness monitoring for the NVIDIA persistence daemon.
  - Added Node Problem Detector configurations for reporting GPU service health conditions.

- **Documentation**
  - Added a comprehensive tutorial covering installation, configuration, verification, recovery, troubleshooting, and optional policy integration for GPU system-service monitoring.


health-monitors/system-services-monitor/system_services_monitor/checkers/watcher.py
```diff
@@ -0,0 +1,238 @@
...
+    ) -> None:
+        self._poll_interval = poll_interval
+        self._callbacks = callbacks
+        self._node_name = node_name
+        self._boot_grace_period = boot_grace_period
+        self._start_time = time.monotonic()
+        self._callback_thread_pool = ThreadPoolExecutor()
+
+        # Initialize checkers and build the check list based on enabled flags
+        self._checkers: List[tuple[str, callable]] = []
```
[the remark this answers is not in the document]
dmvevents: Fixed in d5e0c57 — the registry is now typed `List[tuple[str, Callable[[], List[CheckResult]]]]` (both registered checkers return `List[CheckResult]`), with `Callable` imported from `typing`.

health-monitors/system-services-monitor/system_services_monitor/platform_connector/event_processor.py
```diff
@@ -0,0 +1,190 @@
...
+                stub = platformconnector_pb2_grpc.PlatformConnectorStub(chan)
+                try:
+                    stub.HealthEventOccurredV1(platformconnector_pb2.HealthEvents(events=health_events, version=1))
+                    metrics.events_sent_success.inc()
+                    return True
+                except grpc.RpcError as e:
+                    log.error(f"Failed to send health event to UDS: {e}")
+                    sleep(delay)
+                    delay = min(delay * 1.5, MAX_DELAY)
+                    continue
```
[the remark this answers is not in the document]
dmvevents: Fixed in d5e0c57 — added a per-attempt deadline via a new `GRPC_SEND_TIMEOUT_SECS = 10` constant next to the retry constants. 10s per attempt keeps the worst case (5 attempts + backoff) bounded well under the poll interval, and a hung platform-connector now surfaces as `DEADLINE_EXCEEDED`, which the existing `except grpc.RpcError` retry path already handles. Full module suite: 44 passed.

health-monitors/system-services-monitor/system_services_monitor/protos/health_event_pb2.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
+# NO CHECKED-IN PROTOBUF GENCODE
+# source: health_event.proto
+# Protobuf Python Version: 6.31.1
+"""Generated protocol buffer code."""
```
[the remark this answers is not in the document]
dmvevents: Intentionally not adding headers here: the repo's own license lint exempts generated protobuf bindings — the root Makefile's `license-headers-lint` target runs addlicense with `-ignore '**/*pb2*'`, and the checked-in `gpu-health-monitor` bindings on `main` carry no header either. Adding a header-injection step to just this module would diverge from the sibling monitors. Happy to follow up if maintainers change the lint policy repo-wide.

health-monitors/system-services-monitor/system_services_monitor/checkers/service_check.py
```diff
@@ -0,0 +1,273 @@
...
+                "show",
+                service_name,
+                "--property=NRestarts",
+            ])
+            if result.returncode == 0 and result.stdout.strip():
+                _, _, val = result.stdout.strip().partition("=")
+                return int(val)
+        except Exception:
+            pass
+        return 0
```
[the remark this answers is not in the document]
dmvevents: Fixed in d5e0c57 — the fallback now logs at debug level (`NRestarts query failed for %s (likely unsupported): %s`) so an nsenter/systemctl breakage is distinguishable from an older systemd without NRestarts, while keeping the intended silent-fallback behavior.

health-monitors/system-services-monitor/system_services_monitor/tests/test_cli.py
```diff
@@ -0,0 +1,138 @@
...
+        """An explicit CLI flag wins over the corresponding env var."""
+        with _mock_runtime() as (watcher, processor):
+            result = runner.invoke(
+                climod.cli,
+                ["--platform-connector-socket", "/run/from-cli.sock", "--node-name", "node-a"],
+                env={"PLATFORM_CONNECTOR_SOCKET": "/run/from-env.sock"},
+            )
+
+        assert result.exit_code == 0
+        assert processor.call_args.kwargs["socket_path"] == "/run/from-cli.sock"
```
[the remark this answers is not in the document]
dmvevents: Fixed in d5e0c57 — renamed to `_watcher` per RUF059.

dmvevents:
This PR stack (#1380–#1385) has been refreshed today:

- all six branches are updated to current `main` (none are behind anymore) and report MERGEABLE
- every open CodeRabbit thread across the stack is now resolved with either a fix commit or an evidence-backed rationale: 5 threads here (fixes in d5e0c57), 8 on the ADR (#1380, doc 179212f + matching implementation 12f6c07 here), 3 on #1384 (fixes in 7074580)
- module test suite: 53 passed (9 new tests added for the review-driven behavior changes)

Per copy-pr-bot, workflows have not yet run on NVIDIA's runners for any PR in this stack — only the DCO check has ever executed. Could a maintainer authorize CI validation for the stack (and take a first review pass on the ADR in #1380)? Happy to split, squash, or re-scope if that makes review easier.

health-monitors/system-services-monitor/system_services_monitor/checkers/service_check.py
```diff
@@ -0,0 +1,275 @@
...
+        and ExecMainStartTimestamp. NRestarts is queried separately since
+        older systemd versions don't support it in a combined property list.
+        """
+        try:
+            result = self._run_host_cmd([
+                "systemctl",
+                "show",
+                service_name,
+                "--property=ActiveState,SubState,MainPID,ExecMainStartTimestamp",
+            ])
```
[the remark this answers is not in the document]
dmvevents: Fixed in 1368e49 — Black run over the whole package; this call site now uses the hugged form Black produces.

health-monitors/system-services-monitor/system_services_monitor/checkers/service_check.py
```diff
@@ -0,0 +1,275 @@
...
+            now = time.monotonic()
+            for _ in range(current_restarts - last_count):
+                self._restart_history[service_name].append(now)
+            self._last_restart_count[service_name] = current_restarts
+
+        # Prune entries outside the flap window
+        cutoff = time.monotonic() - self._flap_window
+        history = self._restart_history[service_name]
+        while history and history[0] < cutoff:
+            history.popleft()
```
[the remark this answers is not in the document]
dmvevents: Fixed in 1368e49. A decrease in NRestarts (reset-failed, unit re-creation, reboot) now re-baselines and records one restart event — the restart that produced the reset is itself an observation — so flap detection keeps counting instead of going quiet until the counter passes the old high-water mark. Probe failures are now distinct from a real 0: _get_restart_count returns None (unobserved), which leaves the baseline untouched. Covered by test_counter_reset_rebaselines_and_keeps_counting and test_unobserved_probe_leaves_baseline_untouched.

health-monitors/system-services-monitor/system_services_monitor/checkers/watcher.py
```diff
@@ -0,0 +1,249 @@
...
+        flap_window: int = 600,
+        flap_threshold: int = 3,
+        enable_fabric_check: bool = True,
+    ) -> None:
+        self._poll_interval = poll_interval
+        self._callbacks = callbacks
+        self._node_name = node_name
+        self._boot_grace_period = boot_grace_period
+        self._start_time = time.monotonic()
+        self._callback_thread_pool = ThreadPoolExecutor()
```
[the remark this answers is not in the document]
dmvevents: Fixed in 1368e49: the callback pool is now ThreadPoolExecutor(max_workers=1) — ordering, not throughput, is what this dispatch needs, so one worker turns the pool into a serial queue (submission order == execution order) while keeping callback latency out of the poll loop. A stale unhealthy result can no longer land after a newer healthy one.

health-monitors/system-services-monitor/system_services_monitor/checkers/fabric_state_check.py
```diff
@@ -0,0 +1,236 @@
...
+                                "gpu_index": str(gpu.gpu_index),
+                                "fabric_state": gpu.fabric_state,
+                                "fabric_status": gpu.fabric_status,
+                            },
+                        )
+                    )
+                    continue
+            else:
+                self._in_progress_streak.pop(gpu.gpu_index, None)
+
```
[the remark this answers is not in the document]
dmvevents: Fixed in 1368e49: streaks are now consecutive-by-observation — before evaluating a poll, any GPU absent from statuses has its streak dropped (a gap breaks the run), so an nvidia-smi timeout or partial output can no longer freeze a streak toward FM_REGISTRATION_STUCK. Covered by test_in_progress_streak_drops_when_gpu_absent_from_poll.

health-monitors/system-services-monitor/system_services_monitor/checkers/watcher.py
```diff
@@ -0,0 +1,276 @@
...
+                    f"state={gpu.fabric_state}, status={gpu.fabric_status}"
+                )
+
+        fabric_results = self._fabric_state_checker.to_check_results(statuses, self._node_name)
+        if self._in_grace_period():
+            # Same suppression the service checks apply during node startup:
+            # drop unhealthy results, keep healthy ones (they prime the
+            # transition cache without triggering remediation).
+            fabric_results = [r for r in fabric_results if r.is_healthy]
+        return fabric_results
```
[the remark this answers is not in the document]
dmvevents: Fixed in 1368e49: to_check_results grew an advance_streak flag and the watcher passes advance_streak=False during boot grace, so the streak cannot accrue while its alerts are suppressed — a slow normal startup no longer trips FM_REGISTRATION_STUCK on the first post-grace poll. Covered by test_streak_does_not_advance_when_disabled.

dmvevents:
Status update across the series, correcting my 2026-08-12 comment above, which aged badly on two claims:

- "none are behind anymore" — false as of this week (all six branches had drifted to 18 behind `main`). All six are now merged up to current `main` again as of today's pushes.
- "every open CodeRabbit thread across the stack is now resolved" — that accounting covered the round *before* the 08-12 pushes; CodeRabbit re-reviewed minutes after and that new round went unanswered, and #1383's full round was omitted entirely. As of today every open thread across #1380/#1382/#1385 has a reply citing a fix commit or a reasoned position.

What changed today, per PR:

- **#1380** — the ADR is renumbered **049 → 050** (`35cbf65`): 049 collided with `049-node-validation.md` on `main` (same mistake class as the original 030 collision — my error). Title/body updated; the four open design findings are fixed in-doc (streak reset semantics, enforceable socket-authorization invariants, the 50s-vs-26s retry-bound arithmetic — actual worst case ~66 s — and generation-safe success-path cache commits).
- **#1382** — the three correctness findings are fixed (`1368e49`): callback dispatch is ordered (single-worker executor), flap tracking survives systemd `NRestarts` resets and distinguishes unobserved probes from a real 0, and the registration streak drops for absent GPUs and does not accrue during boot grace. Black applied; docstring coverage is now 100% (was 66.67% vs the 80% gate). 57 unit tests pass.
- **#1384** — docstring coverage 25% → 100% (`3fea98e`).
- **#1385** — the `FabricManagerFlapping` alert is templated from `flapWindow`/`flapThreshold` and uses `>=`, matching the monitor's own threshold semantics (`234f432`); chart values documented. The body no longer says draft. The stale `needs-rebase` label predates the conflict fix — could a maintainer drop it?
- **#1381** — no findings; merged up to `main` only.
- **#1383** — not investing further pending direction: the demo forks #1382's prober and has drifted behind its fixes. Recommendation is to rebuild it as a thin wrapper over the shipped package (or fold the deployment story into the chart) rather than patch the fork.

The `copy-pr-bot` gate note from the earlier comment still stands: only DCO has ever executed on any PR in this series. A vetter authorization would let the lint/build/E2E matrices actually report.

dmvevents:
Thanks for updating the branch and vetting CI, @XRFXLP! Both real failures from that run are fixed in 7d642b98:

- **protos-lint**: root cause was a registration gap — `protos-clean` deletes `*_pb2*.py` repo-wide, but `protos-generate` never regenerated this module's vendored stubs, leaving the tree dirty. This module now has its own `protos-generate`/`protos-clean` targets (mirroring gpu-health-monitor's, including the vendored-crd.proto import rewrite) and is registered in the root `protos-generate`. The stubs were regenerated with the CI toolchain (grpcio-tools 1.75.1) and are byte-identical to gpu-health-monitor's committed copies; regeneration is idempotent and the module's 57 tests pass against them.
- **container-build-test**: poetry rejects the CI-injected `VERSION=pull-request-1382-e2c5ff1` ("Invalid PEP 440 version"). Adopted the same guard nccl-allreduce uses — `poetry version` only runs for semver-shaped VERSIONs. (gpu-health-monitor and dcgm-diag carry the unguarded line and would hit the same error on a PR-triggered container build of those modules — happy to file a small follow-up PR for those two if useful.)
- The **gomod-lint** red in that run was a fail-fast cancellation ("The operation was canceled."), not a gomod issue.

Could you re-vet the new head when you get a chance? `/ok to test 7d642b981c59551ebd1a37fc83b48086044e60c5`

dmvevents:
Reworked for the merged ADR-050 (3550a3ba), as sketched before the merge: the in-tree monitor is withdrawn and the PR now ships the ADR's actual deliverables — the four plugin scripts, one single-condition `CustomPluginMonitor` configuration per check, and the operator tutorial — plus the ADR rename @tanishagoyal2 requested (`050-gpu-system-services-npd-checks.md`). The PR body carries the full semantics mapping from the reviewed module behavior to its new home in the scripts/configs, and the testing notes (shellcheck + behavioral matrix against live systemd, including the reset-vs-restart and hold-state paths).

Companion: #1385 becomes the opt-in `values-npd-gpu-services.yaml` KOM overlay — merging it before this keeps the tutorial's file references live from day one. Ready for review; /ok to test whenever a maintainer gets a chance.

docs/tutorials/assets/npd-gpu-services/check_fm_active.sh
```diff
@@ -0,0 +1,161 @@
...
+  confirmed="none"; fail_count=0
+  save_state_and_exit 0 "${UNIT} is not present on this host; liveness check not applicable"
+fi
+
+case "${active_state}" in
+  active)
+    confirmed="healthy"; fail_count=0
+    save_state_and_exit 0 "${UNIT} is active"
+    ;;
+  inactive|failed)
```
deesharma24: The failure counter is not reset when the service is starting or when a check cannot read its state.

Please reset the pending counter when a check does not confirm inactive or failed. Keep an already-confirmed fault until recovery is confirmed. Apply this to check_gpu_service.sh too.

docs/tutorials/assets/npd-gpu-services/check_gpu_service.sh
```diff
@@ -0,0 +1,173 @@
...
+# reports as this script's deliberate exit, not an NPD plugin timeout.
+PROBE_TIMEOUT_SECONDS=8
+
+# --- state helpers -----------------------------------------------------------
+
+confirmed="none"
+fail_count=0
+probe_fail_count=0
+
+load_state() {
```
deesharma24, line 73: The FM liveness script and generic service script contain almost the same checking and state-saving code. Each fix therefore needs to be made twice.

Could we use the generic script for nvidia-fabricmanager too? Keep separate NPD conditions and state files so the checks remain independent.Both checks would depend on shared code, but fixes and tests would only need to be maintained once.

deesharma24:
@dmvevents Can you please take care of coderabbit comments as well .Will it be possible for you to test this on a cluster with NPD install to see whether it works smoothly or not?

docs/tutorials/monitoring-gpu-system-services.md
```diff
@@ -0,0 +1,452 @@
...
+
+---
+
+## 3. Configure NVSentinel and KOM policies
+
+The GPU-service policies are intentionally excluded from the default KOM
+values, for the same reason as ADR-053's: NVSentinel does not install NPD or
+control how an operator handles its conditions. The opt-in overlay
+`distros/kubernetes/nvsentinel/values-npd-gpu-services.yaml` provides them
+(shipped with the companion chart change; see the ADR's Configuration
```
tanishagoyal2: Where is this values-npd-gpu-services file? I couldn't find it at mentioned location

docs/tutorials/monitoring-gpu-system-services.md
```diff
@@ -0,0 +1,452 @@
...
+
+This integration watches these condition and reason pairs:
+
+| Condition type | Problem reason | KOM policy | Fatal | Action |
+| --- | --- | --- | --- | --- |
+| `FabricManagerDown` | `FabricManagerNotActive` | `NPDFabricManagerDown` | yes | `RESTART_BM` |
+| `FabricManagerFlapping` | `FabricManagerFlapping` | `NPDFabricManagerFlapping` | yes | `RESTART_BM` |
+| `FabricManagerNotInstalled` | `FabricManagerUnitNotFound` | `NPDFabricManagerNotInstalled` | yes | `CONTACT_SUPPORT` |
+| `NvidiaPersistencedDown` | `NvidiaPersistencedNotActive` | `NPDNvidiaPersistencedDown` | no | `CONTACT_SUPPORT` |
+
```
tanishagoyal2, line 79: Why are we considering NvidiaPersistencedDown as non-fatal? Isn't it a genuine error?

docs/tutorials/monitoring-gpu-system-services.md
```diff
@@ -0,0 +1,452 @@
...
+- **NPD restarts.** `skip_initial_status: true` prevents NPD from publishing
+  default healthy conditions at startup, but the first post-restart probe
+  batch can still briefly write a not-yet-probed condition at its default
+  `False` (the batch's completion skew — milliseconds when probes respond).
+  Do not treat a condition transition observed around an NPD restart as proof
+  of recovery, and do not restart NPD mid-remediation.
+- **`Unknown` conditions.** A probe that cannot observe systemd (D-Bus down,
+  timeout) sets its condition to `Unknown`, not `True`; a genuinely stopped
+  service re-reports `True` within one probe cycle. Treat `Unknown` as "could
+  not observe", never as recovery.
```
tanishagoyal2: True -> Unknown transition will make KOM trigger healthy event, right? How we will prevent false recovery in that case?

tanishagoyal2 (changes requested):

docs/tutorials/assets/npd-gpu-services/check_fm_active.sh
```diff
@@ -0,0 +1,161 @@
...
+    case "${key}" in
+      confirmed) confirmed="${value}" ;;
+      fail_count) fail_count="${value}" ;;
+      probe_fail_count) probe_fail_count="${value}" ;;
+    esac
+  done < "${STATE_FILE}"
+  # Invalid content means a fresh baseline: never hold phantom state.
+  if ! [[ "${confirmed}" =~ ^(healthy|unhealthy|none)$ &&
+          "${fail_count}" =~ ^[0-9]+$ &&
+          "${probe_fail_count}" =~ ^[0-9]+$ ]]; then
```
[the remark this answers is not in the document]
dmvevents: Adopted in ac2c2652 — all persisted counters now validate against `^(0|[1-9][0-9]*)$` (a `08` in a tampered/corrupted state file previously passed `^[0-9]+$` and would hit Bash octal arithmetic). Regression: a state file with `fail_count=08` now re-baselines cleanly instead of erroring.

docs/tutorials/assets/npd-gpu-services/check_fm_active.sh
```diff
@@ -0,0 +1,161 @@
...
+  if mkdir -p "${STATE_DIR}" 2>/dev/null && chmod 0700 "${STATE_DIR}" 2>/dev/null &&
+     tmp_file=$(mktemp "${STATE_DIR}/.fm-liveness.XXXXXX" 2>/dev/null); then
+    chmod 0600 "${tmp_file}"
+    {
+      echo "confirmed=${confirmed}"
+      echo "fail_count=${fail_count}"
+      echo "probe_fail_count=${probe_fail_count}"
+    } > "${tmp_file}"
+    mv -f "${tmp_file}" "${STATE_FILE}"
+  fi
```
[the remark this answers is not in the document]
dmvevents: Adopted in ac2c2652 with one deliberate carve-out: every commit step (mkdir/chmod/mktemp/write/mv) is now guarded and the temp file is removed on failure, but a persistence failure must never CLEAR a fault — an unhealthy result is still reported exit 1 (message notes "state persistence failed"), while healthy/unknown results become exit 3, since their debounce accounting could not be persisted. Verified via a stubbed persist_state: confirmed-fault+inactive → rc=1 with the suffix; active → rc=3. (Same treatment in check_fm_installed.sh and check_fm_flapping.sh per your sibling comment.)

docs/tutorials/assets/npd-gpu-services/check_fm_active.sh
```diff
@@ -0,0 +1,161 @@
...
+  probe_fail_count=$((probe_fail_count + 1))
+  if [[ "${confirmed}" == "unhealthy" ]]; then
+    save_state_and_exit 1 \
+      "${UNIT} holding fault: probe failing (${probe_fail_count} consecutive), last confirmed not active"
+  fi
+  if [[ "${confirmed}" == "healthy" && ${probe_fail_count} -lt ${PROBE_FAIL_MAX} ]]; then
+    save_state_and_exit 0 \
+      "${UNIT} holding healthy: probe failing (${probe_fail_count} consecutive)"
+  fi
+  save_state_and_exit 3 "could not observe ${UNIT}: systemctl unavailable (rc=${rc})"
```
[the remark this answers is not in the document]
dmvevents: Adopted in ac2c2652 — when the healthy hold expires (4th consecutive probe failure) the stale confirmation is now discarded (`confirmed=none`), so a post-outage non-running observation reports unknown until the debounce threshold confirms either state, instead of resurrecting a minutes-old healthy hold. Matrix: probe-fail ×4 → Unknown+degrade; following `inactive` → rc=3, not rc=0. Tutorial recovery-timing text aligned.

docs/tutorials/assets/npd-gpu-services/check_fm_flapping.sh
```diff
@@ -0,0 +1,196 @@
...
+done <<< "${show_output}"
+
+if [[ "${load_state}" == "not-found" ]]; then
+  echo "${UNIT} is not present on this host; flap check not applicable"
+  exit 0
+fi
+
+if ! [[ "${n_restarts}" =~ ^[0-9]+$ && "${exec_ts}" =~ ^[0-9]+$ ]]; then
+  echo "could not observe ${UNIT}: unparseable NRestarts/ExecMainStartTimestamp"
+  exit 3
```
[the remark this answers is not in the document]
dmvevents: Adopted in ac2c2652 — the unparseable-NRestarts/ExecMainStartTimestamp branch now routes through the same hold-last-result retention as probe failures (shared `hold_last_result_and_exit`): confirmed flapping holds exit 1 indefinitely, healthy holds through the failure threshold then degrades. Matrix: flapping window + `NRestarts=garbage` → rc=1 "holding fault", was rc=3.

docs/tutorials/assets/npd-gpu-services/check_fm_installed.sh
```diff
@@ -0,0 +1,98 @@
...
+save_state_and_exit() { # $1=exit code, $2=message
+  local rc="${1}" msg="${2}" tmp_file
+  if mkdir -p "${STATE_DIR}" 2>/dev/null && chmod 0700 "${STATE_DIR}" 2>/dev/null &&
+     tmp_file=$(mktemp "${STATE_DIR}/.fm-presence.XXXXXX" 2>/dev/null); then
+    chmod 0600 "${tmp_file}"
+    {
+      echo "confirmed=${confirmed}"
+      echo "probe_fail_count=${probe_fail_count}"
+    } > "${tmp_file}"
+    mv -f "${tmp_file}" "${STATE_FILE}"
```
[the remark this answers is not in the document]
dmvevents: Adopted in ac2c2652 together with the liveness-script fix (same guarded persist_state + fault-preservation rule). Your flap-specific point was the sharpest: an uncommitted baseline double-counts restart deltas next invocation, so a failed commit on the flap main path now reports unknown — unless the computed result is flapping, which is still reported unhealthy (never clear a fault on a commit failure).

docs/tutorials/monitoring-gpu-system-services.md
```diff
@@ -0,0 +1,452 @@
...
+- Downstream remediation enabled: [yes or no]
+
+Follow these requirements:
+
+1. Inspect before changing anything: locate the NPD installation and its
+   existing monitor configurations. Never deploy a second NPD instance.
+2. Install the plugin scripts from docs/tutorials/assets/npd-gpu-services/
+   at /etc/npd-plugins/ (0755) and the four custom-plugin-*.json
+   single-condition monitor configurations in the NPD configuration
+   directory. Include check_fm_installed.sh + custom-plugin-fm-presence.json
```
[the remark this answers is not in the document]
dmvevents: Adopted in ac2c2652 — the baseline expectation and the AI-prompt step now state it per platform: four conditions on required-FM fleets, three where the presence configuration is omitted.

docs/tutorials/assets/npd-gpu-services/check_fm_active.sh
dmvevents: Adopted in ac2c2652 — the consecutive-failure count now tracks strictly consecutive non-running observations: transitional states and probe failures reset it (while a confirmed fault is held until an `active` probe confirms recovery, unchanged). Matrix: inactive ×2 → activating → inactive now reads 1/3, not 3/3.

docs/tutorials/assets/npd-gpu-services/check_gpu_service.sh
dmvevents: Done in ac2c2652 — `check_fm_active.sh` is gone; the FM liveness rule invokes `check_gpu_service.sh` with `nvidia-fabricmanager` (`custom-plugin-fm-liveness.json` updated), keeping its own condition type and per-unit state file (`svc-nvidia-fabricmanager.state`) so the checks stay independent while the logic is maintained once. This round proved your point: every fix above would have been made twice. ADR and tutorial updated to match.

dmvevents:
All eight review comments are addressed in ac2c2652 (each thread has the specifics; the headline changes: the two liveness scripts are consolidated per your suggestion, state-file commits are fully guarded with a never-clear-a-fault rule, a stale healthy hold now degrades after the probe-failure TTL, and the flap check's unparseable-output branch holds the last result).

On cluster testing: yes. I'll run the tutorial's walkthrough against a real NPD install on a Kubernetes cluster — NPD DaemonSet with these four monitor configurations and scripts, validating that the conditions publish with the right reasons, debounce on unit stops, clear on recovery without an NPD restart, hold through an injected D-Bus outage, and that the flap condition fires on a systemd-driven crash loop and drains afterward. I'll post the transcript (commands + observed `kubectl get node` condition outputs) here when done. If you'd also like the KOM half exercised end-to-end (conditions → HealthEvents with these policies), I can follow up with that on a cluster running NVSentinel — say the word.

dmvevents:
Cluster test done, as promised — real NPD v1.36.0 against these exact scripts and the four single-condition monitor configurations. Environment: kind v1.31.0 (Debian 12 node with real systemd), NPD DaemonSet per the tutorial (hostPath state mount at `/var/run/nvsentinel/npd`), controllable `nvidia-fabricmanager`/`nvidia-persistenced` stand-in units (`sleep infinity`, `Restart=always`). Every phase behaved exactly per the ADR contracts:

| Phase | Action | Observed |
| --- | --- | --- |
| Baseline | NPD start | All 4 conditions `False` with the reference healthy reasons/messages; nothing published before first probes (`skip_initial_status`) |
| Liveness debounce | `systemctl stop nvidia-persistenced` | `NvidiaPersistencedDown=True reason=NvidiaPersistencedNotActive` after exactly **3 consecutive probes** (message carries the count) |
| Recovery | `systemctl start` | `False` on the **next probe** — no NPD restart needed (no latching) |
| Fatal path | stop/start `nvidia-fabricmanager` | `FabricManagerDown=True reason=FabricManagerNotActive` debounced; recovered next probe; **flap stayed `False` throughout** (manual stop ≠ crash loop) |
| Crash loop | `systemctl kill -s SIGKILL` ×3, ~15 s apart | `NRestarts=3` → `FabricManagerFlapping=True` ("restarted 3 times in the last 600s") **while `FabricManagerDown` stayed `False`** — flap reports independently of instantaneous liveness, as designed |
| Presence | `rm` unit file + `daemon-reload` | `FabricManagerNotInstalled=True reason=FabricManagerUnitNotFound`; liveness/flap went inert (not-found skip); restored unit → recovered next probe |
| **Hold-state** | break the probe (systemctl unreachable) while `NvidiaPersistencedDown=True` | Condition **stayed `True` through 3+ failed probes** — no `Unknown`/`False` flip, nothing for KOM to misread as recovery; restored probe with unit still down → stayed `True`; started unit → `False` on the confirming probe |
| State files | inspect `/var/run/nvsentinel/npd/` | `0700` dir / `0600` files, canonical contents (`confirmed=healthy`, counters zeroed after recovery) |

One environment note worth having on the record: this NPD image ships no `systemctl`, and an `nsenter -t 1 -m`-style shim hung in this environment — a `chroot /host systemctl` shim (node `/` mounted read-only at `/host`) worked robustly, and query verbs are chroot-safe. I'll add that concrete fallback to the tutorial's DaemonSet section in the next natural push, replacing the vaguer "ensure systemctl works" phrasing.

Full timestamped transcript:

<details>
<summary>walkthrough.log</summary>

```
[16:56:03] === ADR-050 NPD walkthrough — kind v1.31.0 node (Debian 12 systemd), NPD v1.36.0, 4 single-condition CustomPluginMonitor configs, chroot-systemctl shim ===
[16:56:03] --- A0. healthy baseline (all units active) ---
FabricManagerDown=False reason=FabricManagerActive msg=nvidia-fabricmanager is active
FabricManagerFlapping=False reason=FabricManagerStable msg=nvidia-fabricmanager restart rate is normal
FabricManagerNotInstalled=False reason=FabricManagerInstalled msg=nvidia-fabricmanager unit is present
NvidiaPersistencedDown=False reason=NvidiaPersistencedActive msg=nvidia-persistenced is active
[16:56:43] --- A1. systemctl stop nvidia-persistenced (non-fatal path) ---
[16:56:43] stopped at 16:56:43; debounce = 3 consecutive probes @30s
[16:57:48] after ~2 probes (expect still False/holding):
NvidiaPersistencedDown=True reason=NvidiaPersistencedNotActive msg=nvidia-persistenced is not active (state=inactive, sub-state=dead; 3 consecutive probes)
[16:58:23] --- A2. systemctl start nvidia-persistenced (recovery confirms on next probe, no NPD restart) ---
[16:59:03] after ~1 probe:
NvidiaPersistencedDown=False reason=NvidiaPersistencedActive msg=nvidia-persistenced is active
[16:59:03] --- B1. systemctl stop nvidia-fabricmanager (fatal path, debounced) ---
[17:00:43] after ~3 probes:
FabricManagerDown=True reason=FabricManagerNotActive msg=nvidia-fabricmanager is not active (state=inactive, sub-state=dead; 3 consecutive probes)
FabricManagerFlapping=False reason=FabricManagerStable msg=nvidia-fabricmanager restart rate is normal
[17:00:43] --- B2. systemctl start nvidia-fabricmanager (recovery) ---
[17:01:23] after ~1 probe:
FabricManagerDown=False reason=FabricManagerActive msg=nvidia-fabricmanager is active
[17:01:41] --- C1. crash loop: systemctl kill -s SIGKILL x3, ~15s apart (Restart=always; manual restarts would NOT count) ---
[17:01:41] kill #1
[17:01:56] kill #2
[17:02:11] kill #3
NRestarts=3
[17:03:06] after next flap probe (threshold 3 in 600s):
FabricManagerDown=False reason=FabricManagerActive msg=nvidia-fabricmanager is active
FabricManagerFlapping=True reason=FabricManagerFlapping msg=nvidia-fabricmanager restarted 3 times in the last 600s (threshold 3)
[17:03:06] --- D1. presence: rm unit file + daemon-reload (LoadState=not-found) ---
[17:04:16] after ~2 probes:
FabricManagerDown=False reason=FabricManagerActive msg=nvidia-fabricmanager is active
FabricManagerFlapping=False reason=FabricManagerStable msg=nvidia-fabricmanager restart rate is normal
FabricManagerNotInstalled=True reason=FabricManagerUnitNotFound msg=nvidia-fabricmanager unit is not installed on a host that requires Fabric Manager
[17:04:50] --- D2. restore unit file (presence recovers) ---
[17:05:30] after ~1 probe:
FabricManagerNotInstalled=False reason=FabricManagerInstalled msg=nvidia-fabricmanager unit is present
[17:05:30] --- E1. hold-state: stop persistenced, let the fault confirm ---
[17:07:10] fault confirmed:
NvidiaPersistencedDown=True reason=NvidiaPersistencedNotActive msg=nvidia-persistenced is not active (state=inactive, sub-state=dead; 3 consecutive probes)
[17:07:10] --- E2. break the probe (shim exits 1 = systemd unreachable) while fault is active ---
[17:08:50] after ~3 failed probes (condition must STAY True — fault held):
NvidiaPersistencedDown=True reason=NvidiaPersistencedNotActive msg=nvidia-persistenced is not active (state=inactive, sub-state=dead; 3 consecutive probes)
[17:08:50] --- E3. restore the probe; unit still stopped (stays True), then start it (recovers) ---
[17:09:30] probe restored, unit still down:
NvidiaPersistencedDown=True reason=NvidiaPersistencedNotActive msg=nvidia-persistenced is not active (state=inactive, sub-state=dead; 3 consecutive probes)
[17:10:10] unit started (confirming probe recovers):
NvidiaPersistencedDown=False reason=NvidiaPersistencedActive msg=nvidia-persistenced is active
[17:10:10] --- state files on the node ---
total 16
drwx------ 2 root root 120 Sep 17 17:09 .
drwxr-xr-x 3 root root  60 Sep 17 16:51 ..
-rw------- 1 root root 131 Sep 17 17:09 fm-flap.state
-rw------- 1 root root  37 Sep 17 17:09 fm-presence.state
-rw------- 1 root root  50 Sep 17 17:09 svc-nvidia-fabricmanager.state
-rw------- 1 root root  50 Sep 17 17:09 svc-nvidia-persistenced.state
confirmed=healthy
fail_count=0
probe_fail_count=0
```

</details>

Happy to follow up with the KOM half (conditions → HealthEvents with the #1385 policies) on a cluster running NVSentinel if wanted.

docs/tutorials/monitoring-gpu-system-services.md
dmvevents: Apologies for the lag on this one — it ships with the companion chart PR #1385 (`distros/kubernetes/nvsentinel/values-npd-gpu-services.yaml`, four KOM policies in the values-npd-remediation.yaml conventions). That is why the tutorial references it by path without a hyperlink: a relative link would be dead until #1385 merges, which is also why the PR bodies note the merge order — #1385 first keeps the reference live from day one.

docs/tutorials/monitoring-gpu-system-services.md
dmvevents: It's a genuine error worth surfacing — the question is whether it justifies quarantining and draining the node, and we judged not: persistenced's job is keeping the driver initialized between clients (initialization latency, persistence of clock/settings state). A node with active workloads and persistenced dead keeps computing correctly; the cost appears as slower next-job startup and reset settings after the last client exits. `isFatal: true` would trigger the full quarantine path for a restartable convenience daemon, which felt disproportionate against ADR-050's own restart-vs-return taxonomy — versus FM, whose death actively degrades NVLink error-recovery coordination on the fabric.

That said, it's one line in an opt-in overlay, per-fleet tunable — and if the team's operational judgment is that it should be fatal by default, I'll flip it happily; you have far more fleet history with persistenced failures than I do.

docs/tutorials/monitoring-gpu-system-services.md
dmvevents: Right — KOM's transition detector treats any non-matching observation, `Unknown` included, as the healthy edge. Two layers address it, split by who can emit the `Unknown`:

1. **The scripts never emit `Unknown` while a fault is confirmed** — they hold the last confirmed state through probe failures (systemd/D-Bus unreachable, timeouts, unparseable output, even failed state-file commits): a faulted condition keeps reporting `True` until a probe *confirms* recovery. This is now empirically demonstrated, not just designed: in [yesterday's cluster test](https://github.com/NVIDIA/NVSentinel/pull/1382#issuecomment-5718350950), with `NvidiaPersistencedDown=True` I broke the probe entirely for 3+ cycles — the condition stayed `True` with no `Unknown` flip, then cleared only on a confirming probe after the unit actually started.
2. **What the scripts cannot intercept** — an `Unknown` NPD itself generates without running them (plugin exec failure, script crash, outer-timeout kill) — is the named residual in the merged ADR's recovery semantics, and it's exactly what KOM-side three-state handling (`Unknown` ≠ healthy edge) closes for every NPD-sourced policy at once, ADR-053's included. That follow-up issue is drafted and ready to file — from your side of the fence too, say the word and I'll open it.

tanishagoyal2: thanks for the details @dmvevents. Lets file the issue for KOM so that we can track it.

docs/tutorials/monitoring-gpu-system-services.md
tanishagoyal2: makes sense. We can keep it non-fatal

docs/tutorials/monitoring-gpu-system-services.md
dmvevents: Filed: #1876 — covers both failure modes (NPD restart reset + Unknown-as-healthy-edge) with the v1.36.0 source verification, the cluster-test evidence for what scripts can and cannot intercept, the affected checks across both ADRs, and a per-policy opt-in proposal sketch. The recurrence-escalation follow-up is noted there as a related-but-separate concern — say the word if you want that one tracked as its own issue too.

docs/tutorials/monitoring-gpu-system-services.md
dmvevents: Keeping it non-fatal then — thanks for weighing in.

docs/tutorials/monitoring-gpu-system-services.md
```diff
@@ -0,0 +1,458 @@
+# Tutorial: Monitoring GPU System Services with Node Problem Detector
+
+This tutorial extends an existing
+[node-problem-detector (NPD)](https://github.com/kubernetes/node-problem-detector)
+installation with the ADR-050 `CustomPluginMonitor` checks for GPU-critical
```
lalitadithya: we don't have to mention the ADR in user facing docs

docs/tutorials/monitoring-gpu-system-services.md
```diff
@@ -0,0 +1,458 @@
...
+This integration watches these condition and reason pairs:
+
+| Condition type | Problem reason | KOM policy | Fatal | Action |
+| --- | --- | --- | --- | --- |
+| `FabricManagerDown` | `FabricManagerNotActive` | `NPDFabricManagerDown` | yes | `RESTART_BM` |
+| `FabricManagerFlapping` | `FabricManagerFlapping` | `NPDFabricManagerFlapping` | yes | `RESTART_BM` |
+| `FabricManagerNotInstalled` | `FabricManagerUnitNotFound` | `NPDFabricManagerNotInstalled` | yes | `CONTACT_SUPPORT` |
+| `NvidiaPersistencedDown` | `NvidiaPersistencedNotActive` | `NPDNvidiaPersistencedDown` | no | `CONTACT_SUPPORT` |
+
+Unlike ADR-053's log-matched conditions, these are active probes: a healthy
```
lalitadithya: Same comment as earlier, let's not mention ADRs/design documents

docs/tutorials/monitoring-gpu-system-services.md
dmvevents: Done in the latest push — ADR/design-doc references are gone from the tutorial (and the plugin-script headers, same principle): the per-scenario policy sets and recovery caveats it pointed at are now stated inline. The push also lands the concrete systemctl-shim guidance from the cluster test (the upstream NPD image ships no systemctl; a chroot-/host shim is the robust pattern).

docs/tutorials/monitoring-gpu-system-services.md
dmvevents: Done in the latest push — same treatment throughout the doc.

tanishagoyal2 (approved):

lalitadithya (approved):

PR #1384, opened 2026-06-10
feat(preflight): add cuda-validation init-container

dmvevents:
## Summary

One-shot CUDA validation probe (`preflight-checks/cuda-validation/`) that runs as a Pod init-container: verifies GPU visibility and basic CUDA runtime execution on every detected device, and gates the workload pod from starting when CUDA is unhealthy. Sibling of `preflight-checks/dcgm-diag/`, `nccl-allreduce/`, and `nccl-loopback/` (Dockerfile, probe script, Makefile, README). Companion to the #891 split series: per @XRFXLP's #891 review, CUDA memory probing was excluded from the steady-state system-services-monitor daemon (it competes with running-workload GPU memory) — the init-container pattern runs once before the workload starts, eliminating that concern.

Each device probe runs in a child process with a `CUDA_VALIDATION_DEVICE_TIMEOUT` budget (default 60s), so a wedged GPU/driver that blocks `torch.cuda.synchronize()` inside a C call is killed and reported as a per-GPU failure instead of hanging the init-container until the pod deadline. Environment failures (torch not importable / no CUDA) exit distinctly from per-GPU failures.

## Type of Change
- [x] ✨ New feature

## Component(s) Affected
- [x] Other: `preflight-checks/cuda-validation/` (new init-container component)

## Testing
- [x] Tests pass locally — `black --check --line-length 120` (pinned 26.3.1) passes; probe verified against a stub torch that hangs in `synchronize()` (fails in 5.1s under a 5s budget, no leaked children) and against a healthy 2-GPU run
- [x] No breaking changes (new component, nothing existing modified)

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Architecture
Runs as a Pod init-container; the exit code gates the workload pod from starting if CUDA is unhealthy. Workload runtime never sees this probe. All three review threads from @deesharma24 (probe timeout, `(ImportError, OSError)` guard, hard Black gate) are addressed in 7f98e3a with per-thread replies.

## Release Notes

* **New Features**
  * Added a CUDA validation probe deployable as a Kubernetes init-container to verify GPU visibility and basic runtime execution across detected devices.
  * Supports configurable tensor sizes and per-device timeouts.
  * Reports distinct outcomes for environment issues, GPU validation failures, and timed-out checks.

* **Documentation**
  * Added setup, configuration, exit-code, build, publishing, and Kubernetes deployment guidance.

* **Tests / CI**
  * Added automated linting, container builds, and publishing for the CUDA validation component.


preflight-checks/cuda-validation/cuda_validation.py
```diff
@@ -0,0 +1,169 @@
...
+def _tensor_size() -> int:
+    raw = os.getenv("CUDA_VALIDATION_TENSOR_SIZE", str(DEFAULT_TENSOR_SIZE))
+    try:
+        size = int(raw)
+    except ValueError:
+        return DEFAULT_TENSOR_SIZE
+    return size if size > 0 else DEFAULT_TENSOR_SIZE
+
+
+def _probe_device(torch, index: int, size: int) -> GPUResult:
```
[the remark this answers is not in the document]
dmvevents: Fixed in 7074580 — `_probe_device(torch: types.ModuleType, index: int, size: int)` with `import types` added. Kept the module-object parameter (rather than a top-level import) so a missing torch still maps to `EXIT_ENV_FAIL` in `main()`.

preflight-checks/cuda-validation/cuda_validation.py
```diff
@@ -0,0 +1,169 @@
...
+        return GPUResult(index=index, name=name, passed=True, detail="ok")
+    except Exception as exc:  # noqa: BLE001 - intentional broad catch
+        return GPUResult(index=index, name=name, passed=False, detail=str(exc))
+
+
+def main() -> int:
+    log = _setup_logging()
+
+    try:
+        import torch  # noqa: WPS433 - imported here so missing torch maps to EXIT_ENV_FAIL
```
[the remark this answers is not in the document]
dmvevents: Fixed in 7074580 — dropped the invalid `WPS433` code and kept the rationale as a plain comment. Ruff no longer reports RUF102 on the file.

preflight-checks/cuda-validation/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+    CUDA_VALIDATION_TENSOR_SIZE=1024
+
+WORKDIR /app
+
+COPY preflight-checks/cuda-validation/cuda_validation.py /app/cuda_validation.py
+RUN chmod +x /app/cuda_validation.py
+
+# Note: Running as root because CUDA device access typically needs
+# elevated privileges; matches the nccl-allreduce sibling.
+ENTRYPOINT ["python3", "/app/cuda_validation.py"]
```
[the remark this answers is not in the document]
dmvevents: Fixed in 7074580 — the image now creates and switches to a dedicated non-root user using the exact pattern of the `dcgm-diag` sibling (`groupadd -r nvsentinel && useradd -r -g nvsentinel nvsentinel` + `USER nvsentinel`), with `HOME=/tmp` so any torch/JIT cache writes succeed. On standard device-plugin/GPU-operator deployments `/dev/nvidia*` is world-accessible, so the probe keeps working non-root; deployments with restricted device modes can override `runAsUser` in the pod spec.

dmvevents:
Verified this against the PR diff and declining as not-applicable — the module is registered in the correct jobs, and the flagged lines are not this PR's entries:

- **Lint:** this PR adds `cuda-validation` to the `preflight-checks-lint-test` matrix (the job that runs `make -C preflight-checks/${{ matrix.component }} lint-test` and already covers `dcgm-diag`/`nccl-allreduce`/`nccl-loopback`). That is the correct matrix for a `preflight-checks/*` module. `modules-lint-test` (the matrix at lint-test.yml:183) is for top-level modules and would run `make -C cuda-validation`, which is a nonexistent path.
- **Container build:** this PR adds `component: preflight-cuda-validation` with `make_command: 'make -C preflight-checks/cuda-validation docker-build'` to `container-build-test.yml` — exactly the entry the finding asks to add.
- **The `lifecycle-manager` entries** at lint-test.yml:183 and container-build-test.yml:150-151 are upstream `main`'s lines from #1610 ("feat: add lifecycle-manager component scaffolding"), present here only because the branch merged current `main`; they register that component correctly and are untouched by this PR.

Happy to move the entries if a maintainer prefers a different matrix, but as-is the new module is linted and container-built by the same jobs as its sibling preflight checks.

preflight-checks/cuda-validation/cuda_validation.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+    try:
+        torch.cuda.set_device(index)
+        # Allocate a tiny tensor, run a kernel, sync, verify the result.
+        # Small allocation keeps probe footprint negligible — the init
+        # container exits before the workload starts.
+        tensor = torch.randn(size, device=f"cuda:{index}")
+        total = tensor.sum()
+        torch.cuda.synchronize(index)
```
deesharma24, line 144: it can block indefinitely when the GPU/driver is unhealthy.So, it should be timeout based call.

preflight-checks/cuda-validation/cuda_validation.py
```diff
@@ -0,0 +1,176 @@
...
+        return GPUResult(index=index, name=name, passed=False, detail=str(exc))
+
+
+def main() -> int:
+    """Run the per-GPU CUDA validation probe; exit code encodes the failure class."""
+    log = _setup_logging()
+
+    try:
+        import torch  # imported here so missing torch maps to EXIT_ENV_FAIL
+    except ImportError as exc:
```
deesharma24: Should we have both  (ImportError, OSError) as exc:

preflight-checks/cuda-validation/Makefile
```diff
@@ -0,0 +1,91 @@
...
+	python3 -m py_compile cuda_validation.py
+	@if command -v pyflakes >/dev/null 2>&1; then \
+		echo "Running pyflakes..."; \
+		pyflakes cuda_validation.py; \
+	else \
+		echo "pyflakes not installed; skipping (py_compile already gated)"; \
+	fi
+	@if command -v black >/dev/null 2>&1; then \
+		echo "Running black --check (advisory)..."; \
+		black --check --line-length 120 cuda_validation.py \
```
deesharma24: shouldn't it be exited if it fails.something like below

```
  	@black --check --line-length 120 cuda_validation.py || { \
  		echo "ERROR: cuda_validation.py is not Black-formatted"; \
  		exit 1; \
  	}
```

preflight-checks/cuda-validation/cuda_validation.py
dmvevents: Agreed, and fixed in `7f98e3a`. You're right that it can block forever: `torch.cuda.synchronize()` on a wedged GPU/driver blocks inside a C call that holds the GIL, so a signal alarm or thread join can't interrupt it — the only reliable bound is to run the probe in a child process and kill it. Each device now gets a `CUDA_VALIDATION_DEVICE_TIMEOUT` budget (default 60s, `0` restores the old in-process behaviour) and a timeout is reported as a per-GPU failure (exit 1, "GPU or driver not responding"), not an env failure — the runtime came up, one device stopped answering. A child that dies without reporting (driver SIGABRT/SIGSEGV) is reported with its exitcode instead of being misread as a hang.

Verified against a stub torch that hangs in `synchronize()` for one device: the run now fails in 5.1s under a 5s budget with no leaked children, where before it blocked until the pod deadline. Healthy 2-GPU run still passes, `=0` keeps the in-process path, and a non-integer value falls back to the default.

preflight-checks/cuda-validation/cuda_validation.py
dmvevents: Yes — done in `7f98e3a`, now `except (ImportError, OSError) as exc`. The case you're guarding is real: torch's extension modules `dlopen` the CUDA driver/runtime, so a present-but-unloadable torch on a node with a broken or version-mismatched `libcuda` raises `OSError` from the loader, not `ImportError`. Both mean the environment is unusable rather than a specific GPU being bad, so both map to `EXIT_ENV_FAIL` (2). Also reworded the log line from "not available" to "not importable" since it now covers the load-failure case, and added a comment recording why `OSError` belongs there.

preflight-checks/cuda-validation/Makefile
dmvevents: You're right — applied your suggested form verbatim in `7f98e3a` (`Makefile:44-48`). It should be a hard gate: `make/python.mk:21` runs `poetry run black --check .` as a blocking step for every other Python module, and `setup-ci-env` installs the Black pinned in `.versions.yaml` (26.3.1), so the tool is present in `preflight-checks-lint-test`. My earlier advisory-warning comment was wrong about CI not having it.

One thing that came with it: the file was not actually Black-clean, so hard-gating alone would have turned this job red. I reformatted `cuda_validation.py` with the pinned 26.3.1 in the same commit (three collapsed `GPUResult(...)` calls). Verified both directions — `make lint-test` passes on the formatted file, and exits 1 with your error message on a deliberately unformatted copy. Dropped the `pyflakes` block's stale claim while I was there; it still skips cleanly when pyflakes is absent.

dmvevents: Correction to my line reference above: the Black gate is at `Makefile:51-55`, not 44-48 (44-50 is the pyflakes block, which is unchanged — what I removed was the stale *comment* above the target claiming CI has no Black).

dmvevents:
Branch updated to current `main` (clean merge, head `844195c4b806972f38254bdd48548684bfa9852c`); `black --check` with the pinned 26.3.1 passes locally. Retitled to `feat(preflight):` per the repo's conventional-commit titles and brought the body in line with the PR template.

All three of @deesharma24's threads are addressed in 7f98e3a with per-thread replies (child-process timeout for the wedged-GPU case, `(ImportError, OSError)`, hard Black gate) — ready for re-review whenever convenient.

@XRFXLP could you vet CI on this head when you get a chance (`/ok to test 844195c4b806972f38254bdd48548684bfa9852c`)?

dmvevents:
The single red check on this run — `modules-lint-test (labeler)` — is a flake unrelated to this PR (the diff touches only `preflight-checks/cuda-validation/`, no Go): `TestInitializeAll_RateLimitScenarios_InitializedLabelerUsesConfiguredQPS` panics with a nil-pointer dereference, and the same job passed in the #1382, #1383, and #1385 runs vetted in the same window. Flagging the test itself for maintainer awareness since a nil deref in the rate-limit scenarios looks like a real (if racy) bug in the test or initializer.

@XRFXLP could you re-run that job (or re-issue `/ok to test 844195c4b806972f38254bdd48548684bfa9852c`) when convenient? Everything else on this PR is green (79 checks).

dmvevents:
@deesharma24 gentle ping on this one — it's fully green on the CI matrix (the one red last week was a labeler flake that passed on re-run) and all three of your threads are addressed in 7f98e3a with per-thread replies (child-process timeout for the wedged-GPU case, `(ImportError, OSError)`, hard Black gate in your suggested form). It's independent of the ADR scope discussion, so it can merge on its own whenever you're satisfied with the re-review.

dmvevents:
Weekly touch on this one: still fully green (79 checks; the labeler flake cleared on re-run two weeks ago) and unchanged since — no churn pending. @deesharma24's three threads are each addressed in 7f98e3a with per-thread replies, and this PR is fully independent of the NPD/ADR discussion, so it's mergeable on its own. If @deesharma24 is heads-down on the NPD proposal, any maintainer's re-review works equally well — happy to walk through any of it. cc @XRFXLP

preflight-checks/cuda-validation/README.md
```diff
@@ -0,0 +1,116 @@
+# preflight-cuda-validation
```
XRFXLP: nit: we usually keep the docs here: https://github.com/NVIDIA/NVSentinel/tree/main/docs

deesharma24:
> Weekly touch on this one: still fully green (79 checks; the labeler flake cleared on re-run two weeks ago) and unchanged since — no churn pending. @deesharma24's three threads are each addressed in https://github.com/NVIDIA/NVSentinel/commit/7f98e3ad20ecf362a1390b2a087a0e673e4a521d with per-thread replies, and this PR is fully independent of the NPD/ADR discussion, so it's mergeable on its own. If @deesharma24 is heads-down on the NPD proposal, any maintainer's re-review works equally well — happy to walk through any of it. cc @XRFXLP

<<deepak>> preflight check dcgm-diag at the chart default DCGM_DIAG_LEVEL=2 (https://github.com/NVIDIA/NVSentinel/blob/main/distros/kubernetes/nvsentinel/charts/preflight/values.yaml#L151) maps to DCGM_DIAG_LVL_MED (https://github.com/NVIDIA/NVSentinel/blob/main/preflight-checks/dcgm-diag/dcgm_diag/diag.py#L52). What DCGM runs at that level:

- Test selection per level: https://github.com/NVIDIA/DCGM/blob/v4.5.2/nvvs/src/NvidiaValidationSuite.cpp#L1131-L1170 — Software group always, Memory plugin and PCIe plugin at >= MEDIUM.
- Software group "CUDA Main Library" / "CUDA Toolkit Library" is a dlopen of libcuda/libcudart only, no CUDA execution: https://github.com/NVIDIA/DCGM/blob/v4.5.2/nvvs/src/Software.cpp#L364-L381 and #L506.
- Memory plugin does real CUDA work per GPU: cuMemAlloc of most of the framebuffer (https://github.com/NVIDIA/DCGM/blob/v4.5.2/nvvs/plugin_src/memory/memory.cpp#L183) then the L1Tag test which launches CUDA kernels (https://github.com/NVIDIA/DCGM/blob/v4.5.2/nvvs/plugin_src/memory/memory.cpp#L591-L592, https://github.com/NVIDIA/DCGM/blob/v4.5.2/nvvs/plugin_src/memory/L1TagCuda.cpp#L341-L366).
- PCIe plugin does H2D/D2H copies and P2P on each GPU: https://github.com/NVIDIA/DCGM/blob/v4.5.2/nvvs/plugin_src/pcie/Pcie.cpp#L1197-L1210.

So, the above test does cuda related work.

Also, nccl-loopback covers the in-container side: all_reduce_perf -g <N> over every allocated GPU inside the init container (https://github.com/NVIDIA/NVSentinel/blob/main/preflight-checks/nccl-loopback/pkg/benchmark/benchmark.go#L73-L77), same PyTorch base as this PR, with result checking on by default (-c 1, https://github.com/NVIDIA/nccl-tests/blob/v2.13.11/README.md), ~5s, and a fatal HealthEvent on any CUDA/NCCL error (https://github.com/NVIDIA/NVSentinel/blob/main/preflight-checks/nccl-loopback/main.go#L93-L105).

So what #1384 probes (randn/sum/isfinite per GPU) is already exercised by both,  hence, I think we don't need additional  check to do cuda-validation. Please let me know if you disagree.

dmvevents:
I verified each citation before answering, and your analysis holds: at the shipped defaults, `dcgm-diag` level 2 exercises per-GPU CUDA more deeply than this probe does (framebuffer-scale `cuMemAlloc` + L1Tag kernels + PCIe H2D/D2H/P2P, with per-GPU verdicts), and `nccl-loopback` proves exactly the axis this PR claimed as distinctive — CUDA context init and kernel execution *through a workload-representative container* on the same PyTorch base, result-checked, fatal on CUDA/NCCL error.

The only residual I can name honestly: per-GPU fault *identity* through the workload stack specifically — the all-reduce fails as a collective, so a single bad GPU isn't isolated by nccl-loopback alone. But dcgm-diag supplies the per-GPU verdict, and a GPU that passes DCGM's own CUDA work while failing only a torch `randn` is an exotic case (a torch/driver mismatch presents env-wide, which nccl-loopback already catches). That residual doesn't justify a third component and its maintenance surface.

So I agree: **I'd close this PR as covered by the existing preflights** — same reasoning we applied when dropping the DCGM-duplicated checks from #891 and the fabric-state checker from the ADR discussion. Unless you or another maintainer see value in the residual, I'll close it out in a day or two and note the coverage mapping in the closing comment for anyone who searches for cuda-validation later. (The one reusable artifact from this review — the child-process-with-timeout probe pattern for wedged-driver cases, including the bounded-join fix CodeRabbit just suggested — is documented in the threads if it's ever useful elsewhere.) Thanks for the rigorous look — this is the second time your DCGM source-diving has saved the fleet a redundant collector.

dmvevents:
Closing as discussed ([coverage analysis](https://github.com/NVIDIA/NVSentinel/pull/1384#issuecomment-5597447523)): at the shipped defaults, `dcgm-diag` level 2 (per-GPU `cuMemAlloc` + L1Tag kernels + PCIe, with per-GPU verdicts) and the nccl-loopback workload-container check together cover what this init-container probed — @deesharma24's redundancy analysis held up under verification, and the residual gap (per-GPU identity through the torch stack specifically) is too exotic to justify a fourth CUDA-touching preflight. Thanks for the rigorous coverage mapping; this is the right kind of scope pruning.

PR #1388, opened 2026-06-11
feat: add gpu thermal margin runbook

deesharma24:
## Summary
Added runbook for GpuThermalMarginWatch 

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added a runbook for GPU thermal margin / hardware slowdown violations: symptoms, how to confirm and verify using live telemetry and nvidia-smi, checks for per‑GPU threshold metadata, step‑by‑step remediation, reproduce-under-load guidance, and configuration notes for enabling GPU monitoring.
  * Clarified design guidance to explain limits of thermal‑margin detection and avoid implying RMA criteria.

### DIFF
--- a/docs/runbooks/gpu-thermal-margin.md
+++ b/docs/runbooks/gpu-thermal-margin.md
@@ -1,0 +1,128 @@
+# Runbook: GPU Thermal Margin / HW Slowdown Violations (`GpuThermalMarginWatch`)
+
+## Overview
+
+`GpuThermalMarginWatch` monitors each GPU's live thermal margin (DCGM field 153, `DCGM_FI_DEV_GPU_TEMP_LIMIT`) against the per-SKU hardware-slowdown T.Limit offset published by the metadata-collector (NVML field 194, `FI_DEV_TEMPERATURE_SLOWDOWN_TLIMIT`). When a GPU's margin drops below its slowdown threshold, the GPU is at or past the temperature at which the hardware engages thermal slowdown, the fatal event for GpuThermalMarginWatch occurs. The feature is described in [ADR-042: GPU Thermal Margin](../designs/042-gpu-temp-limit-field-monitoring.md).
+
+**Key points:**
+
+- Failure raises node condition `GpuThermalMarginWatch=True` with reason `GpuThermalMarginWatchIsNotHealthy` and error code `GPU_TEMP_HW_SLOWDOWN_VIOLATION`.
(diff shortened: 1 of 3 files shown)


docs/runbooks/gpu-thermal-margin.md
```diff
@@ -0,0 +1,128 @@
...
+```bash
+# Live margin to the slowdown limit (field 153) and current temps
+kubectl exec -n nvsentinel <GPU_MONITOR_POD> -- dcgmi dmon -e 153,150,140 -c 1
+
+# Or via nvidia-smi on the node
+kubectl exec -n nvsentinel <GPU_MONITOR_POD> -- nvidia-smi -q -d TEMPERATURE -i <GPU_INDEX>
+```
+
+- If the GPU is genuinely hot (low/negative margin, temperature crossing the slowdown limit), this is a real thermal fault. Go to step 5.
+- If the GPU is cool and the margin is healthy but the condition still fired, suspect a bad or stale threshold in metadata. Go to step 3.
```
XRFXLP: Is it possible that for few minutes the threshold was crossed and after sometime the temperature went normal?

XRFXLP: Will the condition auto-clear?

deesharma24: yes, it is very much possible. if it happens, then NVSentinel will automatically uncordon the node and we don't need the investigation. if it doesn't happen, then this condition will be there and user will need to investigate it using the steps mentions at the end. Generally, if it is workload behavior , it doesn's cross the h/w slowdown threshold and only when there is something wrong with the h/w, then it will cross the h/w slowdown threshold. We will anyway learn more about it after deployment. Till now, we don't even have visibility , so can't claim anything.

XRFXLP: So should we be conservative on making claim that this is due to staleness? And instead say, run the dcgmproftester anyway if we've a condition.

deesharma24: fair enough. updated the comment

lalitadithya (approved):

PR #1391, opened 2026-06-11
Fix PostgreSQL UpdateDocument placeholder collision

fallintoplace:
## Summary
- Fix direct PostgreSQL `UpdateDocument` so SET parameters are bound before WHERE parameters.
- Shift WHERE placeholders after update args and execute with update args followed by filter args.
- Make placeholder shifting regex-based so multi-digit placeholders are not rewritten incorrectly.

## Testing
- `go test ./pkg/client`
- `go test ./...`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect SQL placeholder numbering for UPDATE statements that include both value assignments and filtering conditions, ensuring parameters are applied in the correct order and updates run reliably.

* **Tests**
  * Added coverage to verify UPDATE query placeholder offsets between SET and WHERE clauses, including correct handling of multi-digit placeholders to prevent renumbering regressions.


lalitadithya:
hey @fallintoplace , can you please look into the failing tests?

KaivalyaMDabhadkar (approved):

PR #1392, opened 2026-06-11
feat(janitor): implement ExternalRemediationRequest reconciler

jtschelling:
## Summary

Second in the ExternalRemediationRequest (ExtRR) series. Builds on #1376 (foundation: CRD + types + RBAC + scheme registration) by adding the controller that drives the node-coordination state machine. Until something starts creating ExtRR objects (a follow-up change wiring fault-remediation), this controller only acts when an operator hand-applies one.

### State machine (6 dispatch branches)

1. **Initialise** — add the cleanup finalizer; seed `NVSentinelOwnershipReleased=Unknown` + `ExternalRemediationComplete=Unknown` on a fresh ExtRR.
2. **Apply path** — strategic-merge-patch the target node with the release taint (taint value = the ExtRR's own metadata.name, so the cleanup path can find only its own taint) and the `managed=false` label, then transition `NVSentinelOwnershipReleased=True`.
3. **True cleanup** — once an external system reports `ExternalRemediationComplete=True`, remove the taint + label, then drop the finalizer so the apiserver can garbage-collect the object.
4. **False no-op (asymmetric)** — `ExternalRemediationComplete=False` does NOT close the ExtRR. The node stays released until an operator deletes the object or the external system retries with `True`. This asymmetry is the core of the ADR-040 contract.
5. **Deletion-with-finalizer cleanup** — `kubectl delete extrr` triggers the same node cleanup before the finalizer drops, so operators can reclaim stalled nodes.
6. **Drift handling** — defends against foreign taints at the same key (left in place, never stomped), missing nodes (waits for apiserver), RBAC-forbidden patches (surfaces a failure condition), and taint already at the correct value (no-op).

### Supporting pieces

- `janitor/pkg/condition` — adapter between the proto `Condition` message and `meta.Condition` so existing controller-runtime helpers (`meta.SetStatusCondition`, `meta.IsStatusConditionTrue`) work against ExtRR status.
- `janitor/pkg/metrics/err_metrics.go` — three Prometheus series in the `nvsentinel_external_remediation_` namespace: lifecycle transitions counter, currently-open gauge, age-at-close histogram.
- `janitor/main.go` — manager registers the new reconciler alongside the existing RebootNode / TerminateNode / GPUReset controllers.

### Things deferred to follow-ups

- fault-remediation producing ExtRRs on the `EXTERNAL_REMEDIATION` recommended action (a separate change).
- node-labeler and cluster-scope monitor gating on `managed=false`.
- Demo GIF + ADR cross-refs.

## Test Plan

- [x] 143 janitor tests pass (124 prior + 19 new), including a full ginkgo envtest suite that exercises every dispatch branch end to end against a real apiserver.
- [x] \`pkg/controller\` coverage rises from 43.7% → 50.0%.
- [x] Foreign-taint test: apply a taint at our release key but with a different value, verify cleanup leaves it in place.
- [x] Asymmetric semantics: \`Complete=False\` must not trigger cleanup; verified in envtest.

## Summary by CodeRabbit

* **New Features**
  * Added support for External Remediation Requests, including controller handling, webhook validation, and lifecycle tracking.
  * External remediation now records completion state so follow-up reconciles can finish cleanly and avoid repeated work.

* **Bug Fixes**
  * Improved node cleanup behavior by removing remediation taints and management labels at the right time.
  * Added safeguards for missing nodes, immutable request fields, and repeated reconcile passes.

* **Tests**
  * Expanded automated coverage for validation, reconciliation, cleanup, and end-to-end remediation flows.


janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,859 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&nvsentinelv1.ExternalRemediationRequest{}).
+		Watches(
+			&corev1.Node{},
+			handler.EnqueueRequestsFromMapFunc(r.mapNodeToERRs),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
+		).
+		Named("externalremediationrequest").
+		Complete(r)
+}
```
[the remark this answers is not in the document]
jtschelling: this needs to be done but not in this PR, its deprecated but we use it in quite a few places so swapping it out here would cause some rbac split issues. imo its better to run with the deprecated call for now and then do a larger sweep to catch them all in one PR later

jtschelling:
@coderabbitai review

jtschelling:
@coderabbitai review

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,755 @@
...
+//
+//nolint:cyclop // distinct apiserver failure modes; inline switch reads cleaner.
+func (r *ExternalRemediationRequestReconciler) reconcileApply(
+	ctx context.Context, extrrObj *nvsentinelv1.ExternalRemediationRequest,
+) (ctrl.Result, error) {
+	nodeName := extrrObj.Spec.HealthEvent.NodeName
+
+	var node corev1.Node
+	if err := r.Get(ctx, client.ObjectKey{Name: nodeName}, &node); err != nil {
+		if apierrors.IsNotFound(err) {
```
lalitadithya, line 306: If the node is not found, do we need to requeue? Won't this be a terminal state?

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,755 @@
...
+	}
+
+	if nodeToUpdate.Labels == nil {
+		nodeToUpdate.Labels = map[string]string{}
+	}
+
+	nodeToUpdate.Labels[managed.ManagedLabelKey] = managed.ManagedLabelValueFalse
+
+	if err := r.Patch(ctx, nodeToUpdate, client.StrategicMergeFrom(&node)); err != nil {
+		if apierrors.IsForbidden(err) {
```
lalitadithya: when would get a forbidden error? Would it be because of a permissions issue?

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,755 @@
...
+		indexExtRRByNodeName,
+	); err != nil {
+		return fmt.Errorf("registering ExtRR node-name field indexer: %w", err)
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&nvsentinelv1.ExternalRemediationRequest{}).
+		Watches(
+			&corev1.Node{},
+			handler.EnqueueRequestsFromMapFunc(r.mapNodeToExtRRs),
```
natherz97: What's the purpose of re-queuing ERRs if the node taint or label changes, it doesn't look like we're re-applying the label or moving the ERR to a failed terminal state on this re-queue?

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,755 @@
...
+				"extrr", extrrObj.Name, "node", nodeName, "requeueAfter", nodeMissingRequeue)
+
+			return ctrl.Result{RequeueAfter: nodeMissingRequeue}, nil
+		}
+
+		return ctrl.Result{}, fmt.Errorf("get node %q for ExtRR %q: %w", nodeName, extrrObj.Name, err)
+	}
+
+	if existing := findTaintByKey(node.Spec.Taints, ReleaseTaintKey); existing != nil {
+		if existing.Value != extrrObj.Name {
```
natherz97: GPUReset had it's own locking implemented within the controller which we removed for the node-level locking across controllers. If we onboard to that we can likely remove the ownership check on the taint since we guarantee the controller will only process 1 ERR at a time. Do you think we need this taint check?

janitor/pkg/controller/externalremediationrequest_controller.go
natherz97: Related to this, what is our expected behavior for when the node is deleted by the external system? NVSentinel should be adding an ownerReference on the ERR for the corresponding node object so that should always cause the ERR to be deleted and we'd never be left in a situation where an ERR in a non-terminal status exists for a deleted node. If that ownerReference wasn't there, it doesn't look like the ERR would have make it to a terminal status.

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,755 @@
...
+) (ctrl.Result, error) {
+	conds := statusConditions(extrrObj)
+
+	switch {
+	case !extrrObj.DeletionTimestamp.IsZero():
+		return r.reconcileCleanupOnDeletion(ctx, extrrObj)
+
+	case meta.IsStatusConditionPresentAndEqual(conds, ConditionNVSentinelOwnershipReleased, metav1.ConditionUnknown):
+		return r.reconcileApply(ctx, extrrObj)
+
```
natherz97, line 276: What happens if ConditionNVSentinelOwnershipReleased is false for example if we fail to apply the taint? Let me know if I'm missing something but it seems like it would fall to the default case below right? I think that's the desired behavior but can you update the comment below since it's assuming that we successfully handed it off to the external system?

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,755 @@
...
+		updated := extrrObj.DeepCopy()
+		controllerutil.AddFinalizer(updated, ExternalRemediationFinalizer)
+
+		if err := r.Update(ctx, updated); err != nil {
+			return ctrl.Result{}, fmt.Errorf("adding finalizer to ExternalRemediationRequest %s: %w", extrrObj.Name, err)
+		}
+
+		slog.InfoContext(ctx, "Added cleanup finalizer to ExternalRemediationRequest", "name", extrrObj.Name)
+		// The own-kind watch re-enqueues on this metadata Update; no requeue needed.
+		return ctrl.Result{}, nil
```
natherz97: Why are we re-queuing after updating the finalizer, shouldn't we continue and call setInitialConditions as part of the same reconcile loop? It looks like setInitialConditions will re-fetch the object prior to the patch

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,755 @@
...
+// the upstream health-monitor trace via annotations stamped by fault-remediation.
+func (r *ExternalRemediationRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	var extrr nvsentinelv1.ExternalRemediationRequest
+	if err := r.Get(ctx, req.NamespacedName, &extrr); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Belt-and-suspenders: the validating webhook normally guarantees these,
+	// but a webhook outage shouldn't crashloop the controller.
+	if extrr.Spec == nil || extrr.Spec.HealthEvent == nil || extrr.Spec.HealthEvent.NodeName == "" {
```
natherz97: The failure policy on the webhook is fail so I don't think we need these duplicated validation checks in the controller.

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,755 @@
...
+//nolint:lll // kubebuilder RBAC marker must stay on one line
+// +kubebuilder:rbac:groups=nvsentinel.dgxc.nvidia.com,resources=externalremediationrequests,verbs=get;list;watch;update;patch
+//nolint:lll
+// +kubebuilder:rbac:groups=nvsentinel.dgxc.nvidia.com,resources=externalremediationrequests/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=nvsentinel.dgxc.nvidia.com,resources=externalremediationrequests/finalizers,verbs=update
+// +kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch;patch
+
+// Reconcile drives the ExtRR through its lifecycle. The OTEL span is linked to
+// the upstream health-monitor trace via annotations stamped by fault-remediation.
+func (r *ExternalRemediationRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
natherz97, line 117: We should onboard this controller to node-level locking right? If NVSentinel creates a RebootNode and then an ERR comes in, fault-remediation will need to create the ERR so we should likely use the same lock we have across RebootNodes + GPUReset to ensure that we don't hand off a node that already has an in-progress remediation action.

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: you're right, fixed

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: yea we can do this in one pass, updated

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: yeah, but looking again i think its unnecessary, we'd catch that in other ways and there's quite a bit of code related to catching rbac issues. removed

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: yeah agreed, implemented

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: no, the lock is better. makes this dead code so removing the check

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: it was for drift detection but now thats not needed with the node locking, removed all this code

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: @lalitadithya i changed it to a timeout, i think you're right. i'm not opposed to simply quitting on the first node not found check, but thought maybe a couple tries could be beneficial? not stuck to it

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: yeah it'll fall to the default case. actually the way the PR is written now the reconciler will never set that condition to false, its either unknown or true. im not sure if thats the best way to handle it, it seems like we'd want to call it failed but i cant figure out if theres actually anything appropriate. even if the taint failed to apply i think we'd retry until it worked

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,728 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	nodeName := extrrObj.Spec.HealthEvent.NodeName
+
+	var node corev1.Node
+	if err := r.Get(ctx, client.ObjectKey{Name: nodeName}, &node); err != nil {
+		if apierrors.IsNotFound(err) {
+			if time.Since(extrrObj.CreationTimestamp.Time) > applyAbandonAfter {
+				return ctrl.Result{}, r.abandonApply(ctx, extrrObj, nodeName)
```
natherz97: I think it's difficult to reason about deriving the state of the ERR using a combination of the NVSentinelOwnershipReleased + ExternalRemediationComplete status conditions and an annotation. Right now we don't have a way to signal the the ERR controller shouldn't continue reconciling indefinitely (and it looks like the same issue exists below with the reconcileCleanupAfterComplete). Can we add a completionTimestamp to signal that an ERR shouldn't be processed at all? Which would capture:
1. The node object not existing before NVSentinelOwnershipReleased
2. ExternalRemediationComplete is true + we already restored the node (left a related comment to this below)

Separately, I think you can also remove the timeout and instantly fail since the node object is guaranteed to exist before any ERR is created so there's no risk of the node object missing from the controller's cache. So if the node is ever missing waiting shouldn't ever result in this call succeeding later.

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,728 @@
...
+// repair) stay observable.
+func (r *ExternalRemediationRequestReconciler) reconcileCleanupAfterComplete(
+	ctx context.Context, extrrObj *nvsentinelv1.ExternalRemediationRequest,
+) (ctrl.Result, error) {
+	changed, err := r.reconcileCleanup(ctx, extrrObj)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if extrrObj.GetAnnotations()[closeRecordedAnnotation] != annotationValueTrue {
```
natherz97: I think we're using an annotation since we don't have a status condition that indicates ExternalRemediationComplete is true + we already restored the node. On controller start-up, won't we re-process all completed ERRs and call reconcileCleanup which will be a no-op? The other Janitor controllers have the completionTimestamp, can we do something like that or add another status condition to tell the controller to short-circuit needing to reconcile?

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,728 @@
...
+
+	if changed {
+		r.emitEvent(extrrObj, corev1.EventTypeNormal, eventReasonReleaseTaintRemoved,
+			fmt.Sprintf("release taint and managed=false label removed (%s)",
+				closeReasonExternalRemediationCompleteTrue))
+	}
+
+	// Release the node-level lock — the external system has finished and the
+	// node is back in NVSentinel's hands. CheckUnlock is idempotent.
+	r.NodeLock.CheckUnlock(ctx, extrrObj, extrrObj.Spec.HealthEvent.NodeName)
```
natherz97: If the unlock fails we should re-queue the object so we don't have to wait for the next controller relist to try unlocking:
```
	retryUnlock := r.NodeLock.CheckUnlock(ctx, &rebootNode, rebootNode.Spec.NodeName)
	if retryUnlock {
		return ctrl.Result{RequeueAfter: time.Second * 2}, nil
	}
```
The location of this call might change if we add a completionTimestamp and short-circuit reconciling in a higher-level function.

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,728 @@
...
+
+	if !changed {
+		return false, nil
+	}
+
+	return r.patchStatusConditions(ctx, extrrObj, conditions)
+}
+
+// dispatch routes to one of the five non-init ADR-040 branches.
+func (r *ExternalRemediationRequestReconciler) dispatch(
```
natherz97, line 264: It looks like the node lock is current acquired in reconcileApply and released in either reconcileCleanupAfterComplete or reconcileCleanupOnDeletion. I'll leave it up to you but this follows a different approach to other controllers which do something like this:

```
	completedReconciling := rebootNode.Status.CompletionTime != nil

	if !completedReconciling {
		locked := r.NodeLock.LockNode(ctx, &rebootNode, rebootNode.Spec.NodeName)
		if !locked {
			return ctrl.Result{RequeueAfter: time.Second * 2}, nil
		}

		result, err := r.reconcileHelper(ctx, &rebootNode)

		if err != nil || result.RequeueAfter > 0 {
			return result, err
		}

		return ctrl.Result{RequeueAfter: time.Second * 2}, nil
	}

	retryUnlock := r.NodeLock.CheckUnlock(ctx, &rebootNode, rebootNode.Spec.NodeName)
	if retryUnlock {
		return ctrl.Result{RequeueAfter: time.Second * 2}, nil
	}
```
The main difference is in the ERR controller we're only calling LockNode once per ERR compared to other controllers which call it on each reconcile loop (where the existing object that holds the lock can continually call LockNode and any objects without the lock will continually fail). I think you can keep the existing implementation especially if it's been tested but the original intent was to have the node lock as a wrapper on top of existing reconcile loops.

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: yeah you're right on both comments. i added completionTimestamp, easier to signal a terminal state and brings it in line with the other custom resources in the project

removed the timeout, i doubt it would be useful so no need to include it

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: added completionTimestamp

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: reworked with the completionTimestamp addition

janitor/pkg/controller/externalremediationrequest_controller.go
jtschelling: reworked

jtschelling:
@coderabbitai review

janitor/pkg/controller/externalremediationrequest_controller.go
```diff
@@ -0,0 +1,732 @@
...
+		}
+
+		return ctrl.Result{}, nil
+	}
+
+	if !r.NodeLock.LockNode(ctx, &extrr, nodeName) {
+		return ctrl.Result{RequeueAfter: nodeLockRequeue}, nil
+	}
+
+	result, dispatchErr := r.dispatch(ctx, &extrr)
```
natherz97, line 166: Overall node-level locking looks good to me. One thing I wanted to call out was that it looks like we might not re-queue a node which has its CompletionTime set but is still holding the node lock. For example, once ExternalRemediationComplete=True and we set the CompletionTime, the reconcileCleanupAfterComplete will return:
```
return ctrl.Result{}, nil
```
This means we aren't re-forcing a re-queue even though the node is still holding the lock. The example implementation in the reboot_controller chose to always force a re-queue in order to call CheckUnlock to prevent needing to re-fetch the CRD to see if the CompletionTime was set (or needing to have the reconciler return this state). The relevant logic is here: https://github.com/NVIDIA/NVSentinel/blob/main/janitor/pkg/controller/rebootnode_controller.go#L111C1-L120C4.

So I think we have 2 options:
1. Always re-queue after reconciling: this will result in us doing unnecessary reconciles when we're waiting on ExternalRemediationComplete to be set to true. Our current implementation is requiring the status condition to trigger the reconcile but we could keep it in this loop continually.
2. Keep the existing behavior: if we keep the existing behavior, can you confirm that the node lock is released after the ERR is completed? For example by checking in the e2e test that the node lock lease has been deleted. What I think is happening is that we're implicitly forcing a re-queue when we write the updated status for the completion timestamp which allows us to remove the lock on that re-queue. If that doesn't happen, my understanding is that we'd be forced to wait for the controller-runtime SyncPeriod which is multiple hours by default.

natherz97 (approved):

PR #1421, opened 2026-06-25
fix: nvsentinel-state label handling on partial recovery

### LINKED ISSUE #1416: [Bug]: remediation-failed node label can remain after the unsupported failing check has recovered
### Prerequisites

### Code of Conduct

### Bug Description

`dgxc.nvidia.com/nvsentinel-state` can remain stuck at `remediation-failed` after the failure that caused that label has recovered.

Scenario:
1. Event A is fatal and has a supported remediation action, so `fault-remediation` creates a remediation CR.
2. Event B is fatal but has an unsupported action, for example `CONTACT_SUPPORT`, so `fault-remediation` sets the node label to `remediation-failed`.
3. Event C is a healthy recovery event for B. `fault-quarantine` removes B from the active quarantine annotation.
4. Event A is still unresolved, so the node remains quarantined and no `UnQuarantined` event is emitted.
5. Because cleanup only happens on `UnQuarantined` / `Cancelled`, the node label stays `remediation-failed` even though B is no longer active.

Expected behavior:
The node state label should reflect the remaining active quarantine/remediation state. If the unsupported failure has recovered, `remediation-failed` should be cleared or recomputed from the still-active events.

### Component

Fault Management

### Steps to Reproduce

1. Create a fatal health event for a node with a supported remediation action, for example `COMPONENT_RESET`, so the node is quarantined and `fault-remediation` creates a remediation CR.
2. While the node is still quarantined, create another fatal health event for the same node with an unsupported remediation action, for example `CONTACT_SUPPORT`.
3. Verify that `fault-remediation` sets:
```text
dgxc.nvidia.com/nvsentinel-state=remediation-failed
```
4. Create a healthy recovery event for the unsupported check from step 2, using the same check name/entities and `RecommendedAction_NONE`.
5. Verify that `fault-quarantine` removes the recovered check from the quarantine annotation but keeps the node quarantined because the first event is still unresolved.
6. Check the node label.
Expected:
```text
dgxc.nvidia.com/nvsentinel-state
```
is recomputed from the remaining active failure and no longer reports `remediation-failed` from the recovered unsupported check.
Actual:
```text
dgxc.nvidia.com/nvsentinel-state=remediation-failed
```
remains on the node even though the unsupported failure is no longer active.

### Environment

- NVSentinel version: v1.9.0
- Kubernetes version:
- Deployment method:

### Logs/Output

_No response_

XRFXLP:
## Summary

Fixes #1416: the `dgxc.nvidia.com/nvsentinel-state=remediation-failed` node label could remain stuck on a node after the unsupported failing check that caused it had recovered, while the node stayed quarantined by a different, still-active failure.

## Root cause

When a node has multiple active failures and one of them has an unsupported remediation action (e.g. `CONTACT_SUPPORT`), fault-remediation writes the terminal `remediation-failed` label. If that unsupported check later recovers but other failures keep the node quarantined, fault-quarantine treats it as a *partial recovery* and (previously) did not propagate anything downstream. Cleanup of the stale label only happened on full `UnQuarantined`/`Cancelled`, so the label stayed at `remediation-failed` even though the failure that set it was gone.

## Changes

**fault-quarantine** — stays domain-agnostic. A partial recovery (a healthy event that clears a tracked failure while the node remains quarantined) is now propagated using the existing `AlreadyQuarantined` status instead of being dropped. FQ does not interpret remediation-action support. The `ProcessEvent` span reports `partial_recovery` for this path for observability.

**node-drainer** — a healthy event on a still-quarantined node is marked `AlreadyDrained` (nothing to drain) so it reaches fault-remediation without performing drain work.

**fault-remediation** — owns the remediation labels and reconciles them. On a partial recovery it recomputes the node label from the remaining active quarantine events, but only when the node currently carries a terminal remediation label (`remediation-failed` or `remediation-succeeded`), since either can be invalidated by a recovery:
- any remaining active event that is unsupported or recorded as remediation-failed → `remediation-failed`
- otherwise any remaining active event already remediated (or covered by the remediation annotation) → `remediation-succeeded`
- otherwise (a supported failure still pending remediation) → no change; the normal remediation flow sets the label

FR only ever writes labels it owns; it never writes node-drainer/fault-quarantine labels (`draining`/`drain-*`/`quarantined`).

## Behavior

- Recovered unsupported failure + remaining remediated failure → label recomputed to `remediation-succeeded`.
- Remaining unsupported failure → label stays `remediation-failed`.
- Stale `remediation-succeeded` with a remaining unsupported failure → recomputed to `remediation-failed`.
- Non-terminal states (`remediating`, etc.) are left to the in-progress flow.

## Testing

- Unit tests in `fault-remediation` covering recompute to succeeded/failed, the skip path for non-terminal labels, and the "leave label alone" path for a pending supported failure.
- E2E (Tilt) test `TestPartialRecoveryClearsStaleRemediationFailed` in `tests/fault_management_test.go` reproducing the full FQ→ND→FR flow, with an idempotent teardown that recovers both entity-level failures. Did the same test manually on a tilt cluster
- `go vet`/build, `golangci-lint` (0 issues), and `gofmt` clean on changed packages.

Closes #1416
## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [ ] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end “partial recovery” handling across fault quarantine and fault remediation, including recomputation of remediation outcomes while a node remains quarantined.

* **Bug Fixes**
  * Partially recovered quarantine events now report `partial_recovery` and continue propagation while the node stays quarantined.
  * Already-healthy drain requests are immediately marked as already drained.
  * Remediation state transitions between succeeded and failed are now treated as valid during partial recovery.

* **Tests**
  * Updated and expanded partial-recovery unit/E2E coverage, including a new test to clear stale `remediation-failed` and a new test helper for healing specific entity failures.


fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -442,6 +446,328 @@ func (r *FaultRemediationReconciler) handleCancellationEvent(
...
+	ctx context.Context,
+	nodeName string,
+	activeEvent *protos.HealthEvent,
+	coveredGroups map[string]struct{},
+	healthEventStore datastore.HealthEventStore,
+	sawRemediationSuccess *bool,
+) (bool, error) {
+	status, err := findHealthEventStatusByID(ctx, healthEventStore, nodeName, activeEvent.Id)
+	if err != nil {
+		return false, err
```
lalitadithya: can we wrap errors consistently please?

lalitadithya (commented):
we have a label state transition diagram in commons/pkg/statemanager/statemanager.go which will need to be updated since based on this change the remedition-* isn't terminal anymore

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -442,6 +446,328 @@ func (r *FaultRemediationReconciler) handleCancellationEvent(
...
+
+	if status != nil && status.FaultRemediated != nil {
+		if !*status.FaultRemediated {
+			return true, nil
+		}
+
+		*sawRemediationSuccess = true
+	}
+
+	if groupConfig != nil && groupConfigMatchesAny(groupConfig, coveredGroups) {
```
KaivalyaMDabhadkar: I think this could cause a problem. `coveredGroups` comes from the remediation annotation, but that annotation is updated when the maintenance CR is created, not when it has succeeded. So group membership means “a CR exists for this equivalence group in this quarantine session,” not necessarily “the remaining active event was successfully remediated.” In a partial recovery, this can turn a stale `remediation-failed` label into `remediation-succeeded` while the remaining supported failure is still in progress: `FaultRemediated` is nil, but the group match sets `sawRemediationSuccess=true`. An example:

1. Event A is supported and creates a remediation CR. The remediation annotation now contains group `restart`.
2. Event B is unsupported and sets the node label to `remediation-failed`.
3. Event B recovers, but Event A, or another active event in group `restart`, still keeps the node quarantined.
4. Partial recovery recompute runs.
5. If that remaining active event has no explicit `FaultRemediated=true`, the group match alone can still set `sawRemediationSuccess=true`.
6. FR writes `nvsentinel-state=remediation-succeeded`.

Can we check the CR/explicit event status before returning `RemediationSucceededLabelValue`?

KaivalyaMDabhadkar:
Since we now make partial recovery intentionally rewrite terminal remediation labels, for example `remediation-failed -> remediation-succeeded` and `remediation-succeeded -> remediation-failed`,  Should we update the state-machine contract here to reflect that? Right now `remediation-succeeded` and `remediation-failed` are still documented/validated as terminal states with no valid forward transitions. The label update still succeeds because `UpdateNVSentinelStateNodeLabel` writes before returning the validation error, and the new partial-recovery path tolerates `(nodeModified=true, err)`, but these expected recomputes will still log `Invalid state transition` and increment `nvsentinel_state_transition_unexpected_total`.

I think the validator should treat the expected recompute transitions as valid so observability does not report them as issues

fault-remediation/pkg/reconciler/reconciler.go
XRFXLP: checking event status now, PTAL.

XRFXLP:
> Since we now make partial recovery intentionally rewrite terminal remediation labels, for example remediation-failed -> remediation-succeeded and remediation-succeeded -> remediation-failed, Should we update the state-machine contract here to reflect that? Right now remediation-succeeded and remediation-failed are still documented/validated as terminal states with no valid forward transitions. The label update still succeeds because UpdateNVSentinelStateNodeLabel writes before returning the validation error, and the new partial-recovery path tolerates (nodeModified=true, err), but these expected recomputes will still log Invalid state transition and increment nvsentinel_state_transition_unexpected_total.
I think the validator should treat the expected recompute transitions as valid so observability does not report them as issues

Handled

fault-remediation/pkg/reconciler/reconciler.go
XRFXLP: this was wrapped internally but wrapper it here as well.

lalitadithya:
@KaivalyaMDabhadkar , can you review please?

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -442,6 +446,308 @@ func (r *FaultRemediationReconciler) handleCancellationEvent(
...
+func isPartialRecoveryRemediationEvent(healthEventWithStatus model.HealthEventWithStatus) bool {
+	if healthEventWithStatus.HealthEvent == nil || healthEventWithStatus.HealthEventStatus == nil {
+		return false
+	}
+
+	healthEvent := healthEventWithStatus.HealthEvent
+	status := healthEventWithStatus.HealthEventStatus.NodeQuarantined
+
+	return healthEvent.IsHealthy &&
+		healthEvent.RecommendedAction == protos.RecommendedAction_NONE &&
```
KaivalyaMDabhadkar: Instead of requiring `RecommendedAction_NONE` to identify a partial recovery, can we identify partial recovery using `IsHealthy` plus `AlreadyQuarantined` instead?

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -442,6 +446,308 @@ func (r *FaultRemediationReconciler) handleCancellationEvent(
...
+	}
+
+	if sawRemediationSuccess {
+		return statemanager.RemediationSucceededLabelValue, true, nil
+	}
+
+	// A remaining failure is supported but not yet remediated. fault-remediation does not own
+	// the pre-remediation labels (quarantined/draining/drain-*), and its normal remediation flow
+	// will set the correct remediation label, so leave the current label untouched.
+	return "", false, nil
```
KaivalyaMDabhadkar, line 606: Is it guaranteed that the normal remediation flow will process this event again? When an event is skipped because an equivalent maintenance CR is already in progress, `handleExistingCRSkip` will advance the resume token without setting FaultRemediated or requeueing. The event can remain `FaultRemediated=nil` permanently. If that event becomes the only remaining active failure after the failure responsible for `remediation-failed` recovers, currently we do no update here and leave the stale remediation-failed label in place

fault-remediation/pkg/reconciler/reconciler.go
XRFXLP: removed `healthEvent.RecommendedAction == protos.RecommendedAction_NONE`, Quarantined/AlreadyQuarantined was already in the check.

fault-remediation/pkg/reconciler/reconciler.go
XRFXLP: Handled: for an active event with `FaultRemediated=nil`, we no longer assume the normal flow will reprocess it. Instead we check the actual CR status for its equivalence group: if it's covered by an in-progress or succeeded CR we treat it as remediated; a failed/absent CR is left for the normal flow to retry. We read the live CR state (not just annotation membership), so a recorded-but-gone/failed CR doesn't count as coverage.

KaivalyaMDabhadkar (approved):

PR #1426, opened 2026-06-26
fix: default kernel-origin syslog checks to SYSLOG_FACILITY=0

### LINKED ISSUE #1417: [Bug]: syslog-health-monitor can miss XIDs when journald rotates before scan
### Prerequisites

### Code of Conduct

### Bug Description

`syslog-health-monitor` can miss XID events when journald rotates or vacuums entries before the monitor reaches them.
In an observed incident, XIDs were still visible through `dmesg` and detected by DCGM, but the retained journald range started after the XID timestamps. Since `syslog-health-monitor` reads `/var/log/journal` via journald rather than `dmesg` or DCGM, it had no retained journal entries to parse.

Expected behavior: XID-related syslog checks should process only kernel-origin journal entries so the monitor can catch up quickly and avoid falling behind due to unrelated audit/container log volume.

Proposed fix: add kernel-only journald filtering for syslog checks, for example `SYSLOG_FACILITY=0`, by default for XID/GPU-fallen/SXID checks.

### Component

Health Monitor

### Steps to Reproduce

1. Run `syslog-health-monitor` on a GPU node where it reads from `/var/log/journal`.
2. Generate enough non-kernel journald traffic to put journald under retention pressure.
3. Trigger or inject an NVIDIA XID kernel log entry.
4. Continue generating journald traffic until the journal segment containing the XID is rotated or vacuumed before `syslog-health-monitor` processes it.
5. Confirm the XID was produced by checking `dmesg -T | grep -i "nvrm: xid"`.
6. Confirm the XID is no longer available through journald using `journalctl | grep -i "nvrm: xid"` or by checking that `journalctl --list-boots` starts after the XID timestamp.
7. Observe that `syslog-health-monitor` does not emit a corresponding health event.

Idea here is to delay syslog HM log processing completion so logs get rotated before HM could process it.

### Environment

- NVSentinel version: v1.9.0
- Kubernetes version:
- Deployment method:

### Logs/Output

_No response_

jackyliusohu:
## Summary

The XID, SXID and GPU-fallen-off-bus checks are built with empty `Tags` in
`buildChecksFromFlag`, so no journald match is applied and they scan every
facility. Under journald retention pressure, unrelated high-volume userspace
logs (audit, containers) can push the read cursor behind the retained window,
so the journal segment holding an XID is vacuumed before the monitor processes
it and the event is silently missed.

This defaults the three kernel-origin checks to the existing `-k`
(`SYSLOG_FACILITY=0`) filter so they only consume kernel entries and keep up.
The Kata path is left filtering by `-u containerd.service`: on Kata nodes XIDs
surface via the guest kernel relayed through containerd, not as host
`SYSLOG_FACILITY=0` entries, and the two matches would AND together and match
nothing.

Verified on a production deployment that neither the regular nor the Kata
DaemonSet currently carries `-k`/`--dmesg`, confirming the gap.

Fixes #1417

## Type of Change
- [x] 🐛 Bug fix

## Component(s) Affected
- [x] Health Monitors

## Testing
- [x] Tests pass locally
- [x] Added unit tests: regular path defaults `-k`, non-kernel checks unfiltered, Kata path uses `-u containerd.service` (never `-k`), and `-k`/`--dmesg` -> `SYSLOG_FACILITY=0` in `configureTagFilters`
- [x] `go build`, `go vet`, `gofmt` clean

## Checklist
- [x] Self-review completed
- [x] Documentation updated (if needed) — N/A
- [x] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved syslog check filtering so kernel-related checks consume the correct kernel-origin log entries.
  * Updated kernel transport filtering to use the journal transport field for more accurate matching.
  * Adjusted Kata mode so containerd filtering reliably overrides any prior kernel filtering.
  * Fixed cursor resume behavior to avoid re-reading a bookmarked journal entry when match filters are active.
* **Tests**
  * Added/extended coverage for kernel tag matching, Kata override behavior, and filtered-journal cursor resume semantics (including improved mock journal behavior).


XRFXLP:
@jackyliusohu Can you please add details of testing in an actual cluster?

jackyliusohu:
> @jackyliusohu Can you please add details of testing in an actual cluster?
 Thanks for the review @XRFXLP. Full end-to-end validation (a real XID being detected by syslog-health-monitor under the new `-k` filter) needs a GPU
  node where the nvidia kernel module emits the XID as a `SYSLOG_FACILITY=0` entry — userspace injection (`/dev/kmsg`/`logger`) is facility=1 and would
  (correctly) be filtered out by this change, so it can't substitute for it.

  I'm bringing up a GB300 cluster and expect to have results in the next day or two. I'll post the pod log showing `Adding kernel log filter ...
  SYSLOG_FACILITY=0` for the three kernel-origin checks, plus a real XID being picked up.

  In the meantime I've confirmed on a live deployment that neither the regular nor the Kata DaemonSet currently carries `-k`/`--dmesg`, which is the gap
  this PR closes.

jackyliusohu:
@XRFXLP  Pushed an update with real-cluster verification on a GB300 (arm64) node.

  **End-to-end test:** I injected a genuine kernel XID via a small `printk` module, so it lands as a real `_TRANSPORT=kernel`, `SYSLOG_FACILITY=0` entry —
  exactly like a driver-emitted XID. syslog-health-monitor with the `-k` filter detects it and reports it to the datastore, confirming the kernel-only filter
  correctly keeps genuine facility=0 XIDs.

  **One issue surfaced during testing, now fixed in this PR:** with the `SYSLOG_FACILITY=0` match active, journald can return the *bookmarked* entry itself on
  the first `Next()` after `SeekCursor` (instead of the entry after it). This caused the last processed kernel XID to be re-read and re-emitted every poll
  cycle — a single fault flooded the datastore with duplicate events. I added `skipBookmarkedEntryIfPresent` in `resumeFromLastCursor` to detect this and
  advance past the bookmark, plus a unit test (FakeJournal simulates the replay).

  Verified on GB300: before the cursor fix the XID was re-sent every 15s; with the fix, the same injected XID is reported exactly once.

  screenshot-1

  screenshot-2

XRFXLP:
@jackyliusohu Can you please check if these tests: https://github.com/NVIDIA/NVSentinel/blob/main/tests/uat/tests.sh#L409 are still valid?

I think they might need the change because we're injecting the error from userspace but we filter those out in in syslog health monitor now.

jackyliusohu:
> @jackyliusohu Can you please check if these tests: https://github.com/NVIDIA/NVSentinel/blob/main/tests/uat/tests.sh#L409 are still valid?
> 
> I think they might need the change because we're injecting the error from userspace but we filter those out in in syslog health monitor now.

 Good catch — you're right. Those tests inject via `/dev/kmsg` from userspace, which the kernel forces to `SYSLOG_FACILITY=1`, so the `-k` filter this PR
  adds
    will now (correctly) drop them. I'll update `tests/uat/tests.sh` to inject facility=0 kernel entries and verify end-to-end on GB300 before pushing. Will
    follow up here.

@XRFXLP  I looked into this further — the original `/dev/kmsg` injection in tests.sh won't work with the `-k` filter, and
  reproducing a genuine facility=0 entry in the test turns out to be tricky (userspace `/dev/kmsg` is forced to facility=1,
  `logger` doesn't reach the host journal from the dcgm pod, and there's no python/systemd in that pod). Wanted to get your advice
  on the direction.

  Two options I can think of:

  1. **Adjust the collection filter** — match on `_TRANSPORT=kernel` instead of `SYSLOG_FACILITY=0`. Real XIDs come from the
  driver's printk → kernel ring buffer, so they're `_TRANSPORT=kernel`; userspace noise is `_TRANSPORT=journal` and still gets
  filtered (so #1417 is still solved). The existing UAT `/dev/kmsg` injections are `_TRANSPORT=kernel`, so they'd pass unchanged.
  Downside: it accepts any `/dev/kmsg` write (which needs privilege anyway).

  2. **Rework tests.sh** — inject a real facility=0 entry at the node level, e.g. a python `journal.send(SYSLOG_FACILITY=0, ...)`
  run on the node (the dcgm pod can't do it). This keeps the strict facility=0 filter, but the injection has to move out of the
  dcgm pod to the node.

  I'm leaning toward option 1 (smaller change, and `_TRANSPORT=kernel` is arguably a more accurate "kernel-origin" definition), but
  wanted to check what you'd recommend before implementing.

XRFXLP:
hey @jackyliusohu,

I think it is better to have `_TRANSPORT=kernel` as that would syslog health monitor to be generic (if we're to capture logs from other device which might not have syslog facility set to `kern`) and at the same time avoid general noise like audits.

jackyliusohu:
> hey @jackyliusohu,
> 
> I think it is better to have `_TRANSPORT=kernel` as that would syslog health monitor to be generic (if we're to capture logs from other device which might not have syslog facility set to `kern`) and at the same time avoid general noise like audits.
 
 Thanks @XRFXLP  — switched the kernel-origin checks to _TRANSPORT=kernel instead of SYSLOG_FACILITY=0 (in beeb2df9). It's more generic (matches kernel-origin
  messages regardless of facility), stays a single AddMatch, and naturally excludes journal/syslog and audit noise. The cursor re-read fix is unchanged.

  Validated end-to-end on a real GB300 (arm64) cluster in dry-run:
  - Detected: injected via the exact UAT command (... NVRM: Xid (PCI:...): 79 ... > /dev/kmsg, which is _TRANSPORT=kernel, SYSLOG_FACILITY=1) → reported once (no
  flooding), persisted to MongoDB, node condition SysLogsXIDError=True. The old SYSLOG_FACILITY=0 filter dropped these, since the kernel forces facility 1 on 
  /dev/kmsg writes — which is why the UAT injections were being missed.
  - Ignored: the same NVRM: Xid text sent as a userspace log (logger, _TRANSPORT=syslog) was correctly not reported.

XRFXLP (approved):

lalitadithya:
we unfortunately had to revert this https://github.com/NVIDIA/NVSentinel/pull/1473 since this is causing failures in GPU reset that weren't being caught in CI

PR #1436, opened 2026-06-29
feat: allow namespace-scoped override for preflight gangDiscovery config

### LINKED ISSUE #1420: [Feature]: Allow preflight tests to have namespace scoped configuration
### Prerequisites

### Code of Conduct

### Feature Summary

In the current preflight tests, we just have one config for gangDiscovery for all the namespace in a cluster. This feature request is scoping it down to a namespace and allowing different namespace to have different config.

### Problem/Use Case

We can have different scheduler system working depending on namespace.

### Proposed Solution

TBD

### Component

Preflight

XRFXLP:
## Summary
Allow preflight gang discovery to be configured **per namespace** instead of only cluster-wide. This adds a namespaced **`PreflightConfig`** CRD (`preflight.nvsentinel.nvidia.com/v1alpha1`): create one in a namespace with a `spec.gangDiscovery` block and pods in that namespace use that discoverer, while every other namespace falls back to the cluster-wide `gangDiscovery` Helm value. It's reconciled at runtime (no Helm upgrade and no controller restart) so a single preflight deployment can serve namespaces running different gang-scheduling systems (e.g. Volcano in one, native Kubernetes or Run:ai/OSMO in another).
Closes #1420.

### What changed
- **New CRD API** (`preflight/api/v1alpha1`): namespaced `PreflightConfig` with `spec.gangDiscovery` (same schema as the Helm value) and a status subresource (`ready`, `discoverer`, `message`, conditions). `spec` is intentionally a container so future per-namespace settings can be added without a new Kind. Adds a `make generate` target (controller-gen, pinned via `.versions.yaml`) producing the DeepCopy code and the chart-shipped CRD manifest.
- **Resolver** (`preflight/pkg/gang/resolver.go`): `DiscovererResolver` is a thread-safe registry holding the cluster-wide default plus a namespace→discoverer map, resolved via `For(namespace)` and mutated via `Set`/`Remove`.
- **Reconciler** (`preflight/pkg/controller`): a `PreflightConfig` controller builds discoverers with the existing `NewDiscovererFromConfig` factory (validating `podGroupGVR`/native GVK against the cluster RESTMapper), updates the resolver, and reports outcome in status. It recomputes the whole namespace per event, so add/update/delete all converge. Edge cases:
  - **Multiple per namespace:** the oldest (tie-broken by name) wins and stays active so a working namespace isn't disrupted; extras are marked not ready (superseded).
  - **Invalid/unresolvable config:** the namespace falls back to the default and the error is surfaced in status — admission is never disrupted.
- **Wiring**: the webhook injector and gang controller resolve the discoverer per `pod.Namespace`; `main.go` registers the scheme and starts the reconciler.
- **Helm chart**: ships the CRD; grants RBAC for `preflightconfigs` (+`/status`); and reads scheduler `PodGroup` CRs cluster-wide via an **aggregated ClusterRole** (built-in contributor for native `scheduling.k8s.io` + the default `podGroupGVR`, extensible by labeling a `ClusterRole` `preflight.nvsentinel.nvidia.com/aggregate-to-gang-discovery: "true"`).
- **Docs**: `docs/configuration/preflight.md` (per-namespace section + RBAC) and `docs/designs/026-preflight-checks.md` (resolver, precedence, conflict, RBAC).
Fully backward compatible: with no `PreflightConfig` objects, every namespace uses the cluster-wide `gangDiscovery` default exactly as before. No changes to gang IDs or coordination ConfigMaps (namespace was already embedded there).

### Example
```yaml
# team-a uses Volcano; every other namespace uses the cluster-wide default.
apiVersion: preflight.nvsentinel.nvidia.com/v1alpha1
kind: PreflightConfig
metadata:
  name: default
  namespace: team-a
spec:
  gangDiscovery:
    name: volcano
    annotationKeys: ["scheduling.k8s.io/group-name"]
    podGroupGVR: {group: scheduling.volcano.sh, version: v1beta1, resource: podgroups}
    minCountExpr: podGroup.spec.minMember
---
# Platform-level RBAC so the controller may read Volcano PodGroups (aggregated in).
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: preflight-gang-discovery-volcano
  labels:
    preflight.nvsentinel.nvidia.com/aggregate-to-gang-discovery: "true"
rules:
  - apiGroups: ["scheduling.volcano.sh"]
    resources: ["podgroups"]
    verbs: ["get", "list", "watch"]
```

### Unit / integration
- CRD reconciler tests: registers the namespace discoverer + sets status ready; invalid config falls back to default with an error status; **oldest-wins** on multiple objects (winner stays ready, extras superseded); removal converges back to the default.
- Resolver `For`/`Set`/`Remove`; `NewResolverFromConfig` against a real RESTMapper (builds + validates the default, fails fast on a bad GVR); injector/handler resolve per namespace.

### Manual (real cluster, Volcano installed)
Built the image, applied the CRD + RBAC, and rolled the preflight deployment (cluster-wide default kept as `osmo-with-kai`):
- **Clean startup:** both the `pod` (gang) and new `preflightconfig` controllers start and watch the CRD; `Gang coordination enabled defaultDiscoverer=osmo-with-kai`.
- **Override path:** a `PreflightConfig` (`spec.gangDiscovery: volcano`) in `volcano-test` reconciled to `READY=true, DISCOVERER=volcano`; a Volcano job there was injected with `GANG_ID=volcano-volcano-test-…`, a `preflight-volcano-…` gang ConfigMap was created, and logs show `discoverer=volcano`.
- **Default path:** a pod in a namespace with no `PreflightConfig` was injected with `GANG_ID=osmo-with-kai-…` — the unchanged cluster-wide default applies.
- **Conflict (oldest-wins):** adding a second `PreflightConfig` left the original `volcano` config `Ready` and active while the new one went not-ready (`… "default" is already active in this namespace; only one is allowed`); deleting it converged back. The working namespace was never disrupted.

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [ ] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added a namespaced `PreflightConfig` CRD to override gang discovery per namespace at runtime.
  * Enabled namespace-aware discovery so pod behavior follows the effective namespace policy.

* **Bug Fixes**
  * Improved scheduler RBAC when enabled, including aggregated access, explicit `PreflightConfig` RBAC (with `/status`), and consistent native scheduling permissions (plus optional PodGroup API access).

* **Documentation**
  * Expanded preflight docs and chart reference with precedence, empty-config behavior, conflict/superseded rules, status visibility, and RBAC aggregation expectations.


preflight/pkg/gang/resolver.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	return NewResolver(defaultDiscoverer, byNamespace), nil
+}
+
+// For returns the gang discoverer for the given namespace, falling back to the
+// cluster-wide default discoverer when the namespace has no override.
+func (r *DiscovererResolver) For(namespace string) GangDiscoverer {
+	if r == nil {
```
lalitadithya: when would r be nil?

XRFXLP: when gang coordination is disabled

lalitadithya: right, but don't we already have a function to check if it enabled? do we have do path where we call this function before the enabled check is done?

preflight/api/v1alpha1/preflightconfig_types.go
```diff
@@ -0,0 +1,140 @@
...
+// PreflightConfigStatus reports the observed state of a PreflightConfig.
+type PreflightConfigStatus struct {
+	// ObservedGeneration is the .metadata.generation last processed by the controller.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// Ready is true when the configuration was successfully applied (e.g. the
+	// gang discovery config resolved into an active discoverer for the namespace).
+	// +optional
+	Ready bool `json:"ready,omitempty"`
```
lalitadithya: I see we have Conditions below shouldn't ready be one of the conditions?

preflight/api/v1alpha1/preflightconfig_types.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// Discoverer is the resolved gang discoverer name in effect for the namespace
+	// (e.g. "volcano", "kubernetes"). Empty when not ready.
+	// +optional
+	Discoverer string `json:"discoverer,omitempty"`
+
+	// Message is a human-readable explanation of the current state, populated
+	// when the configuration is invalid or conflicts with another object.
+	// +optional
+	Message string `json:"message,omitempty"`
```
lalitadithya: I see we have Conditions below can this message be one of the conditions?

lalitadithya (commented):
a braoder question: would this be racy if the configuration in a namespace is changed mid pod group? If yes, how can we guard against that? Should this be documented?

preflight/api/v1alpha1/preflightconfig_types.go
XRFXLP: done

preflight/api/v1alpha1/preflightconfig_types.go
XRFXLP: done

XRFXLP:
> would this be racy if the configuration in a namespace is changed mid pod group? If yes, how can we guard against that? Should this be documented?

Yes, it would be racy but fail-safe i.e. any mess up in gang coordination will lead to gang timeout -- which is a non-fatal error. I've added document advising not to do that.

lalitadithya (approved):

PR #1438, opened 2026-06-30
fix: sentinel message should be matched case sensitivily

XRFXLP:
## Summary

`platform-connectors` could leave a node condition permanently stuck at `Status=True` / `Reason=<Check>IsNotHealthy` with a recovery-looking message (e.g. `"No health failures;"`), even after a healthy event was received and processed.

### Root cause

`parseMessages` recognized the recovery sentinel only via an exact, **case-sensitive** comparison against `NoHealthFailureMsg = "No Health Failures"`. Any stored message that wasn't byte-identical i.e. different casing, a trailing `;`, or a value written by an external tool was parsed as a real fault line. 

`processNodeCondition` then took the `len(messages) > 0` branch and re-asserted `Status=True`. Because the phantom line carries no entity token, the recovery path (`removeImpactedEntitiesMessagesScoped`) can never remove it, so the condition stays wedged while `LastHeartbeatTime` keeps advancing. This makes a one-time, non-canonical recovery message compound into a growing set of false-positive `True` conditions.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review


deesharma24:
LGTM!!

lalitadithya (approved):

PR #1443, opened 2026-07-01
fix: node-drainer cold start replays stale events with ended quarantine session

### LINKED ISSUE #1347: [Bug]: ND cold-start replays stale quarantine events from ended sessions, causing orphaned remediation-failed labels
### Prerequisites

### Code of Conduct

## What happened

When node-drainer restarts, its cold-start logic (`handleColdStart` in `node-drainer/main.go`) queries for health events with `nodequarantined=Quarantined/AlreadyQuarantined` and `userpodsevictionstatus.status=""/"NotStarted"`. However, it does not check whether the quarantine session has already ended (i.e., whether a later `UnQuarantined` event exists for that node).

This causes node-drainer to re-process old health events from quarantine sessions that fault-quarantine already resolved. Node-drainer marks these stale events as `drain-succeeded`, which triggers fault-remediation to process them. If the `recommendedAction` is unsupported (e.g., `CONTACT_SUPPORT`), fault-remediation sets `dgxc.nvidia.com/nvsentinel-state=remediation-failed` on an already-uncordoned, healthy node.

In the observed case, the old events received `drainfinishtimestamp` days after the node had already been uncordoned, and FR labeled the node `remediation-failed` within milliseconds of those drain status updates.

Since the quarantine annotation was already removed by fault-quarantine during the original uncordon, there is no later cleanup path to remove this orphaned label.

## What I expected

Node-drainer cold-start should not replay health events from quarantine sessions that have already ended. Old `Quarantined`/`AlreadyQuarantined` records should be skipped if a later `UnQuarantined` or `Cancelled` event exists for the same node.

## Impact

- Uncordoned, healthy nodes end up with a stale `dgxc.nvidia.com/nvsentinel-state=remediation-failed` label permanently.
- This is a cluster-wide issue: a single ND restart can affect multiple nodes simultaneously.
- The label is never cleaned up automatically because fault-quarantine has already removed the quarantine annotation and fault-remediation's `handleCancellationEvent` does not remove the state label.

#### Root Cause

The cold-start query in `node-drainer/main.go:handleColdStart` selects stale records like this:

```go
query.Or(
    query.Eq("healtheventstatus.userpodsevictionstatus.status", string(model.StatusInProgress)),
    query.And(
        query.Eq("healtheventstatus.nodequarantined", string(model.Quarantined)),
        query.In("healtheventstatus.userpodsevictionstatus.status",
            []interface{}{"", string(model.StatusNotStarted)}),
    ),
    query.And(
        query.Eq("healtheventstatus.nodequarantined", string(model.AlreadyQuarantined)),
        query.In("healtheventstatus.userpodsevictionstatus.status",
            []interface{}{"", string(model.StatusNotStarted)}),
    ),
)
```

This query has no guard against ended quarantine sessions. The live-stream path has cancellation guards (`handleEventCancellation`, `isEventCancelled`) and terminal status checks, but the cold-start path bypasses those checks.

A deeper issue is that ND's live cancellation tracking is in-memory. During normal live processing, an `UnQuarantined` event marks `cancelledNodes[nodeName]`, so queued drain work can be cancelled. After an ND restart, that in-memory cancellation state is lost. Cold-start then relies only on datastore fields and does not reconstruct whether the node had already been unquarantined.

This means old records with `nodequarantined=Quarantined/AlreadyQuarantined` and empty drain status remain eligible, even if the same node has a later `UnQuarantined` event.

Additionally, `handleCancellationEvent` in `fault-remediation/pkg/reconciler/reconciler.go` clears the remediation annotation but does not remove the `dgxc.nvidia.com/nvsentinel-state` label. If fault-remediation processes stale events after fault-quarantine has already uncordoned the node, the label persists.

### Component

Health Monitor

### Steps to Reproduce

1. A node gets quarantined by fault-quarantine for a transient issue, for example `GpuDcgmConnectivityFailure` with `recommendedAction=CONTACT_SUPPORT`.
2. The issue resolves quickly. Fault-quarantine uncordons the node and removes `dgxc.nvidia.com/nvsentinel-state`.
3. The old health event records remain in the datastore with `nodequarantined=Quarantined/AlreadyQuarantined` and `userpodsevictionstatus.status=""` because node-drainer did not process them before the quarantine ended.
4. Node-drainer restarts due to pod restart, rollout, node eviction, etc.
5. Node-drainer cold-start queries the datastore and finds the old records.
6. Node-drainer marks them `drain-succeeded`.
7. Fault-remediation processes them, sees an unsupported action, and sets `dgxc.nvidia.com/nvsentinel-state=remediation-failed` on the already-uncordoned node.
8. The label is never cleaned up.

### Environment

- NVSentinel version: Latest as well as older versions

### Logs/Output

```json
{"level":"INFO","msg":"Querying for events requiring processing","module":"node-drainer"}
{"level":"INFO","msg":"Labeling node","module":"fault-remediation","node":"<node>","from":"drain-succeeded","to":"remediation-failed"}
{"level":"WARN","msg":"Invalid state transition","module":"fault-remediation","node":"<node>","from":"drain-succeeded","to":"remediation-failed","error":"unexpected state transition: drain-succeeded -> remediation-failed (expected one of: [remediating])"}
{"level":"INFO","msg":"Labeling node","module":"node-drainer","node":"<node>","from":"remediation-failed","to":"drain-succeeded"}
{"level":"INFO","msg":"Labeling node","module":"fault-remediation","node":"<node>","from":"drain-succeeded","to":"remediation-failed"}
{"level":"INFO","msg":"Removing label key from node","module":"fault-quarantine","key":"dgxc.nvidia.com/nvsentinel-state","node":"<node>"}
```

XRFXLP:
## Summary

Fixes #1347.

When node-drainer restarts, its cold-start logic re-queries the datastore for quarantine events that still need draining. It never checked whether the quarantine session had already ended, so it would replay stale `Quarantined`/`AlreadyQuarantined` records from sessions that fault-quarantine had already resolved (node uncordoned). Node-drainer marked those replayed records `drain-succeeded`; fault-remediation then processed them and, for an unsupported action (e.g. `CONTACT_SUPPORT`), stamped `dgxc.nvidia.com/nvsentinel-state=remediation-failed` on an already-healthy, uncordoned node. Because fault-quarantine had already removed the quarantine annotation during the original uncordon, there was no cleanup path and the label persisted forever. A single ND restart could do this to many nodes at once.

## Root cause

- Health events are stored one document per emission. Fault-quarantine writes a status only to the document it is currently processing, so on recovery the `UnQuarantined` status lands on the new healthy-event doc.
- Node-drainer's live cancellation tracking (`cancelledNodes`) is in-memory and is lost on restart. Cold start trusted the datastore match alone, with no guard against ended sessions, so it replayed those stale records.

## Fix

In node-drainer cold start, before re-queuing a `Quarantined`/`AlreadyQuarantined` candidate, check whether a later `UnQuarantined`/`Cancelled` event exists for the same node (compared by `createdAt`). If the session has already ended:

- the event is not re-queued, and
- the record is tombstoned (`userpodsevictionstatus.status = Cancelled`) so it drops out of future cold-start queries and does not match fault-remediation's change-stream pipeline (which requires `Succeeded`/`AlreadyDrained`) — so no remediation label is applied.

Fail-safe by design: `UnQuarantined`/`Cancelled` records are never tombstoned (they must run to completion), and if session state can't be determined or the candidate has no usable timestamp, the event is re-queued rather than silently dropped. Session-end lookups are cached per node.

## Changes

- `node-drainer/coldstart.go` (new) — cold-start flow plus `quarantineSessionResolver` (per-node cached lookup of the latest `UnQuarantined`/`Cancelled` marker), the staleness gate, and the tombstone write. `handleColdStart`/`dataStoreAdapter` moved here out of `main.go` (no behavior change).
- `node-drainer/coldstart_test.go` (new) — unit tests for the resolver.
- `node-drainer/main.go` — cold-start logic extracted.

## Manual testing

Common trick for all cases: node-drainer is scaled to 0 during setup and its resume token is deleted before scaling back up, so the change stream can't replay and cold start's DB query is the only path to the events.   `CONTACT_SUPPORT` is an unsupported action, so any node that reaches `drain-succeeded` gets `remediation-failed` from fault-remediation — that's the signal used to detect the bug.

### 1. Core bug: reproduced on `main`, fixed on this branch

Steps: scale ND to 0 → fatal `CONTACT_SUPPORT` event (FQ cordons; record sits `Quarantined` + empty drain status) → healthy event (FQ uncordons + removes the state label; stale record untouched) → delete ND resume token → scale ND to 1.

Before (main): node ended `Ready` (uncordoned) with an orphaned `remediation-failed` label:

```
NAME          STATUS   ROLES   AGE   VERSION       NVSENTINEL-STATE
kwok-node-1   Ready    agent   36m   kwok-v0.8.0   remediation-failed
```

After (this branch): node stayed `Ready` with no state label. The stale record was tombstoned and ND logged the skip:

```
// before scale-up
{"nodequarantined":"Quarantined","userpodsevictionstatus":{"status":""}}          // stale (11:25:17)
{"nodequarantined":"UnQuarantined","userpodsevictionstatus":{"status":""}}        // session-end marker (11:26:06)
// after cold start on the fixed build
{"nodequarantined":"Quarantined","userpodsevictionstatus":{"status":"Cancelled"}} // tombstoned
{"nodequarantined":"UnQuarantined","userpodsevictionstatus":{"status":""}}
```

### 2. Overlapping faults — one recovered, the other still needs processing

Setup on one node (ND down): fatal `faultA` (`GpuXidError`) and fatal `faultB` (`GpuDcgmConnectivityFailure`), then a healthy event for `faultA` only. FQ treats this as partial recovery — the healthy event is recorded with `nodequarantined=""` (no `UnQuarantined`/`Cancelled` session-end marker) and the node stays cordoned. Then delete token + scale ND to 1.

Expected: with no session-end marker, neither fault is stale, so both are processed and the still-needed fault is never dropped.

```
// DB after cold start — both processed, nothing tombstoned
{"checkname":"GpuXidError","message":"A-faultA"}              nodequarantined=Quarantined         userpodsevictionstatus.status=Succeeded
{"checkname":"GpuDcgmConnectivityFailure","message":"A-faultB"} nodequarantined=AlreadyQuarantined  userpodsevictionstatus.status=Succeeded
{"checkname":"GpuXidError","message":"No health failures"}   nodequarantined=""                   userpodsevictionstatus.status=""

// node-drainer log — two Re-queued, zero skips for this node
coldstart.go "Re-queued event from cold start" nodeName=kwok-node-11
coldstart.go "Re-queued event from cold start" nodeName=kwok-node-11
```

Node ended `cordoned + remediation-failed`, which is correct — it's genuinely still quarantined for `faultB` (not the orphan case, which is uncordoned + remediation-failed).

### 3. Multiple quarantine sessions at different times

Setup on one node (ND down): session 1 quarantine → healthy event (session ends → `UnQuarantined` marker), then session 2 quarantine kept active (no recovery). Then delete token + scale ND to 1. The stale session-1 record and the active session-2 record land in the same cold-start batch.

Expected: session 1's record is older than the `UnQuarantined` marker → tombstoned; session 2's record is newer than the marker → processed.

```
// DB after cold start
{"message":"B-session1","createdAt":"...13:58:53Z"} nodequarantined=Quarantined   userpodsevictionstatus.status=Cancelled   // tombstoned
{"message":"No health failures","createdAt":"...13:59:05Z"} nodequarantined=UnQuarantined                                   // session-end marker
{"message":"B-session2","createdAt":"...13:59:18Z"} nodequarantined=Quarantined   userpodsevictionstatus.status=Succeeded   // processed

// node-drainer log — cold start batch of candidates handled in one pass
coldstart.go "Processing cold start batch" count=4
coldstart.go "Skipping stale cold start event from an ended quarantine session" nodeName=kwok-node-10 eventCreatedAt=...13:58:53Z
coldstart.go "Re-queued event from cold start" nodeName=kwok-node-10
coldstart.go "Cold start processing completed"
```

Old session's record cleanly tombstoned while the new session's record is processed. Node ended `cordoned + remediation-failed` (session 2 is an active `CONTACT_SUPPORT` fault), which is the correct state.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added automatic cold-start recovery so node-drainer resumes processing previously missed health events after restarts.
  * Recovery is quarantine-aware: it re-enqueues eligible events and cancels stale quarantine records when the session has ended.
  * Enabled an efficient “latest matching health event” datastore query to improve session-end detection.
* **Bug Fixes**
  * Ensures recovery does not treat stale/resolved quarantine states (including zero/empty timestamps) as still active.
* **Tests**
  * Added unit tests for quarantine-session evaluation and per-node caching; updated datastore mocks/behavioral contracts for the new query.


node-drainer/pkg/coldstart/coldstart.go
```diff
@@ -0,0 +1,344 @@
...
+	dbAdapter := &dataStoreAdapter{DatabaseClient: components.DatabaseClient}
+	resolver := newQuarantineSessionResolver(healthStore)
+
+	err := healthStore.FindHealthEventsByQueryBatched(ctx, q, coldStartBatchSize,
+		func(batch []datastore.HealthEventWithStatus) error {
+			slog.Info("Processing cold start batch", "count", len(batch))
+
+			for i := range batch {
+				processColdStartEvent(ctx, components, healthStore, dbAdapter, resolver, batch[i])
+			}
```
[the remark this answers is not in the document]
XRFXLP: We don't want to stop the node-drainer normal proessing if we encountered some error in cold start.

node-drainer/coldstart.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	return info.latest.After(eventCreatedAt), nil
+}
+
+// lookupLatestSessionEnd finds the newest UnQuarantined/Cancelled event for a node.
+func (r *quarantineSessionResolver) lookupLatestSessionEnd(
+	ctx context.Context, nodeName string,
+) (sessionEndInfo, error) {
+	q := query.New().Build(
```
lalitadithya: would this work for:
1. overlapping events: Say you have two faults on the same node one of them is cancelled while the other still needs to be processed
2. multiple quarantine session at different points in time: First session starts, node is quarantined, draining starts and then is cancelled before completion and then some time a later a new quarantine session starts which will not be cancelled.

node-drainer/pkg/coldstart/coldstart.go
lalitadithya, line 1: this should go into a package -- `node-draining/pkg/coldstart` or something equivalent

node-drainer/coldstart.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	return nil
+}
+
+// quarantineSessionResolver answers, per node, whether a quarantine record predates a
+// later UnQuarantined/Cancelled event (i.e. its quarantine session has already ended).
+// Results are cached per node because a cold-start batch commonly contains multiple
+// records for the same node.
+type quarantineSessionResolver struct {
```
lalitadithya: would recommend making this it's own file inside the package

node-drainer/coldstart.go
XRFXLP: yes, tested and verified. Added details in MR description.

node-drainer/pkg/coldstart/coldstart.go
XRFXLP: done, PTAL

node-drainer/coldstart.go
XRFXLP: done

node-drainer/pkg/coldstart/resolver.go
```diff
@@ -0,0 +1,115 @@
...
+) (sessionEndInfo, error) {
+	q := query.New().Build(
+		query.And(
+			query.Eq("healthevent.nodename", nodeName),
+			query.In("healtheventstatus.nodequarantined",
+				[]interface{}{string(model.UnQuarantined), string(model.Cancelled)}),
+		),
+	)
+
+	events, err := r.finder.FindHealthEventsByQuery(ctx, q)
```
KaivalyaMDabhadkar: `FindHealthEventsByQuery` fetches every past UnQuarantined/Cancelled event for this node, even though we would need only the one event with the latest `createdAt`. As this current query runs once for each distinct node, a node with a large number of such events can load far more than 1000 full events into ND's memory and lead to an OOM during cold start. Can we instead simply write a query like "Find events for this node where nodequarantined is UnQuarantined or cancelled and return only the one with the latest createdat?"

XRFXLP: Handled, PTAL.

KaivalyaMDabhadkar (approved):

PR #1453, opened 2026-07-03
fix: test flake fix for TestNICCounterIBDegradation

### LINKED ISSUE #1446: [Bug]: Flaky E2E: TestNICCounterIBDegradation fails when TestDCGMBootstrapCompletedAnnotation's DCGM pod deletion cordons the shared node
### Prerequisites

### Code of Conduct

## What happened

E2E job [`E2E Tests (AMD64 + PostgreSQL)`](https://github.com/NVIDIA/NVSentinel/actions/runs/28572910522/job/84714597007) failed on `main` (commit `f0a680dc`, an unrelated Dependabot docker bump):

```
--- FAIL: TestNICCounterIBDegradation/NIC_Counter_-_IB_Degradation (5.05s)
    nic-health-monitor.go:461: node nvsentinel-worker is still cordoned
    Error:      Condition satisfied
    Messages:   NIC cleanup should remain stable on node nvsentinel-worker
```

The failure is the `require.Never` stability check in `waitForNICCleanupComplete` (`tests/helpers/nic-health-monitor.go:454`): the node was clean when the test started, then got cordoned mid-way through the 10-second stability window.

## Root cause

The failing test is a result of test pollution from the **preceding** test, `TestDCGMBootstrapCompletedAnnotation` (added in #1425). Both tests pick the same worker node (`nvsentinel-worker`), and NVSentinel's fault pipeline is asynchronous, so a fault triggered by the first test lands as a cordon while the second test is running.

Timeline reconstructed from the job log and the `e2e-kind-container-logs` artifact (fault-quarantine / platform-connector logs):

| Time (UTC) | Event |
|---|---|
| 07:46:44.63 | `TestDCGMBootstrapCompletedAnnotation` deletes the `nvidia-dcgm` pod on `nvsentinel-worker` |
| 07:46:49.65 | Replacement DCGM pod Ready → test asserts annotation persisted, **passes and ends (no teardown)** |
| 07:46:49.68 | `TestNICCounterIBDegradation` starts on the same node; node is clean → `require.Eventually` passes; 10s `require.Never` stability window opens |
| 07:46:53.22 | gpu-health-monitor's periodic health check fires while its DCGM connection is still broken → emits fatal `GpuDcgmConnectivityFailure` (`DCGM_CONNECTIVITY_ERROR`, "Failed to connect to DCGM for health check") |
| 07:46:53.23 | fault-quarantine matches "GPU fatal error ruleset" → **cordons `nvsentinel-worker`** |
| 07:46:54.70 | `require.Never` tick sees `node.Spec.Unschedulable == true` → test fails |

Key fault-quarantine log line:

```json
{"time":"2026-07-02T07:46:53.228574212Z","level":"INFO","msg":"Cordoning node","node":"nvsentinel-worker"}
```
triggered by event `checkName=GpuDcgmConnectivityFailure, isFatal=true, errorCode=["DCGM_CONNECTIVITY_ERROR"], nodeName=nvsentinel-worker` (generated 07:46:53.21 — ~3.5s **after** the DCGM test already passed).

Notes:

- The post-bootstrap DCGM connectivity loss is intentionally fatal (that's the design from #1425). The bug is purely test isolation.
- `TestDCGMBootstrapCompletedAnnotation` only waits for the replacement pod to be Ready; it does not wait for the gpu-health-monitor's DCGM connection to recover, nor for the resulting quarantine to clear. Its sibling `TestGPUHealthMonitorDCGMConnectionError` (same file) *does* wait for the `GpuDcgmConnectivityFailure` condition to return to healthy and has a `Teardown` — the new test lacks both.
- The NIC test's defensive cleanup (`clearNICConditions`) only sends healthy **NIC** events, which cannot clear a **GPU**-caused quarantine, so it can't protect itself from this pollution.
- Flakiness window: it only reproduces when the gpu-health-monitor's periodic check timer fires between the DCGM pod deletion and connection recovery, and the resulting cordon lands inside the next test's 10s `Never` window. All E2E runs on `main` since #1425 merged (2026-06-29) passed until this one (2026-07-02).

## What I expected

E2E tests to be isolated: a passing test should not leave in-flight fatal health events that cordon the node used by the next test.

## Suggested fix

1. **Primary:** Add a recovery wait / `Teardown` to `TestDCGMBootstrapCompletedAnnotation` (`tests/gpu_health_monitor_test.go`): after the replacement DCGM pod is Ready, `require.Eventually` that the `GpuDcgmConnectivityFailure` condition is healthy and the node is not cordoned / has no active quarantine annotation — mirroring the existing pattern in `TestGPUHealthMonitorDCGMConnectionError`.
2. **Secondary hardening:** Make `waitForNICCleanupComplete` (or the NIC tests' setup) tolerate/clear pre-existing non-NIC quarantines before opening the stability window.

### Component

Other

### Steps to Reproduce

Deterministic reproduction (no need to wait for the flake): on a running E2E cluster, delete the `nvidia-dcgm` pod on a worker node and immediately run `TestNICCounterIBDegradation` targeting that node — the fatal `GpuDcgmConnectivityFailure` cordon lands during the stability window.

### Environment

- NVSentinel version: v1.12.0

### Logs/Output

Link to failed run: https://github.com/NVIDIA/NVSentinel/actions/runs/28572910522/job/84714597007

KaivalyaMDabhadkar:
## Summary

Fixes a flaky E2E failure where `TestNICCounterIBDegradation` fails with `node nvsentinel-worker is still cordoned` due to test pollution from the preceding `TestDCGMBootstrapCompletedAnnotation`.

### Root Cause

`TestDCGMBootstrapCompletedAnnotation` (introduced in #1425) deletes the `nvidia-dcgm` pod to verify that the bootstrap annotation persists across pod restarts. However, deleting the DCGM pod breaks the gpu-health-monitor's DCGM connection. When the monitor's next 15-second periodic health check fires against the broken connection, it emits a **fatal `GpuDcgmConnectivityFailure`** event (`DCGM_CONNECTIVITY_ERROR`). This is correct product behavior — post-bootstrap, DCGM connectivity loss is treated as a real node fault.

The problem: the test only waited for the replacement DCGM pod to be Running/Ready, then ended — it never waited for the gpu-health-monitor to reconnect to DCGM. The fatal event propagated asynchronously through the pipeline (gpu-health-monitor → platform-connector → PostgreSQL → fault-quarantine) and arrived **after** the test had already passed, cordoning the node ~3.5 seconds later. When `TestNICCounterIBDegradation` started immediately after on the same node, its `require.Never` stability check saw the unexpected cordon and failed.

The flake was timing-dependent: it only reproduced when (a) the monitor's 15s poll tick fired during the DCGM connection gap, and (b) both tests selected the same worker node.

### Fix

Two changes to `TestDCGMBootstrapCompletedAnnotation` in `tests/gpu_health_monitor_test.go`, following the existing pattern from its sibling test `TestGPUHealthMonitorDCGMConnectionError`:

1. **Setup labels the node `ManagedByNVSentinel=false`** before deleting the DCGM pod. Fault-quarantine's rulesets (CEL expressions in `charts/fault-quarantine/values.yaml`) skip nodes with this label, so the fatal `GpuDcgmConnectivityFailure` cannot cordon the node — neither during this test nor after it ends.

2. **New Teardown waits for the full failure→recovery cycle** before handing the node back:
   - Waits up to 50s (covers three 15s poll cycles plus pipeline latency) for the `GpuDcgmConnectivityFailure` condition to appear. If it doesn't appear, the monitor's connection survived the restart — no cleanup needed.
   - If the condition appeared, waits with `require.Eventually` for it to become healthy (monitor successfully reconnected to DCGM).
   - Runs a 10s `require.Never` stability check to confirm the condition stays healthy.
   - Removes the `ManagedByNVSentinel` label, restoring the node to managed state.

### Test Plan

Tested locally against a Kind cluster (`make cluster-create` → `make tilt-up` → `make -C tests test`):

- **Without the fix (git stash):** Ran `TestDCGMBootstrapCompletedAnnotation|TestNICCounterIBDegradation` 3 times. On the 3rd run, both tests selected `nvsentinel-worker` and `TestNICCounterIBDegradation` failed with `node nvsentinel-worker is still cordoned` — same failure signature as CI.
- **With the fix (git stash pop):** Ran the same test pair 6 times, including runs where both tests landed on the same node. All passed. The teardown log shows the fix in action: `Waiting up to 50s for GpuDcgmConnectivityFailure to appear` → condition detected → `Waiting for GpuDcgmConnectivityFailure to become healthy` → recovery confirmed → label removed.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Tests**
  * Strengthened the GPU health monitor test to more reliably validate connectivity recovery after restarting DCGM.
  * Added a deadline-bounded observation window that polls for connectivity failure recovery before proceeding.
  * Added stability checks for observed healthy connectivity, and now verifies the connectivity failure signal is absent afterward.
  * Preserves and restores the node’s original management label state to avoid side effects in later test runs.

### DIFF
--- a/tests/gpu_health_monitor_test.go
+++ b/tests/gpu_health_monitor_test.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	"tests/helpers"
 
@@ -43,6 +44,14 @@ const (
 	GPUHealthMonitorContainerName = "gpu-health-monitor"
 	GPUHealthMonitorDaemonSetName = "gpu-health-monitor-dcgm-4.x"
 )
+
+// dcgmConnectivityFailureObserveWindow bounds how long we wait for the fatal
+// GpuDcgmConnectivityFailure that the gpu-health-monitor emits on its first
+// poll after its DCGM connection breaks. The monitor polls every 15s
+// (PollIntervalSeconds in the gpu-health-monitor chart configmap), so three
+// poll cycles plus pipeline latency is ample; if nothing shows up by then the
+// connection survived and no fatal is in flight.
+const dcgmConnectivityFailureObserveWindow = 50 * time.Second
 
 const (
 	keyGpuHealthMonitorPodName      contextKey = "gpuHealthMonitorPodName"
@@ -957,6 +966,14 @@ func TestDCGMBootstrapCompletedAnnotation(t *testing.T) {
 		testNodeName = gpuHealthMonitorPod.Spec.NodeName
 		t.Logf("Using node: %s", testNodeName)
 
+		// Deleting the nvidia-dcgm pod below breaks the gpu-health-monitor's DCGM
+		// connection; its next poll emits a fatal GpuDcgmConnectivityFailure, which
+		// would quarantine this node and leak a cordon into whichever test runs
+		// next. Mark the node unmanaged so fault-quarantine ignores that event.
+		t.Logf("Setting ManagedByNVSentinel=false on node %s", testNodeName)
+		err = helpers.SetNodeManagedByNVSentinel(ctx, client, testNodeName, false)
+		require.NoError(t, err, "failed to set ManagedByNVSentinel label")
+
 		node, err := helpers.GetNodeByName(ctx, client, testNodeName)
 		require.NoError(t, err, "failed to get node %s", testNodeName)
 		require.NotEmpty(t, node.Annotations[dcgmBootstrapAnnotation],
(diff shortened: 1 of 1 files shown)


tests/gpu_health_monitor_test.go
```diff
@@ -1001,5 +1024,93 @@ func TestDCGMBootstrapCompletedAnnotation(t *testing.T) {
...
+				}
+			}
+		}
+		require.True(t, hadSuccessfulCheck,
+			"never got a successful condition check for node %s during observe window", testNodeName)
+
+		if sawConnectivityFailure {
+			t.Logf("Waiting for GpuDcgmConnectivityFailure to become healthy on node %s", testNodeName)
+			require.Eventually(t, func() bool {
+				condition, checkErr := helpers.CheckNodeConditionExists(ctx, client, testNodeName,
```
XRFXLP: Can we use `WaitForNodeConditionWithCheckName` and below as well?

KaivalyaMDabhadkar: Done, replaced the Eventually block with WaitForNodeConditionWithCheckName and the Never block with EnsureNodeConditionNotPresent

XRFXLP (approved):

Issue #1379, opened 2026-06-10
[Bug]: NIC monitor emits non-fatal events for expected-down ports after reboot


KaivalyaMDabhadkar:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Code of Conduct

- [x] I agree to follow NVSentinel's Code of Conduct

### What happened

After a node reboot, the NIC health monitor's first-poll suppression logic correctly
identifies expected-down cards via card homogeneity and downgrades them from FATAL to
non-fatal. However, the downgraded event is still emitted as an external `HealthEvent`
with `is_fatal=false`, `is_healthy=false`, `recommended_action=NONE`.

This creates noisy Kubernetes node events (`InfiniBandStateCheckIsNotHealthy`) for ports
that are intentionally disabled, which pollutes NIC health correlation workflows and can
be mistaken for actual NIC degradation.

### Root cause

The issue is in `portTransitionEvents()` in
[`pkg/checks/state/ibstate.go`](https://github.com/NVIDIA/NVSentinel/blob/main/health-monitors/nic-health-monitor/pkg/checks/state/ibstate.go):

```go
func (c *InfiniBandStateCheck) portTransitionEvents(
    st *ibPollState, firstPoll, baselineRun bool, expectedCards map[string]struct{},
) []*pb.HealthEvent {
    var events []*pb.HealthEvent

    for _, port := range st.ibPorts {
        key := portKey(port.Device, port.Port)
        prev, hasPrev := c.previousPorts[key]

        evt := c.evaluatePortTransition(st.currentPorts[key], prev, hasPrev, baselineRun)
        if evt == nil {
            continue
        }

        if firstPoll && !evt.IsHealthy {
            c.applyFirstPollSuppression(evt, port, st.portCard[key], expectedCards)
        }

        events = append(events, evt)  // <-- event is always appended, even after suppression
    }

    return events
}
```

`applyFirstPollSuppression()` mutates the event in-place:

```go
func (c *InfiniBandStateCheck) applyFirstPollSuppression(
    evt *pb.HealthEvent, port discovery.IBPort, card string, expected map[string]struct{},
) {
    if _, ok := expected[card]; !ok {
        return
    }

    slog.Info("Suppressing first-poll fatal for expected-down card",
        "device", port.Device, "port", port.Port, "card", card)

    evt.IsFatal = false
    evt.RecommendedAction = pb.RecommendedAction_NONE
}
```

[...]

Issue #1361, opened 2026-06-04
[Bug]: NIC monitor emits false FATAL for unprovisioned Ethernet/RoCE Aux ports on cloud shapes


KaivalyaMDabhadkar:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Code of Conduct

- [x] I agree to follow NVSentinel's Code of Conduct

## What happened

On cloud bare-metal GPU instances (e.g., OCI BM.GPU.H100.8), the frontend NIC adapter
is a dual-port card where only one port (Prime) is provisioned and active. The second
port (Aux) is intentionally disabled by the cloud provider — it has no VNIC attached,
no IP address, no routes, and `phys_state: Disabled`.

The NIC health monitor emits a **FATAL `REPLACE_VM`** event for this Aux port on every
first poll, because it sees `state: DOWN` and has no way to determine that the port is
intentionally unprovisioned.

## Why existing safety nets don't catch it

1. **Default-route exclusion** correctly removes the active Prime port (which carries
   the host's default route) from monitoring by classifying it as Management. But this
   leaves the Aux port as a **singleton** in the Storage role group.

2. **Card homogeneity** requires ≥2 cards in a role group to infer expected-down
   patterns. With only one card in the Storage group, the check is skipped entirely.
   The `ExpectedDownCards` set is empty, so first-poll suppression never fires.

3. The monitor has no concept of "administratively disabled / unprovisioned port."
   It treats all `state: DOWN` Ethernet/RoCE ports identically regardless of whether
   `phys_state` is `Disabled` (intentional) vs `Polling` (searching for link).

## What I expected

The monitor should not emit FATAL events for ports that are intentionally disabled and
have no provisioning evidence (no admin-up, no carrier history, no routes, no
bond/bridge membership, no configured VFs). These ports should be skipped from
monitoring entirely.

## Impact

- False `REPLACE_VM` recommendation on healthy nodes
- Healthy nodes may be unnecessarily drained/cordoned
- The false positive is currently masked by a separate platform-connector entity-matching
  bug (tracked independently), so the node condition appears healthy for wrong reasons.
  Fixing the connector bug alone would surface this false positive on all affected nodes.

### Component

Health Monitor

### Steps to Reproduce

1. Deploy NVSentinel on a cloud bare-metal GPU node where the frontend NIC is a
   dual-port adapter with only one port provisioned (e.g., Prime active, Aux disabled).
[...]

PR #1462, opened 2026-07-06
fix: require positive evidence for NIC fatals and add explicit inclusion override

KaivalyaMDabhadkar:
## Summary

This PR does three things:

1. **Stops false alarms on ports that are down by design** : A port the
   monitor has never seen healthy is reported FATAL only when there is positive evidence
   it should be up. Without that evidence the port is logged locally and suppressed — no
   HealthEvent, no metrics noise. Runtime detection is unchanged: `ACTIVE → DOWN` is
   always fatal.
2. **Adds `nicInclusionRegexOverride`** — an operational override that pins monitoring to
   an explicit list of NICs, bypassing every automatic filter. Built for debugging a
   suspect NIC when the rest of the stack may be broken.
3. **Fixes the state-handling bugs that restarts and config changes exposed** during
   hardware validation: fabricated "device disappeared" FATALs on scope changes, counter
   latches lost on config changes, and card-level FATAL events that could hold a node
   quarantined forever because no recovery event existed.

## 1. The peer-evidence rule (#1361, #1379)

**Old rule:** a first-sight DOWN port is failed unless proven expected-down.
**New rule:** a first-sight DOWN port is failed only if proven unexpected.

The proof is *peer evidence*: group the node's NIC cards by role, take the most common
active-port count in each group (the "mode"), and flag cards below it. This requires at
least 2 cards and an untied mode. Singletons, tied modes, and all-down groups produce no
verdict — the port is suppressed (still discovered, recorded, and persisted; just no
external event).

Why this matters: on OCI H100 shapes the frontend card has a Prime port (carries the
default route, excluded as management) and an Aux port that is intentionally never
provisioned (`phys_state: Disabled`, no VNIC). Excluding Prime leaves Aux as a singleton,
and the old rule declared it `FATAL REPLACE_VM` on every fresh start of a healthy node.
A second latent path (Prime/Aux tied 1-vs-0 when default-route detection fails) was
closed by making tied modes indecisive instead of breaking ties upward.

(#1379: previously even recognized expected-down ports still published a non-fatal event
every reboot. Suppression is now literal — no event at all.)

| Scenario (first poll, port never seen healthy) | Before | After |
|---|---|---|
| Card below its role group's decisive mode (≥2 cards) | FATAL | FATAL |
| Card at mode (uncabled second port) | Non-fatal event | Logged, no event |
| Singleton role group (OCI Aux port) | **FATAL false positive** | Logged, no event |
| Tied mode (Prime/Aux 1-vs-0) | **FATAL false positive** | Logged, no event |
| All ports down (mode 0) | FATAL per port | Logged, no events |
| Runtime `ACTIVE → DOWN` | FATAL | FATAL (unchanged) |

## 2. The inclusion override

```yaml
nic-health-monitor:
  nicInclusionRegexOverride: "^mlx5_5$"   # comma-separated regexes
```

- **Exclusive:** while set, only matching NICs are monitored; everything else is ignored.
- **Bypasses all filters:** vendor, VF, management/NUMA classification, and
  `nicExclusionRegex` — you can pin devices that normal discovery can never see.
- **Pinned NICs are reported unconditionally.** The peer-evidence gate and the card
  checks do not apply to them: pinning *is* the evidence the device is supposed to be up.
  A pinned port that is DOWN at first sight emits a FATAL.
- **Works without GPU metadata.** If `gpu_metadata.json` is missing while the override is
  active, the monitor starts anyway with a degraded classifier and a loud warning —
  the override exists precisely for situations where the metadata collector is down.
  Normal discovery keeps the hard failure ("no silent fallback").

## 3. State correctness across restarts and config changes

The monitor persists its memory in `state.json` so restarts behave like the monitor never
blinked. Hardware validation of the override exposed three problems in how that memory
interacts with configuration changes:

**a. Scope changes corrupted the seeded memory.** Changing the override (or the exclusion
regex) changes which devices discovery returns. Seeding old state under a new scope
either reported every no-longer-monitored NIC as vanished hardware (12 fatal
`REPLACE_VM`s observed on a real node when enabling the override), or made the monitor
believe it was mid-flight rather than at a first poll — bypassing the suppression gate
and reproducing the #1361 false positive when the override was removed.
*Fix:* the state file records a **scope fingerprint** (override + exclusion regex). On
mismatch, port/device state resets exactly like a reboot, with healthy baselines.

**b. Counters must not reset with it.** Hardware counters don't reset on config changes.
The scope reset preserves counter snapshots and latched breaches, and the state file
reports "scope changed" separately from "rebooted": state checks re-baseline on either;
counter checks react only to real reboots. No synthetic "healthy after reboot" events,
no lost breach latches.

**c. Card-level FATALs had no recovery.** The card-homogeneity event
(`Card X has 1 active ports, expected 2`) had no healthy counterpart. Once
fault-quarantine cordoned a node on the card entity, nothing could ever release it —
reproduced on hardware: ports recovered, node stayed quarantined until manual cleanup.
*Fix:* card anomalies are now a **latch with a full lifecycle** — one FATAL on onset,
silence while latched, one card-healthy event on recovery (card present, group decisive,
back at mode), and card-healthy baselines after reboots and scope changes. The latch is
persisted and survives pod restarts, reboots, and scope changes (including a full
override round trip), and is keyed per link layer so the IB and Ethernet checks can
latch the same physical VPI card without overwriting each other. "No evidence, no
verdict" applies here too: absent cards and indecisive groups hold the latch.

## Behavior changes to approve explicitly

1. **Card homogeneity runs every poll** (was: first poll only). A runtime failure on a
   below-mode card now emits a port FATAL *and* a card FATAL — both with matching
   recovery events.
2. **Changing `nicInclusionRegexOverride` or `nicExclusionRegex` resets port/device
   state** on the next pod start (healthy baselines re-emitted). Counter state and card
   latches survive the reset.
3. **The first deploy of this build performs that reset once** (existing state files have
   no fingerprint). Its card-healthy baselines also clear any stale card conditions left
   behind by the old no-recovery bug.

## Validation

- `go test ./...` and `go vet ./...` pass; regression tests added for: first-poll
  suppression (singleton/tied/below-mode), runtime fatality, both scope-change directions,
  pinned-VF removal, pinned-DOWN reporting (IB + Ethernet), same-scope memory retention,
  counter-latch survival, card lifecycle (onset/silence/recovery), card latch across
  restarts and override round trips, mixed-layer latch keys, and legacy state-file upgrade.
- **Real-cluster validation on OCI `BM.GPU.H100.8`**, all scenarios observed live:
  - Override on/off toggles: no disappearance storm, no false Aux-port FATAL, scope reset
    logged both ways.
  - Pinned `mlx5_11` (the designed-down Aux port): genuine FATAL emitted, as intended.
  - Monitor start with `gpu_metadata.json` removed while pinned: warning + working monitor.
  - Card lifecycle on a dual-port card: FATAL pair on link down, recovery pair 6 seconds
    later on link up; card recovery emitted after a pod restart mid-fault; card recovery
    emitted after a full down → pin → heal → unpin round trip.
  - Scope-change pods emitted zero synthetic counter recoveries.

## Trade-offs and known limitations

- A genuinely dead port in a singleton or tied group at first boot is logged but not
  externally reported — without peer evidence it is indistinguishable from an
  intentionally idle port. It is reported the moment it fails at runtime or peer
  evidence appears.
- Suppressed ports emit no events, so stale *port* conditions from before this change are
  not actively cleared (card conditions now are, via the new baselines).
- Counter checks still evaluate every discovered port, including management ports.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Clarified first-poll/post-reboot baseline and “unhealthy but suppressed” behavior using peer-evidence gating.
  * Updated guidance for tied/indecisive role groups and inclusion-override semantics.

* **New Features / Behavior Changes**
  * Added discovery-scope tracking and card anomaly latching that persists across restarts and recovers/clears appropriately.
  * Improved inclusion-override discovery and event suppression when peer evidence is absent.

* **Bug Fixes**
  * Reduced false alarms by gating first-poll DOWN events on card homogeneity evidence.

* **Tests**
  * Added regression coverage for first-poll suppression, reboot/scope persistence, override behavior, and anomaly latch lifecycles.

* **Chores / Validation**
  * Tightened inclusion override regex validation and updated Helm values documentation.


lalitadithya:
as discussed on call, let's make the nic inclusion regex as the higher priority override.

lalitadithya (approved):

PR #1464, opened 2026-07-07
fix: handle concurrent health events in fault quarantine module

### LINKED ISSUE #1463: [Bug]: fault-quarantine can lose quarantineHealthEvent entries for concurrent unhealthy events
## Bug Description

Fault-quarantine can lose health event entries in the node `quarantineHealthEvent` annotation when multiple unhealthy events for the same node are published within milliseconds of each other.

Observed behavior:

- Two unhealthy events were published for the same node almost at the same time.
- The node condition was updated with both errors.
- Fault-quarantine applied the expected quarantine taint.
- However, the node `quarantineHealthEvent` annotation contained only one of the two unhealthy events.

Expected behavior:

- The first unhealthy event should create the quarantine session and add its event to `quarantineHealthEvent`.
- Any additional unhealthy event for the same already-quarantined node should be merged into the existing `quarantineHealthEvent` annotation.
- The annotation should preserve all active unhealthy events that are keeping the node quarantined.

Actual behavior:

- Only one event is present in `quarantineHealthEvent` even though the node condition contains both errors.
- This can make fault-quarantine lose track of one active failure on the node.
- Follow-up recovery handling may become incorrect because the annotation no longer represents all active unhealthy events.

This appears to be a race in the node annotation update path. Fault-quarantine decides whether the node is already quarantined using the node informer cache. If two events arrive close together, the second event can miss the first annotation update in the cache and enter the fresh quarantine path. The fresh quarantine path builds `quarantineHealthEvent` from only the current event, and `QuarantineNodeAndSetAnnotations` can also skip updates when the live node already appears cordoned with an FQ annotation. In that case the second event is not merged into the annotation.

Relevant code paths:

- `fault-quarantine/pkg/reconciler/reconciler.go`: `handleEvent` calls `hasExistingQuarantine` before choosing the fresh quarantine path vs already-quarantined path.
- `fault-quarantine/pkg/reconciler/reconciler.go`: `getNodeQuarantineAnnotations` reads from `NodeInformer.GetNode`, so the decision depends on informer cache freshness.
- `fault-quarantine/pkg/reconciler/reconciler.go`: `applyQuarantine` creates a new annotation map for the current event.
- `fault-quarantine/pkg/informer/k8s_client.go`: `QuarantineNodeAndSetAnnotations` / `handleCordon` can skip annotation updates when the node is already cordoned and already has the FQ annotation.

## Component

Fault Management

## Steps to Reproduce

1. Configure a fault-quarantine rule that taints or cordons a node for unhealthy node-level events.
2. Publish two unhealthy events for the same node within a very small time window, before the node informer cache observes the annotation update from the first event.
3. Inspect the node condition and verify it contains both errors.
4. Inspect the node annotations:

```yaml
quarantineHealthEvent: '[...]'
quarantineHealthEventAppliedTaints: '[...]'
```

5. Observe that `quarantineHealthEvent` contains only one of the two unhealthy events.

## Environment

- NVSentinel version: unknown
- Kubernetes version: unknown
- Deployment method: Helm/Kubernetes deployment
- Datastore: MongoDB replica set

## Logs/Output

Sanitized example of the inconsistent state:

```yaml
# Node annotation added by fault-quarantine contains only one event
quarantineHealthEvent: '[{"version":1,"agent":"<redacted>","componentClass":"Node","checkName":"<redacted>","isFatal":true,"message":"<redacted>","recommendedAction":5,"errorCode":["<error-code-A>"],"entitiesImpacted":[{"entityType":"v1/Node","entityValue":"<redacted-node>"}],"metadata":{"trace_id":"<redacted>"},"generatedTimestamp":{"seconds":<seconds>,"nanos":<nanos>},"nodeName":"<redacted-node>","processingStrategy":1,"id":"<redacted-event-id>"}]'
quarantineHealthEventAppliedTaints: '[{"Key":"<fq-taint-key>","Value":"true","Effect":"NoSchedule"}]'

# Node condition contains both active errors
Conditions:
  <condition-type>: True
    Message: "ErrorCode:<error-code-A> v1/Node:<redacted-node> <redacted-message-A>; ErrorCode:<error-code-B> v1/Node:<redacted-node> <redacted-message-B>;"

# Taints include the fault-quarantine taint
Taints:
  <existing-taint-key>=<existing-taint-value>:NoSchedule
  <fq-taint-key>=true:NoSchedule
```

## Suggested Fix Direction

Make the annotation update race-resistant and merge-based.

The `Quarantined` vs `AlreadyQuarantined` decision and `quarantineHealthEvent` update should be based on the current live node state inside the per-node update critical section, not only on the informer cache state observed before the update. If the live node already has a non-empty FQ `quarantineHealthEvent` annotation, fault-quarantine should merge the new unhealthy event into that annotation instead of replacing/skipping it. The resulting status can then be classified as `AlreadyQuarantined` when appropriate.

XRFXLP:
## Summary

This PR makes fault-quarantine handling of near-concurrent health events race-resistant.

The race occurred when two unhealthy events for the same node arrived within ~90 ms of each other. Because the informer cache had not yet observed the first node annotation update, fault-quarantine took the "fresh quarantine" path for both events. In that path, the code treated “node is already cordoned with an FQ annotation” as a signal that quarantine updates were already complete, making the second event a no-op and leaving it out of `quarantineHealthEvent`.

This change removes that assumption. The existing live node update path now merges the incoming `quarantineHealthEvent` with the node’s current annotation inside `updateFn`, using the Kubernetes state fetched for the update rather than relying only on informer state. This ensures near-concurrent events are preserved in the annotation even when the informer cache lags behind live Kubernetes state.

## Testing

Added a regression test that delays informer watch updates to reproduce the stale-cache window deterministically. The test validates that near-concurrent unhealthy events for the same node are both preserved in `quarantineHealthEvent`, and that the second event is classified as `AlreadyQuarantined`.

Verified that this test fails without the fix and passes with the fix.

Note: unrelated [change](https://github.com/NVIDIA/NVSentinel/pull/1464/changes/f65e660786966822c23ab7f2ef753586c24135bb) has been added `fault-quarantine/pkg/healthEventsAnnotation/health_events_annotation_map.go` to guard againt integer overflow as per [codeQL sqan ](https://github.com/NVIDIA/NVSentinel/security/code-scanning?query=is%3Aopen+branch%3Apull-request/1464)
## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Node quarantine status now distinguishes first-time quarantine vs already-in-effect, updating “AlreadyQuarantined” appropriately.
  * Unhealthy events are accumulated into quarantine health-event annotations via JSON-aware merging.

* **Bug Fixes**
  * Improved cordon handling: if a node is already unschedulable and the quarantine annotation key exists, updates continue (taints/labels/annotations still apply).
  * Annotation merging now errors on failed parse/merge, and taints are merged with deduplication.

* **Tests**
  * Updated unit tests for the revised quarantine API return value.
  * Added E2E coverage for concurrent unhealthy quarantine under informer watch delay (replacing the concurrent healthy case).


fault-quarantine/pkg/healthEventsAnnotation/health_events_annotation_map.go
```diff
@@ -129,6 +130,18 @@ func createEntityKeys(
 	return keys
 }
 
+func eventKeysCapacity(entityCount, errorCodeCount int) int {
```
XRFXLP, line 133: This is an unrelated [change](https://github.com/NVIDIA/NVSentinel/pull/1464/changes/f65e660786966822c23ab7f2ef753586c24135bb) to guard against integer overflow as per [codeQL sqan ](https://github.com/NVIDIA/NVSentinel/security/code-scanning?query=is%3Aopen+branch%3Apull-request/1464)

fault-quarantine/pkg/informer/k8s_client.go
```diff
@@ -365,22 +375,86 @@ func (c *FaultQuarantineClient) handleCordon(ctx context.Context, node *v1.Node,
...
+		if annotationKey == common.QuarantineHealthEventAnnotationKey {
+			mergedValue, err := mergeQuarantineHealthEventAnnotation(node.Annotations[annotationKey], annotationValue)
+			if err != nil {
+				return fmt.Errorf("failed to merge annotation %q on node %s: %w", annotationKey, nodename, err)
+			}
+
+			annotationValue = mergedValue
+		}
+
 		node.Annotations[annotationKey] = annotationValue
```
KaivalyaMDabhadkar, line 408: Could we also merge the `quarantineHealthEventAppliedTaints` annotation here? If two concurrent events apply different taints, the node keeps both taints, but this annotation is overwritten with only the second events taint. During recovery, cleanup reads this annotation so the first taint would be left on the node. We can do a union of the existing and incoming applied-taint lists and add a recovery test with two different taints

KaivalyaMDabhadkar (approved):

PR #1465, opened 2026-07-07
feat: gate node-labeler and cluster-scope monitors on managed=false label (ADR-040)

jtschelling:
## Summary

Implements **JSC-105 / PR4** of the NVSentinel External Remediations project.

Nodes carrying `nvsentinel.dgxc.nvidia.com/managed=false` are opted out of NVSentinel management while an external breakfix system owns them. This PR gates all emission paths in four modules behind that label check.

### Changes by module

**labeler**
- Strip `DCGMVersion`, `DriverInstalled`, and `KataEnabled` detection labels from opted-out nodes so DaemonSet monitors self-evict via `nodeSelectors`
- Gate `updateNodeLabelsForPod` to skip pod-driven label writes for opted-out nodes
- Trigger `nodeRequiresReconciliation` when the `managed` label changes, so the strip runs promptly on opt-out
- New `labeler_node_labels_skipped_managed_total` Prometheus counter

**csp-health-monitor**
- Gate `processAndSendTrigger` after the empty-NodeName check
- Advance event DB status on skip so the event is not re-fired next poll (no-retry skip)
- New `csp_health_monitor_trigger_skipped_managed_total{trigger_type}` counter
- `setupNodeLister()` helper in `main.go` creates a SharedInformerFactory-backed NodeLister

**kubernetes-object-monitor**
- Gate `PublishHealthEvent` before strategy resolution; return error on lister failure (fail closed)
- New `k8s_object_monitor_emissions_skipped_managed_total{policy_name}` counter
- `buildNodeLister()` helper in initializer uses `ctrl.GetConfig()` + `kubernetes.NewForConfig()`

**slurm-drain-monitor**
- Gate `PublishDrainEvents`; return error on lister failure (fail closed)
- New `slurm_drain_monitor_emissions_skipped_managed_total` counter
- `buildNodeLister()` helper in initializer uses `ctrl.GetConfigOrDie()`

### Design notes

- **Fail-open in labeler**: detection-label stripping is best-effort; a lister error logs a warning and proceeds rather than blocking reconciliation
- **Fail-closed in publishers**: a lister error surfaces as an error return so the caller can decide on retry/backoff rather than silently dropping events
- **Informer-backed lister**: avoids live API calls on the emission hot path; uses `cache.WaitForCacheSync` at startup

## Test plan

- [x] `go build ./...` passes in all four modules
- [x] `go test ./pkg/triggerengine/...` passes (csp-health-monitor) — updated `NewEngine` call sites to pass an empty `NodeLister`
- [x] `go test ./pkg/publisher/...` passes (kubernetes-object-monitor) — updated `New` call site
- [x] All other test failures are pre-existing `envtest` failures (no etcd binary in local env)

Closes JSC-105

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes Node lister/informer support and implemented managed-node opt-out gating across trigger processing, health event publishing, drain event publishing, and label reconciliation.
  * Updated Slurm drain monitor RBAC to allow `nodes` `list`/`watch`.
* **Bug Fixes**
  * Skip trigger/emission/drain publishing and label reconciliation for unmanaged (opted-out) nodes; strip detection/selector-related labels when opted out.
* **Metrics**
  * Added Prometheus counters for skipped triggers, skipped health emissions, skipped drain emissions, and skipped label reconciliation.
* **Tests**
  * Added ARM64 end-to-end coverage for opt-out suppression behavior.

### DIFF
--- a/health-monitors/csp-health-monitor/cmd/maintenance-notifier/main.go
+++ b/health-monitors/csp-health-monitor/cmd/maintenance-notifier/main.go
@@ -27,8 +27,11 @@ import (
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	listersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/nvidia/nvsentinel/commons/pkg/logger"
 	met "github.com/nvidia/nvsentinel/commons/pkg/metrics"
@@ -123,6 +126,18 @@ func setupUDSConnection(udsPath string) (*grpc.ClientConn, pb.PlatformConnectorC
 	slog.Info("Sidecar successfully connected to Platform Connector UDS.")
 
 	return conn, pb.NewPlatformConnectorClient(conn), nil
+}
+
+func setupNodeLister(ctx context.Context, k8sClient kubernetes.Interface) (listersv1.NodeLister, error) {
+	factory := informers.NewSharedInformerFactory(k8sClient, 0)
+	nodeInformer := factory.Core().V1().Nodes()
+	factory.Start(ctx.Done())
+
+	if ok := cache.WaitForCacheSync(ctx.Done(), nodeInformer.Informer().HasSynced); !ok {
+		return nil, fmt.Errorf("timed out waiting for node informer cache to sync")
+	}
+
+	return nodeInformer.Lister(), nil
 }
 
 func setupKubernetesClient() (kubernetes.Interface, error) {
@@ -221,6 +236,11 @@ func run() error {
 			return fmt.Errorf("kubernetes client setup failed: %w", err)
 		}
 
+		nodeLister, err := setupNodeLister(gCtx, k8sClient)
+		if err != nil {
+			return fmt.Errorf("node lister setup failed: %w", err)
+		}
+
 		value, ok := pb.ProcessingStrategy_value[appCfg.processingStrategy]
 		if !ok {
 			return fmt.Errorf("invalid processingStrategy %q (expected EXECUTE_REMEDIATION or STORE_ONLY)",
@@ -231,7 +251,7 @@ func run() error {
 
 		engine := trigger.NewEngine(cfg, store, platformConnectorClient,
 			fmt.Sprintf("unix:%s", appCfg.udsPath),
-			k8sClient, pb.ProcessingStrategy(value))
+			k8sClient, nodeLister, pb.ProcessingStrategy(value))
 
 		slog.Info("Trigger engine starting...")
 		engine.Start(gCtx)
--- a/health-monitors/csp-health-monitor/cmd/maintenance-notifier/main.go
+++ b/health-monitors/csp-health-monitor/cmd/maintenance-notifier/main.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	listersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/cache"
 
 	"github.com/nvidia/nvsentinel/commons/pkg/logger"
 	met "github.com/nvidia/nvsentinel/commons/pkg/metrics"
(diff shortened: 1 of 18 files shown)


lalitadithya:
can we break this into two PRs:
1. one for the labeler change
2. one for the cluster wide health monitors change

health-monitors/csp-health-monitor/cmd/maintenance-notifier/main.go
```diff
@@ -125,6 +128,25 @@ func setupUDSConnection(udsPath string) (*grpc.ClientConn, pb.PlatformConnectorC
 	return conn, pb.NewPlatformConnectorClient(conn), nil
 }
 
+func setupNodeLister(ctx context.Context, k8sClient kubernetes.Interface) (listersv1.NodeLister, error) {
```
natherz97, line 130: Wouldn't it be easier to add a platform-connector transformer which changes the processing strategy of any nodes with the nvsentinel.dgxc.nvidia.com/managed=false from EXECUTE_REMEDIATION to STORE_ONLY? This has benefit that it can be applied to both health + unhealthy events and we don't have to duplicate this behavior per health-monitor. Using the platform-connector transformer also enforces this policy compared to an external health-monitor we don't control or a new one that doesn't opt-in to not sending the event. This should be possible if we add the node label to the HealthEvent metadata and then use a transformer on that metadata property.

natherz97:
Should we also update the default rule-sets for fault-quarantine to use the new nvsentinel.dgxc.nvidia.com/managed=false label value? https://github.com/NVIDIA/NVSentinel/blob/main/distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml#L68

labeler/pkg/labeler/labeler.go
```diff
@@ -706,7 +724,41 @@ func (l *Labeler) desiredNodeLabels(nodeName string) (string, string, error) {
...
+		metrics.NodeLabelsSkippedManaged.Inc()
+	}
+
+	return needsUpdate
+}
+
 func (l *Labeler) reconcileNodeLabelsInPlace(node *v1.Node, driverLabel, dcgmVersion string) bool {
+	// When the node is opted out of NVSentinel management, strip detection labels
+	// so DaemonSet monitors evict via their existing nodeSelectors (ADR-040).
+	optedOut, err := managed.IsNodeOptedOut(l.ctx, l.nodeLister, node.Name)
```
natherz97, line 759: It looks like line 627 has this same behavior were we default to treating as not opted out if we can't check the node label. Shouldn't we not modify the labels in that case and retry until we know the current state of node? Asking because this would restore the health-monitors for a node under an ERR.

labeler/pkg/labeler/labeler.go
```diff
@@ -706,7 +724,41 @@ func (l *Labeler) desiredNodeLabels(nodeName string) (string, string, error) {
...
+	if err != nil {
+		slog.Warn("Failed to check managed label, treating as not opted out", "node", node.Name, "error", err)
+	}
+
+	if optedOut {
+		return l.stripDetectionLabels(node)
```
natherz97, line 769: It looks like we're skipping the l.deviceCounts.ReconcileNodeLabelsInPlace call below if the node is opted out, is that expected? An alternative implementation that would avoid modifying the existing flow would be to update getDCGMVersionForNode and getDriverLabelForNode to return the empty string if the node is opted out and then you don't have to handle this label removal with stripDetectionLabels. I'll leave it up to you but getDCGMVersionForNode already grabs the node from cache so you check there if node is opted out and return "", nil and then you shouldn't have to worry about changing the node label update logic since the labels would've been set correctly initially:

```
func (l *Labeler) getDCGMVersionForNode(nodeName string, excludePod *v1.Pod) (string, error) {
	objs, err := l.podInformer.GetIndexer().ByIndex(NodeDCGMIndex, nodeName)
	if err != nil {
		return "", fmt.Errorf("failed to get DCGM pods by node index for node %s: %w", nodeName, err)
	}

	node, err := l.getNodeFromCache(nodeName)
	if err != nil {
		return "", err
	}

<---------- add check here for if node is opted out and return an empty label value

	requireDCGMReady := l.requireDCGMReadyForBootstrap
	if node.Annotations != nil && len(node.Annotations[DCGMBootstrapCompletedAnnotation]) > 0 {
		requireDCGMReady = false
	}

	for _, obj := range objs {
		isTarget, pod := isTargetPod(obj, requireDCGMReady, excludePod)
		if !isTarget {
			continue
		}

		for _, container := range pod.Spec.Containers {
			if dcgm4Regex.MatchString(container.Image) {
				return dcgmVersion4, nil
			} else if dcgm3Regex.MatchString(container.Image) {
				return dcgmVersion3, nil
			}
		}
	}

	return "", nil
}
```

PR #1469, opened 2026-07-08
fix: postgreSQL change stream race skipping unquarantine events

### LINKED ISSUE #1466: [Bug]: PostgreSQL change stream can skip UnQuarantined events after advancing lastEventID
## Bug Description

The PostgreSQL-backed change stream watcher can skip a relevant `UnQuarantined` health event update after its internal `lastEventID` has advanced beyond that changelog row.

Observed in an e2e run with node-drainer:

- Fault-quarantine recovered a node and wrote an `UnQuarantined` status update.
- Node-drainer should have received that update and cancelled in-progress drain work for the node.
- Instead, node-drainer's PostgreSQL watcher skipped the changelog row as already processed because `lastEventID` was already higher than the NOTIFY payload ID.
- Node-drainer then continued retrying stale drain work and the node remained stuck in `draining`.

This appears PostgreSQL-specific. MongoDB native change streams keep filtering and resume-token advancement within the database change stream, so they do not have the same split between server-side SQL filtering, application-side filtering, and manual changelog cursor advancement.

## Component

Core Service

## Steps to Reproduce

This is timing-dependent and may not reproduce deterministically.

1. Run an e2e flow using PostgreSQL as the datastore.
2. Trigger a quarantine that causes node-drainer to start/retry drain work for a node.
3. Shortly after, recover the node so fault-quarantine writes an `UnQuarantined` status update while other health event updates are also being processed.
4. Observe that fault-quarantine writes `UnQuarantined`, but node-drainer's PostgreSQL watcher skips that changelog row because its `lastEventID` has already advanced past the NOTIFY ID.
5. Observe node-drainer continuing stale drain retries and the node remaining in `draining`.

Expected behavior:

- Node-drainer receives the `UnQuarantined` event.
- In-progress drain events for the node are cancelled.
- The node does not remain stuck in `draining`.

Actual behavior:

- The `UnQuarantined` changelog notification is skipped as already processed.
- Node-drainer does not cancel stale drain work.
- The node can remain stuck in `draining`.

## Environment

- NVSentinel version: dev
- Kubernetes version: kind e2e environment
- Deployment method: Helm/Kubernetes deployment
- Datastore: PostgreSQL

## Logs/Output

Fault-quarantine recovered the node and wrote `UnQuarantined`:

```text
15:26:43.103 All health checks recovered for node, proceeding with uncordon
15:26:43.287 Document updated with status id=<redacted> status="UnQuarantined"
```

Node-drainer's PostgreSQL watcher skipped that changelog row:

```text
15:26:43.287 Received NOTIFY changelogID=1191 table="health_events" operation="UPDATE"
15:26:43.287 Skipping already-processed event from NOTIFY notifiedID=1191 lastEventID=1192
```

After that, stale drain work continued:

```text
15:26:44.519 Processing event node=<redacted-node> eventID=<older-quarantined-event>
15:26:44.519 Evaluated action for node action="CheckCompletion"
15:26:44.540 Pods still running on node, requeuing for later check
```

The test later repeatedly observed the node still in `draining`:

```text
node <redacted-node> still has NVSentinel state label: draining
```

## Suspected Root Cause

The PostgreSQL change stream emulation uses two layers of filtering:

1. A server-side SQL filter generated by `SQLFilterBuilder`
2. Application-side filtering via `PipelineFilter`

The watcher advances `lastEventID` as it iterates returned changelog rows. If the server-side SQL filter is a partial/incomplete translation of the pipeline, it can exclude a relevant row while still returning a later row. The watcher can then advance `lastEventID` past the omitted row.

When the omitted row's NOTIFY arrives later, it is skipped because its ID is lower than `lastEventID`.

In the observed case:

```text
UnQuarantined row: changelogID=1191
Watcher bookmark: lastEventID=1192
```

So the `UnQuarantined` row was treated as already processed even though it was never delivered to node-drainer.

Relevant code:

- `store-client/pkg/datastore/providers/postgresql/changestream.go`
  - `handleNotification`
  - `fetchNewChanges`
  - `sendEventsToChannel`
  - `advancePosition`
- `store-client/pkg/datastore/providers/postgresql/sql_filter_builder.go`
  - `BuildFromPipeline`
  - `$or` handling
  - unsupported operator handling

## Suggested Fix

Make PostgreSQL change-stream SQL filtering fail closed.

The server-side SQL filter should only be used when the entire pipeline can be converted losslessly. If any stage, branch, field, or operator cannot be fully converted, the watcher should not apply a partial SQL filter. It should fetch changelog rows in order and rely on application-side `PipelineFilter` for filtering.

Important invariant:

> The watcher must not advance `lastEventID` past changelog rows it did not fetch, unless the server-side SQL filter is guaranteed equivalent to the application-side filter.

This should prevent relevant events like `UnQuarantined` from being skipped.

XRFXLP:
## Summary

This PR fixes a PostgreSQL change-stream race where node-drainer could miss an `UnQuarantined` event and keep retrying stale drain work.

In the PostgreSQL datastore, we emulate MongoDB-style change streams using a changelog table. The watcher keeps a `lastEventID` bookmark so it knows where to resume. The bug was that the watcher could use a partially converted server-side SQL filter, fetch a later changelog row, advance `lastEventID`, and then skip an earlier relevant row when its NOTIFY arrived later.

In the observed e2e failure, fault-quarantine correctly wrote `UnQuarantined`, but node-drainer skipped the notification because its bookmark had already moved past it:

```text
Received NOTIFY changelogID=452
Skipping already-processed event from NOTIFY notifiedID=452 lastEventID=453
```

Because node-drainer never saw the `UnQuarantined` event, it did not cancel the in-progress drain and the node stayed stuck in `draining`.

This change makes PostgreSQL server-side change-stream filtering fail closed. We now only use the generated SQL filter when the entire pipeline is converted successfully. If any stage, branch, field, or operator cannot be fully converted, the watcher fetches changelog rows in order and relies on the existing application-side `PipelineFilter`.

## Why This Fix

The important invariant is:

> The watcher must not advance `lastEventID` past changelog rows it did not fetch, unless the server-side SQL filter is guaranteed equivalent to the application-side filter.

Partial SQL filtering violates that invariant. Disabling server-side filtering for incomplete conversions preserves correctness while keeping the application-side filtering behavior unchanged.

## Testing

Added a regression test for the node-drainer pipeline showing that incomplete SQL conversion is not used as a server-side filter.

Ran:

```bash
go test ./store-client/pkg/datastore/providers/postgresql -run 'TestSQLFilterBuilder|Test.*Resume|Test.*Filtered|TestMixedFiltered|TestMultipleFiltered|TestTimestampBasedResume'
go test ./store-client/pkg/datastore/providers/postgresql
```

And also validated the fix as flakiness in the test is no longer reproduced after running 7 times.

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL change-stream reliability by preventing partial server-side SQL filters when pipeline-to-SQL conversion is incomplete, ensuring consistent application-side filtering.
  * Enhanced handling of out-of-order change notifications using a bounded recent-event deduplication cache, targeted recovery by exact changelog ID, and safer resume/bookmark advancement without regression.

* **Tests**
  * Added coverage for incomplete “node-drainer” pipelines to verify SQL fallback behavior.
  * Added/updated tests validating out-of-order notification recovery vs deduplication and stable bookmark progression.


KaivalyaMDabhadkar (commented):
This PR is good hardening (fail-closed on partial SQL filters), but I don't think it fixes the race in #1466:

The old partial filter already matched every UPDATE row, the unconvertible `updateDescription.updatedFields` terms were dropped from AND positions, broadening the predicate:

```
(operation = $4 OR (operation = $5 AND ... IN ($6,$7,$8,$9)))
Args: [UPDATE INSERT Quarantined AlreadyQuarantined UnQuarantined Cancelled]
```

So the `UnQuarantined` UPDATE (changelogID=1191) was never excluded by SQL filtering.

The actual race is transaction commit ordering: if txn 1192 commits before txn 1191, the fetch advances the bookmark past 1191 before it's visible. When 1191 commits, its NOTIFY is dropped by `payload.ID <= lastEventID`. The `changed_at DEFAULT NOW()` cursor can't recover it either since `NOW()` is txn-start time, not commit time.

This PR doesn't touch that path. I suggest keeping #1466 open and adding a regression test with two txns committing in reverse ID order.

XRFXLP:
> The actual race is transaction commit ordering: if txn 1192 commits before txn 1191, the fetch advances the bookmark past 1191 before it's visible. When 1191 commits, its NOTIFY is dropped by payload.ID <= lastEventID. The changed_at DEFAULT NOW() cursor can't recover it either since NOW() is txn-start time, not commit time.

Using in-memory cache of size 10k with 5 min TTL:

```golang
	if payload.ID <= lastEventID {
		if w.hasRecentlySeenEventID(payload.ID) {
			slog.Debug("Skipping already-processed event from NOTIFY",
				"client", w.clientName,
				"notifiedID", payload.ID,
				"lastEventID", lastEventID)

			return nil
		}

		slog.Debug("Recovering out-of-order NOTIFY below current bookmark",
			"client", w.clientName,
			"notifiedID", payload.ID,
			"lastEventID", lastEventID)

		if err := w.fetchChangeByID(ctx, payload.ID); err != nil {
			return fmt.Errorf("failed to recover out-of-order changelog event %d: %w", payload.ID, err)
		}

		return nil
	}
```

PTAL.

KaivalyaMDabhadkar (approved):

PR #1472, opened 2026-07-09
feat: add option to reset changeStream to head of eventQueue

### LINKED ISSUE #1261: [Feature]: Option to move to the head of the event queue on deployments
### Prerequisites

### Feature Summary

We would like a way to configure components like node-drainer to start from the head of the event queue on restart or new deployments. We target an eventually correct system and don't necessarily need replay.

### Problem/Use Case

We have had incidents where event replay has taken down many jobs. Specifically, one case where we discovered a component was not working for O(days) and when we turned it back on, many of our nodes were effected in a thundering herd. We would rather control where the replay queue is.

### Proposed Solution

Add an option to values.yaml to control where the event queue is on deployments or restarts.

### Component

Multiple Components

XRFXLP:
## Summary

This change adds a runtime-owned `resume-control` ConfigMap that lets operators perform a one-shot resume-token reset for watcher-based components. Components use `RESUME` for normal behavior and `CREATE` to start from the current stream head. On startup, each component reads its own key from `resume-control`; if the value is `CREATE`, it deletes only its own stored resume token before opening the watcher, then writes its key back to `RESUME`.

Helm does not create or manage this ConfigMap, If the ConfigMap or a component key is missing, the component creates/writes its own key as `RESUME`, so the mapping becomes explicit over time without being GitOps-managed.

### Runtime ConfigMap Creation And Self-Population

Deleted the runtime-owned `resume-control` ConfigMap, then restarted `fault-quarantine`. The module recreated the ConfigMap and wrote its own key as `RESUME`.

```bash
$ kubectl -n nvsentinel delete configmap resume-control --ignore-not-found
configmap "resume-control" deleted from nvsentinel namespace

$ kubectl -n nvsentinel rollout restart deployment/fault-quarantine
deployment.apps/fault-quarantine restarted

$ kubectl -n nvsentinel rollout status deployment/fault-quarantine --timeout=180s
deployment "fault-quarantine" successfully rolled out

$ kubectl -n nvsentinel get configmap resume-control -o yaml
apiVersion: v1
data:
  fault-quarantine: RESUME
  health-events-analyzer: RESUME
kind: ConfigMap
metadata:
  name: resume-control
  namespace: nvsentinel
```

Restarted the remaining watcher deployments. Each module wrote an explicit `RESUME` entry.

```bash
$ kubectl -n nvsentinel rollout restart deployment/node-drainer deployment/fault-remediation deployment/event-exporter deployment/health-events-analyzer
deployment.apps/node-drainer restarted
deployment.apps/fault-remediation restarted
deployment.apps/health-events-analyzer restarted

$ kubectl -n nvsentinel rollout status deployment/node-drainer --timeout=180s
deployment "node-drainer" successfully rolled out

$ kubectl -n nvsentinel rollout status deployment/fault-remediation --timeout=180s
deployment "fault-remediation" successfully rolled out

$ kubectl -n nvsentinel rollout status deployment/health-events-analyzer --timeout=180s
deployment "health-events-analyzer" successfully rolled out

$ kubectl -n nvsentinel get configmap resume-control -o yaml
apiVersion: v1
data:
  fault-quarantine: RESUME
  fault-remediation: RESUME
  health-events-analyzer: RESUME
  node-drainer: RESUME
kind: ConfigMap
metadata:
  name: resume-control
  namespace: nvsentinel
```

### Resume Token Baseline

Checked current resume-token state before testing `CREATE`.

```bash
$ db = db.getSiblingDB("HealthEventsDatabase"); ["event-exporter","fault-quarantine","fault-remediation","health-events-analyzer","node-drainer"].forEach(name => { const doc = db.ResumeTokens.findOne({clientName:name}, {_id:0, clientName:1, resumeToken:1}); if (!doc) { print(name + "|NO_TOKEN"); } else { print(doc.clientName + "|" + doc.resumeToken._data); } });
fault-quarantine|826A4F3464000000032B042C01002B186E5A1004A5CA6EC9A4EA4260BB132D774DAE8303463C6F7065726174696F6E54797065003C696E736572740046646F63756D656E744B65790046645F696400646A4F3464D13C613CC662B57D000004
fault-remediation|NO_TOKEN
health-events-analyzer|826A4F338F000000012B042C0100296E5A1004A5CA6EC9A4EA4260BB132D774DAE8303463C6F7065726174696F6E54797065003C696E736572740046646F63756D656E744B65790046645F696400646A4F338F7A014FF26EE0AD2A000004
node-drainer|NO_TOKEN
```

### CREATE Resets One Component

Set only `health-events-analyzer` to `CREATE`, restarted that deployment, and verified the key was consumed back to `RESUME`.

```bash
$ kubectl -n nvsentinel patch configmap resume-control --type merge -p '{"data":{"health-events-analyzer":"CREATE"}}'
configmap/resume-control patched

$ kubectl -n nvsentinel get configmap resume-control -o jsonpath='{.data.health-events-analyzer}{"\n"}'
CREATE

$ kubectl -n nvsentinel rollout restart deployment/health-events-analyzer
deployment.apps/health-events-analyzer restarted

$ kubectl -n nvsentinel rollout status deployment/health-events-analyzer --timeout=180s
deployment "health-events-analyzer" successfully rolled out

$ kubectl -n nvsentinel get configmap resume-control -o jsonpath='{.data.health-events-analyzer}{"\n"}'
RESUME
```

Confirmed the reset path ran.

```bash
$ kubectl -n nvsentinel logs deployment/health-events-analyzer --since=3m | grep 'Deleting change stream resume token on startup'
{"msg":"Deleting change stream resume token on startup","clientName":"health-events-analyzer","tokenDatabase":"HealthEventsDatabase","tokenCollection":"ResumeTokens"}
```

Confirmed only the selected component’s token changed; existing `event-exporter` and `fault-quarantine` tokens remained unchanged.

```bash
$ db = db.getSiblingDB("HealthEventsDatabase"); ["event-exporter","fault-quarantine","fault-remediation","health-events-analyzer","node-drainer"].forEach(name => { const doc = db.ResumeTokens.findOne({clientName:name}, {_id:0, clientName:1, resumeToken:1}); if (!doc) { print(name + "|NO_TOKEN"); } else { print(doc.clientName + "|" + doc.resumeToken._data); } });
fault-quarantine|826A4F3464000000032B042C01002B186E5A1004A5CA6EC9A4EA4260BB132D774DAE8303463C6F7065726174696F6E54797065003C696E736572740046646F63756D656E744B65790046645F696400646A4F3464D13C613CC662B57D000004
fault-remediation|NO_TOKEN
health-events-analyzer|NO_TOKEN
node-drainer|NO_TOKEN
```

Final `resume-control` state is safe.

```bash
$ kubectl -n nvsentinel get configmap resume-control -o yaml
apiVersion: v1
data:
  fault-quarantine: RESUME
  fault-remediation: RESUME
  health-events-analyzer: RESUME
  node-drainer: RESUME
kind: ConfigMap
metadata:
  name: resume-control
  namespace: nvsentinel
```

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Change stream watchers now resume from the last processed event after restarts.
  * Added an optional one-shot reset workflow to skip accumulated events and begin with new events.
  * Resume-token controls are automatically initialized and restored after use.

* **Documentation**
  * Added configuration guidance and command examples for supported components.

* **Access Control**
  * Updated Kubernetes permissions to support managing resume-token controls.


store-client/pkg/client/resume_token.go
```diff
@@ -0,0 +1,51 @@
...
+		"clientName", tokenConfig.ClientName,
+		"tokenDatabase", tokenConfig.TokenDatabase,
+		"tokenCollection", tokenConfig.TokenCollection)
+
+	if err := dbClient.DeleteResumeToken(ctx, tokenConfig); err != nil {
+		return fmt.Errorf("failed to delete change stream resume token: %w", err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
XRFXLP: should be resolved

docs/configuration/README.md
```diff
@@ -34,6 +34,30 @@ global:
   metricsPort: 2112
 ```
 
+### Change Stream Resume Tokens
```
lalitadithya, line 37: this would mean that even things like OOM kill will result in head of queue behaviour, I'm not sure if this is what we'd want. I'm thinking we go with an intentional user action like we do with circuit breaker.

What do you think @KaivalyaMDabhadkar ?

XRFXLP: This feature is disabled by default, if one-off resume token clearance is needed then we can anyway do it from respective DBs in the current system

XRFXLP: From the issue feature summary:
> We would like a way to configure components like node-drainer to start from the head of the event queue on restart or new deployments

lalitadithya: hmm.. true, but I think the original reason for the github issue was that on upgrade or specific scenarios users wanted an easy option to move to the head of the queue. So, they would flip a key somewhere and the system would automatically move to the head of the queue and skip the cold start.

lalitadithya: > From the issue feature summary:

right, yes, but if we were to do this on every oom kill or pod crash then the system would be unable to progress remediation further.

KaivalyaMDabhadkar: Yeah. I think we need a circuit breaker like mechanism to make sure that we track and persist the state of head of queue behavior. We would only want the "clearing of the older resume tokens" to fire only once and then never fire again during runtime (unless the user manually resets the mechanism for clearing the resume tokens once again similar to CB). We can do something like this:

For each component:

1. At startup, the component reads its replay-control key from the replay state configmap (similar to circuit breaker we have already).
2. If the key is REPLAY, it follows normal behavior and processes backlog from its saved resume token
3. If the key is SKIP, it advances its own resume position to the current head of the stream (by deleting the token)
4. It skips any startup backlog/backfill work older than that new position
5. It writes only its own key back to REPLAY
6. It then starts normally from the new saved position
7. Future restarts use normal REPLAY behavior unless the user sets the key to SKIP again

What do you think @lalitadithya @XRFXLP ?

lalitadithya: we also need to keep in mind that this needs to be a single configmap for all the modules.

XRFXLP: So technically, configmap driven approach can be a superset of the current (in this branch) implementation with a drive moving back to safe config.

lalitadithya: implementation wise, yes. what is key here is that we should not provide footguns to users

XRFXLP: Addressed, PTAL.

docs/configuration/event-exporter.md
```diff
@@ -50,6 +50,19 @@ kubectl create secret generic event-exporter-oidc-secret \
   -n nvsentinel
 ```
 
+### Change Stream Resume Token
+
+To make event-exporter start from the current stream head, set its key in the shared resume-control ConfigMap to `CREATE` and restart the deployment. Event-exporter deletes only its own resume token and resets the key back to `RESUME` during startup. When `exporter.backfill.enabled` is `true`, deleting the token also makes the exporter treat startup like a first deployment for backfill detection.
```
KaivalyaMDabhadkar: This behavior would defeat the purpose of CREATE. The reset deletes event-exporter’s token, after which checkResumeTokenExists returns false and Run treats the startup as a first deployment. Since backfill is enabled by default, this can export up to 30 days of historical events even though the user requested that accumulated events be skipped.
Could we distinguish “token absent because of an intentional CREATE” from “token absent on the first installation,” and suppress backfill for the former? This also matches the earlier agreement that CREATE should skip startup backlog/backfill.

store-client/pkg/factory/client_factory.go
```diff
@@ -171,11 +171,21 @@ func (f *ClientFactory) CreateChangeStreamWatcher(
...
 
-	watcher, err := dbClient.NewChangeStreamWatcher(ctx, client.TokenConfig{
+	clientTokenConfig := client.TokenConfig{
 		ClientName:      tokenConfig.ClientName,
 		TokenDatabase:   tokenConfig.TokenDatabase,
 		TokenCollection: tokenConfig.TokenCollection,
-	}, providerPipeline)
+	}
+
+	if err := client.ResetResumeTokenOnStartIfConfigured(ctx, dbClient, clientTokenConfig); err != nil {
```
KaivalyaMDabhadkar, line 180: This would probably not move PostgreSQL event-exporter to the current head. Event exporter reaches this path through `NewDatastoreClientFromConfig` and then uses `client.PostgreSQLChangeStreamWatcher`. After this call deletes the token, that watcher’s loadResumeToken sets lastSeenID=0, and its query reads unprocessed rows with id>0. As a result CREATE can replay the entire unprocessed changelog instead of skipping it.
Could we either route this path through the newer PostgreSQL watcher, which starts from the current timestamp when no token exists, or explicitly persist/use the current maximum changelog position for an intentional CREATE?

store-client/pkg/client/resume_token.go
```diff
@@ -0,0 +1,210 @@
...
+	slog.InfoContext(ctx, "Deleting change stream resume token on startup",
+		"clientName", tokenConfig.ClientName,
+		"tokenDatabase", tokenConfig.TokenDatabase,
+		"tokenCollection", tokenConfig.TokenCollection)
+
+	if err := dbClient.DeleteResumeToken(ctx, tokenConfig); err != nil {
+		return fmt.Errorf("failed to delete change stream resume token: %w", err)
+	}
+
+	if err := store.SetMode(ctx, tokenConfig.ClientName, ResumeControlModeResume); err != nil {
```
KaivalyaMDabhadkar, line 92: I think CREATE here can race with the documented kubectl rollout restart workflow. A normal rolling restart overlaps the old and new pods, even with one replica. The new pod can delete the token and set this key to RESUME, while the old pod is still processing events and can write its stale token back. The new watcher may then resume from that older token mistakenly. Could we document a scale-to-zero handoff instead of the documented rolling restart?:
1. Record the deployment’s current replica count.
2. Scale it to zero and wait for every old pod to terminate.
3. Set the component’s resume-control key to CREATE.
4. Restore the original replica count.

docs/configuration/event-exporter.md
XRFXLP: Given that event-exporter cannot really affect the workloads or the cluster state, it is better to just exclude it from this changeStream mechanism.

store-client/pkg/factory/client_factory.go
XRFXLP: I've removed event-exporter from this mechanism now.

store-client/pkg/client/resume_token.go
XRFXLP: updated docs.

KaivalyaMDabhadkar (approved):

lalitadithya:
@KaivalyaMDabhadkar , @XRFXLP , does this change also skip the cold starts?

XRFXLP:
>  does this change also skip the cold starts?

no, I'll do follow up MR that is bulit on top of this change in a while

PR #1477, opened 2026-07-11
fix: detect frozen health monitor loops via /healthz liveness probe

shaq918:
## Problem

The `gpu-health-monitor`, `syslog-health-monitor`, and `nic-health-monitor` liveness probes target `/metrics`, which is served by an independent HTTP server thread/goroutine. When the main health check polling loop freezes, `/metrics` keeps responding and Kubernetes never restarts the pod.

**Observed in production:** On a 370-GPU-node cluster, 13 gpu-health-monitor pods showed `Running 1/1` with 0 restarts while their reconciliation loops had been frozen for 7–24 days. These nodes had **zero GPU health monitoring** despite appearing healthy. The `dcgm_reconcile_time_count` Prometheus metric confirmed the loop counter had stopped incrementing.

The existing `WithSimpleHealth()` handler at `/healthz` (used by syslog-health-monitor) always returns 200 regardless of loop state. The gpu-health-monitor and nic-health-monitor had no `/healthz` endpoint at all — their liveness probes pointed directly at `/metrics`.

### `PollingHealthChecker` (commons)

Add a reusable `PollingHealthChecker` to `commons/pkg/server/` that:
- Tracks the last successful poll iteration timestamp via `atomic.Int64` (lock-free, goroutine-safe)
- Implements the existing `HealthChecker` interface
- Returns unhealthy from `/healthz` when `time.Since(lastPoll) > stalenessThreshold`

Each monitor sets the staleness threshold to 3× its poll interval and calls `MarkAlive()` after each **successful** iteration only.

### Per-monitor changes

**gpu-health-monitor (Python):**
- New `healthz.py` module subclasses Prometheus `MetricsHandler` to serve both `/metrics` and `/healthz` on the same port
- `mark_alive()` called only after successful DCGM health check + event publish (not on connectivity failure path)

**syslog-health-monitor (Go):**
- Replace `WithSimpleHealth()` with `WithHealthCheck(pollingHealthChecker)`
- `MarkAlive()` called after successful `monitor.Run()`, not during backoff retries

**nic-health-monitor (Go):**
- Add `PollingHealthChecker` to the metrics server
- Extend `pollingLoop()` with optional `onSuccess func()` callback (avoids code duplication)
- State polling loop marks alive; counter polling loop does not (counter loop runs at 1s and is not the primary health signal)

### Helm templates

- **Liveness probes** → `/healthz` (detects frozen loops, triggers pod restart)
- **Readiness probes** → `/metrics` (unchanged — available immediately at startup, avoids false 0/1 during initial poll interval)

## Evidence

Frozen gpu-health-monitor pod metrics from a production node (pod Running for 24 days, loop frozen since day 1):
```
dcgm_reconcile_time_count 142682.0    # counter stopped incrementing
process_cpu_seconds_total 1187.73     # process alive but idle
```

Pod logs show the last iteration timestamp matching the last event in the database — no new logs or events after that point.

## Test plan

- [ ] Unit test for `PollingHealthChecker` — healthy within threshold, unhealthy after threshold, `MarkAlive()` resets
- [ ] Unit test for Python `healthz.py` — `/healthz` returns 200 when alive, 503 when stale
- [ ] Integration: deploy to test cluster, verify `/healthz` returns 200 during normal operation
- [ ] Integration: pause the DCGM reconcile loop (e.g., freeze DCGM hostengine), verify `/healthz` returns 503 after staleness threshold, verify pod restart
- [ ] Verify readiness probe on `/metrics` passes immediately at startup (no false 0/1)

## Summary by CodeRabbit

* **New Features**
  * Added polling-based health checking for NIC and syslog monitors, using a staleness window derived from their polling interval to mark health unhealthy when iterations stop making progress.
  * GPU health monitoring now exposes a dedicated **`/healthz`** endpoint on the existing HTTP server, returning **503** when stale and **200** when refreshed.
* **Bug Fixes**
  * Updated Kubernetes **livenessProbe** paths for GPU, NIC, and syslog DaemonSets to **`/healthz`** (readiness/metrics endpoints unchanged).
* **Tests**
  * Added unit tests for the new polling health checker behavior (initially healthy, reset on mark, stale threshold, error details).


lalitadithya:
hey @shaq918 , can you please look into the failing lint?

commons/pkg/server/polling_health.go
```diff
@@ -0,0 +1,86 @@
...
+	hc.lastMonoNano.Store(monoNow())
+
+	return hc
+}
+
+// MarkAlive records a successful poll loop iteration. This method is
+// safe to call from any goroutine.
+func (h *PollingHealthChecker) MarkAlive() {
+	h.lastMonoNano.Store(monoNow())
+}
```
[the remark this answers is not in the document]
XRFXLP: There is just one call per health monitor, right?

commons/pkg/server/polling_health.go
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
XRFXLP, line 1: Although PR description lists that we unit test for this utility but they're not added here, can you please add?

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
```diff
@@ -588,6 +589,7 @@ def start(self, fields_to_monitor: list[str], exit: Event) -> None:
                                 types.CallbackInterface.health_event_occurred.__name__,
                                 [health_status, gpu_ids],
                             )
+                            _mark_alive()
```
XRFXLP: `mark_alive()` is only reached on the fully-successful path. The `dcgm_connectivity_failed` branch above never marks alive, so a loop that is iterating fine and *correctly reporting* a persistent DCGM outage will still flip `/healthz` to 503 after 3x poll_interval and get restarted. 

That conflates "loop frozen" (the bug you're fixing) with "dependency persistently down".

XRFXLP (changes requested):

commons/pkg/server/polling_health.go
shaq918: Added 4 deterministic unit tests using an injectable clock (no sleeps) in ae19767.

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
shaq918: Good catch. Moved `_mark_alive()` to the top of every loop iteration so a persistent DCGM outage does not trigger a restart. Fixed in ae19767.

XRFXLP: Same fix is needed in the NIC and syslog health monitors: a failure inside the monitor loop shouldn't make it look silently dead. 

In NIC HM, pollingLoop only marks alive on the success branch instead of that we should mark on every iteration instead. 

In syslog HM, MarkAlive() is inside the inner retry loop and only runs after a successful monitor.Run(), so a sustained failure spins in backoff and never marks (it needs to mark on each attempt, not just on success.)

shaq918: Fixed in 3cd8297. NIC HM pollingLoop now calls the callback on every tick (renamed onSuccess to onIteration). Syslog HM MarkAlive() moved from inner retry success to outer ticker tick.

health-monitors/syslog-health-monitor/main.go
```diff
@@ -436,6 +441,10 @@ func runPollingLoop(
 		case <-ticker.C:
 			slog.Info("Performing scheduled health check run...")
 
+			// Mark alive on every tick so the liveness probe detects a
+			// frozen loop, not a failed dependency retrying in backoff.
+			healthChecker.MarkAlive()
+
 			for {
```
XRFXLP, line 444: If this runs forever then syslog HM on failure then ticker won't be advanced and syslog HM can be marked false positively dead, right?

shaq918: Good catch — you're right. On sustained failure the inner retry loop never returns to the outer ticker, so the per-tick `MarkAlive()` starves and the monitor gets falsely restarted. Fixed in 796faa33 by marking alive on every retry attempt (backoff caps at 30s, well inside the staleness threshold), so only a genuinely hung `monitor.Run()` trips the probe now.

One trade-off worth noting: the mark happens before each attempt, so if `Run()` itself blocks indefinitely, detection is delayed by up to one staleness threshold from the last attempt. That's inherent to any loop-boundary marking scheme (marking on success has the same worst case) — detecting a hang inside `Run()` faster would need a watchdog/timeout around the call, which felt out of scope here.

XRFXLP (approved):

PR #1482, opened 2026-07-13
fix: circuit breaker should only consider GPU nodes for breaking

### LINKED ISSUE #1228: [Question]:  Circuit breaker denominator: all nodes vs. GPU nodes?
### Prerequisites

- [x] I searched existing issues and docs

### Question

The fault-quarantine circuit breaker's percentage setting is documented as _Percentage of nodes that can be cordoned_ (docs/circuit-breaker.md). Looking at the implementation, the denominator (totalNodes) is the count of every Node object the apiserver returns, with no GPU/label filter:
https://github.com/NVIDIA/NVSentinel/blob/00f7a04b19925bdbfa5ef7fbb6aba9332814b424/fault-quarantine/pkg/informer/node_informer.go#L75

However, this produces an asymmetric ratio whenever a cluster contains nodes that fault-quarantine cannot act on:
  - The numerator is GPU-scoped, every shipped ruleset matches GPU/syslog health events, every cordon action targets a node that emitted such an event. Non-GPU nodes don't contribute.
 - The denominator has neither filter, any Node object counts, including ones the breaker can never affect.

A subtle consequence is that the breaker's effective sensitivity drifts with unrelated cluster activity (e.g. a cluster autoscaler spinning non-GPU pools up or down silently changes the threshold).

I also spotted a place in the code that hints the original mental model may have been "GPU nodes":
https://github.com/NVIDIA/NVSentinel/blob/00f7a04b19925bdbfa5ef7fbb6aba9332814b424/fault-quarantine/pkg/breaker/breaker.go#L204-L205
even though the function returns the count of all nodes.

 A few things I'd love clarity on:

 1. Is the all-nodes denominator the intended design, with the assumption that NVSentinel is deployed on (near-)homogeneous GPU clusters where all_nodes ≈ gpu_nodes?
 2. Would adding a circuitBreaker.nodeSelector (label selector string, default empty / current behaviour) be in scope? It would mirror the numerator-side ManagedByNVSentinel exclusion and would allow mixed clusters define a stable denominator.

### Category

Configuration

### Context

- NVSentinel version: v1.1.0

XRFXLP:
## Summary
Small fix so fault-quarantine filters only those nodes that have `nvidia.com/gpu.present=true` label and consequently those are the nodes that circuit breaker would track for breaking.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added configurable GPU node label support via Helm (`fault-quarantine.gpuNodeLabel`) and CLI (`--gpu-node-label`) to define which nodes count toward breaker calculations.
* **Bug Fixes**
  * Circuit-breaker thresholding, trip, and progress calculations now consistently use GPU-present nodes instead of all cluster nodes.
  * Node monitoring/counting now focuses on GPU-labeled nodes.
* **Documentation**
  * Updated circuit-breaker, metrics, and Helm values docs to clarify GPU-node-based threshold behavior.
* **Tests**
  * Updated unit, E2E, and scale tests to enforce deterministic GPU labeling and validate GPU-based thresholds.
* **Chores**
  * Updated local kind setup to apply NVIDIA GPU-related node labels.


fault-quarantine/pkg/informer/node_informer.go
```diff
@@ -34,6 +35,9 @@ import (
 
 const (
 	quarantineAnnotationIndexName = "quarantineAnnotation"
+
+	// GPUNodeLabel identifies nodes with GPUs relevant to NVSentinel.
+	GPUNodeLabel = "nvidia.com/gpu.present"
```
lalitadithya, line 39: Can we externalize this? So that in future if the label change, we don't need a code change? Or if someone is not using GPU operator, they will still be able to use a custom label?

fault-quarantine/pkg/informer/node_informer.go
```diff
@@ -63,14 +67,22 @@ func (ni *NodeInformer) GetInformer() cache.SharedIndexInformer {
...
 		clientset: clientset,
 	}
 
-	informerFactory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
+	gpuNodeSelector := labels.Set{GPUNodeLabel: "true"}.AsSelector()
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(
+		clientset,
+		resyncPeriod,
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.LabelSelector = gpuNodeSelector.String()
```
KaivalyaMDabhadkar: I think one scenario can be problematic:
1. Node A has the nvidia.com/gpu.present=true label
2. FQ quarantines it by cordoning or tainting it
3. The gpu label is then removed or temporarily becomes false.
4. Because node A no longer matches the informer’s selector, k8s sends the filtered watcher a “deleted” event even though the actual node still exists
5. FQ removes Node A from its informer cache
6. Now FQ receives a healthy event for Node A
7. FQ attempts to get Node A and its quarantine annotations but Node A is no longer present in the filtered informer cache
8. Because FQ cannot find the quarantine annotations, it treats Node A as having no active quarantine and skips recovery
9. Node A thus may remain cordoned or tainted indefinitely despite being healthy

Could we keep the informer watching all nodes and apply the gpu label filter only inside GetNodeCounts()?

XRFXLP: yeah, this makes sense.

KaivalyaMDabhadkar (approved):

PR #1487, opened 2026-07-13
fix: bind gpu-health-monitor metrics server for IPv6-only clusters

### LINKED ISSUE #1407: [Bug]: NVSentinel doesn't work in IPv6-only clusters
### Prerequisites

### Code of Conduct

### Bug Description

NVSentinel doesn't work in IPv6-only K8s clusters. There are a few reasons why:

**1. [DCGM client can't connect via hostnames that resolve IPv6 addresses](https://github.com/NVIDIA/DCGM/issues/301)**

**2. The metrics server only binds to `0.0.0.0`, so liveness probes that call it can't reach it in IPv6-only clusters.**
The metrics server inside `gpu-health-monitor` pods is started by calling the `start_http_server` from `prometheus_client` without passing a value for the [optional `addr` argument](https://github.com/prometheus/client_python/blob/master/prometheus_client/exposition.py#L235), so it defaults to `0.0.0.0`, making it unreachable from liveness probes in IPv6 clusters.

Inside the `gpu-health-montior` pods:
```
root@gpu-health-monitor-dcgm-4:/# netstat -tlpn | grep 2112
tcp        0      0 0.0.0.0:2112            0.0.0.0:*               LISTEN      1/python3
```

From the pod events:
```
Events:
  Type     Reason                           Age                      From     Message
  ----     ------                           ----                     ----     -------
  Warning  Unhealthy                        16m (x206 over 5h10m)    kubelet  spec.containers{gpu-health-monitor}: Liveness probe failed: Get "http://[2600:1f18:b39:ec10::611b[]:2112/metrics": dial tcp [2600:1f18:b39:ec10::611b]:2112: connect: connection refused
  Warning  Unhealthy                        6m40s (x868 over 5h11m)  kubelet  spec.containers{gpu-health-monitor}: Readiness probe failed: Get "http://[2600:1f18:b39:ec10::611b[]:2112/metrics": dial tcp [2600:1f18:b39:ec10::611b]:2112: connect: connection refused
  Normal   Killing                          5m14s (x72 over 5h9m)    kubelet  spec.containers{gpu-health-monitor}: Container gpu-health-monitor failed liveness probe, will be restarted
```

**3. No built-in way to make DCGM pods listen on `[::]` instead of `0.0.0.0`**

The only way I could find to make DCGM pods bind on IPv6 interfaces was to extract the hard-coded args from the DCGM pod image and  change the `0.0.0.0` arg to `::` in the GPU operator Helm chart:
```yaml
dcgm:
  args: ["-n", "-b", "::", "--log-level", "NONE", "-f", "-"]
```

This workaround does force DCGM pods to bind on IPv6 interfaces, but it's not particularly pretty. It would be nice if this was made easier, or just worked out of the box.

The end result of these issues is that it's not possible to use NVSentinel in IPv6-only clusters.

### Component

Health Monitor

### Steps to Reproduce

1. Set up an IPv6-only cluster
2. Install Nvidia GPU operator and NVSentinel
3. Notice how multiple pods are stuck in CrashLoopBackoff due to the above issues.

### Environment

- NVSentinel version: v1.9.0
- Kubernetes version: v1.35.2
- Deployment method: Helm

### Logs/Output

_No response_

mvanhorn:
## Summary

Fixes problem #2 of #1407: in IPv6-only clusters the gpu-health-monitor pod
CrashLoopBackoffs because kubelet liveness/readiness probes over IPv6 get
`connection refused`.

Root cause: `gpu_health_monitor/cli.py` calls `start_http_server(port)` without an
`addr`, so prometheus_client binds the metrics server to `0.0.0.0` (IPv4-only). The
daemonset probes hit the pod's IPv6 address, which nothing is listening on.

Changes:
- Add a `--metrics-addr` CLI flag and pass it as `addr` to `start_http_server`. The
  flag defaults to `0.0.0.0`, so non-Helm invocations are unchanged.
- Add a `global.metricsAddress` Helm value (default `::`) and wire `--metrics-addr`
  into the dcgm 3.x and 4.x daemonset templates. `::` binds dual-stack (IPv6 plus
  IPv4-mapped on standard Linux with `net.ipv6.bindv6only=0`), so a freshly installed
  chart works in IPv6-only and IPv4-only clusters and kubelet probes succeed over
  IPv6.

@lalitadithya noted in the issue thread that the probes would still be a problem and
that it was something to work on; this addresses that specific piece.

Scope: this PR only covers the gpu-health-monitor metrics bind (problem #2), which is
the part that lives in this repo. The other two problems in the issue are upstream and
out of scope here: the DCGM client not connecting over IPv6 hostnames (#1, tracked in
NVIDIA/DCGM#301) and making the GPU-operator's DCGM exporter bind on `[::]` (#3, which
requires editing the GPU-operator Helm chart, not NVSentinel).

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

Added `gpu_health_monitor/tests/test_cli.py`: asserts `--metrics-addr ::` reaches
`start_http_server(port, addr="::")`, that the default is still `0.0.0.0`, and that the
option is registered. Full suite passes (67 tests). `helm template` on the chart renders
`--metrics-addr "::"` in both the 3.x and 4.x daemonsets; `helm lint` passes.

Note: the end-to-end probe behavior requires an IPv6-only Kubernetes cluster with NVIDIA
GPUs and DCGM, which is not available here; the dual-stack fix is asserted from
prometheus_client's bind behavior and covered by the unit test.

## Checklist
- [x] Self-review completed
- [x] Ready for review

Refs #1407

## Summary by CodeRabbit

* **New Features**
  * Added configurable metrics-server binding, including IPv4, IPv6, and dual-stack support.
  * Added platform connector token authentication configuration for health-event reporting.
  * Added Kubernetes configuration for authentication token mounting and metrics address settings.

* **Tests**
  * Added coverage for explicit IPv6 and default IPv4 metrics binding.

### DIFF
--- a/distros/kubernetes/nvsentinel/values.yaml
+++ b/distros/kubernetes/nvsentinel/values.yaml
@@ -275,6 +275,7 @@ global:
     enabled: false
   janitor:
     enabled: false
+  # Enables lifecycle-manager reconciliation of maintenance requests.
   lifecycleManager:
     enabled: false
   janitorProvider:
--- a/distros/kubernetes/nvsentinel/values.yaml
+++ b/distros/kubernetes/nvsentinel/values.yaml
@@ -21,6 +21,10 @@ global:
     tag: "12-debian-12-r30"
     pullPolicy: IfNotPresent
   metricsPort: 2112
+  # IPv4 bind address for the GPU health monitor metrics server. Set to "::"
+  # to enable IPv6 / dual-stack binding. This requires an IPv6-capable kernel;
+  # nodes booted with ipv6.disable=1 fail with EAFNOSUPPORT and CrashLoop.
+  metricsAddress: "0.0.0.0"
   # Default Kubernetes API client limits for core controllers. Components can
   # override these with their own qps and burst values.
   # qps: positive values throttle requests, 0 uses the client-go default,
--- a/health-monitors/gpu-health-monitor/gpu_health_monitor/healthz.py
+++ b/health-monitors/gpu-health-monitor/gpu_health_monitor/healthz.py
@@ -23,30 +23,33 @@ false positives from NTP clock adjustments.
 import socket
 import threading
 import time
+from dataclasses import dataclass, field
 from http.server import ThreadingHTTPServer
 
 from prometheus_client import MetricsHandler
 
 
-class _last_reconcile:
+@dataclass
+class _LastReconcile:
     """Thread-safe tracker for the last successful reconcile timestamp.
 
     Uses time.monotonic() for NTP-immune elapsed time measurement.
     Module-scoped singleton — one tracker per process.
     """
 
-    _lock = threading.Lock()
-    _timestamp: float = time.monotonic()
+    lock: threading.Lock = field(default_factory=threading.Lock)
+    timestamp: float = field(default_factory=time.monotonic)
 
-    @classmethod
-    def mark_alive(cls) -> None:
-        with cls._lock:
-            cls._timestamp = time.monotonic()
+    def mark_alive(self) -> None:
+        with self.lock:
+            self.timestamp = time.monotonic()
 
-    @classmethod
-    def seconds_since_last(cls) -> float:
-        with cls._lock:
-            return time.monotonic() - cls._timestamp
+    def seconds_since_last(self) -> float:
+        with self.lock:
+            return time.monotonic() - self.timestamp
+
+
+_last_reconcile = _LastReconcile()
 
 
 # Module-level staleness threshold (seconds). Set by start_server().
@@ -103,13 +106,13 @@ def start_server(
     # Reset the grace period to start from server startup, not module import.
     _last_reconcile.mark_alive()
 
-    server_cls = ThreadingHTTPServer
-    if ":" in addr:
-        # ThreadingHTTPServer defaults to AF_INET, which cannot bind an IPv6 literal.
-        server_cls = type(
-            "_ThreadingHTTPServerV6", (ThreadingHTTPServer,), {"address_family": socket.AF_INET6}
-        )
-    httpd = server_cls((addr, port), _HealthMetricsHandler)
+    family, _, _, _, bind_address = socket.getaddrinfo(addr, port, type=socket.SOCK_STREAM)[0]
+    server_cls = type("_AddressAwareThreadingHTTPServer", (ThreadingHTTPServer,), {"address_family": family})
+    httpd = server_cls(bind_address, _HealthMetricsHandler, bind_and_activate=False)
+    if family == socket.AF_INET6:
+        httpd.socket.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
+    httpd.server_bind()
+    httpd.server_activate()
     t = threading.Thread(target=httpd.serve_forever, daemon=True)
     t.start()
 
--- a/health-monitors/gpu-health-monitor/gpu_health_monitor/healthz.py
+++ b/health-monitors/gpu-health-monitor/gpu_health_monitor/healthz.py
@@ -20,6 +20,7 @@ within the staleness threshold. Uses time.monotonic() to avoid
 false positives from NTP clock adjustments.
 """
 
+import socket
 import threading
 import time
 from http.server import ThreadingHTTPServer
@@ -80,13 +81,17 @@ class _HealthMetricsHandler(MetricsHandler):
             super().do_GET()
 
 
-def start_server(port: int, staleness_seconds: float = 300.0) -> tuple[ThreadingHTTPServer, threading.Thread]:
+def start_server(
+    port: int, staleness_seconds: float = 300.0, addr: str = "0.0.0.0"
+) -> tuple[ThreadingHTTPServer, threading.Thread]:
     """Start an HTTP server serving /metrics and /healthz.
 
     Args:
         port: TCP port to listen on.
         staleness_seconds: Maximum seconds between reconcile iterations
             before /healthz returns 503.
+        addr: Address to bind. Defaults to 0.0.0.0 (IPv4). Pass "::" to bind
+            IPv6, which also accepts IPv4-mapped clients on dual-stack nodes.
 
     Returns:
         (ThreadingHTTPServer, daemon Thread) — same interface as
@@ -98,7 +103,13 @@ def start_server(port: int, staleness_seconds: float = 300.0) -> tuple[Threading
     # Reset the grace period to start from server startup, not module import.
     _last_reconcile.mark_alive()
 
-    httpd = ThreadingHTTPServer(("", port), _HealthMetricsHandler)
+    server_cls = ThreadingHTTPServer
+    if ":" in addr:
+        # ThreadingHTTPServer defaults to AF_INET, which cannot bind an IPv6 literal.
+        server_cls = type(
+            "_ThreadingHTTPServerV6", (ThreadingHTTPServer,), {"address_family": socket.AF_INET6}
+        )
+    httpd = server_cls((addr, port), _HealthMetricsHandler)
     t = threading.Thread(target=httpd.serve_forever, daemon=True)
     t.start()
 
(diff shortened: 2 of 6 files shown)


distros/kubernetes/nvsentinel/values.yaml
```diff
@@ -21,6 +21,9 @@ global:
     tag: "12-debian-12-r30"
     pullPolicy: IfNotPresent
   metricsPort: 2112
+  # Bind address for component metrics servers. "::" binds dual-stack
+  # (IPv6 + IPv4-mapped) so kubelet probes work in IPv6-only clusters.
+  metricsAddress: "::"
```
deesharma24: shouldn't we keep it empty by default  and if someone needs ipv6 , then they can set it to "::"

lalitadithya: I think having the default support both ipv4 and ipv6 is fine, this will ensure the system works out of the box irrespective of the network stack configuration for the most part.

lalitadithya: @deesharma24 , do you think we'd have some backward combability issues with making this the default?

deesharma24: :: is only dual-stack when the kernel has an IPv6 stack loaded. On nodes booted with ipv6.disable=1 (our baremetal fleet), socket(AF_INET6) fails with EAFNOSUPPORT, so it's not "out of the box irrespective of network stack". I verified by building this PR's image and deploying it on a baremetal dev cluster: with --metrics-addr "::" the pod hits OSError: [Errno 97] Address family not supported by protocol  and goes into CrashLoopBackOff; the same image with 0.0.0.0 runs healthy on the same node.for csp, it works . only for baremetal, it is a problem.

lalitadithya: I see, that is good context. Thanks for checking! I think we should leave the default as is (i.e., 0.0.0.0) and document how to enable dual stack / ipv6 support.

lalitadithya:
hey @mvanhorn , just checking in, when you have a min, can you please check conflicts and the review feedback in https://github.com/NVIDIA/NVSentinel/pull/1487#discussion_r3576574836.

mvanhorn:
Rebased and addressed both points in 5a123ac.

@deesharma24 you were right about the default — `metricsAddress` is now empty, which preserves the existing IPv4 behaviour, and `"::"` is the opt-in for dual-stack in IPv6-only clusters. Verified by rendering the chart both ways: the default emits no `--metrics-addr` flag at all, and `--set global.metricsAddress=::` emits it on both gpu-health-monitor DaemonSets.

The rebase needed more than a mechanical resolution. Upstream replaced `prometheus_client.start_http_server` with `healthz.start_server`, which serves `/metrics` and `/healthz` together but had no bind-address parameter — so a naive merge would have silently dropped this PR entirely. `start_server` now takes `addr` alongside upstream's `staleness_seconds`, resolves it with `getaddrinfo` to match what `prometheus_client` did, and for an IPv6 result sets `AF_INET6` with `IPV6_V6ONLY` off so `::` accepts IPv4-mapped connections instead of quietly becoming IPv6-only. Empty normalises to `0.0.0.0`, since an empty passive lookup can prefer IPv6 on a dual-stack host and that would have changed the default behaviour.

`helm lint`, the healthz tests and `black` all pass. No IPv6-only cluster here, so the dual-stack bind is covered by the socket-level code and chart rendering rather than observed end to end.

lalitadithya:
hey @mvanhorn , can you please address the final set of comments -- https://github.com/NVIDIA/NVSentinel/pull/1487#discussion_r3578428444 and we should be good to push further with this PR?

mvanhorn:
@lalitadithya @deesharma24 pushed d4a98ff following the conclusion in the values.yaml thread.

The chart default is back to `0.0.0.0`, so nothing changes for existing clusters. `metricsAddress` is documented as the IPv6 opt-in, with @deesharma24's finding recorded inline: `::` only works where the kernel has an IPv6 stack, and a node booted with `ipv6.disable=1` fails with EAFNOSUPPORT and CrashLoops.

The CLI already defaulted to `0.0.0.0`, and the test asserting that is unchanged and passing. Also took the Ruff S108 nit and moved `--state-file` onto `tmp_path`.

lalitadithya:
thanks @mvanhorn ! can you please rebase and fix conflicts?

mvanhorn:
Rebased onto current main as a3a1f019, squashed to one commit.

Main had replaced the metrics server: `start_http_server(port, addr=...)` is gone and `healthz.start_server(port, staleness_seconds=...)` took its place, binding `("", port)` with the default `AF_INET`. That cannot bind an IPv6 literal at all, so the original approach no longer applied.

`start_server` now takes `addr` (default `0.0.0.0`) and selects `AF_INET6` when the address contains a colon; `cli.py` threads `--metrics-addr` into it. The chart default stays `0.0.0.0` per the conclusion above, with the IPv6 caveat documented next to `metricsAddress`.

Checked on the rebased branch: binding `0.0.0.0` gives an `AF_INET` socket, `::` gives `AF_INET6`, `/healthz` answers 200 on both, and the v6 socket still serves an IPv4-mapped client. The three CLI tests are green, and the `addr` assertions no longer pin the other kwargs so a future parameter will not break them.

deesharma24:
@mvanhorn Can you please work on the coderabbit comments and lint  test failing?

mvanhorn:
Both are in 82ba908. Going through the coderabbit notes one at a time:

`test_cli.py` now uses `tmp_path` rather than a fixed path. `healthz.py` keeps its reconcile state in a `_LastReconcile` dataclass instead of loose locals, and resolves the address and configures dual-stack before it binds. The `cli.py` signatures and the remaining CLI helpers are typed. `values.yaml` documents `lifecycleManager`.

One I did not take: the Kata split suggested on `daemonset-dcgm-3.x.yaml`. The GPU health monitor does not read host journals, so splitting there changes nothing for this component. Happy to revisit if you are seeing something I am not.

Black, Ruff, compilation, the mocked tests, and Helm lint and template all pass locally. I could not run the full Poetry CI here, so that one is on the CI to confirm.

deesharma24:
> Both are in [82ba908](https://github.com/NVIDIA/NVSentinel/commit/82ba90824d0f667802a452f385dda8b1237391ef). Going through the coderabbit notes one at a time:
> 
> `test_cli.py` now uses `tmp_path` rather than a fixed path. `healthz.py` keeps its reconcile state in a `_LastReconcile` dataclass instead of loose locals, and resolves the address and configures dual-stack before it binds. The `cli.py` signatures and the remaining CLI helpers are typed. `values.yaml` documents `lifecycleManager`.
> 
> One I did not take: the Kata split suggested on `daemonset-dcgm-3.x.yaml`. The GPU health monitor does not read host journals, so splitting there changes nothing for this component. Happy to revisit if you are seeing something I am not.
> 
> Black, Ruff, compilation, the mocked tests, and Helm lint and template all pass locally. I could not run the full Poetry CI here, so that one is on the CI to confirm.

@mvanhorn there are merge conflicts here.Can you please resolve it?

mvanhorn:
Rebased onto `main` in 2a850ca, @deesharma24.

The conflict was confined to `test_cli.py`: `main` added `TestParseMinConsecutivePolls` to the same file this PR rewrites. I kept both suites rather than picking one, unioned the imports, and preserved the typing from the review round. `pytest gpu_health_monitor/tests/test_cli.py` is 17 passed, 0 failed.

Still open from your last round is the Kata split on `daemonset-dcgm-3.x.yaml`, which I did not take — the GPU health monitor does not read host journals, so splitting there changes nothing for this component. Happy to be shown otherwise.

health-monitors/gpu-health-monitor/gpu_health_monitor/healthz.py
```diff
@@ -20,32 +20,36 @@
...
 import time
+from dataclasses import dataclass, field
 from http.server import ThreadingHTTPServer
 
 from prometheus_client import MetricsHandler
 
 
-class _last_reconcile:
+@dataclass
+class _LastReconcile:
```
deesharma24, line 33: why we have this change? What value it is adding?

PR #1492, opened 2026-07-14
fix: NIC monitor bug fixes for missing/dropping failure events

### LINKED ISSUE #1488: [Bug]: NIC monitor can miss or drop failure events
### Prerequisites

- [X] I can reproduce this issue

### Code of Conduct

## Summary

A code audit of the NIC health monitor (during the investigation that produced NVIDIA/NVSentinel#1361/NVIDIA/NVSentinel#1379)<br>found six independent reliability gaps in state/counter detection and event delivery. Each can<br>cause a real NIC failure to be under-reported, silently dropped, or a healthy node to be<br>permanently marked FATAL. They are filed together because they share the same subsystem and<br>several interact; they are independent of the fix in PR NVIDIA/NVSentinel#1462 and of the platform-connector<br>entity-matching bug (tracked separately).

> Line references are against `main` @ `4a3f2d54`.

---

## Finding 1 — A flapping link loses fatal protection after its first repair

For runtime IB link loss, fatality is by design delegated to the `link_downed` counter check<br>(`DOWN/Polling` state events are documented as non-fatal; `link_downed` is fatal with a<br>`delta > 0` threshold on the 1s cadence, so a cable pull produces a FATAL within \~1s). The bug<br>is that this safety net works exactly once per boot:

`link_downed` fires FATAL and latches (`pkg/checks/counter/evaluator.go`, `isBreached`),<br>clearing only on an admin counter reset (`perfquery -x`) or reboot — the latch itself is<br>documented design. But fatal counter events are relabeled under the *state* check's name<br>(`fatalCheckName`), with the same NIC+port entities. Downstream, conditions are keyed by<br>check name + entities, so when the port recovers and the state check emits its routine<br>`IsHealthy=true` under that same identity, the platform clears the counter-originated fatal<br>condition — while the breach latch stays set and suppresses every subsequent breach.

**Combined effect:** first cable pull → FATAL via `link_downed`. Cable re-seated → the ordinary<br>state recovery clears the condition (nobody runs `perfquery -x` after fixing a cable). Second<br>pull → state check: non-fatal `DOWN/Polling` only (by design); counter: suppressed by the<br>latch. **No fatal signal anywhere until an admin counter reset or reboot.** A flapping link is<br>exactly the failure mode this creates.

**Expected:** counter-originated conditions carry their own identity (the degradation check<br>name) so a state recovery cannot clear them; the latch then correctly holds the fatal condition<br>until a real counter reset. (Re-arming the latch on state recovery is not a safe alternative:<br>the counter snapshot is frozen while breached, the state and counter checks run on separate<br>goroutines over unsynchronized maps, and it changes the documented counter-reset lifecycle.<br>Note the identity split changes which node-condition name fatal counter breaches appear under.)

---

## Finding 2 — FATAL/recovery events are dropped permanently if publish fails

Events are edge-triggered (emitted once per health-boundary crossing), but delivery is<br>at-most-once: every check updates `previousPorts`/breach flags and persists them to the state<br>file **inside** `Run()`, before `monitor.runChecks` publishes. On publish failure the batch is<br>logged and dropped (`pkg/monitor/monitor.go`). When the platform-connector socket is absent<br>(e.g., during its rolling restart), `healthpub` returns immediately with **zero retries** — and<br>its own doc comment says callers must not advance local state on that error, which the NIC<br>monitor violates.

**Effect:** a port that goes DOWN during a \~30s platform-connector restart produces one FATAL<br>event that is dropped; the next poll sees no boundary crossing and never re-emits. The port can<br>be hard-down indefinitely with the platform never told. The mirror image: a dropped *recovery*<br>event leaves a node stuck FATAL forever.

**Expected:** check state should only advance after a successful publish (or unacked events<br>should be buffered and re-sent next poll). Note the fix must gate **both** the in-memory maps<br>and the on-disk state file, and applies equally to the counter checks (the breach latch is set<br>before publish).

---

## Finding 3 — `ACTIVE → INIT → DOWN` can produce zero events

Transitions are evaluated as a healthy/unhealthy boolean, so only the first crossing emits.<br>On Ethernet/RoCE (`pkg/checks/state/ethstate.go`): `ACTIVE→INIT` returns nil (non-DOWN unhealthy<br>states are ignored) **but still records INIT as previous state**; the subsequent `INIT→DOWN` is<br>then unhealthy→unhealthy and suppressed. Net result: a RoCE port that dips through any<br>intermediate state on its way down emits **no event at all**. On IB the same sequence emits only<br>a single non-fatal INIT event. Whether a dead port produces a fatal depends on whether the 1s<br>poll happened to sample the intermediate state.

**Expected:** emit on severity escalation within unhealthy states (track the unhealthy<br>sub-state, not a binary), or at minimum never let a recorded-but-unreported state consume the<br>boundary crossing.

---

## Finding 4 — Transient sysfs read errors become false disappearance FATALs

Three paths, all producing `REPLACE_VM` for hardware that is still present:

1. `discoverDevice` failure (e.g., `ports/` listing fails mid driver/fw reset) silently drops<br>the device (Debug log only, `pkg/discovery/discovery.go`) → the same poll's disappearance<br>detection raises FATAL "disappeared — hardware failure".
2. A failed `link_layer` read leaves `LinkLayer=""` → the port matches neither state check →<br>FATAL port-disappearance.
3. If `/sys/class/infiniband` is briefly absent (driver unload/reload, mount hiccup),<br>`DiscoverDevices` returns an **empty result with no error** → every previously-known NIC<br>gets a simultaneous FATAL.

There is no N-consecutive-misses debounce and no "all devices vanished at once → systemic, not<br>18 simultaneous hardware failures" guard. Routine `mlx5_core` fw resets / driver reloads<br>transiently produce exactly these read failures.

**Expected:** read errors treated as "unknown — hold previous state"; disappearance requires N<br>consecutive misses (persisted, so the debounce survives pod restarts); whole-class<br>disappearance treated as a monitor-degraded signal (abort the poll loudly, hold state), not<br>per-device hardware failure.

---

## Finding 5 — Disappearance has no recovery path

When a disappeared device re-enumerates (driver reload completes, fw_reset done), its ports are<br>"first-seen" and healthy first-seen ports emit nothing outside a boot-ID baseline<br>(`pkg/checks/state/ibstate.go` / `ethstate.go`). The device-level FATAL from the disappearance<br>is never cleared — the node stays `REPLACE_VM` until the next host reboot. Notably, the<br>*counter* subsystem handles the same admin action gracefully (driver reload resets counters →<br>recovery event); the state side has no equivalent.

**Expected:** when a previously-disappeared device/port re-enumerates healthy, emit a healthy<br>event with the same entities as the disappearance FATAL. The "device was fataled" marker must<br>be persisted (an in-memory marker does not survive a pod restart between disappearance and<br>re-enumeration).

---

## Finding 6 — Cross-layer device tracking produces spurious disappearances that can never recover

Both state checks add **every monitored device** to their tracked device set regardless of link<br>layer (`collectDevicesAndPorts` in `pkg/checks/state/ibstate.go` / `ethstate.go`, lines 172–193<br>and 163–186); port filtering happens later. Removing a pure-IB NIC therefore also produces a<br>spurious `EthernetStateCheck` "RoCE device disappeared from sysfs" FATAL for a device that has<br>no Ethernet ports (and vice versa). Because any recovery path is driven by the check's *own*<br>port events (Finding 5), a device with no Ethernet ports can never clear the Ethernet-side<br>condition: it sticks, and the node stays cordoned indefinitely — even after the device<br>re-enumerates and the IB side recovers.

**Expected:** device lifecycle (disappearance/recovery) is tracked per link layer — a device<br>belongs to a check only while it exposes at least one port of that check's layer. For the<br>residual case (a tracked device re-enumerates with its ports reflashed to the sibling layer),<br>emit a device-level healthy event so the outstanding FATAL cannot be orphaned.

---

## Interactions / severity

Findings 1+2 are the highest real-world risk (permanent false negatives on real failures);<br>findings 4+5 combine into "a routine driver reload permanently fatals a healthy node," and<br>finding 6 makes the stuck condition survive even a successful recovery on mixed-layer nodes.<br>Suggested split if triage prefers smaller PRs: Finding 1 standalone, Finding 2 standalone<br>(architectural), Findings 3+4+5+6 as one state-check batch.

### Component

Health Monitor

### Steps to Reproduce

All on a node with Mellanox IB/RoCE NICs running nic-health-monitor (fault injection on a dev<br>node; findings 2–3 also reproducible in the unit-test harness with the sysfs MockReader).

**Finding 1:**

1. Pull a cable on a monitored IB port → observe the FATAL from `link_downed` is published<br>under check name `InfiniBandStateCheck` (the state check's identity — this is the collision),<br>alongside a non-fatal state event showing `DOWN/Polling`.
2. Re-seat the cable, do NOT reset counters → observe healthy recovery event under the same<br>check name and entities; platform clears the fatal condition.
3. Pull the same cable again → observe: only a non-fatal state event; no FATAL from any source<br>(`link_downed` increments but the latched breach suppresses it).

**Finding 2:**

1. Scale platform-connector to 0 (or delete its pod so the UDS socket is absent).
2. Take a monitored port down → monitor logs "Platform-connector socket missing; skipping send".
3. Restore platform-connector → the FATAL is never re-sent; no condition is created.

**Finding 3 (unit harness or timing-dependent on hardware):**

1. Drive a RoCE port ACTIVE→INIT on one poll, INIT→DOWN on the next (MockReader).
2. Observe zero HealthEvents across both polls.

**Finding 4:**

1. `modprobe -r mlx5_ib mlx5_core && modprobe mlx5_core` (or `mlxfwreset`) on a monitored node.
2. During the reload window, observe FATAL "disappeared from /sys/class/infiniband — hardware<br>failure" events for devices that return seconds later.

**Finding 5:**

1. Continue from Finding 4: after the driver is back and all ports are ACTIVE/LinkUp,
2. Observe that no healthy event is emitted for the disappearance FATAL; the node condition<br>remains until host reboot.

**Finding 6:**

1. On a node with IB NICs and at least one RoCE NIC monitored, remove a single **IB-only**<br>device from sysfs (driver unbind / fault injection) long enough to trigger disappearance.
2. Observe **two** FATALs for the one device: `InfiniBandStateCheck` ("NIC mlx5_X<br>disappeared…") and `EthernetStateCheck` ("RoCE device mlx5_X disappeared from sysfs") —<br>the latter for a device that has no Ethernet ports.
3. Restore the device → the Ethernet-side condition has no possible recovery event (the device<br>never produces an Ethernet port event); on current main both conditions stick per Finding 5,<br>but the Ethernet one remains stuck even once Finding 5 is fixed.

### Environment

* NVSentinel version: v1.13.0

### Logs/Output

```
# Finding 1, first pull — the counter FATAL is filed under the STATE check's identity:
{"msg":"Health event sent","check":"InfiniBandStateCheck","is_fatal":true,
 "message":"Port mlx5_X port 1: link_downed - Port Training State Machine failed - QP disconnect (value=1, delta=1, ...)"}

# Finding 1, second pull — state event is non-fatal, no counter FATAL follows:
{"msg":"Health event sent","check":"InfiniBandStateCheck","is_fatal":false,
 "message":"Port mlx5_X port 1: state DOWN, phys_state Polling"}
# (no subsequent is_fatal=true event; link_downed breach is latched)

# Finding 2 — dropped batch during connector restart:
{"msg":"Platform-connector socket missing; skipping send.","monitor":"nic-health-monitor"}
{"msg":"Failed to send health events","check":"InfiniBandStateCheck"}
# (no re-send after connector returns)

# Finding 4 — driver reload window:
{"msg":"Health event sent","is_fatal":true,
 "message":"NIC mlx5_X disappeared from /sys/class/infiniband/ - hardware failure"}

# Finding 6 — spurious cross-layer disappearance for an IB-only device:
{"msg":"Health event sent","check":"EthernetStateCheck","is_fatal":true,
 "message":"RoCE device mlx5_X disappeared from sysfs"}
```

KaivalyaMDabhadkar:
Fixes #1488

The NIC health monitor could silently lose or never emit failure events, and could fabricate fatal disappearance events from transient sysfs read errors. This PR fixes all six findings from the issue; each fix below maps to its finding number.

### Fixes

**Finding 1 — A flapping link loses fatal protection after its first repair.** Fatal counter events were relabeled under the state check's name (`fatalCheckName`) with the same entities, so an ordinary `ACTIVE/LinkUp` state recovery cleared the counter-originated condition downstream while the breach latch kept suppressing further breaches — the `link_downed` safety net worked exactly once per boot. Fixed by identity separation: fatal counter events now keep their `*DegradationCheck` name, so a state recovery can no longer clear them and the latch correctly holds the condition until a real counter reset. Recovery events use the breach flag's persisted `CheckName`, so conditions filed under the legacy name still clear after upgrade. (Latch re-arming on state recovery was rejected for the reasons noted in the issue: frozen snapshots, cross-goroutine maps, documented lifecycle.)

**Finding 2 — FATAL/recovery events were dropped permanently if publish failed.** Checks advanced and persisted state inside `Run()` before `Publish`, violating `healthpub`'s documented caller contract — any platform-connector outage permanently consumed the health boundary. Checks now implement a transactional `Prepare`/`Commit`/`Discard` interface (`checks.TransactionalCheck`); the monitor commits only after successful publication (immediately for zero-event polls), so undelivered events are re-observed and re-emitted every poll until they land. The gate covers both the in-memory maps and the on-disk state file, and the counter evaluators (cloned per poll, so a discarded poll also rolls back breach latches and reset detection).

**Finding 3 — `ACTIVE → INIT → DOWN` produced zero events.** Both `evaluatePortTransition`s treated health as a boolean and suppressed all unhealthy→unhealthy changes, so the escalation to DOWN was swallowed (on Ethernet, the INIT dip emitted nothing at all). Both checks now emit when the unhealthy sub-state escalates non-fatal→fatal; de-escalation stays silent.

**Finding 4 — Transient sysfs read errors became false disappearance FATALs.** Discovery now reports `Complete`/`UnreadableDevices`: top-level ENOENT is an incomplete observation (checks with prior state abort the poll loudly, holding state — no more mass disappearance from a driver reload), enumerated-but-unreadable devices are retained as *unknown* with their last-known state (card homogeneity is skipped while discovery is uncertain), and genuine absence is debounced over 3 consecutive complete enumerations with the miss counts persisted so the debounce survives pod restarts.

**Finding 5 — Disappearance had no recovery path.** A `disappearedLatch` is now set with the FATAL and **persisted in the state file** (survives pod restarts, reboots, and discovery-scope changes, matching the existing card-anomaly latch rationale). The first healthy port observed on a re-enumerated latched device emits a healthy event with the same NIC entity and consumes the latch.

**Finding 6 — Cross-layer device tracking produced spurious disappearances that could never recover.** Device lifecycle tracking is now scoped per link layer: a device belongs to a check only while it exposes at least one port of that check's layer, so removing a pure-IB NIC no longer raises an `EthernetStateCheck` FATAL. For the residual case (a latched device re-enumerates with its ports reflashed to the sibling layer, or a latch created before this fix), the check emits a device-level healthy event and consumes the latch — which also self-heals any already-wedged node on upgrade.

Finding 7 —homogeneity false positive when exclusion removes a card's only active member (found during L40S validation): a card whose sibling function is the excluded default-route management NIC counted 0 active ports against a decisive peer mode, fataling its uncabled-by-design aux port fleet-wide on OCI BM.GPU.L40S.4. Fixed by exempting management-sibling cards from peer comparison, with latch self-healing for already-affected nodes.

### Behavioral / contract change (from Finding 1)

Fatal counter breaches now surface as `InfiniBandDegradationCheck` / `EthernetDegradationCheck` node conditions instead of `InfiniBandStateCheck` / `EthernetStateCheck`. Fault-quarantine rules (`agent + isFatal`) and the `RepeatedNICDegradation` analyzer rule (non-fatal only) are unaffected; e2e tests and docs are updated in this PR, but external dashboards/alerts keyed on the old condition names need updating.

### Testing

- Unit tests for every finding (escalation, debounce, held-state retention, latch persistence across restart, cross-layer recovery, transactional discard/re-emit, evaluator clone isolation); full suite passes with `-race`; `golangci-lint` clean.
- Validated end-to-end on an Azure A100 (AKS) dev cluster via a fake-sysfs fault-injection harness, following the repro steps from the issue: Finding 1 (counter FATAL survives a full port down/up cycle, clears only on counter reset), Finding 2 (~80s platform-connector outage — state held un-committed, events re-emitted each poll, delivered on socket return), Finding 3 (INIT→DOWN escalation emits FATAL), Finding 4 (sub-threshold absence and whole-tree vanish produce zero events), Finding 5 (confirmed disappearance recovers on re-enumeration), Finding 6 (IB-only device removal produces exactly one FATAL, one latch, full recovery), plus pod-restart persistence with no re-emit storm.
- Real-sysfs soak on the same node: classification (8 compute IB NICs, 1 storage RoCE NIC), PCI card grouping, and homogeneity baselines verified against raw sysfs + `gpu_metadata.json`.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **New Features**
  - NIC Health Monitor now uses a two-step polling flow that stages health transitions and commits them only after successful event publication.
  - Improved device disappearance handling with debounce (requires consecutive missed enumerations), better unreadable/unavailable device behavior, and link-layer scoping.
  - State persistence now retains device-disappearance latches and enumeration-miss counters across restarts.
- **Bug Fixes**
  - Host reboot/reset now preserves outstanding FATAL latches while clearing observational baselines.
  - Fatal severity escalation behavior and health event identity are now consistent across breach/recovery scenarios.
- **Documentation**
  - Updated NIC Health Monitor documentation for the revised reboot/recovery, link-counter, and link-state semantics.
- **Tests**
  - Expanded coverage for transactional polling, evaluator isolation, reboot/disappearance persistence, and persistence retry after failed saves.


docs/nic-health-monitor/link-counter-detection.md
```diff
@@ -623,23 +623,23 @@ T=25s   Poll:  link_downed = 0       (delta=0, not breached, no event)
...
        - **Delta counters** evaluate against the new baseline starting from the **second poll** (one polling interval later) — any increment above threshold triggers an unhealthy event.
        - **Velocity counters** wait for their full `velocityUnit` window (1s / 1m / 1h) to elapse against the new baseline before evaluating; they do not extrapolate from a partial sample.
    - Rationale: the node is effectively a fresh machine after reboot — NICs may have been replaced, firmware updated, cables reseated. The platform must be told that all previously-reported conditions are resolved unless new issues are detected on this boot.
 4. **If boot IDs match** (pod restart, same host boot):
-   - Restore all persisted state (counter snapshots, breach flags, port states, known devices)
+   - Restore all persisted state (counter snapshots, breach flags, port/device snapshots, debounce state, and outstanding latches)
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Fixed

health-monitors/nic-health-monitor/pkg/checks/counter/evaluator.go
```diff
@@ -136,6 +136,63 @@ func NewEvaluator(
...
+	clone := *e
+	clone.snapshots = cloneCounterSnapshots(e.snapshots)
+	clone.breachFlags = cloneBreachFlags(e.breachFlags)
+	clone.lastPollValues = cloneUint64Map(e.lastPollValues)
+	clone.ownedKeys = cloneSet(e.ownedKeys)
+
+	return &clone
+}
+
+func cloneCounterSnapshots(in map[string]statefile.CounterSnapshot) map[string]statefile.CounterSnapshot {
```
lalitadithya: can we use https://pkg.go.dev/maps#Clone here?

health-monitors/nic-health-monitor/pkg/checks/counter/evaluator.go
```diff
@@ -136,6 +136,63 @@ func NewEvaluator(
...
+	out := make(map[string]statefile.CounterSnapshot, len(in))
+	for k, v := range in {
+		out[k] = v
+	}
+
+	return out
+}
+
+func cloneBreachFlags(in map[string]statefile.CounterBreachFlag) map[string]statefile.CounterBreachFlag {
```
lalitadithya: Can we use generics here if the previous suggestion isn't possible?

health-monitors/nic-health-monitor/pkg/checks/state/base.go
```diff
@@ -117,6 +150,230 @@ func (b *baseStateCheck) seedFromPersistedState() {
 	)
 }
 
+func cloneBoolMap(in map[string]bool) map[string]bool {
```
lalitadithya: Same feedback as earlier. If we are doing the same operation in multiple places, we should refactor to a common package at minimum

health-monitors/nic-health-monitor/pkg/checks/counter/evaluator.go
KaivalyaMDabhadkar: Done. All four evaluator clone helpers deleted, Clone() now uses maps.Clone

health-monitors/nic-health-monitor/pkg/checks/counter/evaluator.go
KaivalyaMDabhadkar: Implemented maps.Clone, so not needed

health-monitors/nic-health-monitor/pkg/checks/state/base.go
KaivalyaMDabhadkar: Makes sense, cloneBoolMap/cloneIntMap deleted too so now both state checks now call maps.Clone directly so all the custom copy helpers are gone now

lalitadithya (approved):

PR #1496, opened 2026-07-16
feat: ADR for node validation

natherz97:
## Summary

This ADR covers adding node validation into NVSentinel. The ADR has 5 high-level sections covering:
 
1. Validation API and Configuration: covers which inputs are required from clients to request validation along with how operators can configure a set of validation tests.
2. Triggering Validation: covers how clients will invoke the validation API and gives options for using the node object, the HealthEvent API, or a CRD.
3. Validation Clients: covers how clients will trigger validation for new or existing nodes. This section covers how fault-quarantine will invoke the validation API.
4. State Transitions: covers how nodes transition between validation and remediation. It outlines how a scheduling gate is maintained in these states and when nodes transition between them.
5. Implementation: covers the core logic in the validation-controller and goes through an example for its batching algorithm for running tests across a group of nodes and tests.

The design includes support for the following:

- Validating new nodes or existing nodes post-remediation from NVSentinel
- Blocking validation from starting until readiness gates are satisfied on new nodes
- Adding a validation API with support for clients outside of NVSentinel where clients only need to know test names
- Supporting validation tests which can validate 1 GPU per node, multiple GPUs per node, or multiple nodes 
- Supporting validating partially drained or fully drained nodes
- Choosing which validation tests to run depending on the fault detected by NVSentinel
- Batching together and de-duplicating tests against the same set of GPUs and nodes from across multiple validation requests
- Supporting validation sessions where all tests must succeed prior to the node being uncordoned

## Type of Change

- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [X] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added ADR-045 describing a node validation workflow to confirm GPU readiness before nodes are made schedulable.
  * Documented how validation requests are triggered, including handling new nodes and post-remediation validation.
  * Defined validation configuration options (readiness criteria, default tests, provider retry/batching, and exclusive-access rules).
  * Explained how validation outcomes are converted into requests, coordinated across providers, and integrated with fault quarantine.


docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,436 @@
+# ADR-045: Node Validation
+
+## Context
+
+If a cluster has deployed GPU Operator and NVSentinel, there is no mechanism to ensure that new GPU nodes or existing nodes which underwent remediation are validated prior to being marked as schedulable. New nodes are able to accept GPU workloads after the nvidia-device-plugin advertises GPU capacity on a given node. Additionally, existing nodes which were remediated by NVSentinel are able to accept GPU workloads after the fault-quarantine module unquarantines them. In either case, there is no ability for an operator to declare that a set of GPU performance tests must succeed prior to a node being returned to service.
```
XRFXLP: There is some validation as part of `nvidia-operator-validator` that runs before device-plugin advertisment in GPU operator that validates GPUDirect storage, GDRCopy and few sample cuda programs are working as expected otherwise blocks the later stages.

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,436 @@
...
+
+## Validation API and Configuration
+
+In order to run validation tests against GPU nodes, a cluster operator will need the ability to define a set of supported validation tests. Additionally, a client requesting that a GPU node undergoes validation will need to provide the following inputs:
+
+```
+nodes:
+  - name: gpu-node-01
+    gpusImpacted:
+      - GPU-abc123
```
XRFXLP: Woudl it be more ergonomic to use GPU index instead of GPU UUID?

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,436 @@
...
+  - nccl-loopback
+```
+
+**Validation API overview:**
+
+* **nodes / gpusImpacted:** the client will need to provide the impacted nodes along with the list of impacted GPUs per node as API input. If no GPUs are provided, we will assume that all GPUs on the given node need to be validated.
+* **tests:** the client will need to pass the set of tests which should be executed against the given entities. While the ValidationConfiguration will support a set of default tests, clients will need the ability to specify which tests should be run depending on if a new node is being validated or if a specific fault was encountered.
+    * All entities provided in the validation request will be targeted by the same set of tests. If a client would like to request a different set of tests depending on the impacted entities, they will need to create multiple validation requests.
+* **exclusiveAccess:** a boolean which indicates whether the node is fully drained. If true, the node is idle and any GPU may be validated. If false, workloads may still be running on GPUs not included in the list of impacted entities, so we require at least 1 GPU to be provided in the list of impacted entities.
```
XRFXLP: This looks like a conflicting config, one can have a workload running but with GPUs empty: is it possible to avoid this situation of impasse?

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,436 @@
...
+
+```
+apiVersion: nvsentinel.nvidia.com/v1alpha1
+kind: ValidationConfiguration
+metadata:
+  name: default
+spec:
+  autoCordon: false
+  removeCordon: true
+  validationReadinessCriteria:
```
XRFXLP: Would CEL predicate a better fit here?

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,436 @@
...
+      minimumGPUsPerNodePerBatch: 1 
+      supportsBatchingNodes: false
+      minimumNodesPerBatch: 1       
+      batchFailurePolicy: fail
+```
+
+**Validation configuration overview:**
+
+* **autoCordon**: indicates whether nodes needing validation should be cordoned by the validation-controller if not already cordoned. The cordon should only be applied after the validationReadinessCriteria is satisfied.
+    * NVSentinel will leave nodes cordoned for post-remediation validation so this option would only be needed for cordoning new nodes eligible for validation.
```
lalitadithya: would this mean that the validation controller will uncordon?

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,436 @@
...
+* **minimumNodesPerBatch:** if supportsBatchingNodes is true, this setting indicates the minimum number of nodes that must be provided to the test. Note that the nodes could be provided from the same validation request or from across multiple validation requests.
+    * As an example, nccl-loopback supports 1 node per batch whereas  nccl-all-reduce requires 2 nodes per batch.
+* **batchFailurePolicy:** this setting defines what to do when a test cannot start due to it not meeting its configured minimumGPUsPerNodePerBatch or minimumNodesPerBatch. Possible values include:
+    * **fail:** mark the overall validation request as failed if this individual test cannot be run.
+    * **open:** do not run this individual test if the batch sizes are not satisfied and implicitly consider the test as successful.
+    * **wait:** wait for the next batch window to see if other validation requests can help it meet its minimum requirements. This policy will result in other tests in the same validation request being blocked.
+
+## Triggering Validation
+
+**Option 1: Node object as the validation trigger (not recommended)**
```
lalitadithya: Can we put the recommended option first?

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,436 @@
...
+Rather than require clients to update node object status conditions and annotations or rely on the HealthEvent API, we could provide a ValidationRequest CRD to allow a client to declare that a particular node or a group of nodes should be validated.
+
+```
+apiVersion: nvsentinel.nvidia.com/v1alpha1
+kind: ValidationRequest
+metadata:
+  name: gpu-node-validation
+spec:
+  nodes:
+    - name: gpu-node-01
```
lalitadithya: Can we allow for label selectors here so that if we have large number of nodes, we don't have to list them all one at a time?

natherz97: Yes, we can support label selectors or explicit node names. If a label selector is provided, we won't allow passing the list of impacted GPUs so all GPUs per node would be implicitly targeted. I'll update this section to give an example using a label selector.

docs/designs/045-node-validation.md
natherz97: Done in next revision.

docs/designs/045-node-validation.md
natherz97: Yes, if removeCordon is true, the validation-controller will uncordon any nodes which were cordoned before a node had a ValidationRequest or were cordoned by the controller itself if autoCordon is true. Once we're fine with the high-level approach, I will publish a revision that covers how we plan to maintain the scheduling gate for new or existing nodes undergoing validation along with how state transitions work between maintenance and validation.

docs/designs/045-node-validation.md
natherz97: How common would it be for clients who detect a GPU fault to have the GPU index but not the UUID? For NVSentinel, it looks like gpu-health-monitor will include the index, PCI ID, and GPU UUID whereas syslog-health-monitor will only have the PCI and GPU UUID:
```
  {
    "version": 1,
    "agent": "gpu-health-monitor",
...
    "entitiesImpacted": [
      {
        "entityType": "GPU",
        "entityValue": "1"
      },
      {
        "entityType": "PCI",
        "entityValue": "0009:01:00.0"
      },
      {
        "entityType": "GPU_UUID",
        "entityValue": "GPU-2a16739b-5a94-a53a-44d7-1e7370aa6b23"
      }
    ],
...
},
  {
    "version": 1,
    "agent": "syslog-health-monitor",
...
    "entitiesImpacted": [
      {
        "entityType": "PCI",
        "entityValue": "0009:01:00"
      },
      {
        "entityType": "GPU_UUID",
        "entityValue": "GPU-2a16739b-5a94-a53a-44d7-1e7370aa6b23"
      }
    ],
...
}
```
Comparing to GPU reset, we support the GPU UUID or PCI ID but not index. Let me know your thoughts but it seems like we should support GPU UUID + PCI ID.

docs/designs/045-node-validation.md
natherz97: Great point, we're already using CEL for the validation rule-sets. I updated the ValidationConfiguration to support CEL expressions against either the node object or against the pod objects running on that node.

docs/designs/045-node-validation.md
natherz97: I just checked this, it looks like the nvidia-device-plugin's init container is only checking on the container-toolkit is ready:
```
$ kubectl get pod nvidia-device-plugin-daemonset-tkhsx -n gpu-operator -o yaml
...
  initContainers:
  - args:
    - until [ -f /run/nvidia/validations/toolkit-ready ]; do echo waiting for nvidia
      container stack to be setup; sleep 5; done
```
So even though the nvidia-operator-validator validates in this order for driver -> container-toolkit -> cuda-validation -> device-plugin, wouldn't this mean we're advertising GPUs before the CUDA tests since we're only waiting on the toolkit?

docs/designs/045-node-validation.md
natherz97: The allowed set of inputs include:
- gpusImpacted empty + exclusiveAccess true: empty GPUs means that all need to be validated and the node was fully drained. In this case, we will only allow full node tests.
- gpusImpacted populated + exclusiveAccess true: GPUs passed means that a subset of GPUs need to be validated and the node was fully drained. In this case, we allow tests targeting specific GPUs or full node tests. 
- gpusImpacted populated + exclusiveAccess false: GPUs passed means that a subset of GPUs need to be validated and the node was only partially drained. In this case, we will only allow tests targeting specific GPUs.

Any of the 3 cases exclusiveAccess could fail depending on how the requested tests are configured but all 3 of those should be valid as long as the tests allow it. This last case is not allowed and we'd fail any request:
- gpusImpacted empty + exclusiveAccess false: empty GPUs means that all need to be validated and the node was only partially drained. In this case, we know it's not safe to validate all GPUs since the node is not idle. 

We will prevent case 4 with a mutating webhook or handle it in the controller and mark the request as failed without starting any test. Does that answer your question?

docs/designs/045-node-validation.md
XRFXLP: > but it seems like we should support GPU UUID + PCI ID.

One data point is enough to disambiguate the problematic GPU, right? I was point to GPU index just because it is easy to retrieve and remember on a node.> but it seems like we should support GPU UUID + PCI ID.

One data point is enough to disambiguate the problematic GPU, right? I was point to GPU index just because it is easy to retrieve and remember on a node.

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,455 @@
...
+      gpusImpacted: []
+      exclusiveAccess: true
+  tests: []
+  newNode: true
+```
+
+**Post-remediation validation**
+
+The fault-quarantine module will need the ability to create a ValidationRequest after an unquarantine event if any of the unhealthy events during that quarantine session requested validation. If validation is required, fault-quarantine will create a ValidationRequest, release ownership of the node, but retain the node cordon. The validation-controller will be responsible for removing the cordon after all requested validations complete against the node.
```
lalitadithya: > The validation-controller will be responsible for removing the cordon after all requested validations complete against the node.

won't this result in the fault quarantine module considering this as quarantine node manually uncordoned? Should we add a blurb on how we will handle this?

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,455 @@
...
+    1. Since HealthEvents can be resolved independently during a quarantine session, we will need to either evaluate HealthEvents when they are removed from the quarantineHealthEvent or track all HealthEvents and evaluate all of them when the unquarantine event occurs.
+2. If the drain for the given HealthEvent is completed, accept the ValidationRequest from that fault.
+3. For events with completed drains, set exclusiveAccess to false for COMPONENT\_RESET actions and set it to true for all other recommended actions.
+4. Optionally, for events with cancelled drains, check whether the impacted entities overlap with those of any completed drain. For example, if there was a completed drain against the same GPU or if any event had a full drain, this would permit the ValidationRequest from cancelled events.
+
+In summary, evaluation of a HealthEvent requires both the validation rule-set execution to determine its corresponding tests and checking the drain status for the event. These 2 results can be used to construct a ValidationRequest.
+
+Suppose that the quarantine session for a node had 3 unhealthy events which resulted in the following requests:
```
lalitadithya: what would happen if the events recover at different points of time, would this result in the validation requests being created at different point in time?

docs/designs/045-node-validation.md
natherz97: When NVSentinel is the validation client, we won't always have the index right? Since we use the GPU_UUID for partial drains + GPU resets, should we keep this consistent? Otherwise we'd need to enforce that all health-monitors populate the index and switch our partial drain entity from GPU_UUID over to the index.

docs/designs/045-node-validation.md
natherz97: Fault-quarantine will remove all of it's owning labels and annotations from the node and only leave the cordon in place so it will not detect a manual uncordon. If we went with an alternative design and required fault-quarantine to uncordon after all validations complete, then we would run into the situation you described. I'll add more details when I publish the next revision which covers how we plan to maintain the cordon across validation and remediation.

docs/designs/045-node-validation.md
natherz97: We only send the validation request when the unquarantine event occurs after the last event recovers. If the node was still cordoned by NVSentinel, we'd be blocked from starting validation for any event that was from a partial unquarantine so we should hold off sending any validation request until everything recovers (which also gives us an opportunity to batch across GPUs on the same node or across duplicated tests).

The piece we do need to decide is how we want to batch the ValidationRequests from all unhealthy events that existed during that quarantine session. To keep the client-side validation logic in NVSentinel simple, we could send 1 ValidationRequest per health event. The alternative would be to batch client-side to reduce the number of CRDs but this comes with tradeoff of duplicating batching logic we will have in the validation-controller.

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,455 @@
...
+      supportsBatchingGPUsPerNode: true
+      minimumGPUsPerNodePerBatch: 1 
+      supportsBatchingNodes: false
+      minimumNodesPerBatch: 1       
+      batchFailurePolicy: fail
+```
+
+**Validation configuration overview:**
+
+* **autoCordon**: indicates whether nodes needing validation should be cordoned by the validation-controller if not already cordoned. The cordon should only be applied after the validationReadinessCriteria is satisfied.
```
tanishagoyal2: I think we have to cordon the nodes everytime we are planning to run validation test. In what cases is it possible to have autocordon as false? 🤔

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,455 @@
...
+    * NVSentinel will leave nodes cordoned for post-remediation validation so this option would only be needed for cordoning new nodes eligible for validation.
+* **removeCordon:** indicates whether nodes should be uncordoned after completing validation. A node will only be uncordoned once there are no pending validation requests (a node may be targeted by multiple validation requests).
```
tanishagoyal2: Will validation-controller remove cordoning irrespective of the validation tests pass or failed?

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,455 @@
...
+      gpusImpacted:
+        - GPU-def456
+      exclusiveAccess: false
+    - name: gpu-node-03
+      gpusImpacted: []
+      exclusiveAccess: true
+  tests:
+    - dcgm-level4
+    - nccl-loopback
+```
```
cbumb: The CRD spec is fully described but there's no status field. Clients (fault-quarantine, new-node watcher) have no machine-readable way to know if validation passed, failed, or is in progress.

docs/designs/045-node-validation.md
cbumb: A failed, terminal ValidationRequest is no longer pending so the controller could uncordon a node that failed validation. The cordon-on-failure retention policy needs to be explicitly stated.

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,455 @@
...
+* **minimumNodesPerBatch:** if supportsBatchingNodes is true, this setting indicates the minimum number of nodes that must be provided to the test. Note that the nodes could be provided from the same validation request or from across multiple validation requests.
+    * As an example, nccl-loopback supports 1 node per batch whereas  nccl-all-reduce requires 2 nodes per batch.
+* **batchFailurePolicy:** this setting defines what to do when a test cannot start due to it not meeting its configured minimumGPUsPerNodePerBatch or minimumNodesPerBatch. Possible values include:
+    * **fail:** mark the overall validation request as failed if this individual test cannot be run.
+    * **open:** do not run this individual test if the batch sizes are not satisfied and implicitly consider the test as successful.
+    * **wait:** wait for the next batch window to see if other validation requests can help it meet its minimum requirements. This policy will result in other tests in the same validation request being blocked.
```
cbumb: Lets says nemotron4-15b requires 18 nodes minimum. If only 10 nodes are ever in a validation window, the wait policy blocks indefinitely keeping all those nodes cordoned forever. There's no maximum wait count or deadline defined.

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,455 @@
...
+* **supportsBatchingNodes:** indicates whether this test supports testing multiple nodes in a single test. For example, the k8s-job-provider will only support a single node per test so supportsBatchingNodes will be false. However, the nccl-provider has to support node batching to support multi-node NCCL tests so supportsBatchingNodes will be true.
+* **minimumNodesPerBatch:** if supportsBatchingNodes is true, this setting indicates the minimum number of nodes that must be provided to the test. Note that the nodes could be provided from the same validation request or from across multiple validation requests.
+    * As an example, nccl-loopback supports 1 node per batch whereas  nccl-all-reduce requires 2 nodes per batch.
+* **batchFailurePolicy:** this setting defines what to do when a test cannot start due to it not meeting its configured minimumGPUsPerNodePerBatch or minimumNodesPerBatch. Possible values include:
+    * **fail:** mark the overall validation request as failed if this individual test cannot be run.
+    * **open:** do not run this individual test if the batch sizes are not satisfied and implicitly consider the test as successful.
```
cbumb: If I am reading it right. When `open` fires because minimum batch size wasn't met, the test is "implicitly considered successful." The resulting node condition is indistinguishable from a genuine test pass. nemotron4-15b use `open` today. A node could enter production having skipped validation entirely with no indication.

docs/designs/045-node-validation.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Validation Clients
+
+Assuming that we are proceeding with option 1 above using a CRD as the validation trigger, we will need to ensure that a ValidationRequest CRD is created for nodes in the following contexts.
+
+**New node validation**
+
+The validation-controller can watch for new nodes missing a given node status condition to signal that it should create a ValidationRequest for new nodes. After a ValidationRequest is created, the controller will set the given status condition to true to ensure it does not create duplicated ValidationRequests for new nodes. The name of this status condition will be added to the ValidationConfiguration. Additionally, we can add a list of tests to the ValidationConfiguration that should be executed against new nodes to prevent each ValidationRequest from needing to populate the list of tests. This allows the configuration to specify both a default group of tests and a group of tests for new nodes. In this example, any ValidationRequest for a new node which does not provide an explicit list of tests will run nccl-all-reduce:
+
```
cbumb: AFAIK when first deployed into an existing cluster, every GPU node appears as `new` and will trigger validation simultaneously.

docs/designs/045-node-validation.md
natherz97: It's possible the node will already be cordoned when a client requests that a node needs validated so in that cases autoCordon could be false. For example, nodes returning from remedation from NVSentinel will already be cordoned from fault-quarantine when fault-quarantine requests validation. Additionally, for new nodes an operator may choose to cordon their nodes after launch before they request validation to prevent having to wait for the validation-controller to do it. 

It's also possible an operator may want to run validation without cordoning at all (if they know there's no competing workloads) and making this configurable through autoCordon will leave that as an option.

docs/designs/045-node-validation.md
natherz97: The cordon will persist until all requested validation tests succeed. If validation fails, an operator will either need to trigger a new validation request with the same requests or delete the previous request. The behavior is similar to fault-quarantine where it only removes the cordon if all health events have been remediated, we're doing something similar where the validation-controller will only remove the cordon if all validation requests succeed (which could be split across multiple CRDs).

docs/designs/045-node-validation.md
natherz97: Good callout, I will add the status conditions in the next revision. The structure of what we need to track in the status field depends on how we want to batch tests across multiple ValidationRequests. Since we have consensus that we want the validation-controller to support batching across ValidationRequests (including nodes and tests), I will add the required status conditions to the design.

docs/designs/045-node-validation.md
natherz97: Agree this wording is unclear, I will update it in the next revision. This should say, "A node will only be uncordoned once there are no pending, in-progress, or failed validation requests." In other words, we'll uncordon only after all validation requests have succeeded.

docs/designs/045-node-validation.md
natherz97: To avoid waiting indefinitely, an operator can choose to use the fail or open policies. The deadline in this case comes from the batchPeriod setting we already have in the configuration. So an operator can tell the controller to wait for X number of minutes for 18 nodes and choose to either fail or pass the corresponding validation requests which requested nemotron4-15b. 

Unless we want to add a greater deadline, there's no need to define a separate deadline for the batchFailurePolicy=wait since this is implicitly part of batchPeriod. 

Edit: latest revision is keeping the batchFailurePolicy wait option since there may be cases where we want to use this. For example, if there's a new rack and only 17/18 nodes are launched, we may want to wait indefinitely on the CSP launching the last instance if we are aware of the issue. But agree in general we should choose between fail or open options.

docs/designs/045-node-validation.md
natherz97: Yes, that understanding is correct. If an operator doesn't want to consider a certain test type as best effort, they could use the failure policy to ensure a given test is never skipped when requested. 

If we require a full rack for a given test type, an operator may want to mark the failure policy for new nodes as fail and mark it as open for existing nodes. In some cases, if a rack is already in production, it may not be feasible to run a large training test if a subset of nodes are running workloads. In this case, it may be preferrable to consider the test as best effort. Compared to a new rack where all the nodes would be available for validation and we'd want to consider it blocking.

docs/designs/045-node-validation.md
natherz97: Great callout, let me put more thought into this and add a rollout plan to the ADR. One option would be to check the node object's creation time and have a cutoff on when it's eligible to trigger validation so existing nodes wouldn't be required. Another option would be to require that the node is currently cordoned since new nodes will have a Skyhook that would cordon it but then we could run into issues where old nodes which are manually cordoned could have validation triggered later in lifetime which wouldn't be desirable so the creation time may be the better option.

docs/designs/045-node-validation.md
cbumb: Thanks, the new rack versus in production rack rationale makes sense, and I'm on board with operators choosing fail or open per case.

My main worry was actually the observability side rather than whether open should exist. Even when open is the right call for an existing rack, the skip is currently silent. The node ends up with the same validated condition as one that genuinely passed nemotron, so later on we can't tell which nodes actually ran the test and which just skipped it.

Could we record the skip distinctly when open fires, maybe a separate reason on the condition, an annotation, or surfacing skipped tests in the ValidationRequest status? Best effort is fine, I'd just like it to be auditable rather than indistinguishable from a real pass.

natherz97: Makes sense, I'll make sure that is surfaced in ValidationRequest status. That's the only possible case we skip tests but it's also possible that nodes are skipped if they're part of a ValidationRequest but are deleted so I'll also make sure we're exposing which nodes were skipped as part of any request.

docs/designs/045-node-validation.md
natherz97: We're going to punt on supporting targeting specific GPUs for validation in the initial phase so we can revisit this when that functionality is added. I do think we should be consistent with how we're using the GPU_UUID for partial drains and GPUResets in NVSentinel but I think it's reasonable for the API to also support PCI ID or index as well if there's a client outside of NVSentinel who cannot use the UUID.

docs/designs/045-node-validation.md
natherz97: Latest revision adds the new node agent restriction to prevent running against all nodes on initial deployment of the validation-controller if it's configured to target new nodes.

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+      minimumGPUsPerNodePerBatch: 1
+      supportsBatchingNodes: false
+      minimumNodesPerBatch: 1
+      batchFailurePolicy: fail
+```
+
+**Validation configuration overview:**
+
+* **newNodeValidation:** groups the configuration for detecting and testing new nodes.
+    * **criteria:** a set of CEL expressions evaluated against each node to determine whether it requires new node validation. If all expressions evaluate to true, the validation-controller will create a ValidationRequest for that node. The CEL environment exposes both node and pod data. Once a ValidationRequest is created, the controller sets a condition on the node to prevent duplicate requests from being created on subsequent evaluations.
```
lalitadithya: > The CEL environment exposes both node and pod data.

I presume pod data here indicates only pods on the node under validation right and not all pods in the cluster?

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+      expression: >
+        !node.status.conditions.exists(c,
+          c.type == "NewNodeValidated" && c.status == "True")
+    - name: recently-joined
+      expression: 'now() - timestamp(node.metadata.creationTimestamp) < duration("15m")'
+    - name: gpu-present
+      expression: '"nvidia.com/gpu.present" in node.metadata.labels'
+    tests:
+    - nccl-all-reduce
+  schedulingGate:
```
lalitadithya, line 84: do we also need some "soft" gates like labels?

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+    * **autoTaint:** indicates whether the taints listed in taints should be applied to nodes entering validation if they are not already present. Taints are applied after the validationReadinessCriteria is satisfied, in the same step as autoCordon.
+    * **removeTaint:** indicates whether the taints listed in taints should be removed from a node after all pending validation requests for that node are satisfied.
+    * **taints:** the list of taints the controller manages on nodes under validation. Each taint specifies a key, an optional value, and an effect.
+    * **tolerations:** all test pods created by the validation-controller will include tolerations for the unschedulable taint and for every taint listed in taints, regardless of whether those taints are present on the node. This covers the case where a node was cordoned or tainted by an operator manually, by the validation-controller via autoCordon or autoTaint, or not cordoned or tainted at all. This ensures test pods can always be scheduled on nodes that are gated from regular workloads.
```
lalitadithya: wouldn't it be simpler for the test pods to just tolerate broadly every taint and not just the ones in the list similar to what we have for platform connectors?

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+  schedulingGate:
+    autoCordon: false
+    removeCordon: true
+    autoTaint: true
+    removeTaint: true
+    taints:
+    - key: nvsentinel.nvidia.com/validation-pending
+      value: "true"
+      effect: NoSchedule
+  validationReadinessCriteria:
```
lalitadithya: I think we should make the `AND`/`ALL` relationship here more explicit in the contact, so we could do:

```yaml
  validationReadinessCriteria:
    all:
      - name: ...
         expression
```

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+The provided list of tests in the validation request must map to a supported test in the ValidationConfiguration. Upon receiving a validation request, the validation-controller will reference the active ValidationConfiguration:
+
+```
+apiVersion: nvsentinel.nvidia.com/v1alpha1
+kind: ValidationConfiguration
+metadata:
+  name: default
+spec:
+  newNodeValidation:
+    criteria:
```
lalitadithya, line 76: same comment as in the other case below to make the `AND`/`ALL` explicit in the contract

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+    * **supportsTestBatching:** indicates which test providers support batching their tests into a single test provider request. In the example above, the nccl-provider supports batching, which allows all tests using that provider (dcgm-level4, nccl-loopback, nccl-all-reduce, and nemotron4-15b) to be included in a single CRD. Conversely, the k8s-job-provider does not support batching, so the cuda-smoke-test will not be batched with any other tests using that provider.
+    * **retries:** the number of retries we will allow for all tests referencing this test provider. Note that this setting is at the test provider level and not at the individual test level because if a test provider supports batching, all tests may need to be retried together.
+* **batchPeriod:** the period that allows for batching of tests, nodes, and GPUs across validation requests. This is a global setting that covers all test providers.
+* **maxConcurrentGroups:** the maximum number of test groups that may run concurrently. Groups are additionally constrained by node overlap. Two groups that share a node will never run at the same time regardless of this setting.
```
lalitadithya, line 205: remind me please, what is the use case for this?

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+    * As an example, nccl-loopback supports 1 node per batch whereas nccl-all-reduce requires 2 nodes per batch.
+* **batchFailurePolicy:** this setting defines what to do when a test cannot start due to it not meeting its configured minimumGPUsPerNodePerBatch or minimumNodesPerBatch. Possible values include:
+    * **fail:** mark the overall validation request as failed if this individual test cannot be run.
+    * **open:** do not run this individual test if the batch sizes are not satisfied and implicitly consider the test as successful.
+    * **wait:** wait for the next batch window to see if other validation requests can help it meet its minimum requirements. This policy will result in other tests in the same validation request being blocked.
+* **bandwidthGBps:** an optional pass criterion for bandwidth-oriented tests such as NCCL tests. If set, the test is considered failed if the measured bandwidth falls below this threshold.
+* **goodputRatio:** an optional pass criterion for training-oriented tests such as Nemotron. If set, the test is considered failed if the measured goodput ratio falls below this threshold.
```
lalitadithya: I'd recommend we offload this to the test frameworks to evaluate and fail the tests. I don't think the validation controller will need to be concerned with interpreting tests results and that can become complex very quickly

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+
+## Validation Clients
+
+Assuming that we are proceeding with option 1 above using a CRD as the validation trigger, we will need to ensure that a ValidationRequest CRD is created for nodes in the following contexts.
+
+**New node validation**
+
+The validation-controller will evaluate the newNodeValidation.criteria CEL expressions against each node to determine whether it requires new node validation. The condition name used to prevent duplicate ValidationRequests does not need to be declared separately in the ValidationConfiguration since it is implicit in the CEL expression itself. After a ValidationRequest is created, the controller sets that condition on the node so that subsequent evaluations of the criteria no longer match. The newNodeValidation.tests field specifies which tests to run for new nodes, taking precedence over defaultTests. This allows the configuration to specify both a default group of tests and a group of tests specifically for new nodes. In this example, any ValidationRequest for a new node will run nccl-all-reduce:
```
lalitadithya: > After a ValidationRequest is created, the controller sets that condition on the node so that subsequent evaluations of the criteria no longer match.

this seems a bit leaky because we are now expecting the operators to understand some internal implementation, i.e., this CEL expression will have to always be present in the configuration for the system to work. 

Can we instead make this explicit in the contract, so maybe something like:

```yaml
spec:
  newNodeConditionType: NewNodeValidated
```

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+3. Defer to each test provider to move nodes to remediation. Rather than require the validation-controller to move nodes which fail validation, test providers could independently move nodes to remediation. This could be preferable because test providers will likely have additional information about which group of nodes failed along with the failure reason.
+    - Options 2 and 3 will both send an unhealthy event due to a given validation test failing. However, there is not a clear healthy event that can be leveraged by these health-monitors. Validation passing from a re-run cannot be used as a healthy event since this would violate the validationReadinessCriteria (and complicate our state machine by requiring validation to supersede remediation). As a result, a manual signal to send a healthy event is likely required after manual investigation for options 2 and 3.
+
+## Implementation
```
lalitadithya, line 642: can we add some info about how the validation controller will translate the validation requests to the CRs required by the test providers?

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+    Node3:
+      gpuUUIDs: [GPU1, GPU3]
+      exclusiveAccess: true
+```
+
+**Step 3: Check the exclusiveAccess and gpuUUIDs settings per node per test**
+
+Run the following checks per node per test. If any check fails, the violating ValidationRequest is immediately marked as failed and excluded from the batch.
+
+1. Each test name must exist in the ValidationConfiguration.
```
lalitadithya: can we make this fail quicker? maybe we do it as a part of the initial reconcile or via a webhook depending on what fields are mutable in the CR

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1101 @@
...
+status:
+  phase: Running
+  attempts:
+    - batchExecutionRef: batch-20240714-a3b1
+      phase: Running
+      startTime: "2024-07-14T10:00:00Z"
+
+# BatchExecution
+apiVersion: nvsentinel.nvidia.com/v1alpha1
+kind: BatchExecution
```
lalitadithya: should we call this `ValidationBatchExecution`? `BatchExecution` seems a bit generic and it may have conflicts with other projects and sometimes automation doesn't do well with same Kind under different APIs

docs/designs/images/049-node-validation-statemachine.png
lalitadithya, line 1: nit, maybe we should add a prefix for the file name to avoid conflicts in future `049-state-machine`

docs/designs/049-node-validation.md
natherz97: That is correct, I updated this section to make that explicit.

docs/designs/049-node-validation.md
natherz97: We could definitely support this but I don't think we have a use-case currently. This would be straightforward to add as a follow-up if someone needs this feature. Are you on board if we track this as a follow-up feature?

lalitadithya: yeah, we can do it as a follow up, but I think we should just mention how the API would like when we add it.

docs/designs/049-node-validation.md
natherz97: Good point, I was assuming we'd need to set explicit taintSelectors in the underlying test provider. As long as we have a way to configure these tolerations that sounds reasonable. 

It looks like our NCCL test provider for MPI tests tolerate all taints by default if you don't set a taintSelector and otherwise will have explicit tolerations if a taintSelector is specified. For training tests, it looks like they don't tolerate all taints by default if you don't set a taintSelector and otherwise if will have the same behavior and have explicit tolerations if a taintSelector is specified. On top of that it looks like the catalog will give some explicit tolerations which are appended to the logic above.

Let me know if that's consistent with your understanding. Can we commit to being to tolerating all taints in the training tests by default if a taintSelector isn't specified? In that case I think we should be fine to say we'll tolerate all taints by default.

docs/designs/049-node-validation.md
natherz97: Sounds good, I'll update this in a revision.

docs/designs/049-node-validation.md
natherz97: Will do in the next revision.

docs/designs/049-node-validation.md
lalitadithya: yeah, we can commit to tolerating all taints if no selector is specified

docs/designs/049-node-validation.md
natherz97: This would be to limit the number of test groups we have running at a time. For example, if we have a ValidationRequest which targets all nodes in the cluster with a K8s job, we could use this setting to run on 10 nodes at a time rather than against the all nodes at the same time.

Discussed offline that we won't introduce a setting for maximumNodesPerGroup initially which would allow specifying a limit on the number of nodes per group (or alternatively passing this setting to the given test provider). Currently, our NCCL provider doesn't have a limit on the number of nodes tested together. I'll mention this is something we could support later.

docs/designs/049-node-validation.md
natherz97: Agreed but since this is specified as part of our NCCL test provider CRD, we need a way to pass this from the validation-controller to the provider per test. So this setting will solely be used for passing the bandwidth/goodput thresholds to the test provider and the test provider will determine pass/fail based on that threshold.

I will clarify this in the next revision.

docs/designs/049-node-validation.md
natherz97: Included in the next revision.

docs/designs/049-node-validation.md
natherz97: Agreed, we will require the new node status condition to be set explicitly rather than inferring from CEL expression.

docs/designs/049-node-validation.md
natherz97: Good point, I wanted to include all these checks together in the same section but I will clarify that this property can be checked during admission time to block object creation.

docs/designs/images/049-node-validation-statemachine.png
natherz97: Fixed.

docs/designs/049-node-validation.md
natherz97: Discussed offline that we will use the dynamic client for all test providers regardless of whether they use built-in objects or custom resources. Next revision will have this added to the ValidationConfiguration.

docs/designs/049-node-validation.md
natherz97: I updated this section to call out that all of these checks can be executed during admission. I think this is preferable to instantly marking the CRD as failed and now we don't need to reason about whether those invalid ValidationRequests should contribute to a validation session for a given node or not.

docs/designs/049-node-validation.md
natherz97: I went ahead with ValidationBatch, let me know if you'd prefer the full ValidationBatchExecution name.

lalitadithya:
lgtm

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1168 @@
...
+* **Common validation interface:** a shared interface for requesting validation prevents all clients from needing to maintain logic to work with the different test providers, including our supported K8s job and NCCL test providers. A client who wants to validate a group of nodes or GPUs only needs to know the test name, and we will maintain a common entrypoint in the validation-controller.
+* **Batching multiple validation requests:** if a single client independently creates validation requests across nodes or multiple clients request validation, a dedicated controller provides the opportunity to batch multiple validation requests together. This allows for de-duplicating validation requests across the same node for the same test and also batching nodes together in multi-node tests. This prevents clients from needing to handle batching on their side prior to issuing validation requests.
+    * For example, if NVSentinel remediates 2 nodes and requests that both undergo a multi-node NCCL test, we have the opportunity to include those 2 nodes together in a single test from our NCCL provider.
```
suket22: NVSentinel only remediates a node one at a time right? So if that's the case, why would NVSentinel ever ask for a multi-node NCCL test? 

IIUC, the test request wouldn't be fulfilled until atleast one more node would also be unhealthy at the same time. Unless the validation controller decides to downgrade the multi-node NCCL test to a single-node NCCL test 😅

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1168 @@
...
+  - nccl-loopback
+```
+
+**Validation API overview:**
+
+* **nodes / gpuUUIDs:** the client will need to provide the impacted nodes along with the list of impacted GPUs per node as API input. If no GPUs are provided, we will assume that all GPUs on the given node need to be validated.
+* **tests:** the client will need to pass the set of tests which should be executed against the given entities. While the ValidationConfiguration will support a set of default tests, clients will need the ability to specify which tests should be run depending on whether a new node is being validated or a specific fault was encountered.
+    * All entities provided in the validation request will be targeted by the same set of tests. If a client would like to request a different set of tests depending on the impacted entities, they will need to create multiple validation requests.
+* **exclusiveAccess:** a boolean which indicates whether the node is fully drained. If true, the node is idle and any GPU may be validated. If false, workloads may still be running on GPUs not included in the list of impacted entities, so we require at least 1 GPU to be provided in the list of impacted entities.
```
suket22, line 60: My thinking is that the validation controller should always have exclusive access to the GPUs it needs to test. Fully agree with you that it's the client's responsibility to do the drain before submission to validate but maybe this is something we can infer?

So, 
* If GPU UUIDs is empty, then the whole node should be drained. 
* If a GPU UUID(s) is explicilty mentioned, then only workloads using that device should be drained off. 

And we don't need to expliclty told anything in this boolean. My motivation here is to keep the API surface as small as possible

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1168 @@
...
+    * For example, if NVSentinel encountered an XID which required a GPU reset, only the impacted GPU will be eligible for validation due to the partial node drain executed before the reset. Conversely, if the XID required a node reboot, the node would be eligible for validation against all its GPUs due to the full node drain before the reboot.
+    * This highlights that node draining is not the responsibility of the validation-controller, and validation clients must execute node drains prior to creating requests and notifying the controller if the node is partially or fully drained.
+
+The section below on "Triggering Validation" covers how clients will interact with this API and whether this input is passed as a CRD, an NVSentinel HealthEvent, or as part of the node object.
+
+The provided list of tests in the validation request must map to a supported test in the ValidationConfiguration. Upon receiving a validation request, the validation-controller will reference the active ValidationConfiguration:
+
+```
+apiVersion: nvsentinel.nvidia.com/v1alpha1
+kind: ValidationConfiguration
```
suket22, line 70: I wonder if we could decompose this API surface so it's not one, large flat config? Maybe the test provider configs could be separated out.

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1168 @@
...
+    * **criteria:** a set of CEL expressions evaluated against each node to determine whether it requires new node validation. All expressions must evaluate to true (along with the newNodeCondition check above). The CEL environment exposes both the node being validated and the pods scheduled on it.
+        * In the provided example, a node is eligible for new node validation when it was created within the last 15 minutes (to prevent triggering against existing nodes on the initial deployment of the validation-controller) and has the nvidia.com/gpu.present label.
+    * **tests:** the list of tests to run for new nodes. These take precedence over defaultTests when a ValidationRequest is created for a new node.
+* **schedulingGate:** groups all node scheduling controls applied by the validation-controller during the lifecycle of a validation request.
```
suket22: re: autolabel and autotaint etc - why is labeling and tainting nodes the responsibility of the validation controller anyway? Shouldn't we just enforce our clients to specify this?

I'm wondering if we need so many knobs in general, and if we could safely default some of this.

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1168 @@
...
+* **maximumNodesPerGroup:** this setting is initially out-of-scope and is not included in the example above. There's currently no way to limit the number of nodes per test group if the given test sets supportsBatchingNodes to true. It may be desirable to limit the number of nodes running per test group by adding logic to the validation-controller to not exceed this number of nodes per batch or by deferring to the test provider and passing this value to it. If this is implemented in the validation-controller, we will need to ensure that the minimumGPUsPerNodePerBatch is met per group which may require borrowing nodes between groups.
+* **testGroupTimeout:** the maximum duration allowed for a single test group attempt before it is marked as failed. This is a global setting applied to all test groups. A per-test-provider timeout could optionally be added in the future if individual test providers require different timeouts. If the test provider itself enforces a timeout, this setting could optionally be passed to the provider (for example K8s Jobs support an activeDeadlineSeconds setting).
```
suket22: I'm confused by this one. Won't a reasonable timeout duration be a function of the size of the group? Also why isn't this an attribute of the test config since time to execute should really be dependent on the type of test

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1168 @@
...
+* **tests:** the set of supported tests that can be requested by clients in validation requests. The test name is the only piece of the ValidationConfiguration that clients must be aware of when interacting with the validation API.
+
+**Test configuration:**
+
+* **provider:** indicates which test provider uses the given test. In our example, we support a nccl-provider which is used for the dcgm-level4, nccl-loopback, nccl-all-reduce, and nemotron4-15b tests, and a k8s-job-provider which is used for the cuda-smoke-test.
```
suket22, line 211: Just curious, what other providers do we anticipate? I'd have thought nccl-provider would give me variations of the nccl tests but nemo4-15b doesn't feel like a natural fit IMO.

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1168 @@
...
+3. Defer to each test provider to move nodes to remediation. Rather than require the validation-controller to move nodes which fail validation, test providers could independently move nodes to remediation. This could be preferable because test providers will likely have additional information about which group of nodes failed along with the failure reason.
+    - Options 2 and 3 will both send an unhealthy event due to a given validation test failing. However, there is not a clear healthy event that can be leveraged by these health-monitors. Validation passing from a re-run cannot be used as a healthy event since this would violate the validationReadinessCriteria (and complicate our state machine by requiring validation to supersede remediation). As a result, a manual signal to send a healthy event is likely required after manual investigation for options 2 and 3.
+
+## Implementation
+
+### Batching Algorithm
```
suket22, line 644: This almost deserves its own document because batching this, breaking it into groups, and then getting to a validation execution is fairly complicated. I feel like if we remove the idea of batching, we also maybe don't need another object that a ValidationExecution and instead just rely on the status of a ValidationRequest to communicate when the validation are done.

docs/designs/049-node-validation.md
```diff
@@ -0,0 +1,1168 @@
...
+
+### Optional Functionality
+
+In order to cut scope from this design, we could remove any of the following functionality.
+
+1. Only support ValidationRequests from NVSentinel clients and do not support external clients. This would remove support for new node validation.
+2. Only support tests which require the node to be fully drained and don't support tests which target specific GPUs.
+3. Don't allow clients to specify which tests should be run and always rely on a default set of tests. This would require that NVSentinel would always have to run the same set of tests regardless of the fault needing validation.
+4. Don't support batching across ValidationRequests and require clients to specify their desired node and test batching. This would remove the need to create intermediate ValidationExecution CRDs but defer batching complexity to each client.
```
suket22: Strongly feel we should push this out. Let's scaffold this entire thing e2e and build batching capabilities next. I think that greatly simplifies a lot of what we're doing, reduces our API surface etc

natherz97: I'd be in favor of this. It does defer the batching complexity to clients but that should be restricted to the new node case. This would give us the following behavior:
1. Each ValidationRequest is processed independently and the controller will run tests only for the provided nodes and tests. Since we're not combining multiple requests together, we can remove the ValidationExecution CRD and directly track our test CRDs in the status of the ValidationRequest.
3. For post-remediation validation, we will create 1 ValidationRequest per quarantine session which targets 1 node.
4. For new node validation, the client will handle the batching of new nodes into a single ValidationRequest. Since means we'll wait for new nodes joining the cluster for X number of minutes, and then create 1 ValidationRequest for all those nodes.

One other downside of not batching across ValidationRequests is that we can't de-duplicate multiple requests for the same node. For example, if NVSentinel requested a node is validated with DCGM multiple times, we'd have to re-run the same test. We could add de-duplication support later if this becomes a problem (by doing something like having NVSentinel check for in-progress validations from the node validation but we can explore that later).

docs/designs/049-node-validation.md
natherz97: Yes, that's correct NVSentinel will currently only remediate individual nodes and ask for single-nodes to be remediated. So you're right that the test request for any multi-node NCCL test wouldn't be fulfilled unless another node became unhealthy during the controller's batch period. I agree that a client should specify the tests they want to run based on the nodes it has available to test, rather than hoping that a particular test can be run based on the requests from other clients.

Let's discuss on your last comment on if we need to implement this functionality.

docs/designs/049-node-validation.md
natherz97: Suppose we experience an XID error which required a reboot and after the reboot we want to run 2 tests:
1. DCGM diagnostics against the single GPU which experienced it
2. NCCL loopback against all GPUs on the given node

Suppose our DCGM diagnostics tests requires an explicit GPU to be passed and our NCCL loopback test requires no GPUs input because all GPUs on the node are tested. If we removed exclusiveAccess from the API input, there's no way for us to pass the GPU UUID we need for test 1 while also indicating that test 2 is safe because the node is fully drained. So I think we need this second piece of state indicating the drain state of the node outside which GPUs were impacted.

natherz97: Initially the plan is to not support targeting explicit GPUs so we don't need to add the exclusiveAccess nor gpuUUIDs to the ValidationRequest API yet. So clients will only need to pass the list of tests and they'll be responsible for ensuring test provider has exclusive access by fully draining nodes prior to requesting validation. Does that sound reasonable for us to punt on this?

docs/designs/049-node-validation.md
natherz97: +1 to too many settings for the scheduling gate. Let's remove support for labeling for now so we only have to worry about tainting and cordoning.

At a minimum, we need to support removeCordon and removeTaint to allow our support for validation sessions (where we promise a given node will remain cordoned or tainted until all validation requests against it succeed). Internally, we will always apply the cordon/taint before requesting validation so NKX does not have a use-case for the corresponding autoCordon or autoTaint setting. I'm not sure if that's a valid assumption to make for all operators. For example if someone wants to cordon/taint new nodes after launch and they don't have an ability to change Kubelet start-up flags or have a separate entity apply the scheduling gate they would likely want to use this autoCordon or autoTaint feature. This is consistent with how Skyhook allows an operator to apply the runtime-required taint as part of node launch or Skyhook will add it if not already present on an existing node: https://github.com/NVIDIA/nodewright/blob/main/docs/runtime_required.md#auto-tainting-new-nodes.

So in summary we have:
- keep: removeCordon, removeTaint, taints
- remove: removeLabel, autoLabel, labels
- optional: autoCordon, autoTaint

Let me know you're thoughts on the 2 optional fields and I'll update this.

docs/designs/049-node-validation.md
natherz97: We can support any K8s object for a test provider but the examples we're planning on using are the K8s job provider and a multi-node NCCL/training provider. I'll update the provider name to be more accurate after we make it public but our internal "nccl-provider" supports both NCCL and training already.

docs/designs/049-node-validation.md
natherz97: Agreed, this will be simplified if we remove support for batching ValidationRequests (which allows us to remove the ValidationExecution CRD completely). If we decide to go ahead with removing that batching support, I'll update this section in the existing ADR and if we still think it needs an additional review after the reduction in scope I'll make it a follow-up ADR.

docs/designs/049-node-validation.md
natherz97: Good callout, this doesn't make sense as a global setting across all providers. It is possible that tests across a given provider are batched into a single CRD. For example, if we request single-node NCCL and multi-node NCCL tests on the same set of nodes, we'll create a single test provider CRD. So we'll need to enforce this timeout at the test provider level rather than at the test level. I'll update this to be a timeout at the test provider level and later on we can add more complex timeout behavior if this needs to be configured depending on the node count.

docs/designs/049-node-validation.md
natherz97: I had a similar concern, at a minimum we can remove the in-line CRD templates and replace it with a file name. We could also move the tests or testProviders lists to other files but I'll hold off on that for now unless you think otherwise.

docs/designs/049-node-validation.md
lalitadithya: just to make sure we are on the same page, we don't want to push the batching behaviour to clients. What we are saying here is that we will do the first iteration without batching and then we would do a follow up? So for the first iteration, we have to some basic batching as a part of new node validation which we will then move out later

natherz97: We'll have 2 main loops initially:

1. Primary controller loop to process ValidationRequests. 
a. It depends on if the test providers support batching of nodes but for test providers like Excalibur all nodes in the ValidationRequest will be included in the same Certification
2. Controller loop to watch for nodes matching the new node validation condition. This will batch new nodes into a single ValidationRequest which will be processed by the primary controller.

Later if we need to additional batching, we could extend either loop 1 or 2

docs/designs/049-node-validation.md
suket22: > So clients will only need to pass the list of tests and they'll be responsible for ensuring test provider has exclusive access by fully draining nodes prior to requesting validation. Does that sound reasonable for us to punt on this?

Yup exactly. Move complexity on the client for now to simplify what we're building here. Even in the example you've mentioned above, I think we just assume whatever GPUs are passed to us as being free, regardless of this flag.

docs/designs/049-node-validation.md
suket22: >  This is consistent with how Skyhook allows an operator to apply the runtime-required taint as part of node launch or Skyhook will add it if not already present on an existing node:

Maybe it's just me, but I find the premise of Skyhook applying taints so racy. If you don't have access to modifying start up taints, how do you even ensure that only Skyhook lands first before anyone else? And if you already have access to that, why not specify whatever else you need to rather than letting Skyhook apply it?

I'm tempted to remove autoCordon and autoTaint too. Same principle of don't add it unless super compelling / we have signal from operators asking for this.

docs/designs/049-node-validation.md
natherz97: To clarify, any GPU we pass is assumed to be drained regardless of value of exclusiveAccess. So if we pass GPU1 and exclusiveAccess is false, that still means GPU1 is drained but GPU2, GPU3, and GPU4 are not drained. If we pass GPU1 and exclusiveAccess is true, that means GPU1, GPU2, GPU3, and GPU4 are all drained. Our test providers may require a specific GPU UUID as input or they might require all nodes are drained so I do think long term we need both of those inputs to the API.

We can revisit this API contract once we want to allow single-GPU validations since our initial implementation does not need either field provided.

docs/designs/049-node-validation.md
natherz97: You're right, Skyhook having to add taints does introduce a race condition. The node-readiness-controller calls out the same issue with applying taints for new nodes and recommends that you register nodes with the taints that the controller automatically applies. 

We can punt on the autoCordon + autoTaint functionality. Unfortunately, kubelet does not expose an ability to create new nodes that are cordoned so operators would have to rely on an external actor cordoning like Skyhook in all cases if we don't expose this feature on the validation-controller so the race condition will exist for cordons unless someone registers with a taint and replaces that with a cordon.

natherz97: Separately, I wanted to reason about how to roll this out to an existing cluster. If someone chooses to use Skyhook to apply the cordon for new node bootstrapping, I want to make sure we can support rolling this out safely. 

- We have to make sure the validation-controller is deployed before or at the same time as the Skyhook which cordons or else we'd have nothing to remove the cordon.
- On the initial deployment of the validation-controller, we don't want to validate all existing nodes in the cluster. We can ensure that nodes which are missing the validation status condition, are currently cordoned, and were created in the last 15 minutes are eligible for new node validation. If we don't have the creation timestamp check, we'd trigger validation for an arbitrarily old node if someone manually cordoned it during it's lifetime.

lalitadithya: > I find the premise of Skyhook applying taints so racy. If you don't have access to modifying start up taints, how do you even ensure that only Skyhook lands first before anyone else?

it is racy for certain, but I think they accepted that risk because not every hosted service offers the option of startup taints. 

I think for cordons, since nodes can't join with a cordon it might still be worthwhile to have it

lalitadithya (approved):

docs/designs/049-node-validation.md
suket22: > not every hosted service offers the option of startup taints.

Yeah it sure is a gap most folks should fix. I'm actually pushing for this in K8s06 in the AI Cloud Req docs - https://docs.nvidia.com/dsx/ncp/nvidia-requirements-for-ai-clouds/home#kubernetes-as-a-service-kaas-requirements fwiw

suket22:
LGTM.

docs/designs/049-node-validation.md
natherz97: This will be a trivial follow-up to add. I'll track the automatic cordoning/tainting as follow-ups along with the ability to test individual GPUs but for an MVP I think we can punt on those 2 items.

suket22 (approved):

PR #1499, opened 2026-07-16
Collect bug reports when GPU resets fail

### LINKED ISSUE #1494: [Feature]: Add support to automatically collect an nvidia-bug-report when GPU reset fails
### Prerequisites

### Code of Conduct

### Feature Summary

This feature would allow the GPU reset script to also trigger collection of an nvidia-bug-report if the nvidia-smi fails to reset the target GPU. This feature should also upload the report to the incluster-file-server if enabled. 

### Problem/Use Case

We have seen occurrences of GPU resets failing in response to XID errors which have a RESET_GPU resolution. This includes bursts of XIDs which all recommend reset or an XID 154 recommending a reset be triggered in response of an XID 79 needing a reboot.

### Proposed Solution

To prevent needing to enable the log collection component in fault-remediation to always collect the nvidia-bug-report prior to all remediation actions, triggering the bug report and upload to the incluster-file-server from the GPU reset script will allow only triggering the upload on failed GPU resets. This ensures we can engage XID catalog owners either for an incorrect recommended action for a given XID or for issues related to nvidia-smi reset functionality.

### Component

Fault Management

AnouarMohamed:
## Summary

- Collect `nvidia-bug-report` when the GPU reset script observes a failed `nvidia-smi --gpu-reset`.
- Upload the collected report to the configured file-server endpoint when `UPLOAD_URL_BASE` is set.
- Wire the upload URL through janitor GPU reset job config and Helm values, with automatic in-cluster file-server defaults.

## Validation

- `docker run --rm -v /home/anouar/NVSentinel/gpu-reset:/mnt koalaman/shellcheck:stable /mnt/gpu_reset.sh`
- `bash -n gpu-reset/gpu_reset.sh`
- `env GOCACHE=/tmp/nvsentinel-go-cache GOMODCACHE=/tmp/nvsentinel-go-mod-cache go test ./pkg/config`
- `env GOCACHE=/tmp/nvsentinel-go-cache GOMODCACHE=/tmp/nvsentinel-go-mod-cache go test ./pkg/controller -run '^$'`
- `make -C distros/kubernetes lint`
- `helm template` with `global.inclusterFileServer.enabled=true` and `false`
- `git diff --check`

Fixes #1494

## Summary by CodeRabbit

- **New Features**
  - On GPU reset failures, the system now collects NVIDIA bug reports using the node name and timestamp, and uploads them on a best-effort basis.
  - Upload destination is configurable; when in-cluster file server is enabled, a default in-cluster upload URL is used unless overridden (Helm values updated).
- **Bug Fixes**
  - Improved failure output formatting for easier reading, and bug-report collection no longer blocks subsequent health checks.
- **Chores**
  - Reduced `gpu-reset` image size via streamlined package installation.


janitor/pkg/config/default.go
```diff
@@ -292,5 +295,21 @@ func getDefaultGPUResetJobTemplate(namespace string, image string, secrets []Ima
 		job.Spec.Template.Spec.RuntimeClassName = &runtimeClassName
 	}
 
+	job.Spec.Template.Spec.Containers[0].Env = append(job.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
```
natherz97: It should be safe to define both of these environment variables in-line right?

gpu-reset/gpu_reset.sh
```diff
@@ -46,6 +50,103 @@ nvidia_smi_helper() {
...
+    log "WARN: No nvidia-bug-report artifact is available to upload."
+    return 1
+  fi
+
+  if [ -z "$upload_url_base" ]; then
+    log "INFO: UPLOAD_URL_BASE is not configured; nvidia-bug-report retained locally at $BUG_REPORT_OUTPUT_FILE"
+    return 0
+  fi
+
+  if ! command -v curl >/dev/null 2>&1; then
```
natherz97: We're not checking for the presence of other binaries in this script and we're installing it at build time so I think we can remove this check for consistency (given we're not checking for the existence of the nvidia-smi binary).

gpu-reset/gpu_reset.sh
```diff
@@ -46,6 +50,103 @@ nvidia_smi_helper() {
...
+  log "INFO: nvidia-bug-report collected: $BUG_REPORT_OUTPUT_FILE"
+}
+
+upload_bug_report() {
+  local upload_url_base="${UPLOAD_URL_BASE:-}"
+  local safe_node_name
+  local file_name
+  local upload_url
+
+  if [ -z "$BUG_REPORT_OUTPUT_FILE" ] || [ ! -f "$BUG_REPORT_OUTPUT_FILE" ]; then
```
natherz97: What do you think about having a single function that handles the bug report generation and upload? I would prevent us from needing to use a global variable to pass BUG_REPORT_OUTPUT_FILE and we wouldn't have to duplicate these checks across the 2 helper functions.

gpu-reset/gpu_reset.sh
```diff
@@ -46,6 +50,103 @@ nvidia_smi_helper() {
...
+    log "WARN: Failed to create nvidia-bug-report output directory: $local_bug_report_dir"
+    return 1
+  fi
+
+  bug_report_base="${BUG_REPORT_DIR}/nvidia-bug-report-${safe_node_name}-${BUG_REPORT_TIMESTAMP}"
+  bug_report_path="${bug_report_base}.log.gz"
+  local_bug_report_path=$(driver_root_path "$bug_report_path")
+
+  log "INFO: Collecting nvidia-bug-report for failed GPU reset..."
+  if ! nvidia_bug_report_helper --output-file "${bug_report_base}.log"; then
```
natherz97: The team which maintains XID catalog suggested also providing the ```--safe-mode``` to prevent the command hanging.

gpu-reset/gpu_reset.sh
```diff
@@ -46,6 +50,103 @@ nvidia_smi_helper() {
...
+  local_bug_report_dir=$(driver_root_path "$BUG_REPORT_DIR")
+
+  if ! mkdir -p "$local_bug_report_dir"; then
+    log "WARN: Failed to create nvidia-bug-report output directory: $local_bug_report_dir"
+    return 1
+  fi
+
+  bug_report_base="${BUG_REPORT_DIR}/nvidia-bug-report-${safe_node_name}-${BUG_REPORT_TIMESTAMP}"
+  bug_report_path="${bug_report_base}.log.gz"
+  local_bug_report_path=$(driver_root_path "$bug_report_path")
```
natherz97: Would it be cleaner to prepend local_bug_report_dir since we already have that stored as a local variable above?

gpu-reset/gpu_reset.sh
```diff
@@ -46,6 +50,103 @@ nvidia_smi_helper() {
...
 
+driver_root_path() {
+  if [ "$DRIVER_ROOT" = "/" ]; then
+    printf "%s\n" "$1"
+  else
+    printf "%s%s\n" "${DRIVER_ROOT%/}" "$1"
+  fi
+}
+
+safe_path_component() {
```
natherz97: Do we need this since node names are guaranteed to be valid DNS subdomain names?

gpu-reset/gpu_reset.sh
```diff
@@ -46,6 +50,103 @@ nvidia_smi_helper() {
   chroot "$DRIVER_ROOT" nvidia-smi "$@"
 }
 
+driver_root_path() {
```
natherz97: Can you name this function to something more descriptive like prepend_driver_root?

gpu-reset/gpu_reset.sh
```diff
@@ -36,7 +36,11 @@
 set -eou pipefail
 
 DRIVER_ROOT="${DRIVER_ROOT:-/}"
+NODE_NAME="${NODE_NAME:-unknown-node}"
 START_TIME=$(date +%s.%N)
+BUG_REPORT_TIMESTAMP="${BUG_REPORT_TIMESTAMP:-$(date +%Y%m%d-%H%M%S)}"
+BUG_REPORT_DIR="${BUG_REPORT_DIR:-/var/tmp/nvsentinel-gpu-reset}"
```
natherz97: Can we use the same path as what log-collector does? https://github.com/NVIDIA/NVSentinel/blob/main/log-collector/entrypoint.sh#L133

natherz97:
Thank you for picking this up! Nice work with the error handling in the bash script.

janitor/pkg/config/default.go
AnouarMohamed: oh yeah thats true, I’ll move both env vars inline.

gpu-reset/gpu_reset.sh
AnouarMohamed: okay

gpu-reset/gpu_reset.sh
AnouarMohamed: agreed, i will combine the collection/upload path so we can avoid the global BUG_REPORT_OUTPUT_FILE and keep the checks in one place

gpu-reset/gpu_reset.sh
AnouarMohamed: ok

gpu-reset/gpu_reset.sh
AnouarMohamed: ok im on it rn

gpu-reset/gpu_reset.sh
AnouarMohamed: ok im on it

gpu-reset/gpu_reset.sh
AnouarMohamed: yes i think thats better

gpu-reset/gpu_reset.sh
AnouarMohamed: Since Kubernetes node names are already valid DNS subdomain names, I’ll remove the extra sanitization and use NODE_NAME directly for the upload path

gpu-reset/gpu_reset.sh
```diff
@@ -46,6 +49,74 @@ nvidia_smi_helper() {
...
+    log "WARN: nvidia-bug-report collection failed."
+    return 1
+  fi
+
+  if [ ! -f "$local_bug_report_path" ]; then
+    log "WARN: nvidia-bug-report completed but output was not found: $local_bug_report_path"
+    return 1
+  fi
+
+  log "INFO: nvidia-bug-report collected: $local_bug_report_path"
```
natherz97, line 96: Might be worth adding a comment that this will be written the /run/nvidia/driver HostPath volume so this will persist after the GPU reset container exits. It will be lost on a reboot after /run is cleared.

gpu-reset/gpu_reset.sh
```diff
@@ -46,6 +49,74 @@ nvidia_smi_helper() {
...
+    return 0
+  fi
+
+  upload_url="${upload_url_base%/}/${NODE_NAME}/${BUG_REPORT_TIMESTAMP}/${bug_report_file}"
+
+  log "INFO: Uploading nvidia-bug-report to $upload_url"
+  if curl -fsS --connect-timeout 10 --max-time 120 -X PUT --upload-file "$local_bug_report_path" "$upload_url"; then
+    log "INFO: nvidia-bug-report upload complete: $bug_report_file"
+  else
+    log "WARN: Failed to upload nvidia-bug-report: $bug_report_file"
```
natherz97, line 109: Nit: should we return 1 here for consistency?

gpu-reset/gpu_reset.sh
```diff
@@ -46,6 +49,74 @@ nvidia_smi_helper() {
...
+  fi
+}
+
+collect_reset_failure_diagnostics() {
+  if [ "${COLLECT_BUG_REPORT_ON_RESET_FAILURE:-true}" != "true" ]; then
```
natherz97: We're not providing a Helm chart option to ever set COLLECT_BUG_REPORT_ON_RESET_FAILURE to false. Should we either make this configurable through the Helm chart or remove this check since right now it will always be true?

AnouarMohamed: done i removed it

gpu-reset/gpu_reset.sh
AnouarMohamed: i agree, done

gpu-reset/gpu_reset.sh
AnouarMohamed: done

natherz97 (approved):
Thanks again for picking this up, nice work.

PR #1508, opened 2026-07-20
feat: detect mlx5 NIC driver TX timeout and NAPI soft lockup via syslog-health-monitor

### LINKED ISSUE #1334: [Feature]: Detect mlx5 NIC driver TX timeout and NAPI soft lockup via syslog-health-monitor
### Prerequisites

### Code of Conduct

### Feature Summary

Add syslog NIC driver detection patterns for mlx5 TX timeout and mlx5 NAPI soft lockup events, and a health-events-analyzer correlation rule to escalate the combination to a fatal NIC driver wedge condition that triggers node quarantine/reboot.

## Problem

A node had its mlx5 NIC driver wedged for ~5 days while remaining `Ready` in Kubernetes.
The failure sequence was:

1. `mlx5_core <BDF> <iface>: TX timeout detected`
2. `BUG: soft lockup - CPU#N stuck for 226906s!` in
   `mlx5e_poll_ico_cq → mlx5e_napi_poll → net_rx_action`
   (CPU stuck in the mlx5 NAPI poll loop for ~63 hours)
3. Network-dependent services (e.g., SMB/CIFS CSI driver) became unresponsive
4. Any pod requiring network-backed volumes scheduled to this node got stuck permanently

The node was manually cordoned days later. NVSentinel did not detect or act on this.

## Current Limitations

- **nic-health-monitor**: If the NIC's RDMA object is not an `mlx5_*` InfiniBand device
  (e.g., `siw_*` Soft-iWARP), the monitor skips it entirely. Port state may remain
  `ACTIVE/LinkUp` and net counters (`tx_errors`, `carrier_changes`, etc.) may stay at zero
  even while the driver is wedged.

- **syslog-health-monitor**: The existing `netdev_watchdog` pattern
  (`NETDEV WATCHDOG.*mlx5_core.*transmit queue.*timed out`) may catch one log line, but:
  - It does not match `mlx5_core ... TX timeout detected` directly.
  - It does not match `BUG: soft lockup` with mlx5 NAPI stack evidence.
  - `netdev_watchdog` is non-fatal and disabled by default.
  - `SysLogsNICDriverError` is not in the default `enabledChecks`.

- **health-events-analyzer**: The `RepeatedNICDriverError` rule only covers existing non-fatal
  error codes (`netdev_watchdog`, `port_module_high_temp`, `pci_power_insufficient`,
  `module_unplugged`). There is no rule to escalate a TX timeout + mlx5 NAPI soft lockup
  combination to a fatal condition.

## 1. New syslog NIC driver patterns (`syslog-health-monitor`)

Add two new pattern definitions in `health-monitors/syslog-health-monitor/pkg/nicdriver/config.go`:

| Pattern Name | Regex | Severity | Recommended Action |
|---|---|---|---|
| `mlx5_tx_timeout_detected` | `mlx5_core.*TX timeout detected` | non-fatal | NONE |
| `mlx5_napi_soft_lockup` | `BUG: soft lockup.*mlx5e_(poll_ico_cq\|napi_poll)` | non-fatal (escalated by analyzer) or fatal | REPLACE_VM |

Notes:
- `mlx5_tx_timeout_detected` catches the mlx5 driver follow-up line even if the `NETDEV WATCHDOG`
  line is absent or dropped.
- `mlx5_napi_soft_lockup` catches the actual smoking gun: a CPU stuck in the mlx5 NAPI poll loop.
  This is distinct from generic soft lockups (which should not be fatal).
- Kernel stack traces may be multiline. If the monitor processes one line at a time,
  `mlx5_napi_soft_lockup` needs the `BUG: soft lockup` line and the mlx5 function names to appear
  in the same log record. If journald delivers them as separate lines, a small stateful detector
  or multiline aggregation may be needed.

## 2. Health-events-analyzer correlation rule

Add a new analyzer rule (e.g., `NICDriverWedge`) that correlates on the same node within a time
window:

- **Trigger event**: `SysLogsNICDriverError` with `errorCode` in
  [`netdev_watchdog`, `mlx5_tx_timeout_detected`]
- **Confirming event**: `SysLogsNICDriverError` with `errorCode` = `mlx5_napi_soft_lockup`
- **Result**: Fatal health event, recommended action = REPLACE_VM / reboot

This avoids making every generic soft lockup fatal while still catching the NIC-attributed wedge.

## 3. Fault quarantine

The existing fault-quarantine rule for `event.agent == 'nic-health-monitor' && event.isFatal == true`
should also cover fatal events from the analyzer or syslog monitor for this class. Verify that the
quarantine ruleset covers `event.agent == 'health-events-analyzer'` for NIC-related fatal events.

### Component

Health Monitor

KaivalyaMDabhadkar:
## Summary

Detects a wedged mlx5 NIC driver from kernel logs and quarantines the node, closing the gap where a NAPI soft lockup left a node degraded for days with no NVSentinel signal.

Fixes #1334

### What happened in the incident

The sequence was: mlx5 TX timeout → ~11 seconds later, `BUG: soft lockup - CPU#94 stuck for 226906s!` with the kernel stack trace pinned inside the mlx5 NAPI poll loop (`mlx5e_poll_ico_cq` → `mlx5e_napi_poll`). The driver's built-in recovery never kicked in, the CPU stayed wedged for days, and the node slowly degraded (CSI socket stalls, pods stuck Pending) until it was manually cordoned and rebooted.

None of the existing NIC driver patterns could catch this. The soft lockup message itself is generic — *every* kernel subsystem can trigger one — and the stack frame that pins it to mlx5 shows up as a **separate journal entry**. No single-line regex can connect the two.

### What this PR does

**New multi-line detector** (`pkg/nicdriver/softlockup.go`)

The core addition is a small state machine that watches the kernel log stream for a two-part signal:

1. **Header**: `BUG: soft lockup - CPU#N stuck for Ns!` — this "arms" a 150-line window
2. **Confirmation**: a non-speculative stack frame containing `mlx5e_poll_ico_cq` or `mlx5e_napi_poll` within that window

When both parts match, the monitor emits a fatal `REPLACE_VM` event with the CPU number and lockup duration extracted as structured metadata.

A few details that matter:

- The window is counted in **kernel log lines, not wall-clock time**. After a monitor restart, the journal backlog replays at full speed — hundreds of entries land with near-identical timestamps — so a time-based window would incorrectly pair headers and frames from completely different lockup dumps.
- Stack frames prefixed with `?` (the kernel's "I'm not sure this was actually on the call stack" marker) are ignored. `net_rx_action` — the generic softirq entry point — is deliberately excluded too, because it shows up in lockups from *any* NIC or network subsystem and would cause false attribution to mlx5.
- The kernel re-reports a persistent lockup roughly every 26 seconds, so there's a 30-minute cooldown to avoid flooding the pipeline with duplicate events.

**Two new non-fatal patterns**

- `mlx5_tx_timeout_detected` — the driver's own "TX timeout detected" message (from `en_main.c`)
- `mlx5_rx_timeout_detected` — the driver's "RX timeout on channel" message (from `reporter_rx.c`)

These fire on their own as non-fatal events. They often show up before or alongside a lockup, and they keep re-firing while a queue stays stuck, so they're useful both as early corroboration and as standalone signals for the analyzer's repetition rule.

**Fixed a broken existing pattern**

The `netdev_watchdog` pattern was silently dead on kernels >= 6.8 (and backports to 6.6.47 / 6.1.107). Kernel commit `e316dd1cf135` changed the message format — the old regex only matched the pre-6.8 wording. The new regex matches both old and new formats.

**Kernel-origin filter for NIC checks**

Added `SysLogsNICDriverError` to the list of checks that filter on `_TRANSPORT=kernel` in journald. Without this, userspace log lines (from applications, container runtimes, etc.) could land inside the detector's 150-line window and either waste window slots or, in a worst case, contain text that accidentally matches the stack frame pattern.

**Analyzer escalation fallback**

The health-events-analyzer's `RepeatedNICDriverError` rule now includes the two new error codes. If a node sees 3+ TX or RX timeout events within an hour, the analyzer fires a derived `CONTACT_SUPPORT` event. This is a backup — it flags repeated non-fatal onsets for human attention even if the full lockup sequence never appears.

**Chart defaults**

All three new patterns ship `enabled: false`, matching the existing eight NIC patterns. Nothing changes for any deployment until a fleet explicitly opts in. The recommended rollout (documented in the values file and in the design doc) is: enable with `processingStrategy: STORE_ONLY` first (events get stored and are visible, but no node conditions or cordons happen), then switch to `EXECUTE_REMEDIATION` after the shadow phase validates that the detector isn't producing false positives on your fleet's kernel logs.

### Design decisions

**Why the monitor makes the fatal call, not the analyzer**

This is the key architectural question and it came up repeatedly during design.

Even with the ongoing work to add HEA check names to FQ CEL rules (which this PR fully supports and doesn't conflict with), the soft-lockup detection itself still has to live in the monitor. The FQ CEL fix solves who can *act* on analyzer events, but the deeper issue is that the analyzer never has the right data to make this particular call. Two reasons:

1. **The analyzer works with health events; this evidence is raw kernel log lines.** A soft lockup shows up as a generic `BUG: soft lockup` header in one journal entry, and the stack frame that tells you it's mlx5 shows up as a completely separate entry. Neither can be shipped as a standalone health event — the header alone fires on every soft lockup in the kernel (most have nothing to do with NICs), and a bare stack frame line is meaningless on its own. The only way to feed this to HEA would be emitting two junk events per node and hoping a downstream join filters them.

2. **The join HEA would need can't be expressed in its time-window rules.** The actual signal is "an mlx5 stack frame within ~150 kernel log *lines* after this specific header, in order." That's about the structure of a kernel dump, not about wall-clock time. When the monitor restarts and replays the journal backlog, entries from completely different CPUs' lockup dumps land with nearly identical timestamps — a time-based window would happily mix them up. Line adjacency only exists where the lines are — in the monitor.

The way to think about the split: a soft lockup report is really one kernel diagnostic that happens to be spread across multiple journal entries. Putting those pieces back together is parsing — that's the monitor's job. Correlating multiple distinct reports over time is HEA's job, which is exactly what this PR uses it for (3+ TX/RX timeouts in an hour triggers the `RepeatedNICDriverError` rule). The monitor-side detector and the HEA escalation path are complementary, not alternatives.

We verified this live on a real cluster: with HEA deployed, fault-quarantine evaluated analyzer-derived events against all 6 rulesets — 0 matches, 0 cordons. Once the HEA-to-FQ CEL work lands, the `RepeatedNICDriverError` escalation this PR extends would automatically benefit from it.

**Why detect on first confirmation instead of waiting for repetition**

The first report of what turns into a multi-day wedge looks identical to a transient lockup at onset (~26 seconds of stuck time either way). There's no reliable way to distinguish "this will self-recover in a minute" from "this CPU is gone forever" at detection time.

We considered requiring N occurrences before firing, but:
- Kernels before 5.13 only warn **once** per lockup episode — a repetition gate would literally never fire on those kernels, silently making the detector useless on older fleets
- Even on newer kernels, the ~26s re-report interval means waiting for 3 occurrences adds over a minute of delay before quarantine starts

The safety net for false positives is the **rollout model**, not a weaker detector: every fleet starts with the pattern disabled, enables it in shadow mode to validate precision against their actual journal output, then graduates to active remediation. This is the same rollout path used for all NIC driver patterns.

**Why line-count windows instead of time-based windows**

In a soft lockup dump, the header and the attributing stack frame are typically 3-6 journal entries apart, but they're part of the same kernel printk burst. On a healthy monitor reading the journal in real time, they arrive within milliseconds of each other.

The problem is restarts. When the monitor starts up (or recovers from a crash), it replays the journal backlog. Hundreds of entries — potentially including interleaved dumps from different CPUs — land with effectively identical timestamps. A "within 5 seconds" window would happily pair a header from CPU 12's lockup with a stack frame from CPU 94's unrelated lockup. A 150-line window respects the actual structure of a kernel dump regardless of how fast the entries are consumed.

**Known trade-off: monitor restart during failed delivery**

If the monitor crashes while holding an undelivered confirmed match, that match is lost from memory. This is intentional — the kernel re-reports persistent lockups every ~26 seconds, so the next poll cycle after restart will re-detect and re-confirm. The cooldown timer is deliberately not persisted for the same reason: it's spam prevention, not correctness-critical state.

## Testing

**Unit tests** — 36 tests in the `nicdriver` package (17 new) plus regression tests in `main`. Coverage includes: header→frame confirmation (via both RIP and Call Trace paths), speculative-frame rejection, evidence-without-header rejection, misattribution guards (e1000 lockup presented to the mlx5 detector), exact window boundary (149th vs 150th line), re-arm after expiry, cooldown behavior (100 kernel re-reports → exactly 2 events), send-failure re-delivery, both old and new NETDEV WATCHDOG formats (with non-mlx5 negative case), and the kernel-origin filter wiring.

**Kind e2e tests** (Tilt cluster):

- `TestSyslogHealthMonitorNICDriverDetection` (extended): all 7 non-fatal patterns fire correctly as k8s Events — including the new post-6.8 watchdog format and the RX timeout — plus the full fatal path (condition → cordon) and recovery (uncordon + annotation cleanup).
- `TestSyslogHealthMonitorNAPISoftLockupDetection` (new): injects TX timeout followed by a realistic 6-line lockup dump; asserts exactly one fatal `mlx5_napi_soft_lockup` condition with parsed CPU and duration fields, the non-fatal TX corroboration event, cordon with quarantine annotation, and recovery.

**Manual end-to-end on real hardware** (internal DGX Cloud dev cluster: OCI bare-metal A100 nodes, kernel 5.15.0-oracle, CRI-O; branch image deployed via the umbrella chart; injection via `echo > /dev/kmsg` so entries carry `_TRANSPORT=kernel`):

- **Shadow mode (`STORE_ONLY`)**: Injected the full lockup sequence. The detector correlated header→frame on real journald, emitted the fatal with correct CPU and duration metadata. Platform-connector logged `Skipping non-remediation health event` (no condition, no cordon). Fault-quarantine never saw it (the changestream filter only passes `EXECUTE_REMEDIATION` events). Event was persisted in the store with `processingStrategy=STORE_ONLY`. Node stayed schedulable. The TX timeout corroboration event (running under the default strategy) flowed normally.
- **Active remediation (`EXECUTE_REMEDIATION`)**: Same injection → `SysLogsNICDriverError=True` condition → matched the Syslog fatal ruleset → node cordoned with the `quarantineHealthEvent` annotation carrying full event metadata. Detection to cordon completed within one 15-second poll cycle.
- **Recovery**: `kubectl uncordon` → fault-quarantine detected the manual uncordon and cleaned up its own annotations.
- **Analyzer escalation**: With health-events-analyzer deployed, three TX-timeout injections within the window fired `RepeatedNICDriverError` on the new error code → derived `CONTACT_SUPPORT` fatal + node condition. Fault-quarantine evaluated it against all 6 rulesets: **0 matches, 0 cordons** — live confirmation of the analyzer-agent gap described above. Once the HEA-to-FQ CEL work lands as part of the HEA prod rollout, this escalation path would automatically gain remediation power.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Expanded syslog NIC driver detection to recognize Mellanox mlx5 TX/RX timeout signals and correlate them with non-fatal evidence events.
  * Added stateful detection for mlx5 NAPI soft lockups, including CPU/duration details and quarantine/lockup remediation behavior.
  * Improved kernel-origin filtering so NIC driver checks apply the correct kernel-only log scope.

* **Documentation**
  * Updated syslog detection correlation docs, pattern examples, severity classifications, and escalation references for the new timeout and soft-lockup signals.

* **Tests**
  * Added/updated unit and end-to-end tests covering timeout patterns and mlx5 NAPI soft-lockup detection.


health-monitors/syslog-health-monitor/pkg/nicdriver/config_test.go
```diff
@@ -156,6 +156,68 @@ enabled = true
 	assert.False(t, patterns[0].HasProcessingStrategy)
 }
 
+func TestLoadConfig_NAPISoftLockupAndTimeoutPatterns(t *testing.T) {
```
XRFXLP, line 159: In the long run, would it be better to have table driven test?

health-monitors/syslog-health-monitor/pkg/nicdriver/softlockup.go
```diff
@@ -0,0 +1,151 @@
...
+		d.linesLeft = softLockupWindowLines
+
+		return nil
+	}
+
+	if !d.armed {
+		return nil
+	}
+
+	d.linesLeft--
```
XRFXLP, line 122: If a different event on this line fails to send, the whole line gets processed again on retry, so this counter drops twice for one log line. Should we make sure it only counts down once per line?

health-monitors/syslog-health-monitor/pkg/nicdriver/softlockup.go
```diff
@@ -0,0 +1,151 @@
...
+
+// observe consumes one kernel log line and returns a non-nil match when the
+// line confirms mlx5 attribution for a recently seen soft-lockup header.
+func (d *softLockupDetector) observe(message string) *softLockupMatch {
+	if d.pending != nil {
+		// The pipeline only moves to a different entry after the previous
+		// one was delivered successfully, so an identical message is a
+		// delivery retry (return the same match, bypassing the cooldown)
+		// and any other message means the match was delivered.
+		if message == d.pendingMessage {
```
XRFXLP, line 101: This retry logic checks if the log line text is the same to decide whether the event already got sent. That feels fragile. Could we track the journal entry we matched instead of comparing the text, so a repeated line doesn't accidentally fire the alert twice?

health-monitors/syslog-health-monitor/pkg/nicdriver/softlockup.go
KaivalyaMDabhadkar: you're right that if observe decrements linesLeft and then a single-line pattern match on the same line fails to send, the retry would decrement the window counter a second time. In practice the impact is negligible (the window is 150 lines and confirmation typically lands within 3-6, so losing 1 line of slack doesn't matter), and the scenario itself requires a send failure that persists through all retries. Adding a "last decremented message" guard would introduce complexity for a case that's functionally invisible

health-monitors/syslog-health-monitor/pkg/nicdriver/softlockup.go
KaivalyaMDabhadkar: The text comparison is intentional narrowly-scoped: it only handles send-failure retries, where the monitor replays the exact same journal entry (same cursor position, same message text). It's not used for general dedup as that's what the 30-minute emission cooldown handles (if the kernel re-reports the lockup as a new entry with the same text 26s later, the cooldown suppresses it regardless of the pending state).
Using a journal cursor instead would mean threading cursor IDs through the handler's ProcessLine API, which currently takes just a message string. That's a contract change across all check handlers for a guard that's already covered by the cooldown. The text check is the minimal mechanism for the specific retry-idempotency guarantee it provides.

health-monitors/syslog-health-monitor/pkg/nicdriver/config_test.go
KaivalyaMDabhadkar: Good point, the existing tests in this file all follow the standalone function pattern so this one is consistent with them, but agree it would be a good idea to write table-driven tests. Maybe we can have a follow-up PR to refactor the tests accordingly?

XRFXLP: yeah, we can do a follow up MR.

XRFXLP (approved):

PR #1533, opened 2026-07-23
fix(syslog-health-monitor): seek to boot start instead of tail after reboot

### LINKED ISSUE #1532: [Bug]: syslog-health-monitor misses XIDs emitted before startup after reboot
### Prerequisites

### Code of Conduct

### Bug Description

After a host reboot, `syslog-health-monitor` detects the boot ID change and clears its persisted journal cursors. When each check runs without a cursor, it seeks to the current journal tail and only processes entries added afterward.
Consequently, XIDs emitted during GPU Operator or driver startup—before `syslog-health-monitor` starts—are missed.
A new pod on the same boot can safely resume from its persisted cursor. A cursor from the previous boot, however, is not a reliable starting point because it may no longer be valid.

### Observed sequence
1. `2026-07-21 15:13:33.339`: XID 79 occurred.
2. `2026-07-21 15:13:33.434`: XID 154 occurred.
3. `2026-07-21 15:13:48.273`: NVIDIA driver DaemonSet became ready.
4. `2026-07-21 15:13:50.758`: `syslog-health-monitor` became ready.
5. Neither XID was reported by `syslog-health-monitor`.
### Expected behavior
After detecting a boot-ID change, `syslog-health-monitor` should process matching entries from the beginning of the current boot before switching to normal cursor-based processing.
### Actual behavior
The monitor clears its cursors and initializes them at the journal tail. Entries from the current boot that predate monitor startup are skipped.

### Component

Health Monitor

### Steps to Reproduce

1. Deploy NVSentinel with `syslog-health-monitor`.
2. Reboot a GPU node.
3. Cause or observe an XID during driver/GPU Operator startup, before `syslog-health-monitor` starts.
4. Wait for `syslog-health-monitor` to become ready.
5. Verify the XID exists in the host journal.
6. Verify no corresponding health event was emitted.

### Environment

- NVSentinel version: v1.12.0
- Kubernetes version:
- Deployment method:

### Logs/Output

XIDs present in the host journal:
```text
2026-07-21 15:13:33.339
NVRM: Xid (PCI:0000:38:00): 79, pid=70183, name=nv-hostengine, GPU has fallen off the bus.
2026-07-21 15:13:33.434
NVRM: Xid (PCI:0000:38:00): 154, GPU recovery action changed from 0x0 (None) to 0x1 (GPU Reset Required)
```
Earlier XIDs were successfully reported, confirming that XID parsing and event publication were operational:
```text
2026-07-21 14:56:26.630
NVRM: Xid (PCI:0000:38:00): 79, pid=64064, name=nv-hostengine, GPU has fallen off the bus.
2026-07-21 14:56:26.878
NVRM: Xid (PCI:0000:38:00): 154, GPU recovery action changed from 0x0 (None) to 0x1 (GPU Reset Required)
```

signalgap9-del:
## Summary

Fixes #1532

After a host reboot, `syslog-health-monitor` detects the boot-ID change and clears persisted cursors. On the next poll, the absence of a cursor triggers `initializeJournalFromTail`, which seeks to the journal's end -- permanently skipping XID entries emitted between boot and monitor startup.

## Root Cause

`handleBootIDChange` clears cursors, then `processJournalEntries` sees `hasLastCursor=false` and calls `initializeJournalFromTail` which does `SeekTail()`. Entries before the tail are never read.

## Fix

Add a `postRebootInit` flag set during boot-ID change handling. When there is no cursor and this flag is active, seek to the **beginning** of the current boot's journal via `SeekHead()` with an explicit `_BOOT_ID=<current>` match filter (applied by `configureBootFilter`), then process all entries forward. After all checks complete successfully in `Run()`, clear the flag so subsequent polls resume normal cursor-based processing.

## Changes

- Add `SeekHead()` to the `Journal` interface + all implementations (real sdjournal, stub, fake, mock)
- Add `initializeJournalFromBootStart` for post-reboot initialization
- Track `postRebootInit` state in `SyslogMonitor`
- Clear flag after all checks succeed in `Run()`

## Testing

- `TestBootIDChange_ProcessesEntriesFromBootStart`: verifies entries from the beginning of the new boot are processed after reboot
- `TestBootIDChange_SecondPollResumesFromCursor`: verifies normal cursor resume after the initial post-reboot scan
- `TestNoBootChange_StillSeeksTail`: verifies first-install behavior is unchanged
- All existing tests pass (280 tests; 7 pre-existing failures on Windows due to `/proc` unavailability -- CI runs on Linux)
- `golangci-lint run ./...`: 0 issues

## Notes

- Unit tests cover the cursor initialization logic using the existing mock journal infrastructure
- E2E validation (GPU node reboot with XID during driver startup) requires hardware not available in this environment -- deferring to CI/UAT
- The boot filter (`_BOOT_ID=<current>`) ensures only current-boot entries are read even though `SeekHead` positions at the journal's absolute beginning

## Summary by CodeRabbit

* **New Features**
  * Added a `--boot-lookback-window` setting for controlling post-reboot journal scan duration, defaulting to two hours.
  * Supports unlimited lookback when configured to zero.
* **Bug Fixes**
  * Improved post-reboot monitoring to capture entries from the current boot, including logs generated before monitor startup.
  * Preserved cursor-based processing during subsequent polls and improved recovery after interrupted scans.
* **Tests**
  * Added coverage for lookback configuration, reboot handling, cursor resumption, and recovery scenarios.


signalgap9-del:
CI failure on 299c313 was caused by postRebootInit being set unconditionally in handleBootIDChange — on Linux, first-install (no state file) has oldBootID="" which differs from the real boot UUID, incorrectly triggering the SeekHead path.
Fixed in 7258a10f: guard with oldBootID != "" so only actual reboots (previous state exists) take the boot-start scan. First-install behavior is unchanged (SeekTail). Full test suite verified on Linux.

sorry for mistake

signalgap9-del:
## Reproduction & Fix Verification (local, no GPU required)

Reproduced the #1532 scenario on WSL2 Ubuntu (systemd + journalctl) and verified the fix.

### Setup

- **OS**: WSL2 Ubuntu 24.04, systemd running, persistent journal at `/var/log/journal/`
- **Build**: `go build -tags systemd` (real sdjournal, not stub)
- **Scenario**: State file contains a **previous boot ID** (`00000000-...`) with a stale cursor, simulating a host reboot. XID entries exist in the journal **before** monitor startup.
- **gRPC**: Minimal `PlatformConnector` server accepting `HealthEventOccurredV1` calls via UDS.

### Key log output (fixed version (this PR))

```
"Detected bootID change" old=00000000-0000-0000-0000-000000000000 new=0eddef43-3a50-4472-b158-3b19525e3560
"Published healthy event after system reboot" check=SysLogsXIDError
"Cleared all cursors due to system reboot"

"Post-reboot: seeking to boot start to process pre-startup entries" check=SysLogsXIDError
"Applying boot filter" filter=_BOOT_ID=0eddef433a504472b1583b19525e3560

"Finished processing journal entries" nextCursor=s=7532b4ed...;i=2dae0;b=0eddef43...
"Post-reboot boot-start scan completed; resuming cursor-based processing"

"Resuming from last known cursor" cursor=s=7532b4ed...;i=2dae0;b=0eddef43...
"No new entries since last cursor"
```

### State file before/after

```json
// BEFORE (simulates previous boot)
{"version":1,"boot_id":"00000000-0000-0000-0000-000000000000","check_last_cursors":{"SysLogsXIDError":"s=abc123fake;i=12345"}}

// AFTER (current boot, real journal cursor)
{"version":1,"boot_id":"0eddef43-3a50-4472-b158-3b19525e3560","check_last_cursors":{"SysLogsXIDError":"s=7532b4edc8e94abebf042339cd3487fd;i=2dae0;b=0eddef433a504472b1583b19525e3560;m=b7b5450c;t=6574d836d8701;x=b74c72eb3ef87639"}}
```

### What this proves

1. **Boot ID change detected** -> cursors cleared -> healthy event published (recovery signal)
2. **`postRebootInit` flag triggers `initializeJournalFromBootStart`** -> `SeekHead` + boot filter -> processes all entries from boot start (the fix)
3. **After first scan completes**, flag clears -> subsequent polls resume normal cursor-based processing (`"Resuming from last known cursor"`)
4. **No duplicate processing** -- second poll reports `"No new entries since last cursor"`

Without this fix, the monitor would call `initializeJournalFromTail` -> `SeekTail`, permanently skipping XIDs emitted between boot and monitor startup (the bug described in #1532).

### Note

XID fault events were not received via gRPC in this test because `logger -t kernel` writes with `_TRANSPORT=syslog`, not `_TRANSPORT=kernel` (the monitor's journal filter). This is a test-environment limitation -- the seek behavior itself is fully verified by the log trace above.

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor.go
```diff
@@ -1067,6 +1088,44 @@ func (sm *SyslogMonitor) initializeJournalFromTail(journal Journal, check CheckD
...
+// initializeJournalFromBootStart seeks to the beginning of the current boot's
+// journal and processes all entries forward. Used after a boot-ID change when
+// persisted cursors have been cleared: unlike initializeJournalFromTail (which
+// skips everything before the tail), this ensures entries emitted between boot
+// and monitor startup are not missed.
+//
+// A boot filter (_BOOT_ID=<current>) is applied explicitly so that only entries
+// from the current boot are read, even though SeekHead positions at the very
+// beginning of the journal.
+func (sm *SyslogMonitor) initializeJournalFromBootStart(journal Journal, check CheckDefinition) error {
```
XRFXLP, line 1126: Can we have some sort of time window for lookback? That would avoid processing ancient XIDs in syslog if syslog health monitor lands on a new node and would avoid us doing duplicated reset/reboots in case manual reset (by definition we're safe with reboot) might have been performed?

signalgap9-del: thanks for review! addressed in 22563ab8.
Instead of SeekHead() (absolute journal beginning), the post-reboot scan now seeks to now - 30m via SeekRealtimeUsec. This means:
XIDs emitted between boot and monitor startup are still captured (the typical gap is seconds to a few minutes)
Ancient XIDs from earlier in the boot that may have already been manually remediated are skipped
The window is defined as defaultBootLookbackWindow = 30 * time.Minute (easy to adjust or make configurable later)
Also added SeekRealtimeUsec to the Journal interface + all implementations (real sdjournal, stub, fake, mock).
Note: postRebootInit is only set when oldBootID != "" and differs from the current boot ID, so pod restarts on the same boot (including rescheduling) do not trigger this path at all.

signalgap9-del: is it good to go?

XRFXLP: Can we make that value configurable? Looks like it is always the default value.

XRFXLP (changes requested):

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor.go
signalgap9-del: Made it configurable via --boot-lookback-window CLI flag (default: 30m). The value is parsed at startup and passed through the constructor chain into SyslogMonitor.bootLookbackWindow. Operators can tune it per deployment (e.g. --boot-lookback-window=1h for environments with longer driver startup times).

signalgap9-del: Made it configurable end-to-end:
--boot-lookback-window CLI flag (default: 30m)
bootLookbackWindow in Helm values.yaml

Wired to the DaemonSet args in _helpers.tpl

Operators can tune per deployment:
helm upgrade --set syslog-health-monitor.bootLookbackWindow=1h

The value is parsed at startup and passed through the constructor chain into SyslogMonitor.bootLookbackWindow.

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor.go
```diff
@@ -1067,6 +1092,60 @@ func (sm *SyslogMonitor) initializeJournalFromTail(journal Journal, check CheckD
...
+	lookback := sm.bootLookbackWindow
+	seekTarget := time.Now().Add(-lookback)
+
+	slog.Info("Post-reboot: seeking to boot start with lookback window",
+		"check", check.Name,
+		"lookbackWindow", lookback,
+		"seekTarget", seekTarget.Format(time.RFC3339))
+
+	if err := sm.configureBootFilter(journal, check.Name); err != nil {
```
XRFXLP: Given that the current filter is 
```
(_TRANSPORT=kernel) OR (SYSLOG_IDENTIFIER=nvsentinel-gpu-reset)
```
Won't this make the filter effectively as 
```
(_TRANSPORT=kernel) OR (SYSLOG_IDENTIFIER=nvsentinel-gpu-reset) AND _BOOT_ID=<current>`
```
thereby:
```
(_TRANSPORT=kernel) OR (SYSLOG_IDENTIFIER=nvsentinel-gpu-reset AND _BOOT_ID=<current>)`  
```

Which would make syslog HM process XIDs before reboot?

distros/kubernetes/nvsentinel/charts/syslog-health-monitor/templates/_helpers.tpl
```diff
@@ -133,6 +133,8 @@ spec:
             - "/etc/syslog-health-monitor/nic-driver.toml"
             - "--sysfs-root"
             - "/nvsentinel/sys"
+            - "--boot-lookback-window"
+            - "{{ $root.Values.bootLookbackWindow }}"
```
XRFXLP: Can we use `{{ $root.Values.bootLookbackWindow | default "2h" | quote }}` so it is gracefully backward compatible?

distros/kubernetes/nvsentinel/charts/syslog-health-monitor/values.yaml
```diff
@@ -49,6 +49,11 @@ tolerations: []
 # Customize this if your journal is stored in a non-standard location
 journalHostPath: /var/log
 
+# How far back to scan the journal after a reboot (e.g. 30m, 1h).
+# Entries older than this window are skipped to avoid re-processing
+# ancient XIDs that may have already been manually remediated.
+bootLookbackWindow: "30m"
```
XRFXLP: Can we make this 2h as default given that in some environment it can take ~1 hour for reboot?

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor_test.go
```diff
@@ -703,7 +731,7 @@ func TestBootIDChange_StateNotPersistedWhenSendSkipped(t *testing.T) {
 		"pendingPostRebootBootID must hold the current bootID for retry")
 
 	// Run() with the socket still missing must NOT persist the new
-	// BootID. Previously this was a regression: executeCheck →
+	// BootID. Previously this was a regression: executeCheck ??
```
XRFXLP: This looks like a typo, can we revert this?

health-monitors/syslog-health-monitor/pkg/syslog-monitor/journal_real.go
```diff
@@ -69,6 +69,16 @@ func (j *RealJournal) SeekCursor(cursor string) error {
 	return j.journal.SeekCursor(cursor)
 }
 
+// SeekHead seeks to the beginning of the journal
+func (j *RealJournal) SeekHead() error {
+	return j.journal.SeekHead()
+}
+
+// SeekRealtimeUsec seeks to the entry closest to the given realtime timestamp
+func (j *RealJournal) SeekRealtimeUsec(usec uint64) error {
```
XRFXLP, line 78: Can we change the meaning so `0` means there we go upto the boot without clipping anything i.e. making the loopback unlimited? So we have configuration to not miss any XIDs at all.

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor.go
```diff
@@ -1067,6 +1092,60 @@ func (sm *SyslogMonitor) initializeJournalFromTail(journal Journal, check CheckD
...
+// and monitor startup are not missed.
+//
+// A boot filter (_BOOT_ID=<current>) is applied explicitly so that only entries
+// from the current boot are read, even though SeekHead positions at the very
+// beginning of the journal.
+func (sm *SyslogMonitor) initializeJournalFromBootStart(journal Journal, check CheckDefinition) error {
+	lookback := sm.bootLookbackWindow
+	seekTarget := time.Now().Add(-lookback)
+
+	slog.Info("Post-reboot: seeking to boot start with lookback window",
```
XRFXLP: Can we also log when window actually binds?

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor.go
```diff
@@ -432,6 +444,15 @@ func (sm *SyslogMonitor) handleBootIDChange(oldBootID, newBootID string) error {
 		delete(sm.checkLastCursors, checkName)
 	}
 
+	// Signal that the next journal scan must start from the beginning of
+	// the current boot rather than the tail, so entries emitted between
+	// boot and monitor startup are not missed. Only set when there was a
+	// previous boot (oldBootID != ""); on first install (no state file)
+	// the normal tail-initialization path is correct.
+	if oldBootID != "" {
+		sm.postRebootInit = true
```
XRFXLP, line 473: `postRebootInit` is in-memory only, so a restart loses it. The checkpoint is committed before the work it guards runs.

`tryFlushPostRebootBootIDClear` (507-515) persists BootID: boot-2 before the boot-start scan runs. If the process restarts between that write and the end of the first Run(), `loadState` sees boot-2 == kernel boot ID, `handleBootIDChange` returns at 432, `postRebootInit` is false, cursors are empty → `SeekTail()`. Pre-fix bug, silently.

Not a narrow window — the write is followed by a full journal scan and handler dispatch, and restarts are likely right after a reboot.

Fix: persist it. Add BootStartScanDone bool to syslogMonitorState and derive postRebootInit from disk. migrateStateVersion already handles the field addition.

Same issue at 274-284: a first-cycle check failure keeps the flag set in memory, but not across a restart.

XRFXLP (changes requested):

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor.go
signalgap9-del: Fixed in c6cbd07b — removed configureBootFilter from initializeJournalFromBootStart entirely. Boot filtering is now done at the application level: each entry's _BOOT_ID field is checked against sm.currentBootID before processing (isStaleBootEntry / skipStaleBootEntry). Stale-boot entries are skipped with cursor advancement.
This avoids the AND/OR precedence issue with the existing tag disjunctions and doesn't touch configureTagFilters (shared with normal polling). The SeekRealtimeUsec lookback already bounds how far back we read, so the application-level filter only needs to skip entries within that window belonging to a previous boot.

distros/kubernetes/nvsentinel/charts/syslog-health-monitor/templates/_helpers.tpl
signalgap9-del: Done in c6cbd07b — changed to {{ $root.Values.bootLookbackWindow | default "2h" | quote }}.

distros/kubernetes/nvsentinel/charts/syslog-health-monitor/values.yaml
signalgap9-del: Done in c6cbd07b — default changed to 2h in both values.yaml and the --boot-lookback-window CLI flag.

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor_test.go
signalgap9-del: Fixed in c6cbd07b — reverted ?? back to →

health-monitors/syslog-health-monitor/pkg/syslog-monitor/journal_real.go
signalgap9-del: Done in c6cbd07b — bootLookbackWindow=0 now means unlimited: the scan uses SeekHead() (absolute journal head) with no time clipping. The application-level boot filter still prevents processing entries from previous boots.
helm upgrade --set syslog-health-monitor.bootLookbackWindow=0

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor.go
signalgap9-del: Done in c6cbd07b — when lookback > 0:

"Post-reboot: lookback window active, entries before seekTarget will be skipped"
When lookback == 0:
"Post-reboot: unlimited lookback, scanning from journal head"

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor.go
signalgap9-del: Fixed across c6cbd07b, 3f2daab1, and 6b81e5fe.

Added BootStartScanDone bool to syslogMonitorState (persisted as boot_start_scan_done).
tryFlushPostRebootBootIDClear persists BootStartScanDone: false alongside the new BootID (checkpoint before scan)
saveCurrentState writes BootStartScanDone: !sm.postRebootInit on each check completion
Constructor recovery: if state.BootID == currentBootID && !state.BootStartScanDone, sets postRebootInit = true from disk
Run() calls saveCurrentState() after clearing postRebootInit to persist the A→C transition (BootStartScanDone: true)
Also removed the len(cursors)==0 guard from the recovery path — it failed on partial-crash with multiple checks (check 1 cursor saved, check 2 crashed mid-scan → guard saw non-empty cursors → skipped recovery → check 2 would SeekTail).
Crash windows verified:
Crash before flush → boot change re-detected → scan re-runs
Crash after flush, before/during scan → Done: false on disk → recovery triggers
Crash after all checks, before final persist → cursors exist → normal resume → final persist writes Done: true
After final persist → Done: true → no recovery
Existing state files without the field get Go's zero value (false), which safely triggers a re-scan. thank you for review!

health-monitors/syslog-health-monitor/main.go
```diff
@@ -174,6 +177,17 @@ func run() error {
 	return g.Wait()
 }
 
+// mustParseDuration parses a duration string and exits on failure.
+func mustParseDuration(s string) time.Duration {
```
XRFXLP, line 196: should it reject negative values?

XRFXLP (changes requested):
We're missing unit tests for most of the newly added function, can we add thorough unit tests for them?

health-monitors/syslog-health-monitor/main.go
signalgap9-del: Added in 2fd2652e and 45812eed. Coverage for all newly added functions:
parseBootLookbackWindow: valid durations, zero (unlimited), negative (rejected), invalid format (8 cases)
isStaleBootEntry: same/different boot, empty entry boot ID, empty current boot ID, GetData error (5 cases)
skipStaleBootEntry: normal advance with cursor update, journal exhausted (2 cases)
initializeJournalFromBootStart: unlimited → SeekHead, windowed → SeekRealtimeUsec, no entries → tail fallback (3 cases)
BootStartScanDone persistence: persisted as true after scan completion, crash recovery restores postRebootInit from disk, normal restart does not re-trigger (3 cases)
All tests verified on Linux. The CrashRecovery test uses a 2-phase pattern (discover actual kernel boot ID first, then write state) to ensure the disk-recovery path is exercised directly rather than being silently skipped.        

Yes — a negative duration would invert the seek direction: time.Now().Add(-(-1h)) = time.Now().Add(1h), which seeks into the future rather than the past. The journal would return zero entries (nothing exists ahead of now), and the scan would silently fall back to initializeJournalFromTail, reproducing the exact bug this PR fixes.

Done: extracted parseBootLookbackWindow(s string) (time.Duration, error) from mustParseDuration for testability. Negative values are rejected at startup with an explicit error. Zero remains valid (means unlimited lookback via SeekHead).

signalgap9-del: sorry for lacking test code. I was focused on the implementation logic too much

distros/kubernetes/nvsentinel/charts/syslog-health-monitor/templates/_helpers.tpl
```diff
@@ -133,6 +133,8 @@ spec:
             - "/etc/syslog-health-monitor/nic-driver.toml"
             - "--sysfs-root"
             - "/nvsentinel/sys"
+            - "--boot-lookback-window"
+            - {{ $root.Values.bootLookbackWindow | default "2h" | quote }}
```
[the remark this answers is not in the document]
signalgap9-del: Fixed  in ecf92d89:
Replaced default "2h" with a hasKey guard so that bootLookbackWindow: 0 (unquoted integer, meaning unlimited) is rendered as "0" instead of being silently replaced by "2h".

health-monitors/syslog-health-monitor/main_test.go
```diff
@@ -131,8 +132,48 @@ func TestApplyKataConfig_OverridesKernelFilterWithContainerdUnit(t *testing.T) {
 
 		for _, tag := range tags {
 			if tag == "-k" {
-				t.Errorf("check %q in Kata mode must not carry -k (would AND with the unit filter)", name)
+			t.Errorf("check %q in Kata mode must not carry -k (would AND with the unit filter)", name)
 			}
 		}
 	}
```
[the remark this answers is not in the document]
signalgap9-del: Fixed in ecf92d89
Fixed gofmt indentation in main_test.go.

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor_test.go
```diff
@@ -936,3 +1191,345 @@ func TestResumeFromLastCursor_SkipsBookmarkReReadWithFilteredJournal(t *testing.
...
+		"", "",
+		nil,
+		"tcp://test",
+		30*time.Minute,
+	)
+	assert.NoError(t, err)
+
+	// Normal restart: postRebootInit must remain false.
+	assert.False(t, sm.postRebootInit, "postRebootInit must be false when BootStartScanDone is true")
+}
```
XRFXLP, line 1571: `isStaleBootEntry` and `skipStaleBootEntry` are well covered as units, but nothing runs a mixed journal through `processBootFilteredEntries`. These entries have no Fields, so `GetData(FieldBootID)` returns "field not found", `isStaleBootEntry` fails open, and the filter never fires here — this test passes whether or not it's wired into the loop correctly.

Can we add Fields: `map[string]string{FieldBootID: "boot-2"}` to these entries, prepend a boot-1 entry, and assert only the boot-2 messages were handled? That covers the seam.

XRFXLP (changes requested):

health-monitors/syslog-health-monitor/pkg/syslog-monitor/syslogmonitor_test.go
signalgap9-del: Good point — the entries had no Fields, so GetData(FieldBootID) returned "field not found" and isStaleBootEntry failed open without the filter ever firing. 

Fixed in d149277f:

Added Fields: map[string]string{FieldBootID: ...} to all entries

Prepended a stale boot-1 entry before the two boot-2 entries

Registered a recordingHandler that captures messages passed to ProcessLine
Asserts: only 2 messages reach the handler (both boot-2), the stale boot-1 entry never reaches it, and the final cursor is at cursor-2

This exercises the full seam: processBootFilteredEntries → isStaleBootEntry → skipStaleBootEntry → handler dispatch.

XRFXLP (approved):

PR #1546, opened 2026-07-28
feat(nic-health-monitor): detect missing InfiniBand character devices

shaq918:
## Summary

Adds `InfiniBandCharDeviceCheck`, a passive state check for the NIC health
monitor that flags an InfiniBand device whose `/dev/infiniband` character
nodes (issm/umad/uverbs) are missing while its ports still read
`ACTIVE`/`LinkUp` — a failure the existing port-state checks report as
**healthy**, yet RDMA workloads cannot start on the node (pods fail with
`lstat /dev/infiniband/issm9: no such file or directory`).

### Motivation

A GPU node came up with only 8 of ~12 expected `/dev/infiniband/issm*` nodes.
The IB ports were `ACTIVE`/`LinkUp` throughout, so no existing check fired; it
required manual triage and a spare migration. This is distinct from device
disappearance (the device is present in `/sys/class/infiniband`) and from
verbs-level resource exhaustion — it is a missing character-device node.

### How it works

udev creates `/dev/infiniband/{issm,umad,uverbs}N` from the sysfs class entries
under `/sys/class/infiniband_mad` and `/sys/class/infiniband_verbs`, so a
missing class entry is an exact proxy for a missing `/dev` node. Those
directories are already inside the host `/sys` the DaemonSet mounts, so **no
`/dev` mount is required**.

The check is a per-device internal-consistency test, not an absolute
expected-count test (to avoid the false-positive class that motivated #1361 /
#1379 / #1462): for each discovered, in-scope device that exposes at least one
InfiniBand-mode port it expects one `uverbs` per device and one `umad` and one
`issm` per InfiniBand-mode port. `issm`/`umad` are gated on the InfiniBand link
layer because RoCE/Ethernet-mode ports legitimately have no `issm`. A device
entirely absent from `/sys/class/infiniband` stays the `InfiniBandStateCheck`'s
device-disappearance responsibility.

Missing nodes are fatal (`REPLACE_VM`); a healthy recovery event is emitted when
they reappear. An absent class directory is treated as an uncertain observation
and held rather than reported as mass failure. Enabled by default in the Helm
`enabledChecks`.

## Type of Change
- [x] ✨ New feature

## Component(s) Affected
- [x] Health Monitors

## Testing
- [x] Tests pass locally (`go test ./...`, race) and `golangci-lint` clean
- [x] Unit tests cover: issm/umad/uverbs missing (fatal), RoCE port expects no
      issm, mixed IB/Ethernet device, transition→recovery→quiet, `abi_version`
      ignored, absent class dir (uncertain/held), incomplete-discovery bail,
      reboot baseline clear+reassert, VF skipped, unsupported vendor excluded
- [x] Manual testing: validated the sysfs layout on a live H100-IB node
      (`mlx5_0` Ethernet mgmt NIC correctly has no issm; `mlx5_1..8` IB each
      have issm/umad/uverbs with `ibdev`/`port` files), plus a filesystem
      integration test that mirrors that node and reproduces the missing-issm
      fault against the production `sysfs.Reader`
- [x] No breaking changes

## Checklist
- [x] Self-review completed
- [x] Documentation updated (Helm values comment)
- [x] Ready for review

## Notes for reviewers
- The missing-set transition latch is in-memory (a reboot re-baselines; a
  non-reboot pod restart re-discovers current state on the first poll). If you
  prefer restart-durable recovery semantics, I can promote it into
  `statefile.Manager` mirroring the disappeared-port latch.
- `umad`/`issm` are gated to InfiniBand-mode ports; `umad` on Ethernet-mode
  ports of mixed cards is intentionally not required (conservative — avoids any
  RoCE-firmware-variant false positive). Happy to broaden if desired.

## Summary by CodeRabbit

- **New Features**
  - Added `InfiniBandCharDeviceCheck` to detect missing InfiniBand character-device nodes (`issm`, `umad`, `uverbs`) with debounced FATAL alerts and recovery events.
  - Updated the default enabled check set and configuration examples to include `InfiniBandCharDeviceCheck`.
  - Extended monitoring state to persist missing-node latches across restarts and handle baseline/boot changes.
- **Bug Fixes**
  - Prevented false `issm` alerts on mixed Ethernet/InfiniBand systems and improved handling of uncertain discovery.
- **Tests**
  - Added regression coverage for realistic sysfs layouts and latch behavior across scope changes.
- **Documentation**
  - Documented the new `InfiniBandCharDeviceCheck` and its alerting behavior.

### DIFF
--- a/distros/kubernetes/nvsentinel/charts/nic-health-monitor/values.yaml
+++ b/distros/kubernetes/nvsentinel/charts/nic-health-monitor/values.yaml
@@ -32,12 +32,16 @@ resources:
 #   - InfiniBandStateCheck: Monitors IB port state (UP/DOWN), physical state, device disappearance,
 #       card homogeneity anomalies; auto-excludes management NICs via NUMA topology
 #   - InfiniBandDegradationCheck: Monitors IB counter thresholds (fatal and non-fatal)
+#   - InfiniBandCharDeviceCheck: Detects missing IB character-device nodes (issm/umad/uverbs)
+#       via /sys/class/infiniband_mad and /sys/class/infiniband_verbs — a device whose ports
+#       read ACTIVE/LinkUp but whose /dev/infiniband/* nodes are missing (pods fail to start)
 #   - EthernetStateCheck: Monitors Ethernet/RoCE interface state (operstate), device disappearance,
 #       card homogeneity anomalies; auto-excludes management NICs via NUMA topology
 #   - EthernetDegradationCheck: Monitors Ethernet/RoCE counter thresholds
 enabledChecks:
   - InfiniBandStateCheck
   - InfiniBandDegradationCheck
+  - InfiniBandCharDeviceCheck
   - EthernetStateCheck
   - EthernetDegradationCheck
 
--- a/health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
+++ b/health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
@@ -50,6 +50,15 @@ const (
 	// noPort is the sentinel port used for device-level entries (uverbs),
 	// which are not associated with a specific port.
 	noPort = 0
+
+	// charDevMissThreshold is the number of consecutive polls a character
+	// device must be expected-but-missing before a FATAL is emitted. The
+	// device list and the mad/verbs class directories are read at slightly
+	// different instants, so a driver teardown/reload landing between the
+	// reads can make every node of a device look missing for one poll;
+	// requiring consecutive misses removes that race, mirroring the device
+	// disappearance debounce (deviceMissThreshold).
+	charDevMissThreshold = 3
 )
 
 // charDevKey uniquely identifies an expected or missing character device.
@@ -59,6 +68,21 @@ type charDevKey struct {
 	kind   charDevKind
 	device string
 	port   int
+}
+
+// flagKey renders the persisted statefile key for a charDevKey.
+func (k charDevKey) flagKey() string {
+	return fmt.Sprintf("%s/%s_%d", k.kind, k.device, k.port)
+}
+
+// flagOf converts a charDevKey to its persisted representation.
+func (k charDevKey) flagOf() statefile.MissingCharDeviceFlag {
+	return statefile.MissingCharDeviceFlag{Kind: string(k.kind), Device: k.device, Port: k.port}
+}
+
+// keyOfFlag converts a persisted flag back to a charDevKey.
+func keyOfFlag(f statefile.MissingCharDeviceFlag) charDevKey {
+	return charDevKey{kind: charDevKind(f.Kind), device: f.Device, port: f.Port}
 }
 
 // InfiniBandCharDeviceCheck detects missing InfiniBand character-device
@@ -81,6 +105,14 @@ type charDevKey struct {
 // device that is entirely absent from /sys/class/infiniband is out of
 // scope here — that is the InfiniBandStateCheck's device-disappearance
 // responsibility.
+//
+// Fault handling is latched and debounced: a node must be missing for
+// charDevMissThreshold consecutive polls before the FATAL fires, the
+// resulting latch persists across pod restarts and reboots via
+// pkg/statefile, and it is released only on a positive observation of the
+// node (recovery) or by the baseline clear. A latched key whose device
+// drops out of discovery is HELD, not recovered: absence of the device is
+// not evidence the character device healed.
 type InfiniBandCharDeviceCheck struct {
 	nodeName           string
 	reader             sysfs.Reader
@@ -95,19 +127,38 @@ type InfiniBandCharDeviceCheck struct {
 	// are wiped before current-boot faults are re-asserted.
 	emitHealthyBaselines bool
 
-	// previousMissing is the committed set of currently-missing character
-	// devices; it drives transition (missing↔present) event emission so
-	// steady states do not re-emit every poll. It is in-memory only: a
-	// reboot re-baselines, and a non-reboot pod restart re-discovers the
-	// current state on the first poll.
-	previousMissing map[charDevKey]bool
+	// latched is the committed set of character devices with a
+	// missing-node FATAL outstanding downstream. Seeded from the
+	// persisted statefile latch at construction so a recovery that
+	// happens while the pod is down is still emitted by the next pod.
+	latched map[charDevKey]bool
+
+	// missStreak counts consecutive polls each expected key has been
+	// missing; a FATAL fires when it reaches charDevMissThreshold.
+	// In-memory only: losing it to a restart merely restarts the
+	// debounce, which is the safe direction.
+	missStreak map[charDevKey]int
+
+	// seeded is true once one complete enumeration has succeeded, making
+	// a later incomplete enumeration an error rather than a quiet skip.
+	seeded bool
+
+	// uncertainWarned suppresses the held-poll warning after its first
+	// emission so an IB node without the ib_umad module does not log at
+	// every poll; it resets when a certain observation succeeds.
+	uncertainWarned bool
+
+	// saveFailed keeps a failed statefile Save retrying on subsequent
+	// commits even when nothing else changed.
+	saveFailed bool
 
 	pending *charDevPollCommit
 }
 
 // charDevPollCommit stages a prepared poll until Commit makes it durable.
 type charDevPollCommit struct {
-	missing     map[charDevKey]bool
+	latched     map[charDevKey]bool
+	missStreak  map[charDevKey]int
 	baselineRan bool
 }
 
@@ -131,6 +182,11 @@ func NewInfiniBandCharDeviceCheck(
 		stateManager.SetPendingBaseline(checks.InfiniBandCharDeviceCheckName)
 	}
 
+	latched := make(map[charDevKey]bool)
+	for _, flag := range stateManager.MissingCharDevices() {
+		latched[keyOfFlag(flag)] = true
+	}
+
 	return &InfiniBandCharDeviceCheck{
 		nodeName:             nodeName,
 		reader:               reader,
@@ -139,6 +195,8 @@ func NewInfiniBandCharDeviceCheck(
 		processingStrategy:   processingStrategy,
 		state:                stateManager,
 		emitHealthyBaselines: pendingBaseline,
+		latched:              latched,
+		missStreak:           make(map[charDevKey]int),
 	}
 }
 
@@ -160,7 +218,7 @@ func (c *InfiniBandCharDeviceCheck) Run() ([]*pb.HealthEvent, error) {
 }
 
 // Prepare observes one poll and stages its candidate state without
-// advancing the committed missing-set or persistent state.
+// advancing the committed latch or persistent state.
 func (c *InfiniBandCharDeviceCheck) Prepare() ([]*pb.HealthEvent, error) {
 	c.Discard()
 
@@ -172,7 +230,7 @@ func (c *InfiniBandCharDeviceCheck) Prepare() ([]*pb.HealthEvent, error) {
 	}
 
 	if !result.Complete {
-		if c.previousMissing != nil {
+		if c.seeded {
 			return nil, fmt.Errorf("device discovery incomplete: InfiniBand sysfs tree unavailable")
 		}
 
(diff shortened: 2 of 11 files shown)


distros/kubernetes/nvsentinel/charts/nic-health-monitor/values.yaml
```diff
@@ -32,12 +32,16 @@ resources:
...
+#   - InfiniBandCharDeviceCheck: Detects missing IB character-device nodes (issm/umad/uverbs)
+#       via /sys/class/infiniband_mad and /sys/class/infiniband_verbs — a device whose ports
+#       read ACTIVE/LinkUp but whose /dev/infiniband/* nodes are missing (pods fail to start)
 #   - EthernetStateCheck: Monitors Ethernet/RoCE interface state (operstate), device disappearance,
 #       card homogeneity anomalies; auto-excludes management NICs via NUMA topology
 #   - EthernetDegradationCheck: Monitors Ethernet/RoCE counter thresholds
 enabledChecks:
   - InfiniBandStateCheck
   - InfiniBandDegradationCheck
+  - InfiniBandCharDeviceCheck
```
KaivalyaMDabhadkar, line 44: could you also add the new check to docs/configuration/nic-health-monitor.md? That page lists enabledChecks and has a short section per check, so it will be out of date after this merges

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
```diff
@@ -0,0 +1,543 @@
...
+	// from a previous pod) so stale FATAL conditions from the prior boot
+	// are wiped before current-boot faults are re-asserted.
+	emitHealthyBaselines bool
+
+	// previousMissing is the committed set of currently-missing character
+	// devices; it drives transition (missing↔present) event emission so
+	// steady states do not re-emit every poll. It is in-memory only: a
+	// reboot re-baselines, and a non-reboot pod restart re-discovers the
+	// current state on the first poll.
+	previousMissing map[charDevKey]bool
```
KaivalyaMDabhadkar: You already called this out in the PR notes, and I would take you up on the offer to move this latch into statefile.Manager. Since node conditions have no TTL, if the char device comes back while the monitor pod happens to be down, the new pod starts with an empty latch and never sends the recovery, so the REPLACE_VM condition stays on the node until the next reboot. The disappeared-port latch already solves this exact problem for the state check, so the pattern is there to reuse. One heads-up: today this gap is partly hidden by the baseline re-running on every pod restart, so if that gets fixed on its own, this one becomes more visible. Probably best to fix the two together.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+
+	expected := buildExpectedCharDevices(result.Devices, c.classifier)
+	metrics.DevicesDiscovered.WithLabelValues(c.nodeName, c.Name()).Set(float64(expected.deviceCount))
+
+	// The baseline reconciliation waits for a complete enumeration with no
+	// unreadable devices: clearing while a device is unreadable would wipe
+	// that device's previous-boot condition with nothing able to re-assert
+	// it. Until then the baseline stays owed.
+	baselineRun := c.emitHealthyBaselines && len(result.UnreadableDevices) == 0
```
KaivalyaMDabhadkar, line 249: Really nice work on this check overall, the per-device consistency idea is an improvement over absolute counts. I did find one issue here that I think needs a fix:

When a device becomes unreadable or drops out of /sys/class/infiniband for a moment (a firmware reset is the typical case), discovery removes it from result.Devices. That means all of its keys silently leave the expected set, so currentMissing no longer contains them, and the recovery loop in buildEvents then sends a healthy "issm is present again" event for a fault that never actually healed. I reproduced this both in a unit test and by running the built binary against a fake sysfs tree: moving the device directory away produced the false recovery two seconds before the state check's debounced disappearance fatal. In a real cluster that healthy event clears the REPLACE_VM condition and fault-quarantine uncordons the node, and then it gets re-cordoned once the device is readable again. So a broken node can briefly look healthy every time sysfs blips.

UnreadableDevices is already checked here for the baseline gate, which is great. Could the recovery path get the same treatment? For example, only emit a recovery when the device was positively observed this poll, and hold the latched keys when the device is unreadable or absent, similar to the hold-last-known-state behavior in the state checks.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
```diff
@@ -0,0 +1,543 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if uncertain {
+		// A class directory we need was entirely absent while devices that
+		// should populate it exist: an uncertain observation, not evidence
+		// of mass failure. Hold state (do not stage) so the baseline stays
+		// owed and no spurious FATALs are emitted.
+		slog.Warn("Holding InfiniBand char-device poll: class directory unavailable",
```
KaivalyaMDabhadkar: on an InfiniBand node where the ib_umad module is not loaded, this warning will fire on every poll, which is once per second by default, forever. Holding is the right behavior, but could this log once per transition instead of every poll? There is a warn-once precedent in the counter code (warnUnreadableOnce) that would fit nicely.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
```diff
@@ -0,0 +1,543 @@
...
+	currentMissing := diffMissing(expected, observed)
+	events := c.buildEvents(currentMissing, baselineRun)
+
+	c.pending = &charDevPollCommit{missing: currentMissing, baselineRan: baselineRun}
+
+	return events, nil
+}
+
+// Commit installs and persists the most recently prepared state.
+func (c *InfiniBandCharDeviceCheck) Commit() {
```
KaivalyaMDabhadkar, line 282: Commit clears the pending baseline in memory but never calls state.Save(), and unlike the other checks this one has no other save path. On a quiet node nothing else writes the state file either, so the owed-baseline flag stays set on disk forever. The effect is that every pod restart within the same boot re-emits the "clearing stale conditions after reboot" event and re-asserts the fatals. I confirmed this by restarting the binary on the same boot ID and watching the duplicate baseline come through. It is harmless downstream but noisy, and it defeats the point of persisting the flag. I suggest calling Save here, ideally with the same retry-on-failure pattern saveState uses in base.go.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
```diff
@@ -0,0 +1,543 @@
...
+			events = append(events, c.missingEvent(key))
+		}
+
+		checks.EnsureClearPrecedesBatch(events)
+
+		return events
+	}
+
+	for key := range currentMissing {
+		if !c.previousMissing[key] {
```
KaivalyaMDabhadkar: A single bad poll is enough to send a REPLACE_VM fatal here. The device disappearance path in the state check requires three consecutive misses for exactly this reason. Since the device list and the mad/verbs directories are read at slightly different times, a driver teardown or reload that lands between those reads can make every char device for a device look missing for one poll, which would cordon the node. Requiring the same key to be missing for two or three consecutive polls before going fatal would remove that race, and it also softens the churn from the unreadable-device issue above

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
```diff
@@ -0,0 +1,543 @@
...
+
+	return missing
+}
+
+// missingEvent builds the FATAL event for a missing character device. A
+// missing node cannot be repaired from inside the workload — it requires
+// host-level driver/udev/reboot intervention and pods hard-fail to start —
+// so it recommends node replacement, matching how the real incident was
+// resolved (spare migration).
+func (c *InfiniBandCharDeviceCheck) missingEvent(key charDevKey) *pb.HealthEvent {
```
KaivalyaMDabhadkar, line 620: Could these events carry a per-kind ErrorCode (for example issm, umad, uverbs) on both the fatal and the recovery? Without one, the platform connector clears any condition message that mentions all of the recovery's entities. The issm and umad fatals for the same port have identical entities, and the uverbs fatal's device entity is contained in the port ones. So if both issm and umad are missing and only umad comes back, the umad recovery also wipes the issm message, and since the check only reports transitions, the issm fault is never re-added until reboot. The platform connector and fault-quarantine both already support code-scoped clearing, and the counter checks (withCounterCode) show the pattern.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
shaq918: Great catch — fixed in dd7f17eb. Recovery is now gated on positive observation: a latched key whose device drops out of the expected set (unreadable, absent, or momentarily gone from /sys/class/infiniband) holds its latch and emits nothing; only an expected-and-observed poll releases it. Added `TestIBCharDev_DeviceBlipHoldsLatch` reproducing your scenario: latch → device removed from enumeration → no recovery → device back with the node still missing → still latched and silent → node restored → recovery.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
shaq918: Done in dd7f17eb — added `charDevMissThreshold = 3`: a key must be expected-but-missing for 3 consecutive polls before the fatal fires, mirroring the disappearance debounce and for exactly the read-skew reason you describe. The streak is in-memory (a restart just restarts the debounce, which is the safe direction). `TestIBCharDev_OnePollBlipIsSwallowed` covers the single-bad-poll race.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
shaq918: Taken up in dd7f17eb: the latch is now persisted as `MissingCharDevices` in the shared state file, preserved across reboot loads alongside the disappearance latches, and seeded at construction. `TestIBCharDev_LatchSurvivesPodRestart` covers the recover-while-down case: fault latched → pod restart on the same boot with the node healed → new pod emits the recovery. Agreed on fixing it together with the baseline persistence — both are in this commit.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
shaq918: Fixed in dd7f17eb — Commit now persists via `state.Save()` with the same changed-or-retry pattern as `saveState` in base.go (a failed Save keeps retrying on subsequent commits). `TestIBCharDev_NoDuplicateBaselineAfterRestart` reproduces your repro: baseline consumed → new manager loads the same state file on the same boot → `PendingBaseline` is false and no duplicate clear is emitted.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
shaq918: Done in dd7f17eb — fatal and recovery events now carry the char-device kind as their ErrorCode (`issm`/`umad`/`uverbs`) via `withCharDevCode`, following the `withCounterCode` pattern, so a umad recovery can no longer wipe a still-open issm condition that shares its entities. The check-scoped baseline clear stays code-less by design (it means "clear everything"). Tests assert the code on every fatal and recovery.

health-monitors/nic-health-monitor/pkg/checks/state/ibdevnode.go
shaq918: Fixed in dd7f17eb — the held-poll warning now fires once per transition into the uncertain state (an `uncertainWarned` flag, reset when a certain observation succeeds), so a node without the ib_umad module logs it once rather than every second.

distros/kubernetes/nvsentinel/charts/nic-health-monitor/values.yaml
shaq918: Done in dd7f17eb — added `InfiniBandCharDeviceCheck` to the enabledChecks list and a per-check section in docs/configuration/nic-health-monitor.md describing the sysfs class-dir detection, the per-device expectation model, the IB-mode issm gating, the 3-poll debounce, and the persisted latch semantics.

health-monitors/nic-health-monitor/pkg/statefile/statefile.go
```diff
@@ -340,6 +357,7 @@ func (m *Manager) Load() error {
 			AnomalousCards:     loaded.AnomalousCards,
 			DisappearedDevices: stripDeviceBootMarkers(loaded.DisappearedDevices),
 			DisappearedPorts:   stripPortBootMarkers(loaded.DisappearedPorts),
+			MissingCharDevices: loaded.MissingCharDevices,
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: @shaq918 could you address this as well? The PR looks great otherwise, thanks for your contribution :)

KaivalyaMDabhadkar (approved):

PR #1548, opened 2026-07-28
fix: correct PG pipeline matching and exporter resume safety

Ayush7614:
## Summary
- Fix PostgreSQL `matchesOperators` to AND all MongoDB operators (e.g. `$gte` + `$lte`) instead of returning after the first map entry, matching MongoDB semantics.
- Fail closed when a provided pipeline stage cannot be parsed, so watchers never silently degrade into match-everything filters.
- Persist resume bookmarks after filtered-only batches, and make event-exporter workers always emit results for dequeued items after cancel (with token-writer drain) so in-flight sequences are not silently dropped.

## Test plan
- [x] `go test ./pkg/datastore/providers/postgresql/ -run 'TestMatchesOperators|TestNewPipelineFilter_FailsClosed|TestResumeTokenAdvanced|TestMultipleFiltered|TestMixedFiltered'`
- [x] `go test ./pkg/exporter/ -run 'TestWorkerPool_'` in `event-exporter`
- [x] CI green on this PR

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved event export reliability during cancellation, preventing hangs and preserving progress safely.
  - Resume positions now persist even when an entire event batch is filtered out.
  - Invalid change-stream filters now fail clearly instead of silently allowing unintended events.
  - Filter conditions now correctly require all specified operators to match.

- **Tests**
  - Added coverage for cancellation handling, filtered-batch resume persistence, invalid filters, and combined filter conditions.

### DIFF
--- a/event-exporter/pkg/exporter/worker_pool.go
+++ b/event-exporter/pkg/exporter/worker_pool.go
@@ -72,6 +72,9 @@ type workerPool struct {
 
 	dispatchCh chan workItem
 	resultCh   chan workResult
+
+	// onResult is an optional test hook invoked for every emitted work result.
+	onResult func(workResult)
 }
 
 func newWorkerPool(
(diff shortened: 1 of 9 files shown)


lalitadithya:
closing as stale, please feel free to reopen anytime. thank you!

PR #1551, opened 2026-07-30
feat: gate node-labeler detection labels on managed=false (ADR-040)

jtschelling:
## Summary

Gates the node-labeler and fault-quarantine on `nvsentinel.dgxc.nvidia.com/managed=false` so a node under active ExtRR ownership stops having detection labels stamped (which would otherwise keep DaemonSet health monitors scheduled on it).

### Changes

**labeler**
- `reconcileNodeLabelsInPlace`: strip `DCGMVersion`, `DriverInstalled`, and `KataEnabled` detection labels when `managed=false`; DaemonSet monitors self-evict via their existing `nodeSelector` without any per-monitor changes
- `updateNodeLabelsForPod`: skip pod-driven detection-label re-stamping for opted-out nodes
- Both paths **fail closed** on lister error — no mutations when the managed-label state is unknown, preventing accidental re-enablement of monitors on a released node
- New `labeler_node_labels_skipped_managed_total` Prometheus counter

**fault-quarantine values.yaml**
- All 6 default rulesets now include a Node CEL expression that skips quarantine when `nvsentinel.dgxc.nvidia.com/managed=false` is set, preventing fault-quarantine from opening a new quarantine session on a node already under ERR/Maestro ownership
- The existing `k8saas.nvidia.com/ManagedByNVSentinel` check is retained for backwards compatibility

## Test plan

- [x] `go build ./...` passes in labeler
- [x] `make lint` passes in labeler (0 issues)

## New Features
- Added support for node management opt-out labels across quarantine rules and label reconciliation.
- Opted-out nodes are excluded from quarantine and no longer receive or retain NVSentinel-managed detection labels.
- Label updates respond when a node’s management status changes.
- Reconciliation safely skips updates when opt-out status cannot be determined.

## Monitoring
- Added a Prometheus metric for label reconciliations skipped on opted-out nodes.

## Bug Fixes
- Improved node recovery so health-monitor workloads are removed and rescheduled when a node returns to service.


natherz97:
Can you add a check to the existing TestExtRRLifecycleHappyPath e2e test to confirm that the syslog-health-monitor and gpu-health-monitor are torn down?

natherz97 (approved):

natherz97 (approved):

PR #1556, opened 2026-07-31
fix(kom): recover deletions missed during downtime

### LINKED ISSUE #1542: [Bug]: Missed delete events will prevent kubernetes-object-monitor recovery
### Prerequisites

### Code of Conduct

### Bug Description

It's possible for the kubernetes-object-monitor to not send a healthy event for a deleted object if it misses the delete event fired from the object's corresponding informer. Specifically, if the kubernetes-object-monitor identifies a matched predicate and sends an uhealthy event, it will not be able to recover and send a healthy event if it misses the corresponding delete event. This is possible if the kubernetes-object-monitor is not running when the delete event occurs. If this occurs, the node will need to be manually unquarantined. 

### Component

Health Monitor

### Steps to Reproduce

1. Modify a K8s object to cause a predicate match that results in an unhealthy event and the k8s-object-monitor-policy-matches node annotation being added.
2. Disable the kubernetes-object-monitor
3. Delete the K8s object that resulted in the predicate match
4. Re-enable the kubernetes-object-monitor
5. Observe that node remains cordoned and that the kubernetes-object-monitor will not send a healthy event

If the delete event occurs before a new instance of an informer is running, there will be no delete event (whereas the same instance of the informer will either directly get the delete event or a delete event with a tombstone containing DeletedFinalStateUnknown indicating that the delete event was missed between the K8s watch closing and the re-list occurring). Most controllers workaround this issue by leveraging a finalizer to ensure that the object exists when the new instance of the informer starts. Since this is not an option, an alternative implementation would be to analyze the k8s-object-monitor-policy-matches annotation on monitor start-up and check if any of the referenced objects have been deleted (and not only rely on the delete event arriving). 

### Environment

- NVSentinel version: v1.12.0
- Kubernetes version: v1.31.6
- Deployment method: Helm

### Logs/Output

_No response_

Pybsama:
## Summary

- warm each Kubernetes object controller source before the final persisted-state scan
- gate reconciliation workers until every controller has finished state recovery
- use the direct API reader for startup reads while keeping the cached client for normal reconciliation and writes
- publish the normal healthy transition for resources deleted during downtime, while preserving recoverable state on read or publish failure
- decode persisted keys from the right using the REST-mapped GVK scope, so slash-containing policy names remain compatible and removed policies cannot be claimed by a shorter prefix

This preserves the existing exported `LoadState(ctx)` API; the manager startup
path uses the new scope-aware recovery method.

Closes #1542.

## Why the startup ordering matters

A scan before the controller source starts still leaves a gap: an object can be
present during the scan and disappear before the informer's initial list,
producing neither `NotFound` nor a delete event. The state-loader runnable
therefore waits for the controller's idempotent source warmup first, performs
the direct-API verification second, and only then releases the shared worker
barrier.

If source warmup, state loading, an API read, or recovery publication fails,
`Manager.Start` returns the error so the workload can retry instead of running
with silently stale quarantine state.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected

- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing

- [x] Tests pass locally

The public DCO-signed head `7915b7cac35663577a9bc858e78ccf9d6e1c4931`
was validated on its original base with Kubernetes envtest 1.36.2:

- full module race suite: 50 tests passed
- correct module-path statement coverage: 64.2%
- repository-pinned `golangci-lint` v2.5.0: 0 issues
- component `make lint-test`, `go vet ./...`, and `go build ./...`

Current-main compatibility was revalidated locally on
`0a95948f5a200c8d08641e4e89b8e3f210ec9fe3` after preserving the newer
cache-sync timeout and platform-connector token-auth changes:

- full module `-race -count=1` suite: 51 tests passed
- correct module-path statement coverage: 68.7% total and 40.6% for the controller package
- four scope/key regressions: 12/12 across three race-enabled repetitions
- barrier success/error paths: 100 assertions across 25 race-enabled repetitions
- `go vet ./...`, `go build ./...`, `gofmt`, and `git diff --check`
- Go 1.27-built `golangci-lint` v2.12.2: 0 issues
- independent final review: 0 Critical and 0 Important findings

The current main branch targets Go 1.27, while the repository's pinned
`golangci-lint` v2.5.0 cannot decode Go 1.27 export data. The newer linter was
used only for the current-main compatibility check; the pinned version had
already passed on the public head's original base.

The regression coverage includes a real Manager lifecycle, a Pod deleted with
zero grace while the monitor is stopped, direct reads before the manager cache
starts, non-`NotFound` API errors, recovery publication failure, Node deletion,
cluster- and namespace-scoped slash/prefix collisions, and state left by a
removed policy.

## Branch status

This PR remains Draft because the public head has not yet been refreshed and is
currently conflicting with `main`. The current-main port is locally verified;
a refreshed DCO-signed commit is still required before marking the PR ready.

## Checklist

- [x] Self-review completed
- [ ] Ready for review

### DIFF
--- a/health-monitors/kubernetes-object-monitor/pkg/annotations/manager.go
+++ b/health-monitors/kubernetes-object-monitor/pkg/annotations/manager.go
@@ -35,11 +35,19 @@ const (
 type PolicyMatchState map[string]string
 
 type Manager struct {
-	client client.Client
+	reader client.Reader
+	writer client.Writer
 }
 
 func NewManager(c client.Client) *Manager {
-	return &Manager{client: c}
+	return NewManagerWithReader(c, c)
+}
+
+func NewManagerWithReader(reader client.Reader, writer client.Writer) *Manager {
+	return &Manager{
+		reader: reader,
+		writer: writer,
+	}
 }
 
 func (m *Manager) AddMatch(ctx context.Context, nodeName, stateKey, targetNode string) error {
@@ -77,7 +85,7 @@ func (m *Manager) RemoveMatch(ctx context.Context, nodeName, stateKey string) er
 
 func (m *Manager) GetMatches(ctx context.Context, nodeName string) (PolicyMatchState, error) {
 	node := &v1.Node{}
-	if err := m.client.Get(ctx, types.NamespacedName{Name: nodeName}, node); err != nil {
+	if err := m.reader.Get(ctx, types.NamespacedName{Name: nodeName}, node); err != nil {
 		if apierrors.IsNotFound(err) {
 			return make(PolicyMatchState), nil
 		}
@@ -85,6 +93,10 @@ func (m *Manager) GetMatches(ctx context.Context, nodeName string) (PolicyMatchS
 		return make(PolicyMatchState), fmt.Errorf("failed to get node: %w", err)
 	}
 
+	return matchesFromNode(node)
+}
+
+func matchesFromNode(node *v1.Node) (PolicyMatchState, error) {
 	if node.Annotations == nil {
 		return make(PolicyMatchState), nil
 	}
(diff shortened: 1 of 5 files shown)


Pybsama:
Yes, this is still active. I resolved the current `main` conflict locally and
completed a fresh compatibility validation on `0a95948f5a200c8d08641e4e89b8e3f210ec9fe3`:

- 51/51 full-module race-enabled envtest tests passed
- 68.7% correct-path statement coverage
- `go vet ./...`, `go build ./...`, format/diff checks, and Go 1.27-compatible lint passed
- four scope/key regressions passed across three repetitions, and the barrier paths passed 25 race-enabled repetitions
- independent final review found 0 Critical and 0 Important issues

I have kept the PR in Draft because the public DCO-signed head
`7915b7cac35663577a9bc858e78ccf9d6e1c4931` is still conflicting with `main`;
publishing the refreshed head requires a new contributor DCO sign-off. Please
keep the PR open while that attestation is pending.

PR #1566, opened 2026-08-04
chore: document Percona backend for ARM64 support and Bitnami to Percona migration

### LINKED ISSUE #1497: [Feature]: MongoDB ARM support
### Prerequisites

### Code of Conduct

### Feature Summary

The current default deployment of MongoDB doesn't run on ARM nodes, but the Percona alternative does. This will need to be documented along with a migration path

### Problem/Use Case

Unable to run the default configuration of NVSentinel in ARM only nodes

### Proposed Solution

1. Update documentation to explain Percona supporting ARM nodes
2. Document migration path from default MongoDB to Percona
3. Consider switching to Percona as the default 

### Component

Fault Management

KaivalyaMDabhadkar:
## Summary

This PR addresses the documentation items of #1497. The default MongoDB backend (Bitnami)
only ships amd64 images, so the default configuration of NVSentinel cannot run on ARM64
nodes. The Percona operator backend is multi-arch and works on ARM64. These docs explain
that, describe how to enable Percona, and provide a tested migration path from Bitnami to
Percona.

This PR intentionally does not change the chart default. Switching the default to Percona
(the third item in #1497) is a separate decision and would be a separate PR. The docs here
describe Bitnami as the default, which matches the current chart.

Changes:

- New runbook `docs/runbooks/mongodb-bitnami-to-percona-migration.md`: the full migration
  procedure, including the complete list of objects that survive `helm uninstall` and must
  be cleaned up (PVCs, the credentials secret, all cert-manager TLS secrets including the
  per-replica server certs, and the runtime-created `resume-control` and `circuit-breaker`
  ConfigMaps), handling of in-flight quarantines and remediations on production clusters,
  cloud provider volume size minimums, verification steps, post-migration expectations,
  a troubleshooting table, and rollback guidance.
- `docs/configuration/mongodb-store.md`: new ARM64 support section with the exact image
  pull error users see with Bitnami on ARM64, a warning that backends cannot be switched
  in place on an existing release, a recommendation to pin the backend flags explicitly,
  and Percona configuration reference for volume size, pod placement, and the connection
  endpoint.
- `docs/designs/013-mongodb-bitnami-migration.md`: corrected the Chart.yaml condition
  snippet to match the actual chart, replaced the outdated migration procedure with a
  validated summary that points to the new runbook, and added ARM64 support to the
  positive outcomes.
- `docs/index.yml`: registered the new runbook in the navigation.

## Validation behind the docs

Every claim in these docs was verified on real clusters:

- Percona selected purely by chart defaults was exercised end to end by the e2e matrix on
  both amd64 and arm64 in #1554 (all 8 legs green).
- The migration procedure was executed on an amd64 OCI dev cluster, including a canary
  document proving the documented data loss behavior, and the leftover inventory in the
  runbook is what was actually observed after `helm uninstall`.
- The Bitnami failure on ARM64 and the working Percona install were both reproduced on an
  arm64 AWS dev cluster. The image pull error quoted in the docs is captured from that
  cluster.
- The warning against in-place backend switching comes from performing one: it fails on
  the immutable `create-mongodb-database` Job after already deploying parts of the second
  backend. The recovery steps in the troubleshooting table are the ones that worked.
- The volume size guidance comes from hitting the OCI 50Gi block volume minimum, which
  stops the operator before the replica set is initialized.
- The production cluster guidance (node annotations, remediation CRs, CSP maintenance
  event replay, event exporter backfill) is based on a review of how each component
  stores state that references database contents.

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for configuring MongoDB with the Percona backend, including ARM64 compatibility, storage requirements, readiness checks, and backend selection warnings.
  * Added a comprehensive Bitnami-to-Percona migration runbook covering prerequisites, cleanup, verification, troubleshooting, rollback, and optional data transfer.
  * Added the migration runbook to the documentation navigation.
* **New Features**
  * Added migration utilities for readiness checks, cleanup, optional data migration, and post-migration verification.
  * Added guided workflows for preparing, executing, and validating migrations.


docs/runbooks/mongodb-bitnami-to-percona-migration.md
```diff
@@ -0,0 +1,204 @@
...
+
+```bash
+kubectl logs -n nvsentinel deploy/health-events-analyzer | grep "Successfully pinged"
+```
+
+## After the migration
+
+- **CSP maintenance events replay.** The CSP health monitor tracks its progress through the provider's maintenance feed using the database itself. With the database wiped, it re-ingests every maintenance event still visible in the provider API, which can re-quarantine nodes for maintenance you already handled. Expect this for one polling cycle after the migration.
+- **Event exporter backfill.** With no resume token in the new database, the event exporter starts from scratch. Against an empty database this is a no-op. If you restore or seed any data, it exports all of it again, so warn the owners of the downstream sink about duplicates.
+- **Quarantine history is gone.** Nodes you cleared in step 3 are back in service. Real faults will be re-detected and re-quarantined by the monitors.
```
lalitadithya: this is not always the case right? xids for example will not be detected again

docs/runbooks/mongodb-bitnami-to-percona-migration.md
```diff
@@ -0,0 +1,204 @@
...
+
+```bash
+kubectl delete configmap resume-control circuit-breaker -n nvsentinel --ignore-not-found=true
+```
+
+Do not skip the secret cleanup. The old TLS secrets belong to the old certificate authority. If they survive, the new installation reuses them, clients present certificates the new database does not trust, and every datastore consumer crash-loops with connection errors that do not mention certificates at all.
+
+Do not delete your image pull secrets or any other secrets you created yourself.
+
+## Step 3: Clear fault-handling state (production clusters)
```
lalitadithya: if we were to migrate over only the active faults into the new database, won't our new cold start logic handle resuming?

lalitadithya:
the runbook makes sense, I'm thinking how much of this can we automate

docs/runbooks/mongodb-bitnami-to-percona-migration.md
KaivalyaMDabhadkar: Good catch, you are right. Persistent faults get redetected but one time events like XIDs will not, since the monitor state files keep the journal cursors past those lines. Will update the runbook

docs/runbooks/mongodb-bitnami-to-percona-migration.md
KaivalyaMDabhadkar: Yes, in theory that should work. Mongo keeps the same ObjectIDs, so the node annotations and CR names would still point at valid events.
A few things could bite us though:
- We must not restore the ResumeTokens collection. Those tokens only make sense on the
  cluster that created them.
- The event exporter would treat all the restored events as new and send them downstream
  again, so consumers of the sink would see duplicates.
- We need a query that picks only the active faults, and getting that filter slightly
  wrong leaves some events half processed.

I would rather keep this doc as the simple clean reinstall and try the selective restore separately later. If it holds up, we can document it as an option if we want to in the future

lalitadithya: > We need a query that picks only the active faults, and getting that filter slightly
wrong leaves some events half processed.

won't this basically be the same as the cold start query we already have? I.e., merge the cold start results from both ND and FR?

> We must not restore the ResumeTokens collection. Those tokens only make sense on the
cluster that created them.

Agreed

> The event exporter would treat all the restored events as new and send them downstream
again, so consumers of the sink would see duplicates.

Agreed. This is something we should document. This seems to be a lower risk operation and something we can handle with a code change or some kind of a metadata label fairly easily.

lalitadithya: I'm all for keeping things simple, but at the same time, I think we should also not leave the cluster with a bunch of nodes that will never get remediated.

lalitadithya:
can we also add some guidance for users who are using ArgoCD or FluxCD or equivalent?

docs/runbooks/mongodb-bitnami-to-percona-migration.md
KaivalyaMDabhadkar: Yeah, that makes sense. Let me test that out on a dev cluster

KaivalyaMDabhadkar: @lalitadithya I tested this on a dev cluster and it works. We restore the whole database (excluding ResumeTokens) and each component's cold start does its own filtering. FR's cold start query only enqueues events whose remediation is incomplete, so restored resolved events are simply non-acting.

The full run I did on the dev cluster: injected a fatal event with a remediation action, FQ cordoned and annotated the node, ND drained it, FR created the RebootNode CR. Then migrated with dump and restore. The CR and the annotation survived, mongorestore keeps the ObjectIDs so the references stayed valid, FR's cold start recognized its existing CR instead of creating a duplicate, and node-drainer resolved the event behind the annotation without errors.

On the event exporter: After a restore the backfill re-exports the restored events exactly once each (a clean 1:1 duplication), which is now documented in the runbook. Agree that a metadata label or small code change could remove even that, I can file a follow-up for it.

The runbook now has a "Preserving fault state (optional dump and restore)" section documenting all of this, plus a migrate-data.sh helper script that handles the dump (always excluding ResumeTokens) and the restore.

KaivalyaMDabhadkar:
> the runbook makes sense, I'm thinking how much of this can we automate

I added these:

scripts/mongodb-migration/ has four helpers: preflight.sh (read-only readiness check with a verdict table: current backend, cert-manager, storage class minimums vs the requested volume size, quarantined nodes, leftover state), cleanup.sh (guarded and self-verifying, with --dry-run and an optional --clear-fault-state), verify.sh (the five post-install gates), and migrate-data.sh (dump/restore for the state-preserving path).

skills/mongodb-migration/bitnami-to-percona/ has three agent skills following the pattern (readiness check, migration, verification) that sequence the scripts with explicit confirmation gates, so the destructive steps always stop for a human decision.

Everything was tested end to end on kind plus two dev clusters (amd64 on OCI and arm64 on AWS), including the failure paths that ended up in the troubleshooting table. The scripts do the mechanical parts; the decisions (accepting data loss, reviewing quarantined nodes) stay with the operator.

KaivalyaMDabhadkar:
> the runbook makes sense, I'm thinking how much of this can we automate

Added a "GitOps-managed installations" section to the runbook

docs/runbooks/mongodb-bitnami-to-percona-migration.md
KaivalyaMDabhadkar: Updated

docs/configuration/mongodb-store.md
```diff
@@ -13,13 +13,41 @@ Two in-cluster backends are supported: **Bitnami** (default) and **Percona Opera
...
+    enabled: true
+
+mongodb-store:
+  useBitnami: false
+  usePerconaOperator: true
+```
+
+## Switching backends on an existing installation
+
+Do not switch backends by changing the two flags on a live release with `helm upgrade`. The upgrade fails partway through with an immutable field error on the database initialization Job, and by then it has already deployed parts of the other backend. The result is two MongoDB clusters running side by side and services pointed at the wrong one.
```
lalitadithya: nit:


```suggestion
Do **NOT** switch backends by changing the two flags on a live release with `helm upgrade`. The upgrade fails partway through with an immutable field error on the database initialization Job, and by then it has already deployed parts of the other backend. The result is two MongoDB clusters running side by side and services pointed at the wrong one.
```

docs/runbooks/mongodb-bitnami-to-percona-migration.md
```diff
@@ -0,0 +1,326 @@
...
+Common reasons to migrate:
+
+- Your cluster has ARM64 nodes. The Bitnami MongoDB images are only published for amd64, so the default backend cannot run there. The Percona images are multi-arch. See [MongoDB Store Configuration](../configuration/mongodb-store.md) for details.
+- You want the operator features described in [ADR-013](../designs/013-mongodb-bitnami-migration.md), such as automated lifecycle management and integrated backups.
+
+The commands below assume the release name `nvsentinel` in the namespace `nvsentinel`. Adjust both if your installation differs.
+
+## What to expect
+
+- **Health event data is not preserved**, unless you use the optional restore path described in [Preserving fault state](#preserving-fault-state-optional-dump-and-restore). The default procedure is a clean reinstall of the datastore, not a data migration. NVSentinel stores operational telemetry (health events, resume tokens, maintenance events). Monitors repopulate the new database as they detect issues, but all history is lost.
```
lalitadithya: can  we make the preserve health event data the default path and then have the optional as wipe everything and start from scratch?

docs/runbooks/mongodb-bitnami-to-percona-migration.md
```diff
@@ -0,0 +1,326 @@
...
+| `migrate-data.sh` | optional | `dump` streams a mongodump archive out of the old backend (always excluding `ResumeTokens`); `restore` streams it into the new one. Document IDs are preserved, so node annotations and remediation resource names stay valid. |
+| `verify.sh` | step 6 | Waits on the five post-install gates and prints a verdict table. |
+
+All four respect `NVSENTINEL_NAMESPACE` and `NVSENTINEL_RELEASE` (default `nvsentinel` for both). The scripts automate the steps; the decisions (acknowledging data loss, reviewing quarantined nodes) stay with you.
+
+For AI-agent-assisted runs of this procedure, `skills/mongodb-migration/bitnami-to-percona/` contains agent skills (readiness check, migration, verification) that sequence these scripts with the required confirmation gates. They work with any agent that reads the SKILL.md format.
+
+## GitOps-managed installations (ArgoCD, Flux)
+
+If NVSentinel is deployed by a GitOps controller, stop reconciliation before you start. Otherwise the controller fights the migration: it re-syncs the resources you just removed and reverts the backend change. These are general guidelines; adapt them to how your applications are structured.
```
lalitadithya:
```suggestion
If NVSentinel is deployed by a GitOps controller such as ArgoCD/FluxCD, stop reconciliation before you start. Otherwise the controller fights the migration: it re-syncs the resources you just removed and reverts the backend change. These are general guidelines; adapt them to how your applications are structured.
```

scripts/mongodb-migration/cleanup.sh
```diff
@@ -0,0 +1,159 @@
...
+#
+# Usage: cleanup.sh [--dry-run] [--clear-fault-state] [--yes]
+#   --dry-run            print what would be deleted, delete nothing
+#   --clear-fault-state  also remove quarantine node annotations, remediation CRs,
+#                        and event-labeled log-collector jobs (production clusters)
+#   --yes                skip the interactive confirmation
+#
+# Exit codes: 0 = cleanup complete and verified, 1 = error or leftovers remain,
+#             3 = refused (release still installed, or not confirmed)
+set -uo pipefail
```
lalitadithya:
```suggestion
set -euo pipefail
```

please update in other places as well

skills/mongodb-migration/bitnami-to-percona/check-mongodb-migration-readiness/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+name: check-mongodb-migration-readiness
+description: >-
+  Readiness check for migrating an NVSentinel installation from the Bitnami
+  MongoDB backend to the Percona Operator backend. Runs the preflight script,
+  interprets the verdict table, and captures the operator decisions (data
+  handling, quarantined nodes, GitOps) that the migration skill requires.
+  Use this first, before any migration step.
+maturity: experimental
+lifecycle: evergreen
+api-version: nke.skills/v1
```
lalitadithya: this should be nvsentinel right?

KaivalyaMDabhadkar: Yeah that was an oversight, fixed it

scripts/mongodb-migration/cleanup.sh
KaivalyaMDabhadkar: Fixed

docs/runbooks/mongodb-bitnami-to-percona-migration.md
KaivalyaMDabhadkar: Fixed

docs/runbooks/mongodb-bitnami-to-percona-migration.md
KaivalyaMDabhadkar: Fixed

docs/configuration/mongodb-store.md
KaivalyaMDabhadkar: FIxed

docs/runbooks/mongodb-bitnami-to-percona-migration.md
```diff
@@ -0,0 +1,326 @@
...
+
+```bash
+kubectl delete externalremediationrequests -l app.kubernetes.io/managed-by=nvsentinel --ignore-not-found=true
+```
+
+These remediation resources are cluster scoped (kubectl ignores any namespace flag for them), so the commands select by the ownership label fault-remediation puts on everything it creates rather than deleting every instance on the cluster. If `kubectl get rebootnodes` still lists `maintenance-*` resources afterwards, they were created without the label (for example by an older release); review them and delete by name if they belong to this installation.
+
+```bash
+kubectl delete jobs -l dgxc.nvidia.com/event-id -n nvsentinel
+```
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Fixed

docs/runbooks/mongodb-bitnami-to-percona-migration.md
```diff
@@ -0,0 +1,260 @@
...
+
+Reconciliation stays off until step 6a, after the new values are in git. One Flux-specific rule for later: do NOT delete the suspended HelmRelease in step 4a; Flux does not finalize suspended objects, so the deletion hangs. Flux installations are real Helm releases, so step 4a uses `helm uninstall` directly instead. Helm-managed installations skip this step.
+
+## Step 3: Capture the data
+
+### Step 3a: Preserve health event data (default)
+
+Stop the components that write references to events, and wait for their pods to terminate. An event written after the dump is absent from the archive; if fault-quarantine then annotated a node for it, that annotation would point at a document the restore does not contain. Scale down each of these that exists in your installation (reconciliation is already off on GitOps-managed clusters, so nothing scales them back up):
+
+```bash
```
lalitadithya, line 94: can we put this into the script as well?

KaivalyaMDabhadkar: Done. The dump now takes a `--stop-writers` flag that scales down fault-quarantine, node-drainer and fault-remediation and waits for their pods to be fully deleted before dumping, and the runbook now uses that as the primary path with the manual loop kept as an alternative.
I kept it behind a flag instead of making it the default because the dump also doubles as a plain backup tool, and the script never scales the writers back up (in the migration they get removed in the next step anyway). Auto stopping would leave fault handling silently off for someone who just wanted a copy of the data.

lalitadithya (approved):

PR #1572, opened 2026-08-06
feat(fault-quarantine): gate events for opted-out nodes via skipNodeL…

cbumb:
## Summary

Add a configurable `skipNodeLabels` list to fault-quarantine that silently drops all events (healthy and unhealthy) for nodes carrying matching labels. This prevents opted-out nodes (e.g. nodes under external remediation via ExtRR) from being quarantined or uncordoned by fault-quarantine. Default config ships with `nvsentinel.dgxc.nvidia.com/managed=false`. Updates ADR-040 to reflect the fault-quarantine approach for cluster-scope monitor gating.

**Known tradeoff:** Because the gate is in fault-quarantine rather than at the monitor/platform-connector level, node status conditions set by platform-connectors may become stale for nodes under an ERR. This is a reasonable tradeoff since events should still be "observed, recorded, and exported as usual" per ADR-040, and the guard only applies to state-changing actions (quarantine, uncordon, annotation updates).

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Refactoring
- [ ] Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added configurable node-label exclusions for fault quarantine.
  * Nodes matching configured labels are excluded from quarantine and uncordoning actions.
  * Added default handling for nodes labeled as unmanaged.
  * Added metrics to track events skipped due to node labels.

* **Documentation**
  * Updated external remediation behavior and failure-mode documentation.

* **Tests**
  * Added end-to-end coverage for label matching and remediation lifecycle changes.


jtschelling (approved):

lalitadithya:
why can't this be dropped at the platform connector layer since that has access to the node labels IIRC?

cbumb:
> why can't this be dropped at the platform connector layer since that has access to the node labels IIRC?

`MetadataAugmentor` caches node data for up to an hour, so the label change wouldn't take effect for up to an hour in either direction. 
`OverrideTransformer` can only set three fields (isFatal, isHealthy, recommendedAction) and cannot set ProcessingStrategy. 
Both pieces would need code changes to actually work.

lalitadithya:
> MetadataAugmentor caches node data for up to an hour, so the label change wouldn't take effect for up to an hour in either direction.

we could always cache this specific label for a shorter TTL. We have to keep in mind that the system will be eventually consistent, and we can't fire off a get call for every health event. The FQ also has an informer which is backed by a cache

cbumb:
> > MetadataAugmentor caches node data for up to an hour, so the label change wouldn't take effect for up to an hour in either direction.
> 
> we could always cache this specific label for a shorter TTL. We have to keep in mind that the system will be eventually consistent, and we can't fire off a get call for every health event. The FQ also has an informer which is backed by a cache

Yes both are eventually consistent. The difference is the lag. If I am not wrong FQ's informer uses a watch, so the API server pushes label changes in sub-second. A TTL cache is poll-based, so the lag equals the full TTL no matter how short we set it. Even at 30s TTL, that's 30 seconds of wrong behavior on every label flip vs under a second with the informer right?

And I also confirmed with Ajay that informers on a DaemonSet are wasteful, the only option in platform-connector would be the TTL cache. FQ is a single Deployment, so one informer is fine.

Both approaches work, but I thought the informer in FQ gives tighter consistency with less infrastructure cost.

lalitadithya:
> If I am not wrong FQ's informer uses a watch, so the API server pushes label changes in sub-second. A TTL cache is poll-based, so the lag equals the full TTL no matter how short we set it. Even at 30s TTL, that's 30 seconds of wrong behavior on every label flip vs under a second with the informer right?

yes, but there is no guarantee. a watch drop can also result in a larger lag. Yes, we can't do informers in the daemonset it will have to be TTL based. 

Now that I think about it more, I'm not sure if the platform connector is the right place either. We also have to keep in mind that the number of cluster wide monitors are limited to currently HEA, CSP and KOM. The HEA case is an easy one because it doesn't publish a healthy event, so dropping the event is good enough. The CSP and KOM will need some thought because as I understand it KOM keeps state on the node object so dropping events out of it will cause the state to go out of sync with reality. The CSP health monitor also has some state that it maintains in the database. So, I would like to understand what the implications of having the checks in the monitor itself and making sure state logic if any accounts for this?

cbumb:
> So, I would like to understand what the implications of having the checks in the monitor itself and making sure state logic if any accounts for this?

Yeah that's the thing, each monitor handles state differently. KOM has state on the node object, CSP has state in the database, and HEA is the only clean one. Accounting for all of that means three separate implementations with different state cleanup logic.

That's why I went with fault-quarantine. Monitors keep running normally and their internal state stays consistent. Events still flow through and get stored for audit. FQ just decides not to act on opted-out nodes. Nothing goes out of sync, and also future monitors get the same behavior without opting in. We also dont need to worry about someone not adding the logic in the future monitors.

lalitadithya:
> Yeah that's the thing, each monitor handles state differently. KOM has state on the node object, CSP has state in the database, and HEA is the only clean one. Accounting for all of that means three separate implementations with different state cleanup logic.

correct, this is exactly why I think it has to be per health monitor. If KOM has state saying that the node is bad and it uses that to dedup and we drop the unhealthy event then the node will never be marked once it returns. Same with the case of the CSP health monitor. I don't think we can account for these types of nuances centrally unless we build a backlog of unprocessed events and then reprocess them like cold start, but that is going to get complex quickly

natherz97:
Adding to platform-connector or fault-quarantine are the only approaches which enforce comprehensive coverage for all health events, otherwise any health-monitor which is added would need knowledge of this label. Right now the fault-quarantine rule-set is the mechanism we have to indicate whether a given unhealthy event should contribute to a node's quarantine session. We do not have an equivalent to apply to healthy events so this PR is the first step that applies filtering across all events (I think we could add a full rule-set evaluation for healthy events or for all events later if we need functionality beyond node labels). Independent of this use-case it does seem like we should have this equivalent for healthy events.

I'm trying to think what we want to do in the short-term since if we added this functionality to the global monitors without any optimizations about when it sends events for a node (such as queuing events or force-sending a healthy event when the label is removed), we'd get the same behavior as keeping it in fault-quarantine with the tradeoff of having to implement it 3 times.

cbumb:
> correct, this is exactly why I think it has to be per health monitor. If KOM has state saying that the node is bad and it uses that to dedup and we drop the unhealthy event then the node will never be marked once it returns. Same with the case of the CSP health monitor. I don't think we can account for these types of nuances centrally unless we build a backlog of unprocessed events and then reprocess them like cold start, but that is going to get complex quickly

That's a fair concern. But I think the window is pretty narrow. After external remediation, the fault should normally be fixed, so the monitor would see the state change and publish a healthy event, which clears the dedup. The case where this breaks is if the fault persists after remediation AND only a cluster-scope monitor can detect it AND its dedup prevents re-publishing. That's a pretty specific edge case right?

Also If I am not wrong DaemonSet monitors cold-start when the managed label is removed, so they re-detect everything fresh. This only applies to the cluster-scope ones.

And if we do put the check in each monitor, we'd need to build that "remember to re-publish later" logic into each one differently, which is the same state complexity you were calling out earlier.

If we're really worried about the re-detection case, I think a simpler path would be to add something in FQ that re-processes stored events when a node's skip label is removed, rather than building custom state management into each monitor. But I'd treat that as a follow-up if it turns out to be a real problem.

cbumb:
> > correct, this is exactly why I think it has to be per health monitor. If KOM has state saying that the node is bad and it uses that to dedup and we drop the unhealthy event then the node will never be marked once it returns. Same with the case of the CSP health monitor. I don't think we can account for these types of nuances centrally unless we build a backlog of unprocessed events and then reprocess them like cold start, but that is going to get complex quickly
> 
> That's a fair concern. But I think the window is pretty narrow. After external remediation, the fault should normally be fixed, so the monitor would see the state change and publish a healthy event, which clears the dedup. The case where this breaks is if the fault persists after remediation AND only a cluster-scope monitor can detect it AND its dedup prevents re-publishing. That's a pretty specific edge case right?
> 
> Also If I am not wrong DaemonSet monitors cold-start when the managed label is removed, so they re-detect everything fresh. This only applies to the cluster-scope ones.
> 
> And if we do put the check in each monitor, we'd need to build that "remember to re-publish later" logic into each one differently, which is the same state complexity you were calling out earlier.
> 
> If we're really worried about the re-detection case, I think a simpler path would be to add something in FQ that re-processes stored events when a node's skip label is removed, rather than building custom state management into each monitor. But I'd treat that as a follow-up if it turns out to be a real problem.

I think the last point coincides with Nathan's comment

lalitadithya:
if we would like to break this into two parts that is fine with me, we could start off with dropping the health events in a central place and then as a follow up we can have per health monitor handling if required. For this as well, I'm not sure if FQ is the right home for this logic because this means we drop the events after some part of the system and some downstream consumers have already consumed it. Specifically, this would mean that node conditions, the event exporter and the health event analyzer and anything downstream of these such as an observability stack or an analytics pipeline. Dropping it in the earliest possible place I think would be the most ideal so that we never end up in a state where part of the system emit signals while others don't. Previously this not much of a concern because the behaviour was consistent between per node health monitors and cluster wide health monitors. But now with the more recent changes the per node health monitor are evicted which would mean they will not suffer the similar issue. 

We could add a delayed processing strategy to the health events which would work as well, but that is something we'd probably want to do as a follow up if we have more use cases outside of this.

I agree that having this per health monitor does complicate the logic, but in an ideal world we should have minimal number of cluster wide health monitors since most of the faults can be detected on a per node level. With that said, I also think it would be simpler than having to do event replays in FQ given that right now the biggest source for event replay would likely be the CSP health monitor since those deals with times for maintenance and it is entirely possible that a future maint notification comes while a node is being rebooted and those can't be dropped. Similar case would be with the KOM which monitoring node conditions that are set by an external entity in response to a change outside of the node/cluster.

PR #1577, opened 2026-08-07
docs: ADR-051 MaintenanceRequest (MR) — entry door for signaling incoming node maintenance

jtschelling:
## Summary

Adds ADR-051, the design for the `MaintenanceRequest` (MR) CRD — the inbound counterpart to `ExternalRemediationRequest` (ADR-040). A requester (an external system or operator) creates an MR to tell NVSentinel that maintenance is incoming for a node, so NVSentinel prepares the node (cordon, drain) before the work begins instead of the requester bypassing the pipeline.

## Key design points

- **Emit as authored.** On create, the reconciler emits `spec.healthEvent` unchanged — the requester picks `recommendedAction`, and the pipeline routes it to the matching remediation (`RESTART_VM` → RebootNode, `CUSTOM`/`external-remediation` → ERR). The MR carries a `startTime` recording when the maintenance window opens. The MR name and UID are stamped into event metadata for traceability only; nothing consumes them.
- **Deletion is the clear, and the requester owns it.** The MR persists until the requester deletes it. NVSentinel does **not** track the maintenance to completion — the reconciler does not watch the remediation it triggered, and there is no automatic cleanup. Deleting the MR runs its finalizer, which emits a matching `isHealthy=true` event to retract the fault and un-cordon the node. Auto-deletion on remediation completion is deliberately deferred.
- **Lives in `lifecycle-manager`.** The reconciler and webhook are added to the existing component (#1610, introduced with the validation-controller), reusing its manager, leader election, webhook server, config, chart, and RBAC scaffolding. MR adds the one capability it lacks: a `healthpub` emitter plus a `hostPath` mount of the platform-connector socket, which is that component's first node-local socket dependency.
- **CRD follows janitor's ExtRR pattern** — proto as the source of truth, `protoc-gen-crd` for the CRD YAML, and a Go wrapper with `protojson` marshalling, because `spec.healthEvent` embeds the proto-generated `HealthEvent`. This means API version `v1` and group `nvsentinel.dgxc.nvidia.com`, so `lifecycle-manager` serves that group alongside `nvsentinel.nvidia.com` for `ValidationRequest`.
- **Simple lifecycle.** Present = maintenance active, absent = cleared. A single `HealthEventEmitted` condition, no `completionTime`, and no owner-references, GC coupling, or cross-CRD watches.
- **Webhook** rejects an `isHealthy=true` opening event, requires `startTime` to be in the future, freezes `spec.healthEvent` after creation, and best-effort-rejects a duplicate MR for a node.

## Main tradeoff to push on

**A forgotten MR leaves a node cordoned indefinitely.** Nothing reclaims it, so MRs become an operational surface that needs monitoring. This is the accepted cost of dropping automatic cleanup in the first iteration.

## Alternatives considered

The most substantial is a **`kubernetes-object-monitor` policy instead of a controller** — KOM already implements MR's exact open/close model generically, so it would need no new reconciler. It was rejected because every field in KOM's `HealthEventSpec` is a static TOML literal: each new fault type would need its own policy and a release, and no single request could carry per-request detail. The ADR records the full comparison, along with the deferred auto-completion design and three other rejected shapes.

## Dependencies

- ADR-040 (`ExternalRemediationRequest`) — shipped.
- A `fault-quarantine` ruleset matching MR-emitted events is the **only hard prerequisite**, and must land with MR or before it.
- The `external-remediation` action that renders an ERR is needed only for the `CUSTOM` case.
- `fault-remediation` should be idempotent on re-publish, since the MR re-emits its opening event on retry.

## Summary by CodeRabbit

- **New Features**
  - Added a new `MaintenanceRequest` API for scheduling maintenance windows with associated health events.
  - Maintenance requests now track lifecycle conditions and automatically emit healthy clearing events when maintenance ends.
  - Added validation, permissions, deployment integration, and testing guidance for the new API.

- **Documentation**
  - Added a detailed architecture decision record describing the maintenance request lifecycle, behavior, and operational considerations.


docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+
+ADR-040 covers the NVSentinel-initiated direction. It does not cover the reverse: an external system that *proactively* needs a node drained before it can begin work. In that case NVSentinel does not generate a fault; the external system has received a signal from outside the cluster — a CSP maintenance notification, a hardware-monitoring alert, an external event source — and needs NVSentinel to prepare the node on its behalf.
+
+Without a formal entry point for this direction, external automation must side-step NVSentinel (cordoning nodes directly, fighting over taints) or rely on custom integrations that duplicate NVSentinel's quarantine and drain logic. Both outcomes undermine the ownership model established by ADR-040.
+
+## Decision
+
+Introduce a new CRD, `ExternalFault` (EF), in the existing `nvsentinel.dgxc.nvidia.com` API group. EF is the **entry door** into NVSentinel ownership transfer from an external system:
```
lalitadithya: nit: should we call this `MaintenanceRequest` so that this can be generalized to other use cases in future if required?

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+
+- Created by an external system or human operator when it determines a node must be drained and released before external repair work can begin.
+- The EF reconciler emits a `CUSTOM` health event to the platform-connector. Provided the pipeline is configured to act on that event (see [Pipeline dependencies](#pipeline-dependencies)), this drives the normal NVSentinel flow: quarantine (cordon), drain, and ERR creation by `fault-remediation`.
```
lalitadithya: why does this have to be custom?

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+- The ERR created from an EF-triggered health event carries an `OwnerReference` pointing to the parent EF, linking the two coordination objects for the lifetime of the repair.
+- The EF reconciler watches its child ERR (when one exists) and uses the ERR reporting its remediation complete as the internal trigger to clear the fault — it does **not** surface the ERR's state as an EF condition. The external system never clears faults in NVSentinel: a node is only un-quarantined when the source that raised the fatal health event submits a matching `isHealthy=true` event. EF *is* that source for an external-origin fault, so when the repair completes it emits a second health event — a plain `isHealthy=true` event for the same node/check — to retract the fault it raised, and sets `FaultCleared=True` once that clearing event is submitted.
```
lalitadithya: I'm not too sure about this. I know we spoke about this earlier, should we offload this to the external system and have it delete the CR to signal completion?

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+
+Human operators follow the same path as automated systems: creating an EF CR is the sanctioned mechanism for requesting a node drain before manual intervention.
+
+## ExternalFault Reconciler
+
+Where the EF reconciler is hosted is a distinct, open decision. It is unusual among NVSentinel controllers because it needs three capabilities that no single existing component provides together:
+
+1. **A controller-runtime manager + validating webhook** — to reconcile the EF CRD (finalizers, status, `Owns(&ExternalRemediationRequest{})`) and to enforce admission checks (duplicate-node, `nodeName` immutability, node existence).
+2. **A health-event emitter** — a `healthpub.Publisher` over the platform-connector's node-local socket, to submit both the opening `CUSTOM` event and the closing `isHealthy=true` event.
+3. **Proximity to the ERR** it watches — the `Owns`-watch reacts to the child ERR's `ExternalRemediationComplete` status, which is set by the ERR reconciler.
```
lalitadithya: this is not really required, as long as we can import the types, we should be able to set up the reconciler anywhere

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+
+### Option 2 — csp-health-monitor
+
+Host the EF reconciler in `csp-health-monitor` (or another health monitor).
+
+- **Pros:** health monitors are native emitters — `csp-health-monitor` already mounts the platform-connector socket and holds a `healthpub.Publisher`, so capability (2) is free. It also already detects CSP maintenance signals (AWS/GCP), the canonical EF trigger, so the component that observes external faults would also own their lifecycle — a good fit with EF's "synthetic monitor" framing.
+- **Cons:** `csp-health-monitor` is a poll/emit loop, **not** a controller-runtime app — no manager, no CRD reconcilers, no webhook server, and no leader election (it runs as a multi-replica Deployment). Hosting EF here means grafting on the entire controller-runtime + webhook stack (capability 1). It also splits the EF↔ERR pair across components: `csp-health-monitor` would import janitor's CRD types and `Owns`-watch a janitor-owned CRD, and the EF webhook would either need a new server here or stay behind in janitor.
```
lalitadithya: agreed, but it also has the ability to poll/watch an API for events, so maybe the way we should think about this is in this case the API we are polling/watching is the Kubernetes API

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+3. **Proximity to the ERR** it watches — the `Owns`-watch reacts to the child ERR's `ExternalRemediationComplete` status, which is set by the ERR reconciler.
+
+No component has all three, so each option grafts on what it lacks. The [Implementation](#implementation) section below is written against **Option 1**.
+
+### Option 1 — janitor (reference implementation)
+
+Host the EF reconciler in `janitor`, alongside the ERR reconciler and the shared validating-webhook server.
+
+- **Pros:** janitor is already a controller-runtime manager with a webhook server (RebootNode/TerminateNode/GPUReset/ERR), and the ERR reconciler — whose status the EF watches — lives here, so capabilities (1) and (3) come for free. Smallest change.
+- **Cons:** janitor emits no health events today; EF makes it an emitter, adding the platform-connector socket mount and a `healthpub` client (see [Health-event emission](#health-event-emission-janitor--platform-connector)). That is a new outbound dependency for a component that previously spoke only to the Kubernetes API.
```
lalitadithya: This also makes the configuration complex, say a customer only wants to use the health monitoring capabilities now they also have to enable a non health monitor component

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+│   └── janitor_webhook.go                 (extended — EF validator added)
+└── main.go                                (extended — EF controller + platform-connector emitter + EF TTL)
+```
+
+### CRD schema
+
+The CRD is proto-generated, matching the ERR pattern:
+
+```proto
+message ExternalFaultSpec {
```
lalitadithya: Can we also add some notion of time window, so that if some system wants the node but at a later date, it can be expressed in the API?

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+
+```text
+# kubebuilder:rbac:groups=nvsentinel.dgxc.nvidia.com,resources=externalfaults,verbs=get;list;watch
+```
+
+No new node-level RBAC is needed; EF does not touch nodes directly.
+
+### Sequence: EF-initiated fault (happy path)
+
+```text
```
lalitadithya: can we make this mermaid diagram?

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+
+**Operator delete** (DeletionTimestamp set):
+1. If child ERR exists and `FaultCleared` has not been set to `True`, record an operator-delete close metric.
+2. Stamp `CompletionTime`.
+3. Remove the cleanup finalizer. Kubernetes GC cascades the OwnerReference deletion to the child ERR (see *OwnerReference design* below).
+
+Operator-delete is a deliberate **force-close**: it does **not** emit the clearing `isHealthy=true` event. If the EF was still open, the synthetic fault is therefore *not* retracted and the node stays cordoned — deleting an open EF means abandoning the drain request, and the node is left cordoned for manual attention (open EFs are filterable right up to deletion). One race is tolerated by design: an opening event still in flight can create an ERR *after* the EF is gone; that ERR carries an `OwnerReference` to a UID that no longer exists, so Kubernetes GC deletes it on discovery — and the ERR's own finalizer scrubs the node — rather than leaving an orphan.
```
natherz97: Should we be sending a healthy event as part of the EF deletion rather than rely on the manual uncordon? Since EF deletion will trigger ERR deletion, we'll ensure that the ERR taint + labels are removed as part of its finalizer so should we be consistent and fully clear the node cordon on EF deletion?

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+
+### Option 2 — csp-health-monitor
+
+Host the EF reconciler in `csp-health-monitor` (or another health monitor).
+
+- **Pros:** health monitors are native emitters — `csp-health-monitor` already mounts the platform-connector socket and holds a `healthpub.Publisher`, so capability (2) is free. It also already detects CSP maintenance signals (AWS/GCP), the canonical EF trigger, so the component that observes external faults would also own their lifecycle — a good fit with EF's "synthetic monitor" framing.
+- **Cons:** `csp-health-monitor` is a poll/emit loop, **not** a controller-runtime app — no manager, no CRD reconcilers, no webhook server, and no leader election (it runs as a multi-replica Deployment). Hosting EF here means grafting on the entire controller-runtime + webhook stack (capability 1). It also splits the EF↔ERR pair across components: `csp-health-monitor` would import janitor's CRD types and `Owns`-watch a janitor-owned CRD, and the EF webhook would either need a new server here or stay behind in janitor.
+
+### Option 3 — standalone component
```
natherz97: I'm working on an ADR which will add a validation-controller that reconciles a ValidationRequest: https://github.com/NVIDIA/NVSentinel/pull/1496. We could combine both of these validation and maintenance controllers into a single node lifecycle component. What do you think?

docs/designs/049-external-fault.md
natherz97: Related to Lalit's 2 comments above, would it be possible to not tie EFs to only support ERRs? It might useful for operators to submit a generic MaintenanceRequest which could cordon, drain, reboot, and uncordon automatically. 

I think we could support 2 modes for completing the maintenance: 
1. Require the creator of the MaintenanceRequest to signal that maintenance is complete by writing to it's status (similar to what we do for ERR) or by deleting the MaintenanceRequest
2. Add the OwnerReference functionality which would allow the controller to automatically complete the maintenance once the resulting ERR, RebootNode, etc. has a successful status.

docs/designs/049-external-fault.md
```diff
@@ -0,0 +1,393 @@
...
+- **RBAC.** No new RBAC for emission itself (the gRPC call is not a Kubernetes API call), but the reconciler needs `externalfaults` (+ `/status`, `/finalizers`) and `get;list;watch` on `externalremediationrequests`.
+
+### Pipeline dependencies
+
+EF only *emits* a health event; whether that event flows through quarantine → drain → ERR depends on the rest of the pipeline being configured to act on it. Two pieces must be in place for the happy path to fire, and neither is implied by the EF reconciler itself:
+
+1. **A `fault-quarantine` ruleset that matches EF-emitted events.** `fault-quarantine` cordons only events matching one of its rulesets; a generic external-remediation event (e.g. a CSP-maintenance `CUSTOM` event) matches none of the default agent/check-specific rulesets and is skipped, so no cordon/drain occurs. The EF-emitted event must carry fields that match an existing ruleset, or a dedicated ruleset for EF-originated events must be added.
```
natherz97: The fault-quarantine module is receiving a new ruleset which maps unhealthy events to a set of validation tests which will be run after an unquarantine event occurs for the given node. We could wire this functionality into EFs to support nodes being re-validated after the given maintenance completes. I think we can track that as a follow-up but would be useful in situations where CSPs perform maintenance without terminating the underlying node objects (such as with GCP GB200 maintenance).

docs/designs/049-external-fault.md
jtschelling: originally this was geared towards matching the EF directly with the ExtRR, but it doesn't have to be. I've updated it to be more universal and not have the field requirements. I think it would've been a bit of a problem anyways since nothing in the ADR here makes a requirement on fault-remediation configurations, which is the dependency for actually going from health event to ExtRR/whichever remediation

docs/designs/049-external-fault.md
jtschelling: yes, implemented

docs/designs/049-external-fault.md
jtschelling: with the move to MaintenanceRequest i've simplified the completion signaling, PTAL

docs/designs/049-external-fault.md
jtschelling: yes - reworked where the reconciler lives

docs/designs/049-external-fault.md
jtschelling: ptal at the reconciler deployment reorganization

docs/designs/049-external-fault.md
jtschelling: agreed

docs/designs/049-external-fault.md
jtschelling: yes, implemented

docs/designs/049-external-fault.md
jtschelling: i reworked the completion/clearing logic, ptal

docs/designs/049-external-fault.md
jtschelling: i think thats a good idea. i've reworked the deployment logic for this ADR, i think it will fit nicely with the ValidationRequest reconciler. Having two open ADRs for this makes merge timing a bit difficult so i've intentionally not mentioned #1496 in this ADR

docs/designs/049-external-fault.md
jtschelling: reworked to no longer tie EF/MR to an ERR

docs/designs/049-maintenance-request.md
```diff
@@ -0,0 +1,341 @@
...
+1. Add cleanup finalizer.
+2. Seed `HealthEventEmitted=Unknown`.
+3. Emit the health event from `spec.healthEvent` to the platform-connector **as authored** — the creator's `recommendedAction`/`customRecommendedAction` are passed through unchanged. Inject the MR's name and UID into `healthEvent.metadata["maintenanceRequestName"]` / `["maintenanceRequestUID"]` so `fault-remediation` can associate the child CR with this MR.
+4. On success: set `HealthEventEmitted=True`. On failure: return error; controller-runtime requeues with backoff. The opening emit is gated on `HealthEventEmitted != True`, so a failed emission is retried rather than stranded.
+
+**Watching** (`HealthEventEmitted=True`):
+- Idle. The controller watches janitor's maintenance CRs (`RebootNode`, `TerminateNode`, `GPUReset`, `ExternalRemediationRequest`) and maps a child back to its MR by the association label (see *Associating the child with its MR*). When the mapped child sets `Status.CompletionTime`, the reconciler **deletes this MR** — which runs the finalizer below. If no child is ever created (e.g. the emitted event matches no remediation action), the MR stays here until an operator deletes it.
```
lalitadithya: why does it need the watch? the operator can also use a system other than janitor for remediation so in that case we now need the mapping that is present in fault remediation to also be in the lifecycle manager

docs/designs/051-maintenance-request.md
```diff
@@ -0,0 +1,341 @@
...
+
+`lifecycle-manager` runs its own validating-webhook server:
+
+| Check | On create | On update |
+|---|---|---|
+| `spec.healthEvent.nodeName` is non-empty | ✓ | ✓ |
+| `spec.healthEvent.isHealthy` is `false` (an opening event must *raise* a fault, not clear one) | ✓ | ✓ |
+| Node named by `nodeName` exists in the cluster | ✓ | — |
+| `spec.healthEvent` is immutable (the whole event is frozen after creation) | — | ✓ |
+| No other MaintenanceRequest for the same node | ✓ | — |
```
lalitadithya, line 190: I'm not sure about these checks, we could have nodes have MRs starting at different times though right?

docs/designs/051-maintenance-request.md
```diff
@@ -0,0 +1,341 @@
...
+
+**Negative / tradeoffs:**
+- Deleting an MR always emits the clear, even if a child remediation is still in flight — the node un-cordons while the child continues its own lifecycle independently. This is a deliberate simplification; coordinating deletion with in-flight children (waiting for the child, or refusing to clear early) is future work.
+- MRs are **not retained** after completion (deleted to clear), so there is no built-in historical record of completed maintenance — deferred.
+- New component to build and operate (image, chart, RBAC, webhook certs, leader election, deployment), plus an outbound dependency on the platform-connector socket.
+- The MR path depends on pipeline wiring that does not exist purely for MR: a `fault-quarantine` ruleset that matches MR-emitted events, and `fault-remediation` association-label propagation on whatever child CR it creates (only the `CUSTOM` case additionally needs the not-yet-built ERR-producing action). These must land with, or before, MR.
+- The webhook's duplicate and node-existence checks use an informer-backed lister — a best-effort guard, not a transactional uniqueness constraint.
+
+## Alternatives Considered
```
lalitadithya, line 267: can we add an alternative to understand why we can't use KOM to handle this use case? Is it because it is too generic and representing this in CEL would be to complex?

nitz2407:
I think this is overlapping with different already available modules:
- csp-health-monitor
- KOM
- ExternalRemediationRequest (ADR-040)

Is this really needed since we already have overlapping modules, can't we extend them?

docs/designs/049-maintenance-request.md
```diff
@@ -0,0 +1,341 @@
...
+
+Introduce a new CRD, `MaintenanceRequest` (MR), in the existing `nvsentinel.dgxc.nvidia.com` API group. MR is the **entry door**: an external system or human operator creates an MR to tell NVSentinel *"maintenance is incoming for this node — prepare it."*
+
+- Created with a `healthEvent` describing the preparation NVSentinel should perform, and a `startTime` recording when the maintenance window opens.
+- On create, the MR reconciler emits the `healthEvent` **as authored** — the creator chooses `recommendedAction`, and the pipeline routes it to the matching remediation (`CUSTOM` / `external-remediation` produces an ERR; `RESTART_VM` produces a RebootNode; etc.). Provided the pipeline is configured to act on the event (see [Pipeline dependencies](#pipeline-dependencies)), this drives the normal NVSentinel flow: quarantine (cordon), drain, and creation of the appropriate maintenance CR by `fault-remediation`.
+- The MR reconciler watches the resulting child maintenance CR (associated by a label the pipeline stamps). When the child reaches its terminal state (`Status.CompletionTime` — the signal shared by all janitor maintenance CRs), the reconciler **deletes the MR**.
```
natherz97: What if there's a case where an operator wants to reboot a node under a MaintenanceRequest but run some manual verification prior to completing the maintenance? Should we make automatically completing the maintenance from the Janitor CRD completion an optional setting on the CRD? So by default we require an operator to manually complete it (regardless of whether the maintenance results in a Janitor CRD) and we could have an optional bool for autoComplete which will complete it on Janitor CRD completion (if one exists and completed successfully).

natherz97: I left a comment below on Lalit's thread regarding if we should have support for automatic completion for phase 1.

docs/designs/049-maintenance-request.md
```diff
@@ -0,0 +1,341 @@
...
+- Deleting the MR — automatically on completion, or manually by an operator at any time — runs the MR's finalizer, which emits a matching `isHealthy=true` health event to **retract the fault** it raised, letting the normal quarantine-recovery path un-cordon the node. **Deletion is the clear.**
```
natherz97: I'm not aware of any controllers deleting an object they own (similar to how controllers typically don't modify the spec). Would it be cleaner for the controller to signal maintenance completion via the MaintenanceRequest status for the Janitor CRD case while we still allowing manual deletion by operators as an alternative signal for manual completion? Later on if we at an end time, it would use the same status completion signal.

docs/designs/049-maintenance-request.md
natherz97: I think we'll have to do one of the following 2 options if we want to support the automatic completion of maintenance:
1. Make the maintenance-controller aware of the mapping from recommended actions to Janitor CRDs. This would require Janitor CRD RBAC permissions + passing group version kind for these resources to the controller like what JT is already calling out as required.
2. Add support to Janitor for it to write back to it's parent status which would require the inverse for Janitor to have knowledge of MaintenanceRequests + RBAC permission.

All examples of parent/child relationships I'm aware of use the first approach but it's complex since we support a dynamic mapping from recommended actions to CRDs. 

What if we punted on this decision for the initial phase and defer to the maintenance requestor to signal maintenance completion manually? I think for our initial use-case we can require the requestor to delete the MaintenanceRequest after it writes the ERR completion and we can figure out how to automate as a follow-up feature.

lalitadithya: I agree with punting this for the initial phase and we can come back to auto completion later. For initial phase, we can require that the entity deletes the MR if they want the node to be marked healthy.

docs/designs/051-maintenance-request.md
jtschelling: > we could have nodes have MRs starting at different times though right?

Do you mean node X has 2 MR's created at the same time, one starting at T and the other starting at T+1?

docs/designs/049-maintenance-request.md
jtschelling: yeah i think that's fine for now, we'll simplify it and make it so the requestor will have to delete the MR object to send the clearing event

docs/designs/049-maintenance-request.md
jtschelling: yeah it was a bit odd to me too, but i've simplified and removed that logic anyways. we'll keep it as requestor needs to delete the MR for now

tanishagoyal2: Also, we can keep TTL same as we have for janitor CR to automatically clean up the created CR. WDYT?

docs/designs/051-maintenance-request.md
```diff
@@ -0,0 +1,320 @@
...
+### Example
+
+```yaml
+apiVersion: nvsentinel.dgxc.nvidia.com/v1
+kind: MaintenanceRequest
+metadata:
+  name: csp-maintenance-ip-10-0-31-7
+spec:
+  startTime: "2026-05-13T03:00:00Z"
+  healthEvent:
```
tanishagoyal2, line 128: why do we need startTime in spec? does it represent when was the maintenance requested? won't that be equal to generatedTimestamp of CR?

docs/designs/051-maintenance-request.md
```diff
@@ -0,0 +1,320 @@
...
+    isFatal: false
+    isHealthy: false
+    message: "AWS scheduled maintenance 2026-05-13T03:00Z; node must be drained."
+    metadata:
+      cspEventId: evt-0a9bc8e74e2c2c10c
+      source: aws-health
+    nodeName: ip-10-0-31-7.us-west-2.compute.internal
+    generatedTimestamp: "2026-05-13T02:00:00Z"
+    id: he-mst-c6d92aa1-2f6e-4e8b-9e3d-b75f86b1aaaa
+    version: 1
```
tanishagoyal2, line 145: I couldn't understand, Why do we need generatedTimestamp, ID and version in spec.healthevent? event IDs are automatically created whenever we insert event in DB

docs/designs/049-maintenance-request.md
```diff
@@ -0,0 +1,320 @@
...
+**Positive:**
+- Very small first iteration: two emissions, one condition, one finalizer — no cross-CRD watches, owner-references, or GC coupling, and no new code paths in any other component.
+- Simple lifecycle (*present = active, absent = cleared*) with exactly one place the clear can happen, no matter who deletes the MR or why.
+- Drives any pipeline-supported remediation, not just the ERR handoff.
+- The requester controls when the node is returned — correct when the external maintenance window, not the in-cluster remediation, determines "done."
+
+**Negative / tradeoffs:**
+- **A forgotten MR leaves a node cordoned indefinitely.** Nothing reclaims it, so MRs become an operational surface that needs monitoring (e.g. alert on MRs well past their `startTime`). This is the main cost of dropping automatic cleanup.
+- Deleting an MR always emits the clear, even if the remediation it triggered is still in flight — the node un-cordons while that remediation continues independently.
+- MRs are not retained after completion, so there is no built-in historical record of completed maintenance.
```
tanishagoyal2: since we are not automatically deleting the MR now, this point is not valid anymore, right?

docs/designs/051-maintenance-request.md
jtschelling: startTime is a time in the future that you want the health event to fire at. its a pre-scheduling feature

docs/designs/049-maintenance-request.md
jtschelling: not really, all those statements should remain true since the operator is required to delete the MR

https://github.com/jtschelling/NVSentinel/blob/ef895f5cec74a9e9802427578ebe397d25d7d4d2/docs/designs/049-maintenance-request.md?plain=1#L17-L18

docs/designs/049-maintenance-request.md
```diff
@@ -0,0 +1,334 @@
...
+distros/kubernetes/nvsentinel/charts/lifecycle-manager/     (existing chart)
+├── crds/                                   (new — MR CRD)
+└── templates/                              (extended — socket mount, clusterrole, webhook)
+```
+
+The `api/`, `internal/controller/`, and `pkg/webhook/v1alpha1/` paths follow the layout the validation-controller established. How the MR CRD is generated is still open: `lifecycle-manager` generates its CRDs with `controller-gen` from Go types, while `spec.healthEvent` embeds the proto-generated `HealthEvent`. See [CRD schema](#crd-schema).
+
+### CRD schema
+
+> **Open question.** The schema below is proto-generated through `protoc-gen-crd`, matching the ERR pattern. `lifecycle-manager`, however, generates its CRDs with `controller-gen` from hand-written Go types and imports no protos, and `ValidationRequest` embeds no `HealthEvent`. Reconciling the two — and the API group and version that follow from it — is still to be decided.
```
lalitadithya: Given our health event is in proto, I think it would make sense to continue using what we did for extrr, it will also keep it in sync, what do you think?

lalitadithya (commented):
I'm leaning towards keeping this in the lifecycle controller as if we need to add any additional logic with respect to maintenance budgets or something like that in future it would be easier to do so.

docs/designs/051-maintenance-request.md
natherz97: This is a good callout, do we want to allow overlapping maintenance requests for the same node? We'd handle that gracefully and each maintenance request would be tracked as part of the quarantineHealthEvent annotation and we'd require that both are completed prior to uncordoning the node.

If we do want to block overlapping maintenance requests, it might be simpler to handle this in the controller directly to avoid race conditions between the controller + webhook. Since the controller will be single-threaded (and we don't need node-level locking across controllers like in Janitor), we could add an annotation to the node object to indicate whether it has an in-progress maintenance and check this prior to starting a new maintenance.

docs/designs/051-maintenance-request.md
```diff
@@ -0,0 +1,334 @@
...
+1. If the reconciler emitted the opening event, emit the clearing event. It carries `isHealthy=true`, the same `agent`, `checkName`, and `nodeName`, and `recommendedAction=NONE`, because it must clear the check rather than trigger a second remediation. If the reconciler never emitted the opening event, skip this step, because there is no fault to retract.
+2. On success, remove the finalizer. On failure, return an error and retry, so the MR remains until the clear succeeds.
+
+For this first iteration NVSentinel performs **no automatic cleanup**. The requester deletes the MR when it wants the node marked healthy again. Auto-deletion on remediation completion is deliberately deferred (see [Alternatives Considered](#alternatives-considered)).
+
+MR and the remediation it triggers are otherwise fully decoupled: no owner-reference, label, or watch links them. The remediation CR runs its own lifecycle, and its own reconciler or TTL cleans it up (ADR-040 for ERR, ADR-037 for the others).
+
+### Validating admission webhook
+
+| Check | On create | On update |
```
natherz97, line 183: Do we need a check that startTime is after the current time?

docs/designs/051-maintenance-request.md
```diff
@@ -0,0 +1,334 @@
...
+    isHealthy: false
+    message: "AWS scheduled maintenance 2026-05-13T03:00Z; node must be drained."
+    metadata:
+      cspEventId: evt-0a9bc8e74e2c2c10c
+      source: aws-health
+    nodeName: ip-10-0-31-7.us-west-2.compute.internal
+    generatedTimestamp: "2026-05-13T02:00:00Z"
+    id: he-mst-c6d92aa1-2f6e-4e8b-9e3d-b75f86b1aaaa
+    version: 1
+status:
```
natherz97, line 146: Do we need to use a completionTimestamp to signal completion? For example, if the node is deleted after webhook admission but before the first reconcile, we should likely mark the MaintenanceRequest as completed without sending the event or updating the status condition right? The other case is do we want to mark MaintenanceRequests as complete if the node is deleted after HealthEventEmitted is true?

docs/designs/051-maintenance-request.md
jtschelling: yeah that'd be a good idea

docs/designs/051-maintenance-request.md
jtschelling: we decided for now that the requestor needs to delete the MaintenanceRequest to clear the fault, so technically the CR never reaches a terminal state when it exists. im not opposed to adding it in the CRD now but it wont be used as the ADR is written

docs/designs/049-maintenance-request.md
jtschelling: yeah thats good with me, updating

docs/designs/051-maintenance-request.md
jtschelling: added KOM details

jtschelling:
@coderabbit

jtschelling:
@coderabbit review

docs/designs/051-maintenance-request.md
natherz97: I see, in this case we're only checking that if the healthy event has already been checked so checking the HealthEventEmitted condition is equivalent to checking a completionTimestamp.

At a minimum, should we skip emitting the healthy event if the node is deleted when we are removing the finalizer from the MaintenanceRequest?

docs/designs/051-maintenance-request.md
jtschelling: I think we can allow multiple MRs be created for the node, but then have a check in the controller code before the health event is emitted that looks for any non-terminal state MR's for the node. If a MR exists for the node and is in a non-terminal state then we can requeue the secondary MR reconciliation

docs/designs/051-maintenance-request.md
jtschelling: oh I understand what you're saying better now, sorry misunderstood. I don't think there's a downside to reacting the same way regardless if the node actually exists or not, if a health event is published to the platform-connector for a node that doesn't exist it still accepts it, and if the event gets through to fault-quarantine it gets dropped there. IMO we should ignore those cases with deleted nodes and just keep the same behavior for the controller in all cases

lalitadithya (approved):

PR #1582, opened 2026-08-11
fix(platform-connectors): dedupe node events without full-range event LISTs

HarshavardhanK:
## Summary

`K8sConnector.writeNodeEvent` (`platform-connectors/pkg/connectors/kubernetes/process_node_events.go`) deduplicates node events by listing every event for the node:

```go
events, err := r.clientset.CoreV1().Events(DefaultNamespace).List(ctx, metav1.ListOptions{
    FieldSelector: fmt.Sprintf("involvedObject.name=%s", nodeName),
})
```

`involvedObject.name` is not an index-served selector, and core events bypass the apiserver watch cache, so every call makes the apiserver read the **entire** default-namespace events range out of etcd and filter it in memory. The call fires once per unhealthy, non-fatal health event per node.

### Impact

On a healthy fleet this is close to zero traffic. During a fleet-wide health-event storm every affected node's connector issues full-range etcd scans, with conflict retries on top — precisely when etcd is already under incident load and the events range is at its largest. Cost is O(nodes x events-range size), i.e. quadratic in cluster size during correlated failures.

Measured on a 1,152-node production cluster during an incident: **468–557 etcd read objects/s per apiserver instance** attributable to these LISTs, roughly a **500x fetched-to-returned amplification**.

### Fix

Cache the name of the last event written per `(node, type, reason, message)` on the connector, and resolve the dedupe read with a `metadata.name` field selector. The apiserver resolves an exact-name selector as `matchesSingle`, turning it into a single-key storage lookup instead of a range scan.

- **Cache hit** → `LIST fieldSelector=metadata.name=<cached-name>`, then bump `Count` + `LastTimestamp` via `Update` (unchanged from before).
- **Empty result** (event deleted or TTL-expired) → drop the cache entry and `Create` a fresh event.
- **Cache miss** → `Create`, and cache the returned name.

The cache is a 1024-entry LRU (the `golang-lru/v2` dependency this module already uses), so overflow evicts one least-recently-used entry rather than flushing. `retry.OnError` semantics, the `nodeEventOperationsCounter` metrics, and the trace spans are all preserved exactly.

### Why a `metadata.name` LIST and not a GET

This is the part worth flagging for review. The obvious implementation is `Events().Get(ctx, name, ...)`, but the shipped chart's ClusterRole grants `events` only `create`, `update`, `list` — **not `get`**:

https://github.com/NVIDIA/NVSentinel/blob/main/distros/kubernetes/nvsentinel/templates/clusterrole.yaml#L37-L44

```yaml
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - create
  - update
  - list
```

We built a GET-based version first and it failed with `403 Forbidden` on a live cluster running the chart's own RBAC. envtest does not catch this — it runs with admin credentials, so the GET version passes the whole suite. The `metadata.name` LIST costs the same as a GET server-side (single-key lookup, no range scan) and requires zero RBAC migration, which also means no version-skew window between chart and image during upgrade.

If you would rather add `get` to the ClusterRole and use a plain GET, happy to do it that way — but the LIST form avoids the chart/image ordering problem entirely.

### Behavior change

After a connector restart the cache is empty, so the first occurrence of a recurring fault creates a fresh event instead of incrementing the pre-restart one. Subsequent occurrences deduplicate as before. This is a bounded, one-event-per-fault-per-restart cost and is covered by a new test.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: platform-connectors

## Testing

**Existing suite, unchanged:**

```
$ GOTOOLCHAIN=auto go build ./...
$ go test ./pkg/connectors/kubernetes/...
ok  	github.com/nvidia/nvsentinel/platform-connectors/pkg/connectors/kubernetes	53.891s
```

`go vet` clean, `gofmt` clean. The existing `TestK8sConnector_WithEnvtest_EventCountIncrement` already covers the dedupe-and-increment path and passes unmodified.

**New test** — `TestK8sConnector_WithEnvtest_EventDedupeCacheRecovery` covers the two paths where the cache does not resolve to a live event: the cached event was deleted out from under the connector (must create a fresh event at `Count=1`, not error), and a restarted connector with an empty cache (must create a second event rather than silently doing nothing).

```
=== RUN   TestK8sConnector_WithEnvtest_EventCountIncrement
--- PASS: TestK8sConnector_WithEnvtest_EventCountIncrement (3.13s)
=== RUN   TestK8sConnector_WithEnvtest_EventDedupeCacheRecovery
--- PASS: TestK8sConnector_WithEnvtest_EventDedupeCacheRecovery (4.01s)
PASS
```

**Live-cluster validation** — the patched image has been running as a production canary. On a 3-node cluster with the real chart RBAC and synthetic non-fatal Xid injection (`logger -p kern.err` on the node):

- First injection: event `CREATE`, no LIST.
- Identical re-injection: `Count` 1 → 2, with the audited request being `LIST ...?fieldSelector=metadata.name%3D<name> -> 200` — a single-key lookup.
- Connector restart + re-injection: fresh event object created, no errors (the documented restart semantics).
- Zero full-range `involvedObject` LISTs after the patch; zero `ERROR` log lines.

**Note on `make lint`:** I could not run it locally. `golangci-lint v2.5.0` (pinned in `.versions.yaml`) is itself built with go1.25, and the modules declare `go 1.26.0`, so it panics with `package requires newer Go version go1.26 (application built with go1.25)` during type-checking. `latest` (v2.12.2) is also go1.25-built and panics identically. As a proxy I measured the complexity linters directly: `writeNodeEvent` cognitive complexity is unchanged at 16, and cyclomatic complexity drops from 10 to ≤8, so `cyclop`/`gocognit` should not regress. No line exceeds the `lll` limit of 120. CI will be the real check here.

## Checklist
- [x] Self-review completed
- [ ] Documentation updated (if needed) — no user-facing docs affected
- [x] Ready for review

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Kubernetes node health event deduplication to prevent duplicate recurring events.
  - Added recovery handling so missing or deleted events are recreated correctly with accurate counts.
  - Ensured restarted connector instances create distinct events when prior deduplication state is unavailable.
  - Improved reliability when processing large numbers of node health events.
  - Prevented update races from causing errors and preserved accurate event tracking during high-volume activity.
  - Updated Kubernetes permissions to support reliable event recovery and updates.


platform-connectors/pkg/connectors/kubernetes/k8s_platform_connector_test.go
```diff
@@ -2265,3 +2267,67 @@ func TestDeduplicationBehavior(t *testing.T) {
...
+	require.NoError(t, connector.processHealthEvents(localCtx, healthEvents),
+		"A racing deletion should not surface as a write error")
+
+	events, err = localClientSet.CoreV1().Events(DefaultNamespace).List(localCtx, metav1.ListOptions{})
+	require.NoError(t, err)
+	require.Len(t, events.Items, 2, "A racing deletion should be recovered by creating a fresh event")
+
+	cachedName, ok = connector.getCachedNodeEventName(dedupeKey)
+	require.True(t, ok, "The recreated event name should be cached")
+	assert.NotEqual(t, firstName, cachedName, "The stale cache entry should have been replaced")
```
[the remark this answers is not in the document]
HarshavardhanK: @Claude

Addressed in 2a61f675 (crossed with this comment in flight). The reactor now deletes the event from the tracker before returning NotFound, records that the update ran, and the assertions became: update attempted, exactly one replacement event remains, it is not the deleted one, and the dedupe cache holds the replacement name.

-tsarai

XRFXLP:
hey @HarshavardhanK, thanks for taking this up!


Did you consider using `client-go/tools/record.EventRecorder` here? It would eliminate the LIST entirely — deduplication is in-memory with no API calls — whereas this approach still makes one etcd round-trip per event (O(1), which is fine). Curious if there was a specific reason not to use it: e.g. the `involvedObject` structure doesn't fit the recorder's interface, the aggregation behaviour wasn't desirable, or you intentionally chose the more surgical fix to minimise blast radius?

HarshavardhanK:
> hey @HarshavardhanK, thanks for taking this up!
> 
> Did you consider using `client-go/tools/record.EventRecorder` here? It would eliminate the LIST entirely — deduplication is in-memory with no API calls — whereas this approach still makes one etcd round-trip per event (O(1), which is fine). Curious if there was a specific reason not to use it: e.g. the `involvedObject` structure doesn't fit the recorder's interface, the aggregation behaviour wasn't desirable, or you intentionally chose the more surgical fix to minimise blast radius?

Thanks for the response, @XRFXLP! Yes, I did try that. With defaults, `record.EventRecorder` dropped 35 of a 60 event burst and rewrote aggregated messages. `tools/events` merged distinct-message events outright (the note isn't in its dedup key). Our messages are the diagnostic payload and storms are when they matter. So we kept the surgical fix. The remaining read is a single key on dedup hits only.

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -462,6 +462,43 @@ func messageMatchesAnyErrorCode(msg string, errorCodes []string) bool {
...
+	return fmt.Sprintf("%s\x00%s\x00%s\x00%s", nodeName, event.Type, event.Reason, event.Message)
+}
+
+func (r *K8sConnector) getCachedNodeEventName(key string) (string, bool) {
+	r.nodeEventMu.Lock()
+	defer r.nodeEventMu.Unlock()
+
+	name, ok := r.nodeEventNames[key]
+
+	return name, ok
```
XRFXLP: nit: why not just `return r.nodeEventNames[key]`?

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -462,6 +462,43 @@ func messageMatchesAnyErrorCode(msg string, errorCodes []string) bool {
...
+}
+
+func (r *K8sConnector) setCachedNodeEventName(key, name string) {
+	r.nodeEventMu.Lock()
+	defer r.nodeEventMu.Unlock()
+
+	if r.nodeEventNames == nil || len(r.nodeEventNames) >= maxCachedNodeEventNames {
```
XRFXLP: Can we be graceful in cache flush? Although this is not achievable much in practice unless the health events are relayed over a PC running on a CPU node from a health monitor like kubernetes-object-monitor but if we exchaust 1024 unique key limit then all the keys would be cleared at once. Should we have sliding window here or LRU cache thing?

XRFXLP:
@HarshavardhanK by any chance, did you forgot to push the changes? I got a re-review request but there are no new changes.

HarshavardhanK:
> @HarshavardhanK by any chance, did you forgot to push the changes? I got a re-review request but there are no new changes.

Not yet. Must have clicked it by accident. Pushing the changes now!

HarshavardhanK:
The changes for both comments are up now @XRFXLP

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -471,28 +505,39 @@ func (r *K8sConnector) writeNodeEvent(ctx context.Context, event *corev1.Event,
...
-			FieldSelector: fmt.Sprintf("involvedObject.name=%s", nodeName),
-		})
-		if err != nil {
-			return fmt.Errorf("failed to list events for node %s: %w", nodeName, err)
-		}
-
-		// Check if any event matches the new event
+		// An involvedObject LIST is an unindexed full-range etcd scan, so
+		// dedupe via the cached name with a metadata.name selector: a
+		// single-key read that stays within the chart's RBAC (no get verb).
```
lalitadithya: I think it would be okay to add GET permissions so that we avoid a LIST call. @XRFXLP , thoughts?

lalitadithya: ah, I see 

>If you would rather add get to the ClusterRole and use a plain GET, happy to do it that way — but the LIST form avoids the chart/image ordering problem entirely.

in the description, helm will make sure the clusterroles are applied before the pods so I think we should be good with ordering

XRFXLP: yeah, GET is logically correct than LIST even if they've same latency on server-side.

HarshavardhanK: yeah, good callout

lalitadithya:
sorry, accidental close

XRFXLP (approved):

PR #1584, opened 2026-08-13
fix(labeler): detect drivers preinstalled in the node image

### LINKED ISSUE #1583: labeler: driver.installed never set when no driver pod exists (preinstalled/host-managed drivers)
> **RESOLVED — Option A (documentation only).** Maintainer decision: `labeler.assumeDriverInstalled` is the recommended mechanism for host-installed drivers, and no detection fallback will be added — host-installed-driver users are typically advanced deployments that may not run GFD or other GPU Operator operands, so the labeler stays simple and documentation guides the advanced scenario. #1584 therefore stays closed. The docs PR implementing option A is #1586 — it points the manual-labeling guidance at the flag, documents `assumeDriverInstalled` in the Helm configuration reference, and notes that a bare `kubectl label` cannot be relied on to persist. Note: the option-A row below mentions a note in design 018's Manual Labeling Procedure; per maintainer review on #1586, design docs are historical records and are not updated — the user-facing docs are the guidance surface.

### What happened

On platforms where the NVIDIA driver is preinstalled in the node image rather than installed by the GPU Operator, the labeler never applies `nvsentinel.dgxc.nvidia.com/driver.installed=true`. As a result, every DaemonSet that selects on that label gets 0 desired pods and silently does not run.

This is not specific to one cloud. Any configuration where the GPU Operator runs with `driver.enabled=false` is affected, because there is no driver DaemonSet pod for the labeler to observe:

| Platform | GPU Operator `driver.enabled` | driver supplied by |
|---|---|---|
| AKS, "Driver only" node image | false | Azure node image |
| GKE on COS, driver installed at nodepool creation | false | Google, baked into the node — no driver pod exists |
| OKE | false | OKE node image |

A clarification on GKE-COS, since the two install paths behave differently. The GPU Operator never installs the driver on COS, but where Google's standalone `nvidia-driver-installer` DaemonSet supplies it, the labeler's existing driver-pod detection already covers the node — that path is NOT affected. Only the variant where the driver is installed at nodepool creation, leaving no driver pod at all, hits this defect. The discriminator throughout is whether a driver pod exists for the labeler to observe, not who supplied the driver.

Reproduced and diagnosed on AKS (`Standard_ND96isr_H100_v5`, Ubuntu 24.04, Kubernetes 1.35.6) with NVSentinel v1.9.0, where the GPU node pool is created with Azure's "Driver only" preinstall (`gpuProfile.driver: Install`) and the GPU Operator runs with `driver.enabled=false` and `toolkit.enabled=false`. The GKE-COS and OKE rows are from the same GPU Operator configuration, verified from their deployed values.

### Impact

Three DaemonSets never schedule:

```
metadata-collector             0/0   selector: nvidia.com/gpu.present=true,
                                               nvsentinel.dgxc.nvidia.com/driver.installed=true
syslog-health-monitor-regular  0/0   selector: driver.installed=true, kata.enabled=false
syslog-health-monitor-kata     0/0   selector: driver.installed=true, kata.enabled=true
```

GPU health monitoring itself is unaffected (`gpu-health-monitor-dcgm-4.x` is 2/2, since it selects on the DCGM label), so the failure is partial and easy to miss. There is no error, warning, or event — the DaemonSets simply report 0 desired.

### Root cause

The labeler derives driver presence from the existence of a driver pod. Its own startup log states the detection method:

```
"Labeler created, watching DCGM and driver pods, and nodes for kata detection"
```

The reconciliation completes without error and applies the other labels correctly:

```
Setting DCGM version label on node   node=aks-gpuworker1-...  version=4.x
Setting Kata enabled label on node   node=aks-gpuworker1-...  kata=false
Completed initial node label reconciliation   nodeCount=6
```

`driver.installed` is simply never set, because the signal it looks for is "GPU Operator installed a driver" rather than "a driver is present on this node".

### Suggested fix

Treat a host-installed driver as installed. The evidence is already on the node — GPU Feature Discovery publishes the driver version regardless of who installed it:

```
nvidia.com/gpu.present=true
nvidia.com/cuda.driver-version.full=580.159.04
nvidia.com/cuda.driver-version.major=580
```

Falling back to `nvidia.com/cuda.driver-version.*` (or `nvidia.com/gpu.deploy.driver` in combination with GFD labels) when no driver pod is found would cover preinstalled-driver platforms without weakening detection where the operator does install the driver.

`nvidia.com/cuda.driver-version.full` is present on the node in every one of the configurations in the table above, regardless of who installed the driver, so a single fallback covers all of them.

**Important refinement discovered while implementing this (#1584):** the fallback must be gated on the absence of a driver *pod source*, not on the absence of driver *pods*. Gating on "no pod is currently indexed for this node" is not safe, for two reasons:

1. A pod is only indexed once `spec.NodeName` is set, so a replacement DaemonSet pod that is still unscheduled is invisible to the check.
2. `UpdateFunc` (labeler.go, pod event handlers) returns early when `oldReady == newReady`. A replacement pod being bound to a node while still unready is therefore dropped, so no later event re-evaluates the node.

Together those mean that on an operator-managed cluster, deleting the driver pod — which is exactly what the GPU Operator's supported upgrade flow does, since its driver DaemonSet uses `updateStrategy: OnDelete` and the driver image's termination path unloads the driver — can leave `driver.installed=true` standing on stale GFD labels. If the replacement pod then cannot start (ImagePullBackOff, an untolerated taint, insufficient resources), nothing ever corrects it: the only event that would is the readiness transition that never arrives.

Checking for the existence of the driver DaemonSet itself is stable across pod churn and closes all three cases.

### Workaround

Label the GPU nodes directly:

```bash
kubectl label node -l <gpu-node-selector> \
  nvsentinel.dgxc.nvidia.com/driver.installed=true
```

Verified on the affected cluster: `syslog-health-monitor-regular` went 0/0 to 2/2 immediately, and the label survives labeler reconciliation (the labeler is additive, not authoritative). The workaround is lost when a new GPU node joins the pool.

### Environment

- NVSentinel v1.9.0 (commit `30ee12da2bd0247266a10171e2bbd408a98c5b99`)
- AKS 1.35.6, Ubuntu 24.04, 2x H100 nodes, driver 580.159.04, DCGM 4.x
- GPU Operator v26.3.3 with `driver.enabled=false`, `toolkit.enabled=false`

---

## Design 018 already considered this case

Raising this before proposing any implementation, because `docs/designs/018-syslog-monitor-preinstalled-driver-support.md` covers the exact scenario and answers it differently:

- **Lines 52-54** put "custom machine images with pre-baked drivers (no installer pods)" and "GPU Operator with all components disabled" under **Scenarios requiring manual labeling**.
- **Line 62** names the gap in its own Cons: *"Doesn't automatically handle custom images with pre-baked drivers"*.
- **Line 97** is a dedicated **Manual Labeling Procedure**.
- **Alternative Approaches Considered** explicitly **rejects GPU Feature Discovery (GFD) Label Detection**, on two grounds: GFD is optional, and it validates driver *functionality* rather than *installation* — and for syslog monitoring you want the label present even when a driver is broken, so error logs are still captured.

So this is not an oversight — it is a documented scope boundary. Automatic detection would reverse it, which is a maintainer decision rather than a bug fix.

Note `labeler.assumeDriverInstalled` (chart value, `--assume-driver-installed`) is the chart-level automation of that Manual Labeling Procedure. It post-dates design 018 and is not mentioned there.

## Implementation experience, offered as evidence

We did build the automatic-detection version, and the honest result is worth sharing: the fallback itself is small, but the guard that stops it masking a pod-managed driver needed **four successive corrections**, each found in review rather than by us.

| Guard | Missed |
|---|---|
| no driver pod indexed for the node | delete events; unscheduled and stuck replacement pods |
| a driver DaemonSet exists (cluster-wide) | mixed GKE node pools — install mode is per node pool |
| the DaemonSet targets this node (nodeSelector + required affinity) | DaemonSets whose selector value is generated (`app: {{ .Driver.AppName }}` on the NVIDIADriver CRD path) |
| + `app.kubernetes.io/component` matching | — current state |

The pattern is that the guard must model how every vendor deploys driver pods, and each round found another shape modelled wrong. That is the strongest argument for deciding the design question first.

## Options, and what each would require in design 018

| Option | Change to the labeler | Change to design 018 |
|---|---|---|
| **A. Documentation only** | none | Additive only: note in the Manual Labeling Procedure that `labeler.assumeDriverInstalled` automates it. No decision reversed. |
| **B. Node-scoped `assumeDriverInstalled`** | extend the existing flag from a cluster-wide toggle to a node/label selector | Additive only: same section gains a selector-scoped form. The GFD rejection and the "scenarios requiring manual labeling" list both stand, since this is configuration, not inference. |
| **C. Automatic detection** | GFD fallback tier, gated on no pod source targeting the node | Material amendment, or supersede with a new design doc: add Tier 3 to Detection Tiers; move GFD Label Detection from rejected to accepted-as-last-resort with its two objections addressed; shrink "scenarios requiring manual labeling"; update the Cons; record the new invariant that the fallback must never fire while a pod source targets the node. |
| **D. Opt-in automatic detection** | C, behind a flag defaulting off | Narrower amendment: GFD Label Detection moves from *rejected* to *rejected as the default, available opt-in*. Documented default behavior is unchanged. |

Under **C** and **D**, 018's functionality-vs-installation objection does not fully go away and should be recorded rather than dropped: `nvidia.com/cuda.driver-version.full` is NVML-derived, so for the fallback tier the label's meaning shifts from "installed" toward "installed and queryable".

## The question

Is automatic detection wanted at all, given the Manual Labeling Procedure and `assumeDriverInstalled` already cover this case explicitly?

We have a complete implementation of **C** — PR #1584, closed pending this decision, full package tests and lint green at its head, reopenable in one click if the answer is yes. We are equally happy to send the much smaller **B**, or nothing at all if **A** is the preferred answer. We would rather have your decision before anyone spends review time on it.

Downstream note: this does not block us, but to be accurate about the state — our recipes do **not** set `assumeDriverInstalled` yet. The change that would is held pending an internal design decision. What unblocks affected clusters today is the documented per-bundle override (`--set nv-sentinel:labeler.assumeDriverInstalled=true`), which we have verified end to end. So the remedy is a single flag whichever way you answer this issue.

yuanchen8911:
## Summary

The labeler derived driver presence solely from the existence of a driver pod. On platforms where the driver ships in the node image and the GPU Operator runs with `driver.enabled=false` — AKS with the "Driver only" node image, GKE on Container-Optimized OS, and OKE — no driver pod ever exists, so `nvsentinel.dgxc.nvidia.com/driver.installed` was never applied.

Reconciliation completed without error and set the DCGM and kata labels correctly, so the failure was silent and partial: `metadata-collector`, `syslog-health-monitor-regular` and `syslog-health-monitor-kata` all select on that label and sat at 0 desired pods, while `gpu-health-monitor-dcgm-4.x` kept running because it selects on the DCGM label.

This adds a third detection tier to `getDriverLabelForNode`:

1. `nvidia-driver-daemonset` pods (GPU Operator managed) — unchanged
2. `nvidia-driver-installer` pods in `kube-system` (GKE managed) — unchanged
3. **New:** GPU Feature Discovery node labels, when no DaemonSet owning driver pods schedules onto the node and the node has no driver pod

Tier 3 sets `driver.installed=true` when the node carries both `nvidia.com/gpu.present=true` and a non-empty `nvidia.com/cuda.driver-version.full`. GFD reads the driver version through NVML regardless of who installed the driver, so it is the only pod-free signal available on these platforms.

### Why the gate is DaemonSet targeting, not pod presence

Absence from the pod index is not evidence that no pod source exists, and a tier that assumed otherwise would set a stale `driver.installed=true` on nodes whose driver is genuinely gone:

- A driver pod is indexed only once it is bound to a node — `podNodeIndexerByLabel` indexes nothing while `spec.NodeName` is empty.
- The pod `UpdateFunc` returns early unless readiness *changed*, so the scheduler binding a still-unready replacement pod is never observed.
- The GPU Operator's driver DaemonSet uses `updateStrategy: OnDelete`, so an upgrade deletes the driver pod outright, and the image's termination path unloads the driver. The driver really is absent during that window.
- If the replacement never becomes ready — ImagePullBackOff, insufficient resources, a taint it no longer tolerates — no further watched event fires, because the only event that would is the readiness transition that never comes. The stale-`true` window is unbounded, not brief.

A DaemonSet object is unaffected by that pod churn, so gating on it closes all three paths (delete, unscheduled replacement, permanently stuck replacement). A driver pod with no DaemonSet behind it gates the fallback the same way, and the delete-event pod exclusion is deliberately not applied to that check.

### How a driver DaemonSet is recognized

The GPU Operator has two render paths, and only one of them is matchable by a fixed label set:

- `assets/state-driver/0500_daemonset.yaml` (legacy) hardcodes `app: nvidia-driver-daemonset` — the selector the pod informers already watch.
- `manifests/state-driver/0500_daemonset.yaml` (NVIDIADriver CRD) sets **both** `metadata.name` and `spec.selector.matchLabels.app` to the generated app name `nvidia-<driverType>-driver-<os>-<hash>` (`getDriverAppName`, `internal/state/driver.go`). No fixed `app` value matches it.

Both paths label the pod template with `app.kubernetes.io/component` (`nvidia-driver`, or `nvidia-vgpu-host-manager` for the vGPU host manager), so the gate uses that as its second signal. Without it, a cluster on the CRD path would fall through to the GFD fallback whenever no driver pod was indexed — turning the pre-PR fail-closed outcome (label absent) into a fail-open one (label wrongly `true`). Note that tiers 1–2 are already inert on such clusters, since the labeler's pod informer selector does not match generated app names either; that pre-existing gap is untouched here.

### Why targeting is evaluated per node

Driver install mode is not always a cluster-wide choice. On GKE it is selected per node pool: a cluster can run automatic driver installation on one pool while another pool runs Google's manual installer DaemonSet from container-engine-accelerators, whose `requiredDuringSchedulingIgnoredDuringExecution` affinity excludes nodes carrying `cloud.google.com/gke-gpu-driver-version`. Asking only whether a driver DaemonSet exists *somewhere* would let that DaemonSet suppress detection on the automatic-install pool it never schedules onto — leaving exactly the #1583 failure unfixed on GKE.

So the gate asks whether a driver DaemonSet targets *this* node, evaluating the pod template's `nodeSelector` and required `nodeAffinity` against the Node object via `nodeaffinity.GetRequiredNodeAffinity(pod).Match(node)` — the same helper the upstream DaemonSet controller uses (`pkg/controller/daemon/daemon_controller.go`), rather than a hand-rolled term evaluation. This adds `k8s.io/component-helpers` as a direct dependency: one `require` line and two `go.sum` lines, no new transitive modules. Both node lookup failure and an unparseable selector fail closed, treating the driver as pod-managed.

Tier 3 is additionally gated on `allInformersSynced()`: it infers installation from an absence of evidence, so it must not run while caches are still filling. Reconciliation now re-runs when a driver DaemonSet appears or disappears and when a node's GFD driver version label changes. The labeler's ClusterRole gains `get`/`list`/`watch` on `apps/daemonsets`.

### Known limitation

`nvidia.com/cuda.driver-version.full` comes from an NVML read, so by the taxonomy in `docs/designs/018-syslog-monitor-preinstalled-driver-support.md` tier 3 keys on driver *functionality*, not driver *installation*. On a pre-installed-driver platform whose driver is installed but too broken for GFD to query, the label is not set and the syslog monitors that would capture those errors do not schedule; those nodes still need a manual label. Tiers 1 and 2 are unaffected — they key on pod existence and readiness. This is recorded in `docs/labeler.md`.

Design 018 also rejected GFD label detection as a *sole* mechanism because GFD is optional (`gfd.enabled=false`). That objection does not apply to a last-resort tier evaluated only after both pod tiers find nothing: clusters without GFD behave exactly as they do today, and no pod-based signal is displaced. The existing `--assume-driver-installed` flag still works as a stronger, cluster-wide override.

## Motivation / Context

Fixes #1583

Found on a live AKS cluster: `Standard_ND96isr_H100_v5`, Kubernetes 1.35.6, Ubuntu 24.04, 2x H100, driver 580.159.04, DCGM 4.x, GPU Operator v26.3.3 with `driver.enabled=false` and `toolkit.enabled=false`. Manually applying the label took `syslog-health-monitor-regular` from 0/0 to 2/2 immediately, and the label survived labeler reconciliation, confirming the labeler is additive rather than authoritative for this label.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [x] Documentation/CI
- [x] Health Monitors
- [ ] Janitor
- [x] Other: labeler, labeler RBAC

## Testing
- [x] Tests pass locally

`TestGetDriverLabelForNode_HostInstalledDriver_UsesGFDLabels` is a table-driven test over `getDriverLabelForNode` covering both paths and every gate between them:

- ready driver pod, with and without GFD labels → `true` (existing behavior preserved)
- unready driver pod, and unready GKE installer pod, with GFD labels present → no label
- deleted driver pod (delete-event exclusion) with GFD labels present → no label
- driver DaemonSet present with no driver pod indexed for the node — the stuck-replacement case — → no label
- driver DaemonSet rendered by the NVIDIADriver CRD path (generated name *and* generated `app` selector, recognized by the component label), the vGPU host manager variant, and Google's installer DaemonSet on a manual-install node → no label
- generated-selector driver DaemonSet whose `nodeSelector` excludes the node → fallback still fires
- unscheduled driver pod (indexed to no node) with its DaemonSet present → no label
- mixed GKE pools, modelled with Google's official installer affinity: an automatic-install node carrying `cloud.google.com/gke-gpu-driver-version` → fallback still fires
- driver DaemonSet whose `nodeSelector` excludes the node → fallback still fires; the same DaemonSet on a node it targets → no label
- unrelated DaemonSet (`nvidia-dcgm`) present → fallback still fires
- no driver pod or DaemonSet, GFD labels present → `true` (the fix)
- driver version without `gpu.present`, `gpu.present` without driver version, whitespace-only driver version, node absent from the informer cache → no label

`TestGetDriverLabelForNode_CachesNotSynced_DoesNotUseGFDLabels` covers the sync gate in both directions, and `TestNodeRequiresReconciliation_DriverVersionLabelChanged_ReturnsTrue` covers the new reconciliation trigger.

Each gate was verified to be load-bearing by reverting it and re-running:

- Against no gate at all (pre-PR behavior), seven cases fail: `deleted driver pod`, `stuck replacement`, `generated name`, `GKE installer DaemonSet`, `DaemonSet that targets the node`, `unscheduled driver pod`, and the caches-not-synced test.
- Against a cluster-wide DaemonSet gate, exactly the two per-node cases fail: `GKE installer DaemonSet does not gate an automatic-install node` and `driver DaemonSet that does not target the node does not block the fallback`.
- Against fixed-label-set matching only, exactly the two generated-selector cases fail: `driver DaemonSet with generated selector` and `vGPU host manager DaemonSet`.

In both runs every other case passes, so no case is inert.

Verification run locally:

- `go test ./...` in `labeler/` (envtest, full package) — all pass, 506s
- `golangci-lint run -c .golangci.yml ./...` in `labeler/` (pinned v2.5.0) — 0 issues

**Pre-existing issue noted, not addressed here.** The manual-labelling workaround documented for unsupported environments is undone by the next reconciliation, because `updateDriverLabel` deletes the label whenever the computed desired value is empty. That behavior predates this PR — a node with no driver pod already computed an empty desired value on `main` — and this change does not make it worse, so it is left for a follow-up rather than folded in here.

**Not verified at runtime.** The change has not been exercised against the AKS cluster from the issue; doing so would require publishing a multi-arch labeler image to that cluster. Runtime confirmation on an affected platform would be a useful follow-up before release.

## Checklist
- [x] Self-review completed
- [x] Ready for review


labeler/pkg/labeler/labeler_test.go
```diff
@@ -2071,3 +2071,187 @@ func TestUpdateNodeLabelsForPod_ManagedGate(t *testing.T) {
...
+
+	return pod
+}
+
+// TestGetDriverLabelForNode_HostInstalledDriver_UsesGFDLabels covers driver detection when the
+// driver is installed by the node image rather than by a driver pod. On those
+// platforms the GPU Operator runs with driver.enabled=false, so no driver pod
+// ever exists and the labeler must fall back to the driver version GPU Feature
+// Discovery publishes. Pod-based detection must keep its existing precedence.
+func TestGetDriverLabelForNode_HostInstalledDriver_UsesGFDLabels(t *testing.T) {
```
yuanchen8911, line 2200: Addressed. Renamed both tests to the `TestFunctionName_Scenario_ExpectedBehavior` format from `.github/copilot-instructions.md`:

- `TestGetDriverLabelForNodeHostInstalledDriver` → `TestGetDriverLabelForNode_HostInstalledDriver_UsesGFDLabels`
- `TestNodeRequiresReconciliationDriverVersionLabel` → `TestNodeRequiresReconciliation_DriverVersionLabelChanged_ReturnsTrue`

Doc comments updated to match. Both pass: `go test ./pkg/labeler/ -run ...` — 11 subtests green, `gofmt` clean.

yuanchen8911:
Closing this in favour of settling the design question first, in NVIDIA/NVSentinel#1583.

While implementing this, we found that design 018 already covers this deployment shape and answers it differently: pre-baked-driver images and "GPU Operator with all components disabled" are listed under *Scenarios requiring manual labeling*, there is a dedicated Manual Labeling Procedure, and GPU Feature Discovery label detection is explicitly rejected under *Alternative Approaches Considered*. Automatic detection would reverse that decision, which is yours to make rather than something to settle inside a pull request.

The implementation experience is also worth weighing. The GFD fallback itself is small, but the guard that stops it masking a pod-managed driver needed four successive corrections — no-driver-pod-indexed, then a cluster-wide DaemonSet check, then per-node `nodeSelector`/affinity evaluation, then matching DaemonSets whose selector value is generated on the NVIDIADriver CRD path. Each was found in review. The guard has to model how every vendor deploys driver pods, and it was wrong about that four times.

So rather than ask you to review ~750 lines that may not be wanted, #1583 now lays out four options — documentation only, a node-scoped `assumeDriverInstalled`, full automatic detection, or opt-in automatic detection — along with what each would require in design 018.

This branch is complete and green (full labeler package tests under envtest, golangci-lint clean). **Update: #1583 has been decided — documentation only.** `assumeDriverInstalled` is the recommended mechanism for host-installed drivers and no detection fallback will be added, so this PR stays closed permanently; the documentation change implementing that decision is #1586. Nothing downstream is blocked meanwhile — though to be precise, our recipes do not set `assumeDriverInstalled` yet; that change is held pending an internal design decision. Affected clusters are unblocked by the documented per-bundle override (`--set nv-sentinel:labeler.assumeDriverInstalled=true`), verified end to end.

PR #1587, opened 2026-08-13
fix(platform-connectors): recover after GPU replacement

### LINKED ISSUE #1585: [Bug]: Node Condition recovery fails after GPU replacement because GPU_UUID changes
### Prerequisites

### Code of Conduct

### Bug Description

  The Kubernetes platform connector cannot clear an existing GPU Node Condition after the physical GPU is replaced when the healthy event contains the replacement GPU's new UUID.

  A GPU HealthEvent may contain identifiers with different identity scopes:

  - `nodeName`: node identity
  - `GPU`: logical GPU index within the node
  - `PCI`: PCIe location
  - `GPU_UUID`: physical GPU identity

  For example, the original unhealthy event contains:

  ```text
  nodeName: hd02-gpu1-0054
  checkName: GpuPowerWatch
  ErrorCode: DCGM_FR_POWER_UN
  GPU: 0
  PCI: 0000:18:00.0
  GPU_UUID: GPU-old-uuid

  After replacing the GPU module, the logical GPU index and PCI address usually remain unchanged, while the UUID changes:

  GPU: 0
  PCI: 0000:18:00.0
  GPU_UUID: GPU-new-uuid

  The platform connector's scoped Node Condition recovery requires the stored Condition message to match all entities supplied by the healthy event.

  Therefore, a healthy event containing GPU + PCI + GPU_UUID cannot match the historical Condition because it contains the old UUID. The Node Condition remains unhealthy even though
  the faulty physical GPU has been replaced.

  The current operational workaround is to omit GPU_UUID from the healthy event and recover the fault using only the stable slot identifiers:

  GPU + PCI

  However, callers currently have to know which entities to omit, and there is no configurable recovery identity policy.

  Expected behavior:

  The platform connector should support matching recovery events using stable, configurable entity identifiers. For GPU replacement workflows, a recovery matching GPU + PCI should
  clear the historical Node Condition even when the physical GPU_UUID has changed.

  The old and new UUID values should remain available in persisted HealthEvents for hardware replacement auditing, but UUID changes should not prevent recovery when the configured
  stable identity matches.

### Component

Core Service

### Steps to Reproduce

Steps to Reproduce:

  ```markdown
  1. Send an unhealthy GPU HealthEvent:

     ```text
     nodeName: node-1
     componentClass: GPU
     checkName: GpuPowerWatch
     errorCode: DCGM_FR_POWER_UN
     isHealthy: false
     entitiesImpacted:
       - GPU=0
       - PCI=0000:18:00.0
       - GPU_UUID=GPU-old-uuid

  2. Verify that the platform connector creates a Node Condition containing:

     ErrorCode:DCGM_FR_POWER_UN GPU:0 PCI:0000:18:00.0 GPU_UUID:GPU-old-uuid ...

  3. Simulate replacing the physical GPU while keeping the same logical GPU index and PCI address.
  4. Send a healthy event for the replacement GPU:

     nodeName: node-1
     componentClass: GPU
     checkName: GpuPowerWatch
     errorCode: DCGM_FR_POWER_UN
     isHealthy: true
     entitiesImpacted:
       - GPU=0
       - PCI=0000:18:00.0
       - GPU_UUID=GPU-new-uuid

  5. Inspect the Node Condition.
  6. The historical fault message remains because GPU-new-uuid does not match GPU-old-uuid.
  7. Send the same healthy event again, but omit GPU_UUID and include only:

     GPU=0
     PCI=0000:18:00.0

  8. The historical Condition can now be matched and cleared.

### Environment

- NVSentinel version: v1,8.0
- Kubernetes version: v1.34.2
- Deployment method: helm


### Logs/Output

  Historical Node Condition before GPU replacement:

  ```text
  ErrorCode:DCGM_FR_POWER_UN GPU:0 PCI:0000:18:00.0 GPU_UUID:GPU-old-uuid ...

  Healthy event after GPU replacement:

  ErrorCode:DCGM_FR_POWER_UN
  GPU:0
  PCI:0000:18:00.0
  GPU_UUID:GPU-new-uuid
  IsHealthy:true

  Actual result:

  The historical Node Condition remains active because the replacement GPU UUID does not match the UUID recorded in the existing Condition message.

  Operational workaround:


  Introduce a configurable recovery matching policy for Node Conditions.

  For example:

    GPU:
          - GPU
  - Use the configured entities to identify the historical Condition message.
  - Add tests covering physical GPU replacement where the UUID changes but GPU index and PCI location remain stable.

iacker:
## Summary

Fixes #1585.

GPU recovery now ignores `GPU_UUID` when a stable GPU index or PCI address is present. This lets a healthy event clear the old node condition after a physical GPU replacement, while preserving UUID-only and non-GPU matching behavior.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU replacement recovery so historical node-condition messages are cleared when stable GPU or PCI identities match.
  * Preserved GPU UUID matching when no stable hardware identity is available.
  * Ensured non-GPU event handling remains unchanged.
* **Tests**
  * Added coverage for GPU identity handling, replacement recovery, UUID-only events, non-GPU entities, and clearing stale conditions after healthy hardware replacement.


platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -249,6 +249,38 @@ func (r *K8sConnector) aggregateEventMessages(messages []string, events []*proto
...
+// index or PCI address but receives a new UUID, so requiring the UUID would
+// leave the historical node condition active after the hardware is healthy.
+func recoveryEntities(event *protos.HealthEvent) []*protos.Entity {
+	if !strings.EqualFold(event.ComponentClass, "GPU") {
+		return event.EntitiesImpacted
+	}
+
+	hasStableGPUIdentity := false
+
+	for _, entity := range event.EntitiesImpacted {
```
XRFXLP, line 265: Do we need two `for` loops? If we loop over `EntitiesImpacted` just once we can collect all the "stable identities", and on exit it that is not empty then return?

iacker:
yeah, good call, one loop is enough. Done in 2fe000d.

I collect the entities and set `hasStableGPUIdentity` in the same pass, then return the original slice if the flag never got set:

```go
hasStableGPUIdentity := false
entities := make([]*protos.Entity, 0, len(event.EntitiesImpacted))

for _, entity := range event.EntitiesImpacted {
    if strings.EqualFold(entity.EntityType, "GPU") || strings.EqualFold(entity.EntityType, "PCI") {
        hasStableGPUIdentity = true
    }

    if !strings.EqualFold(entity.EntityType, "GPU_UUID") {
        entities = append(entities, entity)
    }
}

if !hasStableGPUIdentity {
    return event.EntitiesImpacted
}

return entities
```

Only cost is we build the slice even when we end up not using it, which felt cheaper than walking the list twice.

`TestRecoveryEntities` still passes on all three cases with `-race`, including the one where the UUID has to stay.

XRFXLP (approved):

PR #1602, opened 2026-08-14
fix(labeler): reconcile device count labels on allocatable changes

cbumb:
## Summary

The labeler's `nodeRequiresReconciliation` only checked for node label changes, but the NIC count CEL expression reads from `node.status.allocatable`. When MOFED driver restarts temporarily zeroed allocatable, the labeler wrote `nic.count.current=0` and never corrected it after recovery because no labels changed, leaving stale labels on 20+ nodes.

This adds `NodeResourcesAffectDeviceCounts` to also trigger reconciliation when `status.allocatable` or `status.capacity` changes, for classes whose CEL expression references `node.status`. Label-only classes (e.g., GPU) pay zero cost.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: Labeler

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **Bug Fixes**
  - Node label reconciliation now responds to relevant allocatable and capacity resource changes.
  - Detects added, removed, and modified resources referenced by device-count configurations while ignoring unrelated changes.
  - Driver pod updates are now reliably incorporated into label detection and reconciliation.
  - Driver and device-count labels can be updated together in a single reconciliation.

- **Tests**
  - Expanded coverage for resource changes, driver pod events, cache synchronization, and combined label updates.


labeler/pkg/devicecounts/device_counts.go
```diff
@@ -577,6 +623,14 @@ func (class compiledClass) referencesResourceSlices() bool {
 	return strings.Contains(class.CurrentExpression, "resourceSlices")
 }
 
+// referencesNodeStatus is a cheap heuristic mirroring referencesResourceSlices.
+// It detects dot-style access (e.g. node.status.allocatable) which covers all
+// shipped and documented CEL expression forms. NodeResourcesAffectDeviceCounts
+// uses this to decide whether allocatable/capacity changes need reconciliation.
+func (class compiledClass) referencesNodeStatus() bool {
```
XRFXLP: Would this be too blunt given that it can also capture `node.status.condition` which is kinda noisy with all the heartbeats?

XRFXLP (changes requested):
Can you please add testing details in the MR on a kubernetes env?

labeler/pkg/devicecounts/device_counts.go
cbumb: Good catch. Updated in the latest commit renamed `referencesNodeStatus()` to `referencesNodeResources()` and narrowed the check to match only `node.status.allocatable` or `node.status.capacity` (the two fields we actually compare), rather than the broad `node.status`. 

A `node.status.conditions`-based expression will no longer trigger allocatable/capacity comparisons on node updates. Added a test case explicitly verifying this.

This also stays consistent with the sibling `referencesResourceSlices()` heuristic which uses the same substring matching approach.

XRFXLP (approved):

PR #1615, opened 2026-08-17
feat(node-drainer): reduce node-drainer informer cache footprint by removing unused fields from node/pod

### LINKED ISSUE #1606: [Feature]: Apply SetTransform to Node Drainer pod and node informers
### Prerequisites

### Code of Conduct

### Feature Summary

Node Drainer holds a cluster-wide pod informer caching every pod regardless of drain eligibility. DaemonSet pods and system-namespace pods that can never be drained consume the majority of that memory. A `SetTransform` hook replaces ineligible pods with identity-only stubs before they enter the cache, cutting pod informer memory by ~60–90%.

### Problem/Use Case

**Pod informer — cluster-wide, no transform**

`NewInformers` creates a cluster-wide `corev1.Pod` informer (`node-drainer/pkg/informers/informers.go:63`). Every pod in the cluster is stored as a full object. In a typical 100k-node GPU cluster with 14 DaemonSet pods per node (1.4M pods), only ~2 drain-eligible pods per node matter for drain decisions.

`isDaemonSetPod` (`informers.go:385`) already filters DaemonSet pods out at drain time by checking `pod.OwnerReferences`. The `systemNamespaces` config already excludes system-namespace pods from eviction (`evaluator.go:192`). Both checks happen in-memory after the full object is fetched from cache — ND caches the full 14 pods/node to use fewer than 2.

**Node informer**

Node Drainer reads only `node.Annotations[QuarantineHealthEventAnnotationKey]` from the node informer (`evaluator.go:563`). All other fields — status, conditions, addresses, images, capacity, spec — are never read.

### Proposed Solution

**Pod informer — SetTransform**

Add a `SetTransform` to the pod informer that intercepts each pod before it enters the cache. Pods in system namespaces or owned by DaemonSets are replaced with minimal identity stubs; drain-eligible pods are stored in full:

```go
func excludedPodTransform(systemNamespacesRegex string) cache.TransformFunc {
    return func(obj interface{}) (interface{}, error) {
        pod := obj.(*corev1.Pod)
        if matchesSystemNamespace(pod.Namespace, systemNamespacesRegex) ||
            isDaemonSetOwned(pod.OwnerReferences) {
            return &corev1.Pod{
                ObjectMeta: metav1.ObjectMeta{
                    Name: pod.Name, Namespace: pod.Namespace,
                    UID: pod.UID, ResourceVersion: pod.ResourceVersion,
                },
            }, nil
        }
        return pod, nil
    }
}
```

The `systemNamespaces` config field already exists — pass it into the transform at construction time. Clear `NodeName` on stubs so excluded pods do not appear in the `spec.nodeName` index.

**Node informer — SetTransform**

Strip the node object down to annotations only:

```go
Transform: func(obj interface{}) (interface{}, error) {
    node := obj.(*corev1.Node)
    return &corev1.Node{
        ObjectMeta: metav1.ObjectMeta{
            Name: node.Name, ResourceVersion: node.ResourceVersion,
            Annotations: node.Annotations,
        },
    }, nil
},
```

**Expected improvement**: Pod informer drops from ~43 MB/1k pods to ~2–3 MB/1k drain-eligible pods. At 100k nodes with 14 pods/node: from ~58 GiB → ~2–3 GiB. Node informer: ~43 MB/1k → ~5 MB/1k.

### Component

Fault Management

XRFXLP:
## Summary

Reduces the Node Drainer node and pod informer caches to the minimum fields required for drain decisions.

#### Nodes
| Kept | Dropped |
|---|---|
| `name`, `uid`, `resourceVersion` | all labels |
| `annotations[quarantineHealthEventAnnotationKey]` | all other annotations |
| — | `spec` (taints, unschedulable, …) |
| — | `status` (conditions, capacity, images, nodeInfo, …) |

#### System-namespace pods + DaemonSet-owned pods
Replaced with a bare identity object: `name`, `namespace`, `uid`, `resourceVersion` only. These pods are not drain targets so no further fields are needed.

#### Drain-eligible pods (non-system-namespace, non-DaemonSet)
| Kept | Dropped |
|---|---|
| `name`, `namespace`, `uid`, `resourceVersion` | all labels |
| `annotations[PodDeviceAnnotationName]` (GPU UUID, for partial drain) | all other annotations |
| `ownerReferences[*].kind` | all other owner fields |
| `deletionTimestamp` | — |
| `spec.nodeName` | all other spec fields |
| `spec.terminationGracePeriodSeconds` | — |
| `spec.containers[*].resources.limits` | container names, images, env, ports, … |
| `spec.initContainers[*].resources.limits` | — |
| `status.phase` | all other status fields |
| `status.conditions` where `type == Ready` (Type, Status, LastTransitionTime) | all other conditions |

### Memory impact at 99k nodes + 299k pods (scale benchmark)

| Build | Settled working set |
|---|---|
| v1.18.0 — no transforms | 8,270 MiB (8.1 GiB) |
| **This PR** | **744 MiB (0.73 GiB)** |
| **Reduction** | **11.1×** |

> Measured on a KWOK-dominated cluster (99,373 KWOK nodes + 5 EC2 nodes, 299k pods). Peak before GC was 1,540 MiB; settled value after GC is 744 MiB. Production nodes carry significantly more labels, annotations, and status conditions — the absolute saving on real hardware is expected to be larger.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Performance**
  * Reduced cached workload data to retain only information needed for node draining and indexing.
  * Preserved essential node health and identity details for evaluation and event handling.

* **Configuration**
  * System namespace exclusion patterns are now applied consistently during workload processing.
  * DaemonSet-managed workloads receive streamlined handling.

* **Reliability**
  * Improved validation and error handling for namespace patterns and workload transformations.
  * Added coverage for filtering, indexing, node events, and workload metadata preservation.


node-drainer/pkg/informers/informers.go
```diff
@@ -111,6 +125,153 @@ func NewInformers(clientset kubernetes.Interface, resyncPeriod time.Duration,
...
+			InitContainers:                cacheContainers(pod.Spec.InitContainers),
+		},
+		Status: v1.PodStatus{
+			Phase:      pod.Status.Phase,
+			Conditions: cachePodReadyConditions(pod.Status.Conditions),
+		},
+	}
+}
+
+func cacheContainers(containers []v1.Container) []v1.Container {
```
lalitadithya: nit: a better name could be


```suggestion
func trimContainers(containers []v1.Container) []v1.Container {
```

node-drainer/pkg/informers/informers.go
```diff
@@ -111,6 +125,153 @@ func NewInformers(clientset kubernetes.Interface, resyncPeriod time.Duration,
...
+			limits[resourceName] = quantity.DeepCopy()
+		}
+
+		cached[idx].Resources.Limits = limits
+	}
+
+	return cached
+}
+
+func cachePodReadyConditions(conditions []v1.PodCondition) []v1.PodCondition {
```
lalitadithya: nit:


```suggestion
func trimPodReadyConditions(conditions []v1.PodCondition) []v1.PodCondition {
```

lalitadithya:
> Replaced with a bare identity object: name, namespace, uid, resourceVersion only. These pods are not drain targets so no further fields are needed.

Shouldn't these be filtered out using a filter function?

XRFXLP:
> Shouldn't these be filtered out using a filter function?

We don't get ready-made filter function, would it be okay to use custom `ListerWatcher` for that? It would add lot of additional complexity though.

lalitadithya:
> We don't get ready-made filter function, would it be okay to use custom ListerWatcher for that? It would add lot of additional complexity though.

I was thinking about https://pkg.go.dev/k8s.io/client-go/tools/cache#FilteringResourceEventHandler, but it looks like that won't help us since that doesn't do what we need for memory savings.

lalitadithya (approved):

PR #1618, opened 2026-08-17
feat(mongodb-store): expose HealthEvents TTL and PVC size as Helm values

### LINKED ISSUE #1601: [Feature]: Expose HealthEvents TTL and MongoDB PVC size as Helm values
### Prerequisites

### Code of Conduct

### Feature Summary

The HealthEvents collection TTL and MongoDB PVC size are both hardcoded in the chart. At production write rates, the defaults exhaust storage in days and the TTL window is far longer than necessary. Neither can be changed without modifying the chart source.

### Problem/Use Case

**TTL**: `MONGODB_COLLECTION_EXPIRY_SECONDS` is hardcoded to `2592000` (30 days) in `charts/mongodb-store/templates/configmap.yaml`. At 500 events/s with 350 bytes/doc, a 30-day TTL requires ~454 GB of storage. Reducing TTL to 7 days cuts this by 4.3× and proportionally reduces HEA query latency (P50 scales with collection size). Operators have no way to change this without patching the chart.

**PVC**: `storage: "8Gi"` is hardcoded in `charts/mongodb-store/values.yaml`. This is sized for development only. At any production write rate it exhausts in hours.

### Proposed Solution

Expose both as Helm values with sensible defaults:

```yaml
# mongodb-store/values.yaml
collectionExpirySeconds: 2592000  # 30 days default, operators should reduce to 604800 (7 days)
persistence:
  size: "8Gi"  # operators must size per fleet write rate
```

Wire `collectionExpirySeconds` into `configmap.yaml` and `persistence.size` into the PVC spec.

**Expected improvement**: Operators can reduce TTL to 7 days (4.3× smaller collection, proportional HEA latency reduction) and provision PVC appropriate for their fleet size — without touching chart source.

### Component

Core Service

rupalis-nv:
## Summary

Fixes #1601. HealthEvents TTL and MongoDB PVC size were hardcoded; this exposes them as Helm values so operators can set them without editing chart source.

**TTL** (`mongodb-store.collectionExpirySeconds`, default `2592000` / 30 days):
- Applied to in-cluster MongoDB and external Mongo (`nvsentinel.collectionExpirySeconds`).
- Nil/empty falls back to `2592000` before `int` (MongoDB `expireAfterSeconds: 0` deletes immediately).
- Validated as an integer in `0..2147483647` (MongoDB int32); out of range fails Helm render.
- Changing TTL creates Job `create-mongodb-database-<seconds>` so Helm/ArgoCD run `collMod` instead of patching a completed Job. Label: `app.kubernetes.io/name=create-mongodb-database`.

**PVC size** (default `8Gi`):
- Bitnami: `mongodb-store.mongodb.persistence.size`
- Percona: `mongodb-store.psmdb-db.replsets.rs0.volumeSpec.pvc.resources.requests.storage`
- First-party `templates/pv.yaml` reads the same keys. Vendored Bitnami/Percona templates are not patched (they would be overwritten on chart bump).

**Upgrade behavior:**
- TTL-only change: `helm upgrade` / Argo sync is enough.
- PVC size change: StatefulSet `volumeClaimTemplates` is immutable, so Helm upgrade fails if the live size differs. Keep the size the same on upgrade, or delete the app (and PVCs) and reinstall. Expand-in-place needs StorageClass expansion plus an orphan STS recreate; shrink is not supported.

### Helm CLI (osmogroot-dgxc-k8s-dev1)

Fresh install of this chart with TTL `604800` (7 days) and PVC `12Gi`:

```bash
helm upgrade --install nvsentinel distros/kubernetes/nvsentinel \
  --namespace nvsentinel \
  --set mongodb-store.collectionExpirySeconds=604800 \
  --set mongodb-store.mongodb.persistence.size=12Gi
```

Verified:
- ConfigMap `mongodb-config`: `MONGODB_COLLECTION_EXPIRY_SECONDS=604800`
- Job `create-mongodb-database-604800` Complete
- PVCs `datadir-mongodb-{0,1,2}` Bound at `12Gi`
- StatefulSet `mongodb` 3/3 with `volumeClaimTemplates` `12Gi`

### Argo CD (same cluster)

Recreated/synced Application `nvsentinel` with cluster-spec overlay `collectionExpirySeconds: 604800` and `persistenceSize: "12Gi"`.

Verified:
- Application **Synced / Healthy**
- TTL ConfigMap `604800`
- PVCs remain `12Gi` (same size as the live STS, so sync did not hit the immutable VCT error)

### Upgrade notes from this test
- Same PVC size on upgrade/sync: no issue.
- Different PVC size on upgrade: Helm/Argo fail; delete the app (and PVCs) and reinstall.
- TTL change on upgrade: new Job name applies without deleting the app.
## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: mongodb-store Helm chart

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * MongoDB HealthEvents retention is configurable, defaults to 30 days, preserves zero, and updates existing TTL indexes when changed.
  * Initialization Jobs now use generated names that reflect configuration and script changes.
  * MongoDB storage sizing is configurable across supported deployment options.

* **Bug Fixes**
  * Improved discovery, validation, cleanup, and rerun handling for initialization Jobs.

* **Documentation**
  * Added guidance for retention settings, storage sizing, external MongoDB, Job identification, and migration troubleshooting.


distros/kubernetes/nvsentinel/charts/mongodb-store/values.yaml
```diff
@@ -31,6 +31,28 @@
 useBitnami: true
 usePerconaOperator: false
 
+# HealthEvents / MaintenanceEvents TTL index expireAfterSeconds.
+# Default 2592000 = 30 days. Operators should reduce for production write rates
+# (e.g. 604800 = 7 days) to shrink collection size and HEA query latency.
+collectionExpirySeconds: 2592000
```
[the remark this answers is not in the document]
rupalis-nv: Fixed

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/configmap.yaml
```diff
@@ -48,4 +48,4 @@ data:
   MONGODB_CHANGE_STREAM_RETRY_DEADLINE_SECONDS: "300"
   MONGODB_CHANGE_STREAM_RETRY_INTERVAL_SECONDS: "5"
   UNPROCESSED_EVENTS_METRIC_UPDATE_INTERVAL_SECONDS: "25"
-  MONGODB_COLLECTION_EXPIRY_SECONDS: "2592000"
+  MONGODB_COLLECTION_EXPIRY_SECONDS: {{ printf "%d" (.Values.collectionExpirySeconds | int) | quote }}
```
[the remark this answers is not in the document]
rupalis-nv: Fixed

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
```diff
@@ -1,3 +1,10 @@
+{{/*
+MongoDB PVC / manual-PV size. Prefers persistence.size, then global.persistenceSize.
+*/}}
+{{- define "mongodb-store.persistenceSize" -}}
+{{- coalesce .Values.persistence.size .Values.global.persistenceSize "8Gi" -}}
```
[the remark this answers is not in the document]
rupalis-nv: Fixed

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
```diff
@@ -1,3 +1,24 @@
...
+Validated HealthEvents/MaintenanceEvents TTL expireAfterSeconds.
+Accepts integers from 0 through 2147483647 (MongoDB int32); fails render otherwise.
+*/}}
+{{- define "mongodb-store.collectionExpirySeconds" -}}
+{{- $raw := .Values.collectionExpirySeconds -}}
+{{- $v := int $raw -}}
+{{- if or (lt $v 0) (gt $v 2147483647) -}}
+{{- fail (printf "mongodb-store.collectionExpirySeconds must be an integer from 0 through 2147483647, got %v" $raw) -}}
+{{- end -}}
+{{- $v -}}
```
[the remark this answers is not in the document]
rupalis-nv: mongodb-store.collectionExpirySeconds already ends correctly: line 18 closes the if, line 20 closes the define. Helm render with 604800 succeeds. No change needed.

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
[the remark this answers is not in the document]
rupalis-nv: @coderabbitai Please review

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
```diff
@@ -58,6 +58,10 @@ metadata:
   name: create-mongodb-database
   annotations:
     argocd.argoproj.io/sync-options: Force=true,Replace=true
+    # Re-run on install/upgrade so collectionExpirySeconds changes apply to existing TTL indexes.
+    helm.sh/hook: post-install,post-upgrade
```
lalitadithya: @XRFXLP , AFAIK I remember ArgoCD has issues with deleting application that make use of this helm hook. I think this was either related to Skyhook/GPU Operator back when we did a workaround for this in the controller that did the application install. Do we know if that is still the case?

XRFXLP: Last I checked that was still the case even though version was upgraded to post-fix.

XRFXLP:
persistenceSize is duplicated in 5 places (global.persistenceSize, persistence.size, mongodb.persistence.size, the psmdb-db storage value, and all 4 again in values-full.yaml). Three are commented "keep equal to global.persistenceSize" — which just means someone has to remember to update multiple copies by hand. Since global.persistenceSize always has a default, the fallbacks never actually get used today, so they're just dead config waiting to drift. Can we drop the nested copies and let global.persistenceSize be the only source of truth?

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
```diff
@@ -1,3 +1,24 @@
...
+{{- .Values.global.persistenceSize | default "8Gi" -}}
+{{- end }}
+
+{{/*
+Validated HealthEvents/MaintenanceEvents TTL expireAfterSeconds.
+Accepts integers from 0 through 2147483647 (MongoDB int32); fails render otherwise.
+*/}}
+{{- define "mongodb-store.collectionExpirySeconds" -}}
+{{- $raw := .Values.collectionExpirySeconds -}}
+{{- $v := int $raw -}}
```
KaivalyaMDabhadkar, line 29: int turns nil into 0 and expireAfterSeconds: 0 means mongo deletes everything immediately. Can we check for nil before the cast or default to 2592000 first?

distros/kubernetes/nvsentinel/charts/mongodb-store/charts/mongodb/templates/replicaset/statefulset.yaml
```diff
@@ -546,7 +546,7 @@ spec:
         {{- end }}
         resources:
           requests:
-            storage: {{ .Values.persistence.size | quote }}
+            storage: {{ .Values.global.persistenceSize | default "8Gi" | quote }}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Also, this drops .Values.persistence.size with no fallback, so mongodb.persistence.size is now silently ignored. Rendering with --set mongodb.persistence.size=100Gi gives an 8Gi PVC and no warning for example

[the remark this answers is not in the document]
KaivalyaMDabhadkar: Could we fall back instead? `.Values.global.persistenceSize | default .Values.persistence.size | default "8Gi"`
Same for standalone/dep-sts.yaml:492 and standalone/pvc.yaml:29

distros/kubernetes/nvsentinel/charts/mongodb-store/charts/psmdb-db/templates/cluster.yaml
```diff
@@ -253,7 +253,11 @@ spec:
         {{- end }}
       {{- else if $replset.volumeSpec.pvc }}
       persistentVolumeClaim:
-{{ $replset.volumeSpec.pvc | toYaml | indent 8 }}
+{{- $pvc := deepCopy $replset.volumeSpec.pvc }}
+{{- if not $pvc.resources }}{{- $_ := set $pvc "resources" (dict) -}}{{- end }}
+{{- if not $pvc.resources.requests }}{{- $_ := set $pvc.resources "requests" (dict) -}}{{- end }}
+{{- $_ := set $pvc.resources.requests "storage" ($.Values.global.persistenceSize | default "8Gi") }}
```
KaivalyaMDabhadkar: This overwrites a user-set volumeSpec.pvc.resources.requests.storage unconditionally. Could we only set storage when the user hasnt? If the override stays, these 4 lines collapse into one mergeOverwrite and the requests: {} stub in values.yaml can go.

distros/kubernetes/nvsentinel/values-full.yaml
```diff
@@ -1464,12 +1464,28 @@ janitor:
 # Required by: fault-quarantine, node-drainer, fault-remediation
 ################################################################################
 mongodb-store:
+  # HealthEvents TTL (expireAfterSeconds). Default 30 days; reduce for production
+  # write rates (e.g. 604800 = 7 days) to limit collection growth / HEA latency.
+  # Valid range: 0–2147483647 (invalid values fail Helm render / schema validation).
+  collectionExpirySeconds: 2592000
```
KaivalyaMDabhadkar: This setting has no effect if you're using an external mongo. In that mode the mongodb-store subchart isnt deployed at all, so nothing reads the value. The external path has its own copy of the config in templates/configmap-datastore.yaml:180, where the TTL is still hardcoded to 2592000. That path also still uses plain createIndex in job-external-mongodb-setup.yaml, so it keeps the IndexOptionsConflict problem that this PR fixes for the internal job. I think a simple fix is a note here saying this is for the internal mongodb only

KaivalyaMDabhadkar:
I don't think we should edit the Bitnami and Percona chart files at all for this PR.

The copies of the Bitnami and Percona charts checked into the repo under charts/mongodb-store/charts came from upstream, and the way we update them is to replace the whole folder with a newer release. This PR edits four template files inside those copies so they read the new global.persistenceSize. The next time someone replaces the folder, those edits are gone and the setting quietly stops working. Nothing marks the files as modified, and no test would catch it, because our default and the upstream fallback are both 8Gi, so the rendered output looks identical with or without the edits.

We can avoid this as both charts already support setting the volume size through values. I rendered the chart from main, with completely untouched templates:
```
mongodb.persistence.size=100Gi                                        -> storage: 100Gi  (replicaset and standalone)
psmdb-db.replsets.rs0.volumeSpec.pvc.resources.requests.storage=100Gi -> storage: 100Gi  (percona)
```

Both work today with no code change. They're also the exact settings our docs, the percona migration runbook, and values-tilt.yaml already use, and this PR breaks all of them: on this branch those settings render 8Gi with no warning, and for bitnami an existing install that upgrades with a custom size hits the StatefulSet immutability error and the upgrade fails outright.

The values route also covers more than the current patch does. With global.persistenceSize=100Gi, bitnami hidden nodes still render 8Gi and percona nonvoting members 3Gi, because only the main data volumes were patched. The native settings handle those too, I checked both: mongodb.hidden.persistence.size and the nonvoting volumeSpec come through fine on unmodified templates.

I know a single shared key was suggested in the comments, but I don't think we should treat that as a requirement here. useBitnami and usePerconaOperator are mutually exclusive, so any given operator only ever sets one key anyway. And if we want the chart's own defaults tied together in one place, a YAML anchor in values.yaml does that:
```
persistenceSize: &persistenceSize "8Gi"
mongodb:
  persistence:
    size: *persistenceSize
psmdb-db:
  replsets:
    rs0:
      volumeSpec:
        pvc:
          resources:
            requests:
              storage: *persistenceSize
```
So my suggestion is to drop the four vendored template edits and the global.persistenceSize key, set the sizes through the charts' own values instead, and document those two keys. The templates/pv.yaml change is our own file and templating it is exactly right, keep that. The TTL half of the PR is all first-party and unaffected by any of this.

That one change also resolves several other issues in one go: existing installs keep upgrading cleanly, our docs and tilt config stay correct, hidden and non-voting volumes get covered, and there's nothing to lose in the next chart bump

distros/kubernetes/nvsentinel/values-full.yaml
KaivalyaMDabhadkar, line 1516: this setting only works when we deploy mongo ourselves. If someone runs with an external mongo (global.mongodbStore.enabled: false with global.datastore.provider: mongodb), the mongodb-store chart isn't deployed at all, so nothing ever reads this value. The external path has its own copy of this config in templates/configmap-datastore.yaml (line 180), and the TTL there is still hardcoded to 2592000.

The external setup job (job-external-mongodb-setup.yaml) also still uses plain createIndex, so it keeps the exact problem this PR fixes for the internal job: if the TTL ever differs from what's already on the index, the job fails with IndexOptionsConflict instead of updating it.

Since this comment block is where people will learn about the knob, we should either wire the external path to the same value and give it the same ensureTTL logic or add a line here saying it only applies to the internal mongo.

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
```diff
@@ -57,7 +57,14 @@ kind: Job
...
   name: create-mongodb-database
   annotations:
+    # ArgoCD: Force/Replace recreates this immutable Job when the template
+    # changes (e.g. checksum/collection-expiry). Avoid helm.sh/hook here —
+    # Helm hooks have caused ArgoCD Application delete/sync stuck states
+    # (hook-finalizer / BeforeHookCreation). Match live nvsentinel pattern.
     argocd.argoproj.io/sync-options: Force=true,Replace=true
+    # Changes when collectionExpirySeconds changes so ArgoCD/Helm see a diff
+    # and Replace recreates the Job to collMod TTL indexes.
+    checksum/collection-expiry: {{ include "mongodb-store.collectionExpirySeconds" . | toString | sha256sum }}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Adding the other half of this: plain helm users have the opposite problem. Changing collectionExpirySeconds only changes this annotation, since the value reaches the pod through envFrom and the pod spec stays identical. Annotations on a finished Job get patched in place, so helm upgrade succeeds, the job never re-runs, and the collMod never happens. The ConfigMap shows the new TTL while mongo keeps the old one, and nothing else in the system reconciles it. The only workaround (delete the job before upgrading) is written in this chart's values.yaml but not in values-full.yaml, which is the file operators actually edit.

One change fixes both this and the every-sync recreation: put the checksum in the job's name and add ttlSecondsAfterFinished. A TTL change then makes a new Job name, which Helm and argo  both handle as a fresh create, and force/replace stops being needed for this. The fixed name is referenced in scripts/validate-nvsentinel.sh, scripts/mongodb-migration/verify.sh, and the tiltfile, so those would need to find it by label instead.

distros/kubernetes/nvsentinel/charts/mongodb-store/charts/mongodb/templates/replicaset/statefulset.yaml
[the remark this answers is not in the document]
rupalis-nv: No longer applicable. We reverted all Bitnami/Percona template edits, including this StatefulSet. PVC size is set with mongodb.persistence.size (and Percona volumeSpec) on unmodified upstream templates, so there is no global.persistenceSize and no duplicate storage key from our patch.

distros/kubernetes/nvsentinel/charts/mongodb-store/values.schema.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "collectionExpirySeconds": {
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 2147483647,
+      "title": "HealthEvents / MaintenanceEvents TTL expireAfterSeconds",
+      "description": "MongoDB TTL index expireAfterSeconds for HealthEvents.createdAt and MaintenanceEvents.actualEndTime. Must be an integer from 0 through 2147483647 (int32). Unset/null defaults to 2592000; do not render as 0 accidentally (MongoDB would delete documents immediately).",
```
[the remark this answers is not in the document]
rupalis-nv: Dropped values.schema.json. Nil/empty TTL defaults to 2592000 and the 0–2147483647 range is enforced in the Helm helper, including when mongodb-store is not installed (external Mongo).

distros/kubernetes/nvsentinel/templates/job-external-mongodb-setup.yaml
```diff
@@ -63,14 +63,16 @@ inside the mongodb-store subchart instead.
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ .Release.Name }}-external-mongodb-setup
+  name: {{ printf "%s-external-mongodb-setup-%s" .Release.Name (include "nvsentinel.collectionExpirySeconds" . | toString) | trunc 63 | trimSuffix "-" }}
```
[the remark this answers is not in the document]
rupalis-nv: Acknowledged. trunc 63 can drop the TTL suffix only if the release name is very long. Default nvsentinel is fine (nvsentinel-external-mongodb-setup-2592000 is 42 chars). Out of scope for this PR; we can follow up if a long release name becomes a real install path.

docs/configuration/mongodb-store.md
```diff
@@ -123,6 +123,38 @@ global:
...
+
+`mongodb-store.collectionExpirySeconds` is MongoDB `expireAfterSeconds` for HealthEvents (`createdAt`) and MaintenanceEvents (`actualEndTime`). Default `2592000` (30 days). Unset/null falls back to that default; do not leave the value missing in a way that renders as `0` — MongoDB treats `expireAfterSeconds: 0` as “delete immediately”.
+
+The same key applies to **external** MongoDB (`global.mongodbStore.enabled: false` with `global.datastore.provider: mongodb`). The mongodb-store chart is not deployed in that mode, but parent templates still read this value.
+
+Changing the TTL creates a new setup Job named `create-mongodb-database-<seconds>` (external: `<release>-external-mongodb-setup-<seconds>`). Helm and ArgoCD create the new Job; they do not patch a completed Job in place. Find it with:
+
+```bash
+kubectl get job -n <namespace> -l app.kubernetes.io/name=create-mongodb-database
+```
```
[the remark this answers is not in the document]
rupalis-nv: Intentional. That selector is for the in-cluster Job (create-mongodb-database-<seconds>). External Mongo uses a different Job (<release>-external-mongodb-setup-<seconds>, label app.kubernetes.io/name=external-mongodb-setup), documented in docs/external-datastore.md.

tilt/Tiltfile
```diff
@@ -229,6 +229,7 @@ k8s_yaml(yaml)
 
 # Check if using Percona operator (only relevant when using MongoDB)
 use_percona = effective_values.get('mongodb-store', {}).get('usePerconaOperator', False) if not use_postgresql else False
+mongodb_init_job = 'create-mongodb-database-%s' % effective_values.get('mongodb-store', {}).get('collectionExpirySeconds', 2592000)
```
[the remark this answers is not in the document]
rupalis-nv: Tilt either omits the key (.get(..., 2592000) already defaults) or sets an integer. We do not set YAML null in Tilt values. Helm still maps nil/empty to 2592000. Leaving this as-is.

distros/kubernetes/nvsentinel/templates/job-external-mongodb-setup.yaml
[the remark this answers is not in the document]
rupalis-nv: no

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
rupalis-nv: Agreed — we dropped helm.sh/hook from the in-cluster setup Job.

TTL changes now create a new Job named create-mongodb-database-<seconds>, with ttlSecondsAfterFinished: 86400. Helm/Argo therefore create a new Job instead of patching a completed one, without a helm hook.

The external Mongo Job still has the hooks it already had on main; this PR did not add new ones there.

rupalis-nv:
> I don't think we should edit the Bitnami and Percona chart files at all for this PR.
> 
> The copies of the Bitnami and Percona charts checked into the repo under charts/mongodb-store/charts came from upstream, and the way we update them is to replace the whole folder with a newer release. This PR edits four template files inside those copies so they read the new global.persistenceSize. The next time someone replaces the folder, those edits are gone and the setting quietly stops working. Nothing marks the files as modified, and no test would catch it, because our default and the upstream fallback are both 8Gi, so the rendered output looks identical with or without the edits.
> 
> We can avoid this as both charts already support setting the volume size through values. I rendered the chart from main, with completely untouched templates:
> 
> ```
> mongodb.persistence.size=100Gi                                        -> storage: 100Gi  (replicaset and standalone)
> psmdb-db.replsets.rs0.volumeSpec.pvc.resources.requests.storage=100Gi -> storage: 100Gi  (percona)
> ```
> 
> Both work today with no code change. They're also the exact settings our docs, the percona migration runbook, and values-tilt.yaml already use, and this PR breaks all of them: on this branch those settings render 8Gi with no warning, and for bitnami an existing install that upgrades with a custom size hits the StatefulSet immutability error and the upgrade fails outright.
> 
> The values route also covers more than the current patch does. With global.persistenceSize=100Gi, bitnami hidden nodes still render 8Gi and percona nonvoting members 3Gi, because only the main data volumes were patched. The native settings handle those too, I checked both: mongodb.hidden.persistence.size and the nonvoting volumeSpec come through fine on unmodified templates.
> 
> I know a single shared key was suggested in the comments, but I don't think we should treat that as a requirement here. useBitnami and usePerconaOperator are mutually exclusive, so any given operator only ever sets one key anyway. And if we want the chart's own defaults tied together in one place, a YAML anchor in values.yaml does that:
> 
> ```
> persistenceSize: &persistenceSize "8Gi"
> mongodb:
>   persistence:
>     size: *persistenceSize
> psmdb-db:
>   replsets:
>     rs0:
>       volumeSpec:
>         pvc:
>           resources:
>             requests:
>               storage: *persistenceSize
> ```
> 
> So my suggestion is to drop the four vendored template edits and the global.persistenceSize key, set the sizes through the charts' own values instead, and document those two keys. The templates/pv.yaml change is our own file and templating it is exactly right, keep that. The TTL half of the PR is all first-party and unaffected by any of this.
> 
> That one change also resolves several other issues in one go: existing installs keep upgrading cleanly, our docs and tilt config stay correct, hidden and non-voting volumes get covered, and there's nothing to lose in the next chart bump

@KaivalyaMDabhadkar Done. Reverted the four vendored Bitnami/Percona template edits and dropped global.persistenceSize.

PVC size is set only through the charts’ own values (mongodb.persistence.size / psmdb-db ... storage), with a YAML anchor for the shared 8Gi default. templates/pv.yaml stays templated. Docs/Tilt keep the native keys. TTL half is unchanged and first-party only.

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
rupalis-nv: done

distros/kubernetes/nvsentinel/values-full.yaml
rupalis-nv: done

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
```diff
@@ -55,10 +55,14 @@ subjects:
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: create-mongodb-database
-  annotations:
-    argocd.argoproj.io/sync-options: Force=true,Replace=true
```
lalitadithya, line 60: why did we remove `argocd.argoproj.io/sync-options: Force=true,Replace=true`?

rupalis-nv: We removed Force/Replace because it only helps Argo, and it recreates the Job on sync. Helm still would not rerun collMod on a TTL change. Kaivalya asked us to put the TTL in the Job name instead, so both Helm and Argo create a new Job only when TTL changes. TTL on upgrade still works without that annotation.

lalitadithya: correct, the force/replace will also make sure the job reruns if there are failures right in the first run when ArgoCD is used? It would help cases when TTL doesn't change and the job needs to be rerun

rupalis-nv: Yes. Force/Replace would let Argo rerun the same Job after a Failed first run, even if TTL didn’t change. Helm still wouldn’t. First-run connection failures are already retried in the Job (until + backoff + 10 min deadline). We removed Force/Replace so we don’t delete+create the init Job on every sync after it succeeded. If we want Failed-job retry on Argo, we can add Force/Replace back on top of the TTL Job name — that’s optional,

lalitadithya: yes, let's keep it

rupalis-nv: PTAL

distros/kubernetes/nvsentinel/templates/job-external-mongodb-setup.yaml
```diff
@@ -63,14 +63,16 @@ inside the mongodb-store subchart instead.
...
   labels:
     {{- include "nvsentinel.labels" . | nindent 4 }}
+    app.kubernetes.io/name: external-mongodb-setup
+    app.kubernetes.io/component: mongodb-init
   annotations:
-    argocd.argoproj.io/sync-options: Force=true,Replace=true
```
lalitadithya, line 70: can we restore it here as well?

rupalis-nv: sure .. done!

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
```diff
@@ -55,10 +55,18 @@ subjects:
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: create-mongodb-database
+  # Name includes TTL so a change creates a new Job. Label: create-mongodb-database
+  name: {{ include "mongodb-store.initJobName" . }}
```
XRFXLP, line 59: Can you do a sweep in runbooks like `docs/runbooks/mongodb-connection-error.md` and fix the stale reference to the old job name?

Also should we do a more general strategy here? Right now, we've changed just the TTL as per that comes inside the immutable job spec. What if we add a new index? Or any other change in the job spec that need the job to be re-installed?

distros/kubernetes/nvsentinel/templates/_helpers.tpl
```diff
@@ -1,3 +1,28 @@
+{{/*
```
XRFXLP, line 1: this is duplicate of distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl, right? can we refer the internal helpers.tpl and extract out the keys in parent chart?

XRFXLP (changes requested):

distros/kubernetes/nvsentinel/templates/_helpers.tpl
rupalis-nv: It's the same validation on purpose, not a leftover copy we can delete.

Helm subcharts cannot include parent templates, and when mongodb-store is disabled  the subchart helpers are not loaded at all. The parent helper reads `.Values.mongodb-store.collectionExpirySeconds`; the subchart helper reads `.Values.collectionExpirySeconds`. Both are required. A shared library chart would be the only way to have one copy, which is altogether different and big change

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
rupalis-nv: For other new index changes : Argo already recreates via Force/Replace. Helm still cannot patch a completed Job — if TTL is unchanged, delete by that label and upgrade. TTL stays in the Job name so a TTL-only change is a new Job without a delete. A checksum in the name would cover Helm script changes too; happy to add that in a follow-up if you want it in this PR.

XRFXLP: we can do that in this PR itself.

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_init-eval.tpl
```diff
@@ -0,0 +1,111 @@
...
+  var secs = Number($MONGODB_COLLECTION_EXPIRY_SECONDS);
+  var key = {};
+  key[field] = 1;
+  var existing = db.getCollection(collName).getIndexes().find(function(idx) {
+    return idx.key && idx.key[field] === 1 && Object.keys(idx.key).length === 1;
+  });
+  if (!existing) {
+    db.getCollection(collName).createIndex(key, { expireAfterSeconds: secs });
+    print('Created TTL index ' + collName + '.' + field + '=' + secs);
+  } else if (existing.expireAfterSeconds != secs) {
```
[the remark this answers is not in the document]
rupalis-nv: Done in both ensureTTL copies (_init-eval.tpl and _external-mongo-init-eval.tpl).

The env expansion is now quoted ('$MONGODB_COLLECTION_EXPIRY_SECONDS'). If it is empty or not a non-negative integer, the script throws before createIndex / collMod, so we never write expireAfterSeconds: 0 / NaN from a bad or missing ConfigMap value. Helm still defaults nil/empty chart values to 2592000 before the Job runs.

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
rupalis-nv: Updated PTAL

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
```diff
@@ -55,10 +55,18 @@ subjects:
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: create-mongodb-database
+  # Name is create-mongodb-database-<ttl>-<scriptHash>. Label: create-mongodb-database
```
XRFXLP, line 58: do we need to mention `ttl` at all, given that script hash will already provide the uniqueness?

KaivalyaMDabhadkar: I think we need it as ttl isn't part of the script text, the script reads it from the configmap at runtime. So if somone only changes the ttl, the script and its hash would still stay exactly the same

KaivalyaMDabhadkar (approved):

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
rupalis-nv: Yes — the hash is only of the init script. TTL is an env var, so it does not change the hash. We rendered both 2592000 and 604800 and got the same f38c611f. Without TTL in the name, a TTL-only upgrade would not create a new Job and collMod would not run.

Happy to fold TTL into the hash input and drop it from the visible name (create-mongodb-database-<hash>) if you prefer that. Functionally it is the same.

XRFXLP: What I thought I had in mind is: given the spec of this job is not immutable and whenever we change anything in the job spec, we need a way to automate the deletion -> creation of the job. So if the TTL is going in the spec of the job then changing TTL should change the hash, right?

XRFXLP: Or are you saying the expansion of env var is being done in runtime so hash cannot change if input (env var name) is not changing?

rupalis-nv: Yes. Expansion is at runtime. Helm hashes the script text, which always has $MONGODB_COLLECTION_EXPIRY_SECONDS, so 7d and 30d produce the same hash.

TTL is also not in the Job spec — only in ConfigMap mongodb-config via envFrom. A completed Job is immutable, so a TTL-only change would not recreate the Job unless TTL is in the name (or we fold that value into the hash).

Keeping <ttl>-<scriptHash> as-is.

XRFXLP (approved):

PR #1636, opened 2026-08-21
feat(auth): authenticate health event publishers and bind events to the reporting node

KaivalyaMDabhadkar:
## Summary

Health events are submitted to platform-connector over a Unix domain socket on each node, and every event carries a `nodeName` saying which machine it is about. Until now the connector had no way to establish which component sent an event, so it took that `nodeName` at face value. Health events drive cordon, drain and reboot, so getting the node right matters: an event carrying the wrong node name causes disruptive work to land on the wrong machine.

This change gives every publisher an identity and teaches platform-connector to check it. Each publisher mounts a short lived projected ServiceAccount token, minted for a dedicated audience and rotated by the kubelet. A server side interceptor validates the token with a TokenReview and then decides which nodes that caller is allowed to report on.

Everything is contained to the health event path. There is no change to how events are stored, evaluated or remediated once accepted.

### On the client side

Each publisher gets a projected token volume with the audience `platform-connector.nvsentinel.nvidia.com` and a one hour lifetime. The chart writes the mount, and the component is given the path with a `--platform-connector-token-path` flag.

The token is attached as a bearer credential on every gRPC call by a small shared interceptor in `commons/pkg/grpcclient`. It re-reads the file on each call rather than caching the string in memory, because the kubelet rewrites that file as it rotates the token. Caching it would work until the first rotation and then start failing for no visible reason. If the path is configured but the file is missing or empty, the call fails with an error naming the path rather than quietly publishing without a credential.

### On the server side

The connector runs a unary interceptor backed by `commons/pkg/grpcauth`, which does the TokenReview and answers two separate questions.

**Is this token being presented on the node it was issued for?** Since Kubernetes 1.32 the API server records the caller's node inside the token at issue time, so the token carries a statement about where its holder runs that the holder cannot write itself. The connector's socket is only reachable from its own node, so a token arriving with a different node recorded in it has travelled between machines. That is refused for every caller, including ones allowed to name other nodes.

**May this identity speak for nodes other than its own?** Most publishers only ever report on the node they run on. A few are cluster scoped and legitimately report on other nodes, so those are listed explicitly.

| What the caller presents | What it may report on |
|---|---|
| No token | Only the connector's own node |
| Valid token, ServiceAccount not on the cross-node list | Only the connector's own node |
| Valid token, ServiceAccount on the cross-node list, token bound to a running pod | Any node in the cluster |
| Valid token, ServiceAccount on the cross-node list, token not bound to a pod | Refused |

A caller restricted to its own node that submits an event naming a different node is refused rather than silently corrected. Being refused is easier to notice and diagnose than an event that quietly changed meaning.

That last row is worth explaining. A token created by hand with `kubectl create token <sa> --audience=...` and no `--bound-object-ref` authenticates as that ServiceAccount with the right audience, but the API server ties it to no pod, so it carries neither a pod nor a node. It would therefore skip the node check entirely and stay usable from anywhere for its whole lifetime. That would make an unbound token more capable than a correctly bound one, which is backwards, so callers on the cross-node list must present a pod bound credential even for events about their own node.

### Two questions, deliberately kept separate

Authentication (is this a valid token for this audience, and who does the API server say it is) lives in the shared `commons/pkg/grpcauth` package. Authorisation (may that identity do what it is asking) lives in `platform-connectors/pkg/auth`. Keeping them apart is what allows the node check to apply to everyone while the cross-node list applies to only a few.

## Caching and control plane load

Checking a token means calling the API server. Doing that for every batch from every publisher on every node would be significant load, so successful results are cached for a flat **two minutes** (`cacheTTL`), bounded at 4096 entries.

The cache deliberately remembers only **who the caller is**, never whether they are allowed. Permission is re-evaluated on every request, which is a map lookup against an in memory list and costs essentially nothing. The benefit is that editing the cross-node list takes effect immediately, rather than leaving a removed entry working for up to two minutes.

Two minutes rather than "until the token expires" is a deliberate trade. TokenReview is the only step that notices a caller's pod has been deleted, so whatever the cache lifetime is, that is exactly how long a deleted pod's token keeps working. Tokens last an hour by default, so caching until expiry would mean up to an hour. Two minutes bounds it, costs roughly one TokenReview per token per two minutes, and matches the default the API server itself uses for caching webhook authentication results.

## When the API server is unreachable

A TokenReview that cannot reach the API server is retried with exponential backoff (500ms base, factor 2, jitter, up to 8 steps) within an 8 second window. If it still fails, the call is rejected with `Unavailable` and tagged with structured error details so callers can tell an authentication backend outage apart from an ordinary refusal. That distinction matters for callers that need to know whether retrying is safe.

## Which publishers are converted

All of them, in one change:

`gpu-health-monitor`, `syslog-health-monitor`, `nic-health-monitor`, `slurm-drain-monitor`, `csp-health-monitor`, `kubernetes-object-monitor`, `health-events-analyzer`, the three preflight checks (`dcgm-diag`, `nccl-allreduce`, `nccl-loopback`), and connector to connector forwarding via the gRPC sink.

`event-exporter` loses its host path socket mount in this change. It consumes health events from the datastore change stream and does not publish any, so it never needed a route to the connector socket. That is a pure removal.

### Notes on Preflight

Preflight's checks do not run in a pod this project owns. They run as init containers that the mutating webhook injects into a user's workload pod, so there is no chart template to add a volume to. The webhook adds the projected token at admission time instead, next to the injected check containers.

Two consequences worth knowing. The token is minted for the **workload's own ServiceAccount**, not for an NVSentinel identity, and that ServiceAccount is not on the cross-node list, so a check can only ever report about the node its pod landed on. And a workload that declares its own volume named `nvsentinel-connector-token` is refused at admission rather than silently overwritten, because the injected checks read their credential from that name.

## Observability

Three counters are exported by platform-connector and documented in `docs/METRICS.md`:

| Metric | Labels | Meaning |
|---|---|---|
| `platform_connector_auth_decisions_total` | `decision` = `node_local` or `cross_node` | Which scope a caller resolved to |
| `platform_connector_auth_node_claim_total` | `result` = `verified` or `absent` | Whether the token carried a node and whether it matched |
| `platform_connector_auth_violations_total` | `reason` | Why a caller was refused |

The `reason` values are `node_mismatch`, `node_claim_mismatch`, `unbound_cross_node_token`, `cross_node_claim_absent`, `missing_node_name`, `token_invalid`, `malformed_credentials`, `validator_unavailable`, `validator_timeout` and `validator_error`.

A healthy cluster reports zero violations. A sustained non zero `validator_unavailable` points at the API server rather than at any caller.

The audit trail logs the full caller identity (user, user UID, pod, pod UID, node, node UID and audiences) on both accept and refusal. No token text is ever logged.

## Configuration

```yaml
global:
  platformConnectorAuth:
    enabled: true
    audience: "platform-connector.nvsentinel.nvidia.com"
    tokenExpirationSeconds: 3600
    tokenMountPath: "/var/run/secrets/nvsentinel/platform-connector"
    crossNodeServiceAccounts: []
```

The four bundled cluster scoped monitors (`csp-health-monitor`, `kubernetes-object-monitor`, `slurm-drain-monitor` and `health-events-analyzer`) are **derived from the release namespace** for whichever of them are enabled. You do not list them, and installing into a namespace other than `nvsentinel` needs no edit.

Deriving them matters because the username TokenReview reports carries the real namespace. A hardcoded namespace would render cleanly into any other namespace and then have every one of those events refused at runtime, so the chart computes it from the release instead.

`crossNodeServiceAccounts` is therefore only for publishers this chart does not ship, for example an out of tree monitor. Entries are fully qualified `system:serviceaccount:<namespace>:<name>` usernames, matched verbatim. Nothing is prefixed, trimmed or normalised, because an entry that does not match exactly can never match anything at all.

There is no `mode` and no `requireNodeClaim`. A setting that is "on but not enforcing" cannot be reasoned about from its configuration alone, so enforcement is not separable from enablement. `global.platformConnectorAuth.enabled=false` turns the whole thing off in one step.

Full operator reference is in `docs/configuration/authentication.md`.

## Misconfiguration now fails at install time

These settings are checked at render time, with a message naming the offending key, rather than being accepted and then misbehaving at runtime in ways that are hard to trace back to a values file:

- `enabled` that is not a real boolean. Go template truthiness reads the string `"false"` as **true**, so a quoted value silently did the opposite of what it said.
- A token lifetime outside `600 <= x <= 2^32` seconds, or a fractional one. Kubernetes rejects these when the pod is created, a long way from the values file.
- A username TokenReview could never report: wrong shape, stray whitespace, uppercase, or segments over the DNS 1123 length limits (63 for the namespace, 253 for the name).
- A missing or `null` cross-node list. Both are refused rather than read as an empty list.
- Preflight's three connector token settings (`connectorTokenAudience`, `connectorTokenMountPath`, `connectorTokenExpirationSeconds`) set partially rather than all or none. Preflight also refuses a mount path that is not absolute.

At the ConfigMap level, relevant only when bypassing the chart, a config that omits `enableNodeBindingAuth` entirely fails startup rather than defaulting. Such a config is also missing the audience and the cross-node list, so guessing the intended behaviour from it would be worse than refusing.

## Rolling this out

platform-connector and the publishers are subcharts of one release, so a single `helm upgrade` moves them together and you cannot sequence one before the other. What you can control is when enforcement starts.

**Images must roll out before or with the chart.** The chart passes a `--platform-connector-token-path` flag that older images do not recognise, and Go's flag parser rejects unknown flags at startup, so those pods would go into `CrashLoopBackOff`. The reverse does not start either: the connector requires `enableNodeBindingAuth` to be present, and an old chart does not write it. Both directions are documented in `docs/platform-connectors.md`.

Pods restart on their own schedule after an upgrade, so there is a brief window where a connector that has already restarted sees a publisher pod that has not yet been replaced and is therefore still sending no token. That exposure is narrow. A publisher with no token is not refused, it is pinned to its own node, so node local publishers are unaffected either way. Only a cluster scoped publisher is affected, and only for the seconds between the connector restarting and its own pod being replaced. Its cross-node events are refused with `node_mismatch` for that window, and its next attempt succeeds.

### Migration

| Old | New |
|---|---|
| `crossNodeServiceAccounts` listing the bundled monitors | Remove them. They are derived from the release namespace. |

### Unit and integration

All affected Go modules pass with `-race`, including the envtest backed suites: **2355 tests across 12 modules, 0 failures**. Python modules pass as well: **275 tests** across `gpu-health-monitor`, `dcgm-diag` and `nccl-allreduce`.

New tests cover each row of the rule table, the verdict cache (hit, miss, rotation, expiry, and the cross-node list still being enforced on a cache hit), the token client (attach, re-read on every call, whitespace handling, missing file, no token configured), the reason label mapping, and a regression test for a retry window that expires before its first attempt.

Helm: **31 tests across 6 charts**, including new `pc_auth_strictness` suites for gpu, nic, preflight, slurm-drain and syslog monitors, plus renders of the default, Tilt, full and remediation profiles and the three UAT profiles, in both the default namespace and a non-default one.

### End to end on real clusters

Five clusters spanning Kubernetes 1.32 and 1.34, each running images built from this branch with authentication enforcing:

|  | OCI | on-prem datacenter | AWS | GCP | Azure |
|--|-----|--------------------|-----|-----|-------|
| DCGM fault injected, node cordoned, cleared, uncordoned | Yes | Yes | Yes | Yes | Yes |
| Node local publish accepted, node recorded in token verified | Yes | Yes | Yes | Yes | Yes |
| Cross-node publish authorised | Yes | Yes | Yes | Yes | Yes |

Cross-node was driven by `kubernetes-object-monitor` running on one node and publishing an event about a different node, confirmed by `platform_connector_auth_decisions_total{decision="cross_node"}` on the connector that accepted it, with `platform_connector_auth_node_claim_total{result="verified"}` alongside.

On OCI the full refusal matrix was run as seven credential types against both the connector's own node and another node. All fourteen outcomes matched expectation: no token, a garbage token, an unbound `kubectl create token`, a token bound to a pod on another node, the wrong audience, and the cluster's own default API server audience. The metric deltas reconciled exactly with the fourteen requests.

Preflight was verified separately on OCI: a GPU workload in a labelled namespace received the injected check container and a projected volume with the configured audience, `path: token` and a 3600 second lifetime, mounted at the configured path; the token was pod bound and carried the node; the workload started once the check passed; and a pod declaring a conflicting `nvsentinel-connector-token` volume was refused at admission.

Also confirmed on the real clusters: the audit trail logs the full caller identity with no token text anywhere, a token minted for a different audience is refused, and a pod bound token stops working once its pod is deleted.

### Local e2e

The suite was run against a kind cluster with the default topology (50 simulated GPU nodes, 5 kata nodes). Every failure observed was traced to a pre-existing test defect or to leaked cluster state from an earlier run rather than to this change, and no authentication refusal was logged in any run.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added optional ServiceAccount-token authentication for platform connector communications.
  * Added node-binding authorization for local and approved cross-node health-event publishing.
  * Added projected-token support across Kubernetes deployments, Helm charts, monitors, and preflight checks.
  * Added token rotation, caching, authentication metrics, and transient-failure handling.

* **Bug Fixes**
  * Improved validation and retry handling for permanent versus transient authentication errors.
  * Improved test-node selection by excluding unreadable, cordoned, or quarantined nodes.
  * Removed an unnecessary event-exporter host volume mount.

* **Documentation**
  * Added configuration, migration, monitoring, and troubleshooting guidance for authentication and node binding.


distros/kubernetes/nvsentinel/charts/slurm-drain-monitor/templates/_helpers.tpl
```diff
@@ -38,3 +38,85 @@ Selector labels
...
+in tests/pc_auth_strictness_test.yaml is what catches the two drifting again.
+*/}}
+{{- define "slurm-drain-monitor.pcAuth.enabled" -}}
+{{- $auth := ((.Values.global).platformConnectorAuth) | default dict -}}
+{{- $enabled := $auth.enabled -}}
+{{- if not (kindIs "bool" $enabled) -}}
+{{- fail (printf "global.platformConnectorAuth.enabled must be a boolean (true or false), got %s %#v. Quoted strings, null and numbers are refused because they would silently enable or disable authentication." (kindOf $enabled) $enabled) -}}
+{{- end -}}
+{{- if $enabled -}}true{{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Good catch that it fails without the global, but I would rather leave it as is.
All six of these charts have a test called "refuses a null rather than silently disabling authentication". Adding enabled: false as a chart default is that exact silent disable. If the parent chart ever stopped passing the global down, the subchart would quietly render with auth turned off instead of failing the install, which is the thing those tests exist to prevent.
Rendering a subchart on its own isn't really a supported mode either. Nothing in the Makefile or CI does it, and we don't package or publish these charts separately. make helm-lint renders them with the parent chart's values, and I have fixed that loop in this PR (it was silently skipping every chart before), so they are covered.

preflight-checks/dcgm-diag/dcgm_diag/tests/test_config.py
```diff
@@ -91,3 +91,14 @@ def test_invalid_status_retry_interval(self, valid_env: None, monkeypatch: pytes
         monkeypatch.setenv("DCGM_DIAG_STATUS_RETRY_INTERVAL_SECONDS", "0")
         with pytest.raises(ValueError, match="DCGM_DIAG_STATUS_RETRY_INTERVAL_SECONDS must be > 0"):
             Config.from_env()
+
+    def test_token_path_is_none_when_unset(self, valid_env: None) -> None:
+        assert Config.from_env().token_path is None
+
+    def test_token_path_loaded_from_env(self, valid_env: None, monkeypatch: pytest.MonkeyPatch) -> None:
+        monkeypatch.setenv("PLATFORM_CONNECTOR_TOKEN_PATH", "/var/run/secrets/nvsentinel/token")
+        assert Config.from_env().token_path == "/var/run/secrets/nvsentinel/token"
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: S105 isn't switched on for this package. There's no select in preflight-checks/dcgm-diag/pyproject.toml turning on flake8-bandit, and preflight-checks-lint-test (dcgm-diag) is green on this PR. I would rather not add a noqa for a rule that isn't running.

preflight-checks/nccl-loopback/pkg/health/reporter_test.go
```diff
@@ -0,0 +1,208 @@
...
+// temporary directory and returns the socket path.
+func startTestConnector(t *testing.T) (string, *capturingConnector) {
+	t.Helper()
+
+	socketPath := filepath.Join(t.TempDir(), "pc.sock")
+
+	lis, err := net.Listen("unix", socketPath)
+	if err != nil {
+		t.Fatalf("failed to listen on %s: %v", socketPath, err)
+	}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: noctx is enabled in .golangci.yml, but it isn't flagging this file. I ran golangci-lint on the module locally and the only findings are 4 gosec ones in pkg/config/config.go, which are already there on main today, so they aren't from this PR. preflight-checks-lint-test (nccl-loopback) passes here too.

tests/main_test.go
```diff
@@ -28,6 +28,13 @@ import (
 	kwokv1alpha1 "sigs.k8s.io/kwok/pkg/apis/v1alpha1"
 )
 
+// gpuOperatorNamespace is declared here, in the only file without a build
+// tag, so that both the amd64_group and arm64_group suites can use it and
+// the two tags can still be combined in one `go test` invocation. Declaring
+// it in a tagged file in each group made the combined build fail with a
+// redeclaration error.
+const gpuOperatorNamespace = "gpu-operator"
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: modules-lint-test (tests) is green on this PR, and golangci-lint on the tests module locally gives 0 issues, so nothing is flagging it.

The const is where it is on purpose: it lives in the only file without a build tag so that both amd64_group and arm64_group can use it, and so the two tags can still be combined in a single go test run. Putting it in a tagged file broke that with a redeclaration error. A //nolint:unused for a warning that isn't firing would just be noise

docs/platform-connectors.md
```diff
@@ -212,6 +212,221 @@ Transformer failures log warnings but don't block event processing. If a transfo
...
+helm upgrade nvsentinel <chart> --reuse-values --set global.platformConnectorAuth.enabled=false
+```
+
+That drops the flag, the token volume and the connector's enforcement in one
+step, leaving the pre-feature behavior. Re-enable it once the images are current.
+
+The other direction — old chart, new images — is safe. The flag defaults to
+empty, which disables the client-side token interceptor, and an old chart also
+leaves connector-side enforcement off, so both ends agree that no token is in
+play.
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Fixed

commons/pkg/grpcauth/cache.go
```diff
@@ -0,0 +1,96 @@
...
+	//
+	// It is a fixed value rather than the token's own expiry. TokenReview is the
+	// only check that notices the bound pod being deleted, so the TTL is the
+	// window in which a deleted pod's token still works. Honouring the token's
+	// exp would make that window the token lifetime — an hour at the chart's
+	// default — for every ordinary publisher. A flat two minutes bounds it for
+	// everyone, costs about one TokenReview per token per two minutes, and
+	// needs no assumption about what kind of credential TokenReview accepted.
+	// It matches the API server's own webhook-authenticator cache default.
+	cacheTTL = 2 * time.Minute
```
neerajnv, line 42: it can be as big as the time kubelet takes to rotate a token. it's an optimisation though, has no effect on functionality. However, consider moving the value to configuration.

commons/pkg/grpcauth/tokenreview.go
```diff
@@ -0,0 +1,539 @@
...
+
+	result, err := CreateTokenReview(ctx, v.client, review, v.retryWindow, v.backoff)
+	if err != nil {
+		return nil, err
+	}
+
+	if !result.Status.Authenticated {
+		slog.WarnContext(ctx, "Token authentication failed", "error", result.Status.Error)
+
+		return nil, status.Errorf(codes.Unauthenticated, "token not authenticated: %s", result.Status.Error)
```
neerajnv, line 205: should we have a failure metric when authentication does not succeed? consider adding if this is thought to be an uncommon occurrence.

commons/pkg/grpcauth/tokenreview.go
```diff
@@ -0,0 +1,539 @@
...
+}
+
+// soleExtra returns the single value recorded for key in a TokenReview's extra
+// map, or "" if the key is absent entirely. It is safe on a nil map.
+//
+// These fields identify who the caller is, so an ambiguous answer is refused
+// rather than resolved: Kubernetes records exactly one value for each of them,
+// and picking one of several would let whatever produced the extra list decide
+// which identity this connector enforces against.
+func soleExtra(extra map[string]authv1.ExtraValue, key string) (string, error) {
```
neerajnv: can we change the name to help interpretation without going through the entire comment

commons/pkg/grpcauth/tokenreview.go
```diff
@@ -0,0 +1,539 @@
...
+) (*authv1.TokenReview, error) {
+	opCtx, cancel := context.WithTimeout(callerCtx, window)
+	defer cancel()
+
+	var (
+		result *authv1.TokenReview
+		err    error
+	)
+
+	waitErr := wait.ExponentialBackoffWithContext(opCtx, backoff, func(attemptCtx context.Context) (bool, error) {
```
neerajnv, line 280: consider adding a latency metric if you think this can be prone to frequent retries.

commons/pkg/grpcauth/tokenreview.go
```diff
@@ -0,0 +1,539 @@
...
+//
+// It exists because Unavailable alone is ambiguous to a client: a connection
+// that dropped after the server acted looks identical to one that never
+// arrived. For an idempotent call that does not matter, but a client driving a
+// destructive RPC — terminating a node — cannot safely retry an ambiguous
+// failure, because the node may already be gone.
+//
+// A status carrying this reason is unambiguous: the server-side interceptor
+// produced it before dispatch, so the handler never ran and no CSP action was
+// taken. Retrying is then provably safe.
```
neerajnv, line 397: is this the right place for this documentation?

commons/pkg/grpcclient/token_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	err := interceptor(context.Background(), "/test.Method", nil, nil, nil, fakeInvoker)
+	require.NoError(t, err)
+
+	md, ok := metadata.FromOutgoingContext(capturedCtx)
+	require.True(t, ok, "outgoing metadata should be present")
+
+	authValues := md.Get("authorization")
+	require.Len(t, authValues, 1)
+	assert.Equal(t, "Bearer test-token-value", authValues[0])
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: fatcontext isn't one of the linters we enable in .golangci.yml, and golangci-lint on commons/pkg/grpcclient comes back with 0 issues, so nothing is flagging these closures. Leaving them as they are.

commons/pkg/grpcclient/token.go
```diff
@@ -0,0 +1,80 @@
...
+		method string,
+		req, reply any,
+		cc *grpc.ClientConn,
+		invoker grpc.UnaryInvoker,
+		opts ...grpc.CallOption,
+	) error {
+		//nolint:gosec // G304: tokenPath is operator-controlled config, not user input.
+		tokenBytes, err := os.ReadFile(tokenPath)
+		if err != nil {
+			return fmt.Errorf("reading SA token from %q: %w", tokenPath, err)
```
neerajnv, line 49: is this error non-recoverable?

health-monitors/gpu-health-monitor/gpu_health_monitor/tests/test_platform_connector/test_platform_connector.py
```diff
@@ -1884,3 +1885,226 @@ def test_recovery_resets_the_escalation_counter(self):
...
+    def test_transient_status_is_retried(self):
+        for code in (grpc.StatusCode.UNAVAILABLE, grpc.StatusCode.DEADLINE_EXCEEDED):
+            with self.subTest(code=code):
+                processor = self._make_processor(token_path=self._write_token("projected-token"))
+
+                result, stub = self._send_with_failing_stub(processor, RpcErrorWithCode(code))
+
+                assert result is False
+                assert stub.HealthEventOccurredV1.call_count > 1
+                assert processor.entity_cache == {}
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Added -> None to the 9 new tests in test_platform_connector.py. The other three files were already annotated, including the tests this PR adds. The test_health.py one has a multi-line signature so the annotation is on the closing line

neerajnv (commented):
left some comments

commons/pkg/grpcauth/tokenreview.go
XRFXLP: we've `platform_connector_auth_violations_total` for that, it is being called a layer above in `platform-connectors/pkg/auth/nodebinding.go`

XRFXLP: https://github.com/NVIDIA/NVSentinel/pull/1636/changes#diff-4ad8ced54658e1d3d6224a8971b5a7ac01137405b74cbddc530dc48b87112f1bR315

commons/pkg/grpcauth/cache.go
KaivalyaMDabhadkar: good point and you are right that it is an optimisation and does not change behaviour. But I would still rather keep it as a constant, because this value quietly controls something else as well. The cache stores the result of a successful TokenReview and TokenReview is the only step that ever asks the apiserver whether the caller still exists. So once a pod is deleted, its token keeps being accepted until the cached entry expires. Whatever this number is set to is exactly how long that lasts.
At two minutes that window is small. If it were configurable, someone could reasonably raise it to an hour to cut apiserver calls, without realising they had also made a deleted pod's token usable for an hour. Two minutes is also what the API server itself uses for caching webhook authentication results, so that is the more compelling reason for hardcoding it to that specific value.

commons/pkg/grpcauth/tokenreview.go
KaivalyaMDabhadkar: +1 on what Ajay said. In platform-connector, a failed authentication comes back as Unauthenticated, and violationReasonsByCode turns that into platform_connector_auth_violations_total{reason="token_invalid"}.
The same map covers the other outcomes too: Unavailable becomes validator_unavailable, and DeadlineExceeded or Canceled become validator_timeout

neerajnv:
`AuthBackendUnavailableReason`/`IsAuthBackendUnavailable` (tokenreview.go) look like a real gap: this mechanism exists specifically so a caller driving a non-idempotent RPC can distinguish "auth backend unavailable, safe to retry" from an ambiguous transport failure, but nothing in the codebase actually calls `IsAuthBackendUnavailable` — not `commons/pkg/healthpub/publisher.go`, not `platform-connectors/main.go`. Right now a sustained TokenReview/API-server outage surfaces as a generic `Unavailable` indistinguishable from "connector pod is down," so the disambiguation this was built for isn't available to operators or callers during an incident. Worth wiring this up (e.g. have `healthpub.Publisher.finalize` check it and label errors distinctly) before this ships, given health events drive cordon/drain/reboot.

commons/pkg/grpcauth/tokenreview.go
KaivalyaMDabhadkar: Yes that's a good idea. Today you can see retries that give up, because they end as platform_connector_auth_violations_total{reason="validator_unavailable"} once the 8 second window expires. What you cannot see is retries that eventually succeed, so an apiserver that is slow but not failing stays invisible until it tips over.
I have left it out of this change because it means adding instrumentation to the shared library, which currently has none at all by design. But I can pick it up as a follow-up.

commons/pkg/grpcclient/token.go
KaivalyaMDabhadkar: Not permanently. It fails that one call, and the next publish tries again, because the token file is re-read on every call rather than cached. So a transient read error clears itself without anything having to restart.
What it deliberately will not do is carry on and send the request without a credential. If a publisher has been given a token path, failing to read it means the mount is wrong, and publishing anyway would turn a clear local error into a confusing rejection at the other end.

commons/pkg/grpcauth/tokenreview.go
lalitadithya: yes, it shows up on the go package website and it also keeps it from drifting when implmentation changes

commons/pkg/grpcauth/tokenreview.go
KaivalyaMDabhadkar: Done, the old name needed the comment to make sense. Renamed it to exactlyOneExtraValue which states the rule it enforces.

commons/pkg/grpcauth/tokenreview.go
KaivalyaMDabhadkar: @lalitadithya makes sense, leaving it on the exported const in that case

KaivalyaMDabhadkar:
> `AuthBackendUnavailableReason`/`IsAuthBackendUnavailable` (tokenreview.go) look like a real gap: this mechanism exists specifically so a caller driving a non-idempotent RPC can distinguish "auth backend unavailable, safe to retry" from an ambiguous transport failure, but nothing in the codebase actually calls `IsAuthBackendUnavailable` — not `commons/pkg/healthpub/publisher.go`, not `platform-connectors/main.go`. Right now a sustained TokenReview/API-server outage surfaces as a generic `Unavailable` indistinguishable from "connector pod is down," so the disambiguation this was built for isn't available to operators or callers during an incident. Worth wiring this up (e.g. have `healthpub.Publisher.finalize` check it and label errors distinctly) before this ships, given health events drive cordon/drain/reboot.

Good catch, and you are right that it was a real gap rather than just unused code. The mechanism existed but nothing consumed it, so during an apiserver outage the publisher metric reported code="Unavailable", which is exactly what a dead platform-connector looks like.
I now wired it into healthpub.errorCodeLabel, which is where that label is chosen. An authentication backend outage now reports code="AuthBackendUnavailable" while an unreachable connector still reports code="Unavailable", so the two are separable on the same counter. Added a test that builds both error shapes and asserts the labels differ, so it cannot regress quietly. healthpub now imports grpcauth, which adds hashicorp/golang-lru/v2 as an indirect dependency to the five monitor modules that publish events, so their go.mod and go.sum are updated too.

XRFXLP (approved):

lalitadithya (approved):

PR #1638, opened 2026-08-21
fix(labeler): kata label update should not short-circuit driver/dcgm label update

XRFXLP:
## Summary

This PR fixes a short-circuit bug:
```golang
needsUpdate = needsUpdate || l.updateDriverAndDCGMLabels(node, driverLabel, dcgmVersion)
```

here if `needsUpdate` was true before this assignment then it can short-circuit the function call ` l.updateDriverAndDCGMLabels(node, driverLabel, dcgmVersion` making it a no-op even if the there is a genuine diff resulting with respect to dcgm and driver labels. The fix is to avoid the trick altogether:

```golang
	if l.updateDriverAndDCGMLabels(node, driverLabel, dcgmVersion) {
		needsUpdate = true
	}
```

Added a regression test `TestReconcileAppliesKataAndDetectionLabelsInOnePass` for this fix, verified that the test fails is the fix is undone.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed node label reconciliation so both Kata and driver labels are applied during a single update pass.
  * Prevented driver label updates from being skipped when another label change is already detected.

* **Tests**
  * Added focused coverage for applying both labels when they are initially missing.


labeler/pkg/labeler/labeler.go
```diff
@@ -930,7 +930,9 @@ func (l *Labeler) reconcileNodeLabelsInPlace(node *v1.Node, driverLabel, dcgmVer
 		return needsUpdate
 	}
 
-	needsUpdate = needsUpdate || l.updateDriverAndDCGMLabels(node, driverLabel, dcgmVersion)
+	if l.updateDriverAndDCGMLabels(node, driverLabel, dcgmVersion) {
```
deesharma24, line 933:
```suggestion
	needsUpdate = l.updateDriverAndDCGMLabels(node, driverLabel, dcgmVersion)
```

deesharma24: not that it matters much but it can be handled in a single line.

XRFXLP: it can make true value of needsUpdate to false, right?

lalitadithya: won't this mean that is `l.updateDriverAndDCGMLabels` returns false it will set `needsUpdate` to false which we don't want?

deesharma24: ahh.got it.

deesharma24 (approved):

lalitadithya (approved):

PR #1639, opened 2026-08-21
feat: add max remediation attempts per equivalence group

### LINKED ISSUE #1543: [Feature]: Enforce a maximum number of remediation attempts per node
### Prerequisites

### Code of Conduct

### Feature Summary

This feature would enforce a maximum number of GPU reset of reboot attempts per node during a given quarantine session. Currently, there is no limit on the number of GPU resets or reboots attempted per node and fault-remediation will process all unhealthy events with a supported remediation action (as long as there's no InProgress CRD for the remediation action's equivalence group). 

### Problem/Use Case

We have observed remediation loops continually occur with both GPU resets and reboots where a node will retry the same remediation action which does not resolve the underlying fault. If the remediation action itself results in a new unhealthy event being generated, a loop can occur where the node stays cordoned and the same remediation action is attempted against the duplicated event.

While all health-monitors can experience this behavior, the gpu-health-monitor is most susceptible because both GPU resets and reboots will result in the gpu-health-monitor pod being recreated. This results in the pod losing its cache of unhealthy DCGM checks which will cause it re-fire events from all failing health checks after the GPU reset or reboot attempt. This can result in the following behavior:
1. The gpu-health-monitor fires an event from a DCGM health check needing a GPU reset or reboot
2. The fault-remediation module creates the RebootNode or GPUReset CRD which succeeds and recreates the gpu-health-monitor pod
3. The gpu-health-monitor re-fires the same event since the remediation action did not result in the DCGM health check passing
4. Steps 2 and 3 repeat until an operator manually remediates 

### Proposed Solution

The fault-remediation module could maintain state related to the number of remediation actions attempted per equivalence group and either enforce a maximum number of remediation actions across all groups per node or enforce a maximum per equivalence group. Once the limit is reached, fault-remediation could react with the same behavior as if it received a CONTACT_SUPPORT event and mark the state as remediation-failed without proceeding with the remediation action.

An alternative option would be to implement unhealthy event de-duplicating at either the health-monitor or platform-connector level. Note that we have an XID de-duplication mechanism on a time-based window in the platform-connector which could be re-used for gpu-health-monitor: https://github.com/NVIDIA/NVSentinel/pull/1283

### Component

Health Monitor

iacker:
This PR implements a configurable maximum number of remediation attempts per equivalence group to prevent infinite remediation loops.

## Changes

- Added `MaxRetryAttempts` field to `TomlConfig` (default 0 = unlimited, preserving current behavior)
- Added `RetryCount` field to `EquivalenceGroupState` in the node annotation
- The reconciler now increments the retry count on each remediation attempt
- Remediation is skipped when the retry limit is exceeded, marking the event as failed
- Added unit tests verifying retry count increments, independence per group, and persistence across pod restarts

## Implementation

The retry counter persists across pod restarts because it's stored in the node's annotation (`nvidia.com/fault-remediation-state`) rather than in-memory state. Each time a remediation CR is created for an equivalence group, the counter increments and is written back to the annotation.

When `MaxRetryAttempts` is configured and the limit is reached, the remediation is skipped with the same behavior as unsupported events (marked as failed, event processed, but no CR created).

## Testing

```bash
cd fault-remediation
go test ./pkg/annotation/... -run TestRetry -v
go test ./pkg/annotation/...
go test ./pkg/config/...
```

All tests pass.

Fixes #1543

## Summary by CodeRabbit

- **New Features**
  - Added configurable maximum remediation retry attempts, with unlimited retries retained as the default.
  - Remediation retry counts now persist across restarts and are tracked independently for each equivalence group.
- **Bug Fixes**
  - Remediation events exceeding the configured retry limit are now safely marked as failed and skipped.
  - Improved GPU recovery matching to recognize replacement hardware and clear stale fault conditions.
  - Preserved existing behavior for non-GPU events and GPUs without stable identity information.

### DIFF
--- a/fault-remediation/pkg/annotation/annotation.go
+++ b/fault-remediation/pkg/annotation/annotation.go
@@ -104,11 +104,19 @@ func (m *NodeAnnotationManager) UpdateRemediationState(ctx context.Context, node
 			return err
 		}
 
+		// Increment retry count if this group already exists
+		existingGroup, exists := state.EquivalenceGroups[group]
+		retryCount := 0
+		if exists {
+			retryCount = existingGroup.RetryCount + 1
+		}
+
 		// Update state for the group
 		state.EquivalenceGroups[group] = EquivalenceGroupState{
 			MaintenanceCR: crName,
 			CreatedAt:     time.Now().UTC(),
 			ActionName:    actionName,
+			RetryCount:    retryCount,
 		}
 
 		// Marshal to JSON
@@ -131,7 +139,8 @@ func (m *NodeAnnotationManager) UpdateRemediationState(ctx context.Context, node
 		slog.InfoContext(ctx, "Updated remediation state annotation for node",
 			"node", nodeName,
 			"group", group,
-			"crName", crName)
+			"crName", crName,
+			"retryCount", retryCount)
 
 		return nil
 	})
--- a/fault-remediation/pkg/annotation/annotation_interface.go
+++ b/fault-remediation/pkg/annotation/annotation_interface.go
@@ -47,4 +47,8 @@ type EquivalenceGroupState struct {
 	// Action that created the CR (e.g., "RESTART_BM")
 	// Required to look up the corresponding MaintenanceResource from the TomlConfig
 	ActionName string `json:"actionName"`
+
+	// RetryCount tracks how many remediation attempts have been made for this equivalence group.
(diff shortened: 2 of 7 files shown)


iacker:
Closing this in favor of #1640 which contains only the intended commit for #1543. This PR accidentally included two commits from the old #1587 work.

PR #1640, opened 2026-08-21
feat: add max remediation attempts per equivalence group

### LINKED ISSUE #1543: [Feature]: Enforce a maximum number of remediation attempts per node

iacker:
This PR implements a configurable maximum number of remediation attempts per equivalence group to prevent infinite remediation loops.

## Changes

- Added `MaxRetryAttempts` field to `TomlConfig` (default 0 = unlimited, preserving current behavior)
- Added `RetryCount` field to `EquivalenceGroupState` in the node annotation
- The reconciler now checks retry count before creating a remediation CR and increments the counter on each attempt
- The counter persists across pod restarts via the node annotation

## Testing

- Added unit tests verifying retry count increments correctly, stays independent per group, and persists across simulated pod restarts
- Existing annotation and config tests pass

Closes #1543

## Summary by CodeRabbit

* **New Features**
  * Added an optional per-group remediation-attempt limit, configurable through `maxRemediationAttempts`.
  * Attempt counts persist across failed remediation attempts and service restarts.
  * Counts reset when a quarantine session ends.

* **Bug Fixes**
  * Nodes exceeding the configured limit are marked `remediation-failed` and no longer retried.
  * Failed resource creation attempts are now counted consistently.
  * Remediation state is correctly preserved and cleaned up across group changes.


fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -1065,6 +1065,19 @@ func (r *FaultRemediationReconciler) handleRemediationEvent(
 			nodeName)
 	}
 
+	// Check if we've exceeded the maximum retry attempts
+	maxRetries := r.Config.RemediationClient.GetConfig().MaxRetryAttempts
+	if maxRetries > 0 {
+		res, err, done := r.trySkipMaxRetriesExceeded(ctx, nodeName, groupConfig.EffectiveEquivalenceGroup,
```
KaivalyaMDabhadkar: I don't think this check can ever fire in the scenario it's meant for. checkExistingCRStatus runs first (line 1042), and every branch that allows a new CR also marks the group for removal (evaluateExistingCR lines 1737-1745, evaluateSucceededCR line 1765). So RemoveGroupsFromState (line 1700) deletes the group entry, including its RetryCount, right before this function reads the annotation. The check then sees "first attempt", and after the new CR is created UpdateRemediationState writes retryCount=1 again. The counter can never get past 1, so with any maxRetryAttempts >= 2 nothing is ever capped and the loop from #1543 is still there. Even if the check did fire once, the next event would delete the failed CR's entry and remediation would resume past the limit.

Suggestion: do the cap check inside evaluateExistingCR's CRStateFailed branch instead, where the old RetryCount is still in hand, and carry the count over when the group is recreated (or keep the exhausted entry) so the give-up actually holds.

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -1161,6 +1174,56 @@ func (r *FaultRemediationReconciler) trySkipResolvedEvent(
...
+func (r *FaultRemediationReconciler) trySkipMaxRetriesExceeded(
+	ctx context.Context,
+	nodeName string,
+	effectiveEquivalenceGroup string,
+	maxRetries int,
+	eventWithToken datastore.EventWithToken,
+	watcherInstance datastore.ChangeStreamWatcher,
+	healthEventStore datastore.HealthEventStore,
+) (ctrl.Result, error, bool) {
+	remediationState, _, err := r.annotationManager.GetRemediationState(ctx, nodeName)
```
KaivalyaMDabhadkar: There's a cache problem here as well. This GetRemediationState goes through the manager's cached client, right after RemoveGroupsFromState wrote the removal to the api server, so this read will often still see the old entry. With maxRetryAttempts=1 that can misfire badly: the previous CR succeeded, a genuinely new fault shows up later, checkExistingCRStatus removes the old group to start a new session, but this stale read still sees retryCount=1, so 1 >= 1 triggers. The event gets marked faultRemediated=false and its resume token is consumed. The change stream won't re-deliver it and cold start skips events that already have faultRemediated set, so if the fault doesn't fire again (a one-shot fault on an already drained node), the node just sits quarantined with no CR until someone notices.

Reusing the state checkExistingCRStatus already fetched (one read, taken before the removal) fixes this and removes the extra API read at the same time.

fault-remediation/pkg/config/config.go
```diff
@@ -85,6 +85,10 @@ type TomlConfig struct {
 
 	// Common configuration
 	UpdateRetry UpdateRetry `toml:"updateRetry"`
+
+	// MaxRetryAttempts is the maximum number of remediation attempts per equivalence group
+	// before marking the remediation as failed. Zero means unlimited retries (legacy behavior).
+	MaxRetryAttempts int `toml:"maxRetryAttempts"`
```
KaivalyaMDabhadkar: This field isn't wired up anywhere, so nobody can actually turn the feature on. The fault-remediation chart's configmap.yaml doesn't render maxRetryAttempts, and there's no values.yaml key or docs entry, so every install runs with 0 (unlimited).

One thing to watch when wiring it: this is a top-level TOML key, so it must go before the first [section] header in config.toml. If it's added after [updateRetry], the parser treats it as an updateRetry sub-key and our loader silently ignores unknown keys, so it would look configured but do nothing.

fault-remediation/pkg/annotation/annotation.go
```diff
@@ -104,11 +104,19 @@ func (m *NodeAnnotationManager) UpdateRemediationState(ctx context.Context, node
 			return err
 		}
 
+		// Increment retry count if this group already exists
+		existingGroup, exists := state.EquivalenceGroups[group]
+		retryCount := 1
+		if exists {
+			retryCount = existingGroup.RetryCount + 1
```
KaivalyaMDabhadkar: There is a gap in the counting: RetryCount only goes up after a CR is created successfully (updateRemediationAnnotationIfNeeded runs after createMaintenanceCR succeeds). If CR creation itself keeps failing, say the CRD isn't installed or a webhook or RBAC rejects it, no entry is ever written, so those attempts are invisible to the cap and can loop forever even with maxRetryAttempts set

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -1161,6 +1174,56 @@ func (r *FaultRemediationReconciler) trySkipResolvedEvent(
...
+		slog.WarnContext(ctx, "Maximum retry attempts exceeded for equivalence group",
+			"node", nodeName,
+			"group", effectiveEquivalenceGroup,
+			"retryCount", groupState.RetryCount,
+			"maxRetries", maxRetries)
+
+		metrics.EventsProcessed.WithLabelValues(metrics.CRStatusSkipped, nodeName).Inc()
+
+		// Mark as failed since we cannot remediate
+		if err := r.updateNodeRemediatedStatus(ctx, healthEventStore, eventWithToken, false); err != nil {
```
KaivalyaMDabhadkar: Could we also set the node label to remediation-failed when the cap is hit? Right now we only write faultRemediated=false. #1543 asks for the same behavior as CONTACT_SUPPORT, and the unsupported-action path already does this (lines 292-294). Without it, the node keeps the label from the last attempt, usually remediation-succeeded, and since the runbook tells operators to look for remediation-failed, these nodes would be easy to miss.
One thing to watch: the label would need to be set every time the cap fires, not just once. Nothing protects the label value, and fault-quarantine and node-drainer re-stamp it as new events come through, so a one-time write would get overwritten. This also depends on the retry counter surviving between events (see my other comment).
small note: these lines do the same thing as markEventTerminalAndProcessed (line 1578). A shared helper that writes the status and the label together would keep all of this in one place.

fault-remediation/pkg/annotation/retry_count_test.go
```diff
@@ -0,0 +1,128 @@
...
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        nodeName,
+			Annotations: map[string]string{},
+		},
+	}
+
+	client := fake.NewClientBuilder().WithObjects(node).Build()
+	annotationManager := NodeAnnotationManager{client: client}
+
```
KaivalyaMDabhadkar: Could you add integration coverage in reconciler_e2e_test.go rather than more annotation-level unit tests? The tests here call UpdateRemediationState back to back, which is a sequence the reconciler never produces (it always removes the group entry in between), so they pass while the end-to-end cap never engages.

Scenarios that would have caught this:
1. Failed CR, then next event with maxRetryAttempts=2: assert the count carries over and no CR is created once the cap is reached.
2. After the cap fires, more equivalent events keep it stopped and re-set the remediation-failed label each time.
3. A fresh reconciler over the same cluster still enforces the cap (real restart persistence).
4. A cancellation event resets the budget for the next session.
5. A succeeded CR followed by an out-of-session event is NOT capped.

TestEventSequenceWithAnnotations_Integration is a good template. Related: MockNodeAnnotationManager's RemoveGroupsFromState and UpdateRemediationState are no-ops and MockK8sClient.GetConfig never sets MaxRetryAttempts, so unit tests against the mocks can't catch any of this. Also TestRetryCountPersistsAcrossPodRestarts doesn't add coverage: the manager is stateless, so a second instance over the same fake client is the same as calling the first one twice.

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -1161,6 +1174,56 @@ func (r *FaultRemediationReconciler) trySkipResolvedEvent(
...
+	nodeName string,
+	effectiveEquivalenceGroup string,
+	maxRetries int,
+	eventWithToken datastore.EventWithToken,
+	watcherInstance datastore.ChangeStreamWatcher,
+	healthEventStore datastore.HealthEventStore,
+) (ctrl.Result, error, bool) {
+	remediationState, _, err := r.annotationManager.GetRemediationState(ctx, nodeName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
KaivalyaMDabhadkar, line 1223: If the node is deleted between the earlier read and this one, we return false and go on to run log collection and CR creation against a node that doesn't exist. The same situation two steps earlier (lines 1045-1048) marks the event terminal and processed instead. I suggest mirroring that here or just dropping this branch and letting the normal error path requeue.

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -1161,6 +1174,56 @@ func (r *FaultRemediationReconciler) trySkipResolvedEvent(
...
+		return ctrl.Result{}, fmt.Errorf("failed to get remediation state for retry check: %w", err), true
+	}
+
+	groupState, exists := remediationState.EquivalenceGroups[effectiveEquivalenceGroup]
+	if !exists {
+		// First attempt for this group
+		return ctrl.Result{}, nil, false
+	}
+
+	if groupState.RetryCount >= maxRetries {
```
KaivalyaMDabhadkar: Small naming thing: RetryCount is 1 after the first attempt and the check is >=, so maxRetryAttempts=N means N total attempts (N-1 retries), and maxRetryAttempts=1 means no retries at all. The field name, the "unlimited retries" comment, and the "Maximum retry attempts exceeded" log (which fires when the count is reached, not exceeded) all suggest otherwise. Something like maxRemediationAttempts plus a sentence on how counting works would help, and it also avoids confusion with the existing updateRetry.maxRetries in the same config.

fault-remediation/pkg/annotation/annotation.go
```diff
@@ -104,11 +104,19 @@ func (m *NodeAnnotationManager) UpdateRemediationState(ctx context.Context, node
 			return err
 		}
 
+		// Increment retry count if this group already exists
+		existingGroup, exists := state.EquivalenceGroups[group]
```
KaivalyaMDabhadkar: the exists check isn't needed, a missing key gives the zero struct, so this whole block can just be retryCount := state.EquivalenceGroups[group].RetryCount + 1.

KaivalyaMDabhadkar:
@iacker the lint checks are failing in the CI, Can you also check and fix them too?

fault-remediation/pkg/reconciler/reconciler.go
iacker: yeah, you're right. checkExistingCRStatus deletes the group entry before this reads it, so the count never survives past 1 and nothing gets capped. I'll move the cap into evaluateExistingCR's CRStateFailed branch where the old RetryCount is still in hand, and carry it over on recreation. That also drops the stale cached read at 1188 since it reuses the state already fetched.

iacker:
Fixed the lint failures in fault-remediation. gocyclo tripped on handleRemediationEvent, so I moved the CR-status handling into its own helper, plus the wsl whitespace fixes. golangci-lint v2.5.0 runs clean locally and the reconciler and annotation tests still pass.

KaivalyaMDabhadkar:
> Fixed the lint failures in fault-remediation. gocyclo tripped on handleRemediationEvent, so I moved the CR-status handling into its own helper, plus the wsl whitespace fixes. golangci-lint v2.5.0 runs clean locally and the reconciler and annotation tests still pass.

@iacker just a friendly reminder, when you have a chance, could you please address the remaining comments as well? Thank you!

iacker:
Thanks for the detailed review, all nine comments are addressed in 7650094.

You were right that the cap could never fire, and the root cause made most of the other comments collapse into one fix. Rather than carrying the count through `evaluateExistingCR`, I made `RemoveGroupsFromState` drop the CR reference but keep `AttemptCount`. Groups are removed exactly when their CR failed or vanished, which is the loop the cap exists to stop, so the budget has to outlive that removal. `ClearRemediationState` still wipes it when the quarantine session actually ends, which keeps the reset-per-session behaviour you and @lalitadithya agreed on in #1543.

That also resolved the two comments that depended on it: the stale cached read is gone (the check now uses `RecordRemediationAttempt`, no second `GetRemediationState`), and the node-deleted branch now mirrors the terminal path at lines 1045-1048 instead of falling through to remediate a node that no longer exists.

On counting: `RecordRemediationAttempt` runs *before* the CR is created, so attempts that never produce one (missing CRD, RBAC denial, rejecting webhook) consume budget too.

The node is now labelled `remediation-failed` on every capped event, not once, since fault-quarantine and node-drainer re-stamp it. I routed it through the existing `markEventTerminalAndProcessed` rather than adding a shared helper, so the status write and the label already live together.

Renamed to `maxRemediationAttempts` / `AttemptCount`, and the doc comment now states plainly that N means N attempts, not N retries.

Wiring: added to `values.yaml` and `configmap.yaml`, above the first section header. I added a config test that pins this, one case asserts that the same key placed after `[updateRetry]` decodes to 0 — the silent failure you warned about now breaks a test instead of shipping.

On tests, you were right that the annotation-level ones were misleading, so I deleted `retry_count_test.go` (including the restart test, which as you noted only called a stateless manager twice). Replaced with `max_attempts_e2e_test.go` covering your five scenarios against envtest, and `MockNodeAnnotationManager`'s `RemoveGroupsFromState` / `UpdateRemediationState` are no longer no-ops, so unit tests can now catch this class of bug.

Verified locally: full `fault-remediation` suite passes, and golangci-lint v2.13.1 (the pinned version) reports 0 issues.

One judgement call worth flagging: keeping a counter-only annotation entry means a group can exist with an empty `MaintenanceCR`. I guarded the two places that would misread it — the CR evaluation loop skips it, and `closeStaleEquivalentEvents` no longer treats it as covered, which would otherwise have closed events this session never remediated. Happy to switch to carrying the count through `existingCRDecision` if you'd rather not have that state shape.

fault-remediation/pkg/reconciler/reconciler.go
iacker: Fixed at the root. RemoveGroupsFromState now drops the CR reference but keeps AttemptCount, so the count survives the removal that the cap exists to stop, and ClearRemediationState still wipes it when the session ends. No plumbing through evaluateExistingCR.

fault-remediation/pkg/reconciler/reconciler.go
iacker: Gone. The check reads the count from RecordRemediationAttempt instead of a second GetRemediationState, so there's no stale cached read anymore.

fault-remediation/pkg/config/config.go
iacker: Wired up in values.yaml and configmap.yaml, above the first section header. Added a config test where the same key placed after [updateRetry] decodes to 0, so the silent-ignore trap you flagged breaks a test now instead of shipping.

fault-remediation/pkg/annotation/annotation.go
iacker: RecordRemediationAttempt runs before createMaintenanceCR now, so attempts that never produce a CR (missing CRD, RBAC denial, rejecting webhook) consume budget too.

fault-remediation/pkg/reconciler/reconciler.go
iacker: Node is labelled remediation-failed on every capped event now, routed through the existing markEventTerminalAndProcessed so the status write and the label stay together, and re-stamped each time since fault-quarantine and node-drainer overwrite it.

fault-remediation/pkg/annotation/retry_count_test.go
iacker: Deleted retry_count_test.go and added max_attempts_e2e_test.go covering your five scenarios against envtest. MockNodeAnnotationManager's RemoveGroupsFromState and UpdateRemediationState are no longer no-ops, so the mocks can catch this now.

fault-remediation/pkg/reconciler/reconciler.go
iacker: Now mirrors the terminal path at 1045-1048 instead of falling through to remediate a node that's already gone.

fault-remediation/pkg/reconciler/reconciler.go
iacker: Renamed to maxRemediationAttempts / AttemptCount, and the doc comment now says plainly that N means N attempts, not N retries.

fault-remediation/pkg/annotation/annotation.go
iacker: Done, it's just groupState := state.EquivalenceGroups[group]; groupState.AttemptCount++ now, no exists guard (annotation.go:164).

fault-remediation/pkg/reconciler/reconciler.go
```diff
@@ -1039,13 +1039,64 @@ func (r *FaultRemediationReconciler) handleRemediationEvent(
...
+		watcherInstance, healthEventStore, nodeName)
+	if done {
+		return res, err
+	}
+
+	// Stop remediating this group once it has spent its attempt budget for the session.
+	// groupConfig is non-nil here: trySkipEvent finalizes events without a group config.
+	maxAttempts := r.Config.RemediationClient.GetConfig().MaxRemediationAttempts
+	if maxAttempts > 0 {
+		res, err, done := r.tryStopAtMaxAttempts(ctx, nodeName, groupConfig.EffectiveEquivalenceGroup,
```
KaivalyaMDabhadkar: Really nice fix overall, I think just one interaction slipped through: the attempt counter increments on every pass through handleRemediationEvent, but while a log-collector job is running, RunLogCollectorJob requeues the same event every 10 seconds, and each requeue counts as a new attempt. That means with logCollector.enabled=true, a slow log collection could use up the whole budget and mark the node remediation-failed before any CR is even created. Would it make sense to record the attempt after log collection finishes but still before performRemediation, and maybe add an e2e case with EnableLogCollector=true to cover it?

KaivalyaMDabhadkar: @iacker friendly reminder regarding the above

iacker:
Fixed in 3b17b89. The remediation attempt is now recorded only after log collection completes. Log-collector polling requeues no longer consume the per-equivalence-group budget. Added an envtest regression covering two collector requeues with EnableLogCollector=true. Focused reconciler tests pass locally.

KaivalyaMDabhadkar:
> Fixed in [3b17b89](https://github.com/NVIDIA/NVSentinel/commit/3b17b89ebf0615821c4f9c15057d3ad3f1280c94). The remediation attempt is now recorded only after log collection completes. Log-collector polling requeues no longer consume the per-equivalence-group budget. Added an envtest regression covering two collector requeues with EnableLogCollector=true. Focused reconciler tests pass locally.

Thanks! The PR looks good to me now. The only thing remaining thing is the DCO check failing because the newest commit 3b17b89ebf0615821c4f9c15057d3ad3f1280c94 is missing its Signed-off-by line. Since it is the head commit, a quick git commit --amend -s --no-edit and a force push will fix it. Can you please push it?

KaivalyaMDabhadkar:
@iacker Friendly reminder regarding the unsigned commit ^^

KaivalyaMDabhadkar (approved):

PR #1646, opened 2026-08-23
fix(labeler): serialize per-node label reconciliation and fix other CI flakes

XRFXLP:
### Problem
During labeler startup, `reconcileAllNodes` can read that a driver pod exists, then race with its deletion. The delete handler correctly removes `driver.installed`, but the startup sweep can subsequently restore `driver.installed=true` using its earlier result. Because this output label is not a watched reconciliation input, it can remain stale when no replacement pod event occurs. This caused [`TestLabeler_handlePodEvent/driver_pod_deletion_removes_driver_label`](https://github.com/NVIDIA/NVSentinel/actions/runs/32498643196/job/96822964327?pr=1635) to fail.

### What is the fix?
Serialize reconciliation for the same node across startup/node reconciliation and pod add, update, and delete handlers. The shared lock covers both desired-state calculation and the write, so whichever path runs second recalculates from the latest informer state and produces the final result. A fixed set of striped locks keeps memory bounded while allowing different nodes to reconcile concurrently.

This replaces the earlier periodic-resync approach, which could not repair the observed test failure before its timeout and only bounded the stale-label duration instead of preventing the lost update.

As a small CI-stability adjustment, expand window for valid throughputs on raising QPS from 4 -> 40 to [6, 14] instead of [8, 11] as we were getting lot of glitches for ratio of high-qps-throughput / low-qps-throughput outside of the range.

### Validation of the fix
- Added a deterministic regression test that pauses node reconciliation after it reads the pod cache, starts the pod deletion, and verifies the deletion result wins.
- The deterministic regression test passed 10/10 under `-race`.
- The original failing `driver_pod_deletion_removes_driver_label` envtest passed 10/10.

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved driver-label reconciliation during concurrent node and pod changes.
  * Ensured label updates remain consistent when pods are added, updated, or removed.
  * Reduced unnecessary reconciliation when relevant inputs have not changed.

* **Tests**
  * Added coverage for concurrent reconciliation and pod deletion scenarios.
  * Broadened accepted throughput ranges for node draining, fault quarantine, labeling, and gang coordination tests.


labeler/pkg/labeler/labeler.go
```diff
@@ -823,16 +825,33 @@ func (l *Labeler) handleNodeEvent(obj any) error {
...
-		metrics.NodeUpdateFailures.Inc()
+}
 
-		return fmt.Errorf("failed to update labels for node %s: %w", nodeName, err)
+func (l *Labeler) withNodeLock(nodeName string, fn func() error) error {
+	hash := uint32(2166136261)
+	for idx := range len(nodeName) {
+		hash ^= uint32(nodeName[idx])
+		hash *= 16777619
 	}
```
lalitadithya: Is this FNV-1 or FNV-1a? If yes, can we use https://pkg.go.dev/hash/fnv so that we don't have to worry about if we got the logic correct?

Also, should we consider using https://pkg.go.dev/hash/maphash instead?

XRFXLP: that was fnv-1a, using maphash direclty now.

lalitadithya (approved):

PR #1648, opened 2026-08-24
feat(labeler): replace GET + PUT to apiserver with PATCH

### LINKED ISSUE #1596: [Feature]: Replace node GET + full PUT with PATCH in Fault Quarantine and Labeler
### Prerequisites

### Code of Conduct

### Feature Summary

Fault Quarantine and Labeler each write node objects with a GET followed by a full PUT. The GET exists solely to fetch `resourceVersion` for conflict detection. PATCH doesn't need it — it applies only the changed fields directly against etcd. Switching halves the API calls per node write and doubles throughput at any QPS setting.

### Problem/Use Case

Both components follow the same pattern: GET the node to obtain `resourceVersion`, mutate the object in memory, then PUT the entire object back. This costs 2 API calls per node write.

`PUT` (client-go's `Update()`) requires a `resourceVersion` to detect write conflicts. Without a prior GET, the API server rejects the PUT. `PATCH` with `application/merge-patch+json` applies only the changed fields — no `resourceVersion` needed, no prior GET.

FQ cordons nodes by setting taints and `Unschedulable`. Labeler writes GPU labels. Both own distinct field namespaces — nothing else writes FQ's taints or Labeler's `nvidia.com/*` labels concurrently — so PATCH conflict detection is not needed.

At default QPS=5 with 2 calls per node:
- FQ cordons at 2.5 nodes/s
- Labeler labels at ~1 node/s → cold-relabel of a 100k node fleet takes ~28 hours

### Proposed Solution

Replace `Nodes().Get()` + `Nodes().Update()` with `Nodes().Patch()` using `types.MergePatchType`. Send only the fields being changed — taints + `spec.unschedulable` for FQ; the label map for Labeler.

**Relevant files**:
- `fault-quarantine/pkg/informer/k8s_client.go:61-68`
- `labeler/pkg/labeler/labeler.go:626,682`

**Expected improvement**: API calls per node write drop from 2 to 1. FQ: 2.5 → 5 nodes/s. Labeler: ~1 → ~2 nodes/s, cutting cold-relabel time from ~28h to ~14h at 100k nodes.

### Component

Fault Management

XRFXLP:
### What was happening?

The Labeler reconciles node labels in response to pod events, node events, and periodic reconciliation. Before this change, each reconciliation fetched the latest Node directly from the API server and then sent a full-object \`PUT\` when labels needed to change. That meant a changed Node cost two API calls (\`GET\` + \`PUT\`), and even a reconciliation that discovered no changes still paid for the \`GET\`.

This was unnecessary because the Labeler already maintains informer caches containing the pod and Node data needed to calculate its desired labels. At scale, repeatedly bypassing those caches added avoidable API-server traffic to a path that runs frequently.

### What is the fix?

The reconciliation flow now behaves as follows:

1. The Labeler calculates the desired labels from its informer-backed view, just as before.
2. Once all informers are synced, it supplies the cached Node to a shared \`kubeclient.NodePatcher\`. If the cache is not ready or the Node is unavailable there, the patcher safely falls back to one live \`GET\`.
3. The patcher deep-copies that Node and lets the Labeler mutate only the desired labels.
4. \`NodeMergePatch\` builds metadata-only projections of the original and desired Nodes, then uses Kubernetes' \`strategicpatch.CreateTwoWayMergePatch\` utility to calculate the difference. Added or changed keys carry their new value, removed keys are emitted as \`null\`, and unrelated labels and annotations are omitted.
5. If the comparison finds no change, the operation stops without writing to the API server. With a healthy cache, this means a no-op reconciliation makes zero API calls.
6. If labels changed, the Labeler sends one merge \`PATCH\` instead of a full-object \`PUT\`. With a healthy cache, a changed reconciliation therefore makes exactly one API call.

The patch deliberately compares metadata-only projections. Informer transforms may remove large or irrelevant Node fields, and diffing the complete projected Node could treat those omissions as desired changes. Supplying only labels and annotations to the Kubernetes patch utility guarantees that fields outside Labeler's scope remain untouched. Spec fields such as taints are excluded because patching a list replaces it as a whole.

### Why not Server-Side Apply with Force?

Server-Side Apply would add field-ownership tracking, but it would not improve the API-call behavior targeted by this change: without retaining the same local diff, even an unchanged reconciliation would still send an APPLY request. Existing installations also created these labels through UPDATE or PATCH rather than a stable SSA field manager, so adopting SSA would require an ownership-migration policy. Using \`Force\` during normal reconciliation could take a label key away from another manager instead of reporting the conflict.

This concern is stronger for the stacked Fault Quarantine change. Kubernetes treats \`spec.taints\` as one atomic field, so an SSA manager cannot own only the FQ taint; forcing ownership can replace the complete taint list or compete with another actor's cordon state. A merge patch also replaces the taint list, so FQ protects that operation with \`resourceVersion\`: if another controller changes the Node after FQ reads it, the patch conflicts, FQ refreshes the live Node, reapplies only its intended mutation, and retries with the other taints preserved.

### How does the cache remain safe after a write?

A successful PATCH can briefly be newer than the informer cache. \`NodePatcher\` remembers the resource version returned by the API server. Until the informer observes that version, a subsequent reconciliation refreshes the Node with a live \`GET\` instead of computing another patch from stale cached data. The marker is retained if that refresh fails, so a transient read failure cannot cause the next attempt to trust stale state.

When Kubernetes reports a conflict, the patcher fetches the current live Node, runs the Labeler mutation again, rebuilds the diff, and retries. Other retryable API errors use bounded backoff. This keeps retry and cache-freshness mechanics in the shared helper while the Labeler remains responsible only for calculating desired labels.

Pod-driven and node-driven updates for the same Node are also serialized through striped per-node locks. This prevents a reconciliation based on an older pod view from racing with a pod deletion and restoring a label that the deletion correctly removed.

### Resulting API behavior

- Cached Node with no label changes: **0 API calls**
- Cached Node with label changes: **1 PATCH**
- Cache unavailable or known to lag a completed write: **1 GET**, followed by a PATCH only when needed
- Conflict: **refresh live Node, recompute the diff, and retry**

### Validation of the fix

The tests verify the behavior at each layer:

- Kubernetes strategic-patch generation for added, changed, and removed label/annotation keys
- No patch for an unchanged Node
- Preservation of fields omitted by informer transforms
- One PATCH and no GET/PUT when a synced cached Node needs a change
- Zero API actions for a cached no-op reconciliation
- Live refresh when a completed write has not yet appeared in the informer cache
- Preservation of the pending-write marker when that live refresh fails
- Conflict refresh and retry from the latest Node
- Eventual convergence when informer delivery is deliberately delayed
- Correct ordering when a driver-pod deletion races with reconciliation

### Benchmark

Cluster: ~99,000 KWOK nodes (lightweight simulated nodes registered with the API server, used for scale testing). Both labelers ran at `--kube-api-qps=5 --kube-api-burst=10`. 1,500 nodes were pre-labeled with an incorrect kata label to force a write on reconcile. Throughput was measured by counting label corrections over a 5-minute window during the cold-start sweep.

The node informer fires one `AddFunc` reconcile per node on startup. On main, each no-op reconcile (nothing to change) still pays a `GET`. With 99k nodes at QPS=5, the no-op sweep consumes most of the budget and writes trickle in at a rate proportional to `bench_nodes / total_nodes`. The 1596-patch eliminates the `GET` for no-ops entirely, leaving the full budget for writes.

**Throughput**

| branch | corrections / 5 min | rate |
|---|---|---|
| main (v1.20.0) | 45 | 0.15 /s |
| 1596-patch | 261 | 0.87 /s (5.8x) |

**Memory**

| branch | cold-start peak | steady-state |
|---|---|---|
| main (v1.20.0) | ~2.9 GB | 1.34 GB |
| 1596-patch | ~2.06 GB | 1.25 GB |

The cold-start peak gap (~840 MB) reflects main's GET+discard pattern: each no-op fetches a full node object (~50 KB) and immediately discards it, generating sustained GC pressure across 99k nodes. The 1596-patch reads from the cached object and allocates nothing for no-ops.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: Labeler and shared Kubernetes client utilities

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review


commons/pkg/kubeclient/nodepatch.go
lalitadithya, line 1: Can SSA + Force be used here by leveraging https://pkg.go.dev/k8s.io/client-go/applyconfigurations? If not possible, can we document somewhere (PR description is fine) as to why?

commons/pkg/kubeclient/nodepatch.go
```diff
@@ -0,0 +1,217 @@
...
+// The patch is assembled key by key rather than by marshalling modified, because
+// marshalling a Node emits every populated field. A caller that read original from an
+// informer cache holding a projected Node would then patch the projection's gaps back
+// over the real object. Emitting only keys that differ means fields missing from both
+// sides are left untouched.
+//
+// Spec fields such as taints and unschedulable are deliberately out of scope: a merge
+// patch replaces a list wholesale, so patching taints from a projected Node whose Spec
+// had been cleared would silently drop every taint on the real object.
+func NodeMergePatch(original, modified *v1.Node) ([]byte, error) {
```
lalitadithya, line 171: can we reuse something from https://pkg.go.dev/k8s.io/apimachinery/pkg/util/strategicpatch or https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Patch to avoid having custom logic to build the patch?

lalitadithya (commented):
Can we quantify what is the difference at scale like we have been doing in the previous perf improvement PRs thus far?

commons/pkg/kubeclient/nodepatch.go
XRFXLP: done.

commons/pkg/kubeclient/nodepatch.go
XRFXLP: we can do in labeler, but FQ would be a bit restrictive but both won't have any gain. Added that in description on why we're not using it.

lalitadithya (approved):

PR #1654, opened 2026-08-24
feat: add audit mode and flags to fail open when auth backend is ureachable

### LINKED ISSUE #1653: [Feature]: platformConnectorAuth needs an audit-only mode, and arguably should not default to enabled in its first release
### Prerequisites

### Code of Conduct

### Feature Summary

`platformConnectorAuth` (#1636, new in v1.20.0) has no observe-only mode, so operators cannot verify it works in their environment before it starts rejecting health events. Please add an audit-only setting that logs and counts violations without rejecting, and consider defaulting `enabled` to `false` for one release.

### Problem/Use Case

The feature is a genuine improvement and we want it. The concern is purely about how it can be adopted safely.

**There is no way to prove it before trusting it.** Looking through `platform-connectors/pkg/auth/`, there is no `permissive`, `auditOnly` or `warnOnly` option: the interceptor either enforces or is off. So the only way to find out whether audience wiring, TokenReview RBAC and the cross-node publisher list are all correct in a given cluster is to enable enforcement in production and watch whether health events stop.

**The failure mode is silent loss of detection.** The values comment for the disabled path reads "not a supported production configuration", but the feature did not exist before v1.20.0, so neither path has production history yet.

**A control-plane blip rejects events.** In `nodebinding.go`, `codes.Unavailable`, `DeadlineExceeded` and `Canceled` all reject. The code distinguishes them from a forged token only in a metric label:

```go
codes.Unauthenticated:  reasonTokenInvalid
codes.Unavailable:      reasonValidatorUnavailable
codes.DeadlineExceeded: reasonValidatorTimeout
```

That couples detection availability to API server availability. On a large cluster with rolling maintenance, brief API server unavailability is routine, and the result is dropped health events rather than degraded authentication.

**The codebase already anticipates this.** From `platform-connectors/main.go`:

> Silently reading either as "no cross-node publishers" would start the connector in a configuration where every cluster-scoped monitor is pinned to one node and its events rejected — a failure that surfaces far from its cause.

That is exactly the risk operators face on upgrade, and the mitigation currently offered is a startup error on one specific misconfiguration rather than a way to observe the whole feature safely.

**Adoption is not a boolean flip.** It also requires a ConfigMap key holding a JSON array of canonical ServiceAccount usernames, which is a hard error when missing or null. Cluster-scoped publishers must be enumerated correctly or they are pinned to one node and rejected.

### Proposed Solution

1. **Add an audit-only mode**, for example `platformConnectorAuth.mode: audit|enforce` (or `enforcement: false`). In audit mode, validate every request and increment the existing violation counters by reason, but let the event through. Operators can then run it for a period, confirm the violation counters stay at zero, and flip to enforce with evidence.

2. **Consider defaulting `enabled: false` for one release**, then flipping the default once the feature has field history. As written, upgrading to v1.20.0 silently changes the trust model between every monitor and the connector in the same change as everything else in the release, which makes any resulting event loss hard to attribute.

3. **Distinguish validator-unavailable from credential-invalid in behaviour, not only in metrics.** A forged token should be rejected; an unreachable API server is arguably a reason to fail open, or at least a separately configurable choice, since it says nothing about the caller.

For context on why the silent-loss mode concerns us specifically: we recently spent several days diagnosing a case where our health-events-analyzer received zero events for a week because of a `processingStrategy` mismatch. Everything reported healthy throughout. Adding a second, security-shaped way for events to disappear silently is what prompted this request.

We are deploying v1.20.0 with `enabled: false` and enabling it separately once verified, so this is not blocking us. Filing it because an audit mode would make the feature straightforwardly adoptable for everyone else.

### Component

Core Service

lalitadithya:
Closes #1653

## Summary

* Add audit mode option to the authentication system 
* Add flag to fail open if auth backend can't be reached

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [X] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added configurable authentication modes: `enforce` (default) and `audit`.
  - Added optional fail-open handling for unavailable or timed-out credential validators.
  - Unavailable validators may allow degraded node-local access; cross-node requests remain retryable.
  - Audit mode records eligible violations without rejecting requests.
  - Invalid credentials and missing node names remain rejected in all modes.
  - Added Kubernetes configuration settings and strict validation for authentication controls.

- **Documentation**
  - Documented authentication modes, validator availability handling, defaults, configuration options, and metrics.


docs/METRICS.md
```diff
@@ -168,7 +168,7 @@ submitting health events naming another node. See
...
...s_node_claim_absent` (pod-bound, but the pod was never scheduled), `missing_node_name` (no node name on the event and none could be stamped), `token_invalid` (TokenReview rejected the token), `malformed_credentials` (the authorization header was duplicated or did not use the Bearer scheme). Three further reasons mean **no verdict could be reached**, not that the caller was rejected: `validator_unavailable` (the API server was unreachable), `validator_timeout` (the caller gave up or its deadline passed) and `validator_error` (an unexpected failure, typically this component's own RBAC to create TokenReviews). With `failOpenOnUnavailable: true`, `validator_unavailable` and `validator_timeout` still increment this counter but fall back to node-local scope instead of rejecting the request, in enforce mode too — an outage says nothing about the caller's credential. Exclude those three when alerting on suspected credential abuse — a control-plane outage increments them for every in-flight request and would otherwise look identical to an attack: `sum(rate(platform_connector_auth_violations_total{reason=~"node_mismatch\|token_invalid\|node_claim_mismatch\|unbound_cross_node_token"}[5m]))` |
```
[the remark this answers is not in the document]
lalitadithya: done

platform-connectors/pkg/auth/nodebinding.go
```diff
@@ -314,7 +386,15 @@ func (b *nodeBinder) resolveScope(ctx context.Context) (scope, error) {
 
 	identity, err := b.validator.Authenticate(ctx, token)
 	if err != nil {
-		b.recordViolation(violationReasonFor(err))
+		reason := violationReasonFor(err)
+		b.recordViolation(reason)
+
+		if b.failOpenOnUnavailable && isValidatorUnavailable(reason) {
```
KaivalyaMDabhadkar, line 416: I think there is a gap here for crossnode publishers. When the apiserver is down and this flag is on, we return node-local scope before the allowlist is ever checked. So a legitimate batch from csp HM naming another node still gets rejected below in validateBatch but now with PermissionDenied instead of Unavailable. All our publishers treat PermissionDenied as non-retryable, so events that used to survive an outage through retries are now dropped for good. It also records these as node_mismatch, which is one of the reasons METRICS.md tells users to alert on as suspected credential abuse, so an outage would look like an attack on a dashboard.

Could the fail-open path remember that the scope is degraded, and keep returning a retryable Unavailable for events that need cross-node scope instead of PermissionDenied? The current tests only cover fail-open with events naming the local node, so this case is not caught

platform-connectors/pkg/auth/nodebinding.go
```diff
@@ -277,15 +323,15 @@ func (b *nodeBinder) intercept(
...
+	if err := b.auditOrReject(ctx, err); err != nil {
 		return nil, err
 	}
 
 	authDecisions.WithLabelValues(callerScope.String()).Inc()
 
 	// Validate the entire batch before mutating any of it, so a rejected batch
 	// leaves no partially-stamped events behind.
-	if err := b.validateBatch(ctx, events, callerScope); err != nil {
+	if err := b.auditOrReject(ctx, b.validateBatch(ctx, events, callerScope)); err != nil {
```
KaivalyaMDabhadkar: In audit mode this swallows the missing_node_name violation, but the stamping below only runs for node-local scope. So a crossnode caller that sends a blank nodename gets its event forwarded with an empty node name. Nothing downstream can handle that currently: the store inserts it as is, fault-quarantine fails GetNode("") and drops the event, and the k8s connector fails the whole batch without retry. Audit mode is meant to be the safe rollout option, but this lets records into the datastore that enforce mode could never produce. Maybe audit should still reject (or skip) events with no node name, since letting them through would not be useful

KaivalyaMDabhadkar:
Couple of doc fixes needed: docs/platform-connectors.md still says there is no mode (line 300), that the connector fails closed with Unavailable (line 282), and that violations_total counts rejected batches (line 387), and its config table does not list the two new keys. Also the grpcSinkConnector.tokenPath comment in values.yaml (around line 340) argues that chaining connectors cannot authenticate because the receiver rejects mismatched node claims, which is no longer true under audit mode. Since these pages are what people will follow for rollouts, can you update in this PR?

platform-connectors/pkg/auth/nodebinding.go
lalitadithya: good catch, updated, please take a look

platform-connectors/pkg/auth/nodebinding.go
lalitadithya: good catch, updated, please take a look

KaivalyaMDabhadkar (approved):

PR #1657, opened 2026-08-25
feat(fault-quarantine): replace GET + PUT with PATCH to kubernetes apiserver

### LINKED ISSUE #1596: [Feature]: Replace node GET + full PUT with PATCH in Fault Quarantine and Labeler

XRFXLP:
### What was happening?

Fault Quarantine updates several parts of a Node while quarantining or recovering it: taints, `spec.unschedulable`, labels, and annotations. Previously, each update read the live Node from the API server and then sent a full-object `PUT`. A changed Node therefore required a `GET` followed by an `UPDATE`, and even a reconciliation that found no changes still paid for the `GET`.

A full update also makes concurrency broader than necessary. In particular, taints are represented as a list, so an update calculated from stale state can replace a concurrent taint change made by another controller.

### What is the fix?

`FaultQuarantineClient.UpdateNode` now delegates Node mutation mechanics to the shared `kubeclient.NodePatcher` introduced by the Labeler change. The flow is:

1. Serialize FQ operations for the same Node.
2. Use the synced Node informer cache as the starting state.
3. Deep-copy that Node and apply the FQ mutation callback.
4. Project only the fields this helper intentionally supports: labels, annotations, taints, and `unschedulable`.
5. Use Kubernetes' `strategicpatch.CreateTwoWayMergePatch` utility to calculate the difference.
6. Send one PATCH when a change is required.

The explicit patchable-field projection is narrower than FQ's informer transform. FQ's cache removes only `Status` and retains the complete metadata and `Spec`; the projection prevents a future mutation callback from accidentally adding unrelated Node fields to the patch.

### How are concurrent updates protected?

Labels and annotations are map fields, so the patch contains only changed keys and does not require an optimistic-lock precondition. Taints are different: the taint list is replaced as a whole. Whenever taints or `unschedulable` change, the patch includes the original `metadata.resourceVersion`.

If another controller modifies the Node after FQ read it, Kubernetes rejects the stale PATCH with a conflict. `NodePatcher` then fetches the latest live Node, reruns the FQ mutation, rebuilds the patch, and retries. For example, if another controller adds a taint between FQ's read and PATCH, the retry starts from that new taint list and preserves both changes instead of overwriting the concurrent taint.

After a successful write, `NodePatcher` remembers the returned resource version. Until the informer observes that version, the next mutation uses a live read rather than trusting stale cached state. A failed refresh retains this marker so a transient read failure cannot make a later attempt reuse the stale cache.

### How are taints reconciled?

Kubernetes defines taint identity by `Key` and `Effect`; `Value` is mutable data and is not part of uniqueness. FQ now indexes taints by that pair:

- An incoming taint with the same Key/Effect updates the existing value in place.
- Duplicate Key/Effect entries are collapsed rather than appended.
- A taint with a genuinely new Key/Effect pair is appended.
- Existing metadata such as `TimeAdded` is preserved when the value changes.
- The same key with a different effect remains a distinct taint.

This prevents configurations such as `fault=old:NoSchedule` and `fault=new:NoSchedule` from producing two entries for what Kubernetes considers one taint.

### Why not Server-Side Apply with Force?

Kubernetes treats `spec.taints` as one atomic SSA field, not as independently owned taint entries. An FQ field manager therefore cannot own only the taint it added. Using `Force` could take ownership of and replace the complete taint list maintained by the kubelet or another controller. `spec.unschedulable` is also shared with tools such as `kubectl cordon` and `kubectl drain`. The resourceVersion-guarded PATCH approach detects concurrent changes and recomputes from current live state instead of overriding another manager's intent.

### Resulting API behavior

- Cached Node with no changes: **0 API calls**
- Cached Node with changes: **1 PATCH**
- Cache unavailable or known to lag a completed write: **1 GET**, followed by a PATCH only if needed
- Concurrent spec change: **conflict, live refresh, recompute, and retry**

### Validation

Coverage includes:

- Strategic patch generation for metadata, taints, and scheduling state
- Conditional `resourceVersion` inclusion for spec changes
- Patchable-field projection and preservation of unrelated fields
- Cached PATCH-only behavior and zero-request no-op behavior using the client-go fake client
- A deterministic envtest conflict caused by a concurrent taint update, verifying both taints survive the retry
- Key/Effect taint deduplication, value updates, distinct effects, new pairs, and `TimeAdded` preservation
- Full Fault Quarantine informer tests and lint

### Scale benchmark

Measured on a KWOK cluster: 1000 GPU nodes (`nvidia.com/gpu.present=true`), `--kube-api-qps=5 --kube-api-burst=10`, 1000 fatal GPU health events injected simultaneously into MongoDB. Wall-clock time measured from event injection until all 1000 nodes had `spec.unschedulable=true`.

| | v1.20.0 | this PR | improvement |
|---|---|---|---|
| Time to cordon 1000 nodes | 418 s (7.0 min) | 203 s (3.4 min) | **2.1× faster** |
| Peak memory | 1.55 G | 1.32 G | **15% lower** |
| Settled memory (idle) | ~990 M | ~960 M | ~same |
| Peak CPU | ~40 m cores | ~10 m cores | **~4× lower** |

The speedup matches the theoretical model exactly: v1.20.0 performs `GET + UPDATE` per node (2 API calls × 200 ms/call at QPS=5 = ~400 ms/node × 1000 nodes ≈ 400 s). This PR performs a single `PATCH` per node (1 API call × 200 ms = ~200 ms/node × 1000 nodes ≈ 200 s).

The lower CPU peak reflects the same reduction: half the HTTP round-trips means half the serialization work. The lower memory peak comes from eliminating the GET response allocation (one full deserialized Node object per event, ~50–100 KB each) that v1.20.0 held alongside the update body.

**Memory** (v1.20.0 left, this PR right):
image

**CPU** (v1.20.0 left, this PR right):
image

## Type of Change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [x] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: shared Kubernetes client utilities

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node updates by retrying transient Kubernetes service failures and resolving concurrent update conflicts more reliably.
  * Preserved accumulated annotations and taints across successive updates.
  * Prevented unnecessary updates when no node changes are required.
  * Improved taint handling by matching taints by key and effect, updating changed values, removing duplicates, and preserving existing metadata.
  * Added support for safely updating node labels, annotations, taints, and scheduling state.


fault-quarantine/pkg/informer/k8s_client.go
```diff
@@ -376,29 +362,25 @@ func (c *FaultQuarantineClient) applyTaints(
...
-			existingTaints[key] = v1.Taint{
+			node.Spec.Taints = append(node.Spec.Taints, v1.Taint{
 				Key:    taintConfig.Key,
 				Value:  taintConfig.Value,
 				Effect: v1.TaintEffect(taintConfig.Effect),
-			}
+			})
+			existingTaints[key] = struct{}{}
 		}
 	}
```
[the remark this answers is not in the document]
XRFXLP: should be handled

fault-quarantine/pkg/informer/k8s_client.go
```diff
@@ -376,27 +362,45 @@ func (c *FaultQuarantineClient) applyTaints(
...
-			slog.InfoContext(ctx, "Tainting node", "node", nodename, "taintConfig", taintConfig)
-			existingTaints[key] = v1.Taint{
-				Key:    taintConfig.Key,
-				Value:  taintConfig.Value,
-				Effect: v1.TaintEffect(taintConfig.Effect),
+		identity := taintIdentity{key: taintConfig.Key, effect: v1.TaintEffect(taintConfig.Effect)}
+		if index, exists := existingTaints[identity]; exists {
+			if node.Spec.Taints[index].Value != taintConfig.Value {
+				slog.InfoContext(ctx, "Updating node taint", "node", nodename, "taintConfig", taintConfig)
+				node.Spec.Taints[index].Value = taintConfig.Value
```
KaivalyaMDabhadkar, line 390: Updating the value in place puts the node and FQ's bookkeeping out of sync in a way the untaint detectors misread. After a v1 -> v2 update the node carries v2, but the applied-taints annotation keeps both entries, because mergeAppliedTaints keys on the full key/value/effect triple. hasTaint in node_informer.go also matches the full triple, so the stale v1 entry looks like a manually removed taint: detectAndHandleManualUntaint can fire on the update event, and checkStaleUntaintState fires deterministically on every FQ restart. handleManualUntaint then wipes all quarantine annotations and labels and cancels the quarantining events, while the node stays cordoned and tainted. With the annotations gone, prepareUncordonParams has nothing to work from, so no recovery event can ever clean the node up.
Two rulesets sharing a taint key and effect with different values is legal config today (there is no validation against it), so a single event matching both rulesets is enough to create the mismatch. Before this PR the same input failed loudly at the apiserver (duplicate key+effect validation) so the silent teardown is new.
I think the value update needs to also replace the old entry in the applied-taints annotation, and the untaint detection should match on key+effect, so the update stays consistent end to end.

fault-quarantine/pkg/informer/k8s_client.go
```diff
@@ -161,50 +162,35 @@ func (c *FaultQuarantineClient) UpdateNode(ctx context.Context, nodeName string,
...
 	defer mu.(*sync.Mutex).Unlock()
 
-	// Increased retry attempts to handle node update conflicts when multiple modules
-	// attempt concurrent updates, preventing nodes from remaining cordoned with stale annotations.
-	backoff := wait.Backoff{
-		Steps:    10,                    // Increased from default 5
-		Duration: 20 * time.Millisecond, // Increased from default 10ms
-		Factor:   2.0,
-		Jitter:   0.1,
+	changed, err := c.nodePatcher.Patch(
```
KaivalyaMDabhadkar, line 165: the old UpdateNode ran its GET inside retry.OnError with 10 attempts, so a transient 429/timeout on the read was absorbed. NodePatcher.currentNode does its live GET once, before the retry loop, so a single transient error now fails the whole UpdateNode.
This can cause an issue at two places: during startup, checkStaleStateOnAdd runs before the cache is synced, so every cleanup call GETs live, and a failure there is only logged and not retried until the next restart. And in the event path, an UpdateNode error makes applyQuarantine return nil while the event watcher marks the event processed anyway, so the quarantine is dropped for good
I think wrapping the currentNode GETs in the same retry/backoff as the patch loop would restore the old resilience.

fault-quarantine/pkg/informer/k8s_client.go
```diff
@@ -161,50 +162,35 @@ func (c *FaultQuarantineClient) UpdateNode(ctx context.Context, nodeName string,
...
+		ctx,
+		c.Clientset.CoreV1().Nodes(),
+		nodeName,
+		c.cachedNodeForPatch(nodeName),
```
KaivalyaMDabhadkar, line 169: I think there's a window here where FQ can lose its own writes. The pendingVersions marker gets deleted after a single live GET (nodepatch.go:128, :138), but a new one is only stored when a patch is actually sent. So if a no-op update lands between two real writes (a duplicate event is enough), it burns the marker, and the next write trusts the informer cache even though the cache hasn't caught up with the previous write yet.
Annotation patches carry no resourceVersion, so that stale merge simply wins. It can drop health events written moments earlier, and removeEventFromAnnotation can come back empty and uncordon the node while another fault is still active (reconciler.go:1417). The old GET+PUT flow always read live, so this couldn't happen before.
I suggest to only delete the marker once the cache has actually caught up to the written version, not after the fallback GET. A test replaying this write/no-op/write sequence with a lagging cache would pin it

KaivalyaMDabhadkar (approved):

PR #1659, opened 2026-08-25
feat(mongodb-store): set oplog size with oplogSizeMB

### LINKED ISSUE #1594: [Feature]: Increase MongoDB oplog to cover 24-hour recovery window for change-stream consumers
### Prerequisites

### Code of Conduct

### Feature Summary

Four NVSentinel components resume from MongoDB change-stream tokens after a restart. If the token falls off the oplog before the component restarts, those components silently lose every event that occurred during the outage. At production write rates, the default oplog fills in under 2 hours.

### Problem/Use Case

FQ, Node Drainer, Fault Remediation, and Event Exporter all maintain change-stream resume tokens. A restart within the oplog window means seamless recovery. A restart after the token has been overwritten means `CappedPositionLost` — missed quarantines, missed drains, missed remediations, missed exports, with no alert and no retry.

At 500 health events/s (production rate at 100k nodes), the default 990 MB Percona oplog fills in approximately 1.57 hours. Any outage longer than that — a rolling update, a node eviction, a network partition — loses events.

The failure is silent. Components log a recovery message and start fresh from "now", with no indication of how many events were skipped.

### Proposed Solution

Set `replication.oplogSizeMB: 15120` in the Percona Helm chart. At 500 events/s and ~350 bytes/event, a 15 GB oplog provides a 24-hour recovery window. Apply equivalent sizing for Bitnami via PVC expansion.

**Expected improvement**: All four change-stream consumers can safely resume after outages up to 24 hours. `CappedPositionLost` errors are eliminated at production write rates.

### Component

Core Service

rupalis-nv:
Fixes #1594

## Summary
- Size the replica-set oplog with `mongodb-store.oplogSizeMB` (integer MiB, minimum 990, default 990 = MongoDB's floor). Helm does not derive this from the PVC.
- The same integer must be on mongod startup: Bitnami `mongodb.extraFlags --oplogSize=`, Percona `psmdb-db.replsets.rs0.configuration` `replication.oplogSizeMB`. Helm fails if they differ. Do not mount a snippet `mongodb.conf`.
- Init Job runs `replSetResizeOplog` on each member (`directConnection`). It does not shrink a larger live oplog unless `oplogAllowShrink` is true. Changing `oplogSizeMB` (or replica members / init script) creates a new Job `create-mongodb-database-<ttl>-<scriptHash>`.
- Percona resize uses `clusterAdmin` (`databaseAdmin` is not authorized for `replSetResizeOplog`).

### Bitnami — Helm CLI upgrade

| Check | Result |
| --- | --- |
| Helm | revision 10, deployed |
| ConfigMap `MONGODB_OPLOG_SIZE_MB` | **1400** |
| STS `MONGODB_EXTRA_FLAGS` | includes `--oplogSize=1400` (no snippet ConfigMap) |
| Mongo pods | rolled 2→1→0, then **3/3 Running**, PRIMARY on mongodb-0 |
| Live `oplog.rs` maxSize | **1400 MiB** on all 3 |

### Bitnami — ArgoCD upgrade

| Check | Before | After |
| --- | --- | --- |
| ConfigMap `MONGODB_OPLOG_SIZE_MB` | 1400 | **1500** |
| STS `--oplogSize` | 1400 | **1500** |
| Init Job | — | `create-mongodb-database-604800-f26fecf8` Complete |
| `replSetResizeOplog` | — | mongodb-0/1/2 logged `currentMB=1400 targetMB=1500` then `Oplog resized to 1500 MB` |
| Live `oplog.rs` maxSize | 1400 MiB | **1500 MiB** on all 3 |
| Mongo pods | 3/3 | rolled, then **3/3 Running**, PRIMARY on mongodb-0 |
| PVC | 12Gi Bound ×3 | **12Gi** Bound (unchanged) |

### Percona — Helm CLI 

| Check | Result |
| --- | --- |
| Percona CR | `status.state=ready` |
| URI | `mongodb-rs0` |
| Helm values | `oplogSizeMB: 1500`, Percona `replication.oplogSizeMB: 1500` |
| ConfigMap | `MONGODB_OPLOG_SIZE_MB=1500` |
| Init Job (grow 1400 → 1500) | `create-mongodb-database-604800-ae77baff` Complete |
| Live resize | rs0-0/1/2 `currentMB=1400` → `1500`; `maxSizeMB: 1500` |
| Init Job (shrink 1500 → 1400) | `create-mongodb-database-604800-0526994d` Complete; required `oplogAllowShrink: true` |
| Live shrink | rs0-0/1/2 `currentMB=1500` → `1400`; `maxSizeMB: 1400` |

### Percona — ArgoCD upgrade 

| Check | Result |
| --- | --- |
| Sync | Succeeded |
| ConfigMap | `MONGODB_OPLOG_SIZE_MB=1500` |
| Init Job | `create-mongodb-database-604800-ae77baff` Complete |
| `replSetResizeOplog` | rs0-0/1/2 `currentMB=1400` → `1500` |
| Live `oplog.rs` maxSize | **1500 MiB** on all 3 |
| Percona / URI | still `ready` / `mongodb-rs0` |

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review


distros/kubernetes/nvsentinel/values-full.yaml
```diff
@@ -1522,6 +1522,10 @@ mongodb-store:
   # In-cluster and external Mongo. Job name includes this value so helm/argo rerun collMod.
   collectionExpirySeconds: 2592000
 
+  # Oplog as a percent of the data PVC (1–50). Default 10.
+  # 8Gi → 990Mi minimum; 32Gi → 3276Mi. In-cluster init Job only.
+  oplogPercentOfVolume: 10
```
lalitadithya: what is the use case for this to be a percentage? Why can't we just ask the user for a size like we do with storage? This will reduce the math complexity we are doing in the helm chart

rupalis-nv: actually @XRFXLP  suggested to use size oplog as a percent of the data volume instead of another hardcoded MB value . That is why the knob is oplogPercentOfVolume

XRFXLP: Yeah, I guess you deleted the earlier comment that has the reason -- which is simple: both PVC size and oplog size are almost linear function of event rate. Which makes oplog size / pvc close to an invariant which we can figure out. Choice is between where do we want complexity: at the user's end to manually calculate the oplog size or at the helm chart end.

lalitadithya: I see, this is good context, can we add this as a comment to the file? Is 10% the number we arrived at after experimentation?

XRFXLP: No, we need to carry the experiment to find out the good default, most likely current 10% is retrofitted with 990 Mi/8 Gi of the existing defaults.

Theoretically:
```
Total storage = event-rate x TTL (30 days) x event-size
oplog-volume-need = individual-oplog-size  x tolerable-window x event-rate x margin-for-fault-handling-module-writes

tolerable-window = downtime-tolerance + backlog-drain-time = downtime-tolerance + max(0, event-rate - consumption-rate)-driven-drain-time
```

- Event-size is not a raw json event-size, we need to account for wiredtiger compression and percona encryption. 
- margin-for-fault-handling-module-writes: this would depend on what is nominal fatal-HE / total-HE.
- event consumption rate must have been drifted from benchmarked state, given the recent changes across all the modules.

lalitadithya: what is `driven-drain-time`, do you mean `data-driven-drain-time`?

XRFXLP: ah, bad hyphenation, I just mean it is driven by event-rate - slowest-consumer-drain-rate

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_init-eval.tpl
```diff
@@ -113,3 +113,23 @@ if (scramUser) {
...
+*/}}
+{{- define "mongodb-store.oplogEval" -}}
+var raw = '$MONGODB_OPLOG_SIZE_MB';
+var mb = Number(raw);
+if (raw === '' || !Number.isInteger(mb) || mb < 990) {
+  throw new Error('MONGODB_OPLOG_SIZE_MB must be an integer >= 990, got ' + JSON.stringify(raw));
+}
+var hello = db.hello();
+print('Oplog resize on ' + (hello.me || 'unknown') + ' targetMB=' + mb);
+var res = db.adminCommand({ replSetResizeOplog: 1, size: mb });
```
KaivalyaMDabhadkar: This resize runs without checking the member's current oplog size, so it can shrink as well as grow. Shrinking truncates the oldest oplog entries immediately, and because the job name hash formula changed in this PR, every existing install will rerun this job on its first upgrade to this chart version.
On local-path/hostpath storage (our kind/Tilt dev flow) mongod's default oplog is 5% of the host filesystem, often several GB, so that first upgrade would cut it down to 990MB. Stored change stream resume tokens can then fall out of the window, and the watcher handles ChangeStreamHistoryLost by dropping the token and resuming from now, which is silent event loss that the exact failure mode #1594 is about. It would also silently revert any manual emergency oplog resize on every rerun.
Could we read the current size first (db.getSiblingDB('local').oplog.rs.stats().maxSize) and skip when it is already at or above the target? If deliberate shrinking should be supported, maybe gate it behind an explicit value like oplogAllowShrink.

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
```diff
@@ -34,12 +34,105 @@ Strings must be base-10 digits; range 0–2147483647 is checked before int for s
...
+{{- $coef = trimSuffix "G" $s -}}{{- $bytesPer = 1000000000 -}}
+{{- else if hasSuffix "M" $s -}}
+{{- $coef = trimSuffix "M" $s -}}{{- $bytesPer = 1000000 -}}
+{{- else if hasSuffix "k" $s -}}
+{{- $coef = trimSuffix "k" $s -}}{{- $bytesPer = 1000 -}}
+{{- else -}}
+{{- fail (printf "mongodb-store persistence size %q must have a unit (Gi, G, Mi, M, Ki, k, Ti, T)" $s) -}}
+{{- end -}}
+{{- if not (regexMatch "^[0-9]+$" $coef) -}}
+{{- fail (printf "mongodb-store persistence size %q must use an integer coefficient" $s) -}}
```
KaivalyaMDabhadkar: This parser rejects some values that k8s itself accepts: fractional coefficients like "1.5Gi" or "0.5Gi", plain byte integers like "8589934592", exponent forms like "8e9", and the Pi/Ei suffixes. Before this PR the chart never parsed persistence.size (persistenceSize was only evaluated in mode=manual for pv.yaml), so anyone running with one of those values today gets a hard helm render failure on upgrade with no change on their side. I verified "1.5Gi" renders fine on main and fails on this branch.
Could we accept the full quantity grammar, or at least fractional coefficients and the remaining suffixes, so existing installs do not break?

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
```diff
@@ -275,6 +276,37 @@ spec:
               sleep 5;
             done
 
+            echo "=============================="
+            echo "Resizing oplog on each replica..."
+            echo "=============================="
+
+            # replSetResizeOplog is per-member; replica-set URI would only hit primary.
+            # Percona databaseAdmin cannot run this; clusterAdmin can. Bitnami uses root.
+            for host in {{ include "mongodb-store.oplogMemberHosts" . }}; do
```
KaivalyaMDabhadkar, line 291: With this loop the job only succeeds if every member is individually reachable and authenticated, all within the existing activeDeadlineSeconds of 600. Before, the job only needed the primary via the replica set URI.
So a replica set that is healthy by majority but has one member down (node maintenance for example) now fails the job, and that also blocks unrelated changes like a TTL update since those rerun the same job. With plain helm a failed job with the same name has to be deleted manually. The until loop also hides permanent errors (bad credentials, cert SAN mismatch) behind "Waiting for mongodb member...".
Maybe we can bound the retries per member and continue (the next rerun can fix a skipped member), or tolerate one unreachable member?

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
```diff
@@ -34,12 +34,105 @@ Strings must be base-10 digits; range 0–2147483647 is checked before int for s
...
+{{- fail (printf "mongodb-store.oplogPercentOfVolume must be an integer from 1 through 50, got %v" $raw) -}}
+{{- end -}}
+{{- $volMB := include "mongodb-store.storageToMB" (include "mongodb-store.persistenceSize" .) | int -}}
+{{- $mb := div (mul $volMB $pct) 100 -}}
+{{- if lt $mb 990 -}}
+{{- $mb = 990 -}}
+{{- end -}}
+{{- $half := div $volMB 2 -}}
+{{- if lt $half 990 -}}
+{{- fail (printf "data volume %d MB cannot reserve 990 MB for oplog and retain 50%% for data; increase mongodb persistence size" $volMB) -}}
```
KaivalyaMDabhadkar: This makes any data PVC under 1980Mi fail rendering. An existing install with a small bound PVC hits a hard render error on upgrade, and since bound PVCs do not grow when the value changes, the only way to unblock is to declare a size bigger than the real disk, which then feeds wrong numbers into the sizing.
Should this be a warning plus clamping instead of a hard fail, or at least come with a documented migration path for existing below-2Gi volumes?

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
```diff
@@ -34,12 +34,105 @@ Strings must be base-10 digits; range 0–2147483647 is checked before int for s
...
+{{- end -}}
+{{- $s := $raw | toString -}}
+{{- if not (regexMatch "^[0-9]+$" $s) -}}
+{{- fail (printf "mongodb-store.oplogPercentOfVolume must be an integer from 1 through 50, got %v" $raw) -}}
+{{- end -}}
+{{- $pct := int $s -}}
+{{- if or (lt $pct 1) (gt $pct 50) -}}
+{{- fail (printf "mongodb-store.oplogPercentOfVolume must be an integer from 1 through 50, got %v" $raw) -}}
+{{- end -}}
+{{- $volMB := include "mongodb-store.storageToMB" (include "mongodb-store.persistenceSize" .) | int -}}
```
KaivalyaMDabhadkar: The sizing uses the declared helm value, not the actual bound PVC, and there are a few cases where those differ:
1. Raising persistence.size does not expand an existing PVC, so the new job would resize the oplog beyond the real disk (replSetResizeOplog does no capacity check) and mongod can hit disk full.
2. Percona with volumeSpec.hostPath renders fine but this falls back to the literal "8Gi" default, so the percent knob is a no-op on the real host disk.
3. Bitnami with persistence.enabled=false sizes from persistence.size while the pods actually run on emptyDir; with size "100Gi" that resizes every member's oplog to 10GiB of node-local disk.

Can you check if we can guard these or call them out next to the PVC expansion note in the docs?

lalitadithya:
@XRFXLP , @KaivalyaMDabhadkar , based on the reviews if the parsing of the values is getting complex to get right and support all use case, I'd recommend we just ask the user to do the math required

XRFXLP:
> @XRFXLP , @KaivalyaMDabhadkar , based on the reviews if the parsing of the values is getting complex to get right and support all use case, I'd recommend we just ask the user to do the math required
 
 yeah, makes sense to document the sizing guidance in the configuration docs.

docs/configuration/mongodb-store.md
```diff
@@ -134,11 +134,46 @@ mongodb-store:
...
+
+The init Job runs `replSetResizeOplog` on every reachable member. It **never shrinks** an existing oplog unless you set `mongodb-store.oplogAllowShrink: true`. Shrinking truncates the oldest entries immediately; change-stream watchers then hit `ChangeStreamHistoryLost` and resume from now (silent event loss — issue #1594). Kind/Tilt hostPath often starts with MongoDB's default of 5% of the node disk (several GiB); skip-shrink leaves that larger window in place.
+
+External Mongo is not resized. Resize is skipped when there is no PVC: Bitnami `mongodb.persistence.enabled=false` (emptyDir) or Percona `volumeSpec.hostPath` / `emptyDir`.
+
+Raising `persistence.size` does not expand an already-Bound PVC. Expand and verify the live volume first, then raise `oplogSizeMB`. `replSetResizeOplog` does not check free disk.
+
+```yaml
+mongodb-store:
+  oplogSizeMB: 3276   # example: ~3.2Gi; compute from the formula, do not copy blindly
```
nicdumz: Putting myself in the shoes of someone trying to deploy this, the formula is not something I can actually use, and this likely needs iteration.

Is `oplogsize = oplog-entry-size × window × event-rate × extra-writes`?

What is `oplog-entry-size`?
How does one estimate `event-rate`?
What is `extra-writes`?

For all of the 3 above, could you provide ways to estimate this or reasonable defaults?

window    ≈ downtime-tolerance + drain-time
downtime-tolerance should probably recommend something reasonable: 24h? 48h?
`max(0, ingest-rate − slowest-consumer-rate)`: no idea and some estimate would be helpful.

Overall I also imagine that PVC size must be > oplogsize but `data PVC  ≈ event-rate × TTL × stored-event-size` does not help me estimate the margin that might be needed.

In order to double check, should you provide an example, e.g. "on a cluster with NN machines of type T, assume 48 downtime tolerance, we recommend oplogsize: AA & persistence size: BB"

rupalis-nv: Updated the oplog sizing docs so each term is defined, measurable, and has a default.

oplogSizeMB ≈ rate/s × ~350 B × window_seconds × extra_writes / 1e6

Window: 24h (48h if you want extra margin). Drain time is ~0 if consumers keep up; the 24h already covers a day of downtime.
Rate: measure inserts/s, or ~5/s per 1k nodes (~500/s at 100k).
extra_writes: 1.5 if unmeasured.
PVC: at least 2× oplog, then add TTL data. Bound volumes do not grow from Helm.
Example (~100k nodes, 500/s, 24h): oplogSizeMB: 15120, PVC ≥ 32Gi. ~10k nodes → ~1512. Kind/idle → leave 990.

If Mongo is already up, size from wrap capacity

distros/kubernetes/nvsentinel/values-full.yaml
rupalis-nv: PTAL

docs/configuration/mongodb-store.md
XRFXLP: @rupalis-nv can you check if exposing https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-storage.oplogMinRetentionHours helps us get around this calculation and helps us directly to solve the issue? We need to make sure there are no associated risks with it.

rupalis-nv: Checked storage.oplogMinRetentionHours (also replSetResizeOplog.minRetentionHours).

It matches the 24h goal: Mongo keeps oplog entries until they are N hours old, even if that means growing past oplogSizeMB. Change-stream resume tokens would survive that long.

It does not replace the sizing work, and on this chart it is less safe than a size cap:

Mongo’s docs: the oplog can grow without a size limit and can fill the disk.
Our PVCs are Bound and do not grow from Helm. Default is 8Gi. 24h at the rate in #1594 is about 15GB of oplog alone — disk full, Mongo down.
With retention hours set, the MB cap is no longer a hard ceiling. A write spike grows the oplog until the volume is full.
Runtime minRetentionHours is lost on mongod restart unless it is also in mongod.conf (Bitnami and Percona). Size from replSetResizeOplog does persist.
Must be set on every replica. Clock skew can change how long entries are kept. Extra space is not returned without compact on oplog.rs (that blocks).
So this changes the knob from MB to hours, but you still need a disk big enough for those hours. Without volume auto-expand, a size cap is the safer fix. I would keep oplogSizeMB in this PR. Retention hours could be a later add-on after the PVC is sized, not a replacement.

XRFXLP: Why are we mixing the requirement from the issue and current default? If you project the size of PVC as per the set of assumption that the issue dictates then it will definitely be a lot of larger than 8Gi, so lets not mix both.

> Clock skew can change how long entries are kept.

What do you mean exactly by this? Is this can be in hours?

> Runtime minRetentionHours is lost on mongod restart unless it is also in mongod.conf (Bitnami and Percona). Size from replSetResizeOplog does persist.

Looks like the doc is pointing something opposite:
> To change the minimum oplog retention period after starting the mongod, use [replSetResizeOplog](https://www.mongodb.com/docs/manual/reference/command/replSetResizeOplog/#mongodb-dbcommand-dbcmd.replSetResizeOplog). replSetResizeOplog enables you to resize the oplog dynamically without restarting the mongod process. To persist the changes made using replSetResizeOplog through a restart, update the value of oplogMinRetentionHours.

===> Doesn't that implies replSetResizeOplog doesn't persist restarts?

rupalis-nv: Agreed — I should not mix those two.

8Gi is only the small dev disk the chart ships with. The issue is about a busy production cluster (~100k nodes, ~500 events/s). That log is about 15GB, so the disk must be much bigger than 8Gi. I will split the docs: one production example, and a separate note that Kind/dev can keep the default 990. No 8Gi on the production row.

Clock skew: Mongo uses each machine’s clock to decide “is this log entry older than N hours?” If clocks are in sync (normal), the error is seconds, not hours. It would only be hours if a machine’s clock is badly wrong.

On restart:

That doc sentence is only about the hours setting, not the size.

Size: if we change it with the resize command, Mongo remembers it after restart. We still put size in the config file so a brand-new empty member does not start at 990.
Hours: the command works until restart. After restart Mongo only keeps hours if it is also in the config file. If it is not, hours is gone.
So size survives restart; hours need the config file too. I will keep this PR on a fixed size, so the log cannot grow until the disk is full.

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/mongodb-mongod-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: {{ include "mongodb-store.mongodConfigMapName" . }}
+  labels:
+    app.kubernetes.io/name: mongodb
+    app.kubernetes.io/component: mongod-config
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+data:
+  mongodb.conf: |
+    replication:
```
KaivalyaMDabhadkar: The bitnami half of this breaks the default deployment unfortunately. I reproduced it with the exact image the chart pins (bitnamilegacy/mongodb:8.0.3-debian-12-r0): running it with the chart's env plus this configmap mounted read only over /opt/bitnami/mongodb/conf/mongodb.conf, the container exits with code 2 on every start. With BITNAMI_DEBUG=true the image explains why:
```
  DEBUG ==> /mongodb.conf mounted. Skipping setting port and IPv6 settings
  DEBUG ==> /mongodb.conf mounted. Skipping setting log settings
  DEBUG ==> /mongodb.conf mounted. Skipping setting storage settings
  DEBUG ==> /mongodb.conf mounted. Skipping keyfile location configuration
  BadValue: --fork has to be used with --logpath or --syslog
```
The bitnami entrypoint treats any mounted config as fully user managed, so it stops writing systemLog, replSetName, storage.dbPath, keyfile and net settings, and its bootstrap mongod (which uses --fork) then refuses to start without a log path. The same container with identical env and no mounted config forms rs0 fine. In the cluster this means every mongodb pod crashloops after the first rollout of this chart version.

Repro:
```
  printf 'replication:\n  oplogSizeMB: 990\n' > /tmp/minimal-mongodb.conf
  docker run --rm -e BITNAMI_DEBUG=true -e MONGODB_REPLICA_SET_MODE=primary -e MONGODB_REPLICA_SET_NAME=rs0 -e MONGODB_ROOT_PASSWORD=rootpass -e MONGODB_REPLICA_SET_KEY=key0123456789 -e MONGODB_ADVERTISED_HOSTNAME=localhost -v /tmp/minimal-mongodb.conf:/opt/bitnami/mongodb/conf/mongodb.conf:ro docker.io/bitnamilegacy/mongodb:8.0.3-debian-12-r0
```
A safer route that also avoids touching any chart internals: pass the size as a flag. mongod accepts --oplogSize, and flags are appended after the generated config, so nothing gets skipped. That can live entirely in our own values as an extra entry in mongodb.extraFlags, with the same keep-in-sync validation you added for percona.

distros/kubernetes/nvsentinel/charts/mongodb-store/charts/psmdb-db/templates/cluster.yaml
```diff
@@ -116,10 +116,12 @@ spec:
     externalNodes:
 {{ $replset.externalNodes | toYaml | indent 6 }}
     {{- end }}
-    {{- if $replset.configuration }}
     configuration: |
+      replication:
+        oplogSizeMB: {{ $.Values.oplogSizeMB | default 990 }}
```
KaivalyaMDabhadkar: This works functionally, but it edits the vendored psmdb-db chart, and we should not modify vendored charts.  The next psmdb-db version bump will silently overwrite it, bringing the new-member gap back with no signal.
The same result is possible from our own chart: put the line into the default rs0 configuration value in mongodb-store/values.yaml, for example
```
  psmdb-db:
    replsets:
      rs0:
        configuration: |
          replication:
            oplogSizeMB: 990
          setParameter:
            authenticationMechanisms: "MONGODB-X509,SCRAM-SHA-256,SCRAM-SHA-1"
```
and extend the validator you already wrote so it fails the render when usePerconaOperator is on and rs0.configuration does not contain an oplogSizeMB matching mongodb-store.oplogSizeMB. That keeps the fail-loud behavior, covers users who override the configuration block, and survives chart bumps. It also removes the duplicate-key risk this template version has when a user's configuration already contains its own replication section, since with the values approach the user owns the whole block.

rupalis-nv: PTAL

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/mongodb-mongod-config.yaml
rupalis-nv: PTAL

KaivalyaMDabhadkar (dismissed):

docs/configuration/mongodb-store.md
rupalis-nv: hey @nicdumz  , when you have a min PTAL at PR

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
```diff
@@ -275,6 +276,63 @@ spec:
...
+                  {{- else }}
+                  --username {{ $rootUsername }} --password "$MONGODB_ROOT_PASSWORD" --authenticationDatabase admin \
+                  {{- end }}
+                  {{- end }}
+                  {{- if .Values.mongodb.tls.enabled }}
+                  --tls \
+                  --tlsCAFile /certs/ca.crt \
+                  --tlsCertificateKeyFile /tmp/tls.pem \
+                  {{- end }}
+                                  --eval "
```
lalitadithya: what is with the spaces before `--eval`? that seems like something that will trip us up in future

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
```diff
@@ -34,12 +66,90 @@ Strings must be base-10 digits; range 0–2147483647 is checked before int for s
 {{- end }}
 
 {{/*
-create-mongodb-database-<ttl>-<scriptHash> so a TTL or init-script change
-(new index, etc.) is a new Job, not a patch on a completed one.
+Oplog size in MiB. The operator sets this; the chart does not derive it from the PVC.
+Nil/empty → 990 (MongoDB minimum). Integer >= 990.
+oplogPercentOfVolume is rejected: Helm no longer multiplies percent × volume.
```
lalitadithya: this changed was never merged to main right? Do we need this?

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
rupalis-nv: Updated PTAL

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
rupalis-nv: right, updated PTAL

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/jobs.yaml
rupalis-nv: addressed

distros/kubernetes/nvsentinel/charts/mongodb-store/templates/_helpers.tpl
rupalis-nv: Right — oplogPercentOfVolume never landed on main. 
i drooped it

XRFXLP (approved):

KaivalyaMDabhadkar (approved):

nicdumz:
@rupalis-nv @lalitadithya, some feedback:

deploying this correctly requires setting as values:

```
mongodb-store:
  oplogSizeMB: 12345
  mongodb:
    extraFlags:
      - "--setParameter"
      - "authenticationMechanisms=MONGODB-X509,SCRAM-SHA-256"
      - "--oplogSize=12345
```

That IMO is repetitive and error-prone: without setting extraFlags.oplogSize an error is brought up. If you set only extraFlags.oplogSize then authenticationMechanisms is dropped

lalitadithya:
> @rupalis-nv @lalitadithya, some feedback:
> 
> deploying this correctly requires setting as values:
> 
> ```
> mongodb-store:
>   oplogSizeMB: 12345
>   mongodb:
>     extraFlags:
>       - "--setParameter"
>       - "authenticationMechanisms=MONGODB-X509,SCRAM-SHA-256"
>       - "--oplogSize=12345
> ```
> 
> That IMO is repetitive and error-prone: without setting extraFlags.oplogSize an error is brought up. If you set only extraFlags.oplogSize then authenticationMechanisms is dropped

Yeah, I agree. A lot of this unfortunately comes down to the limitations we have with helm specially with subcharts and also the fact that helm doesn't support multi layered templating. There is also one more place you have to set this if you are using percona :( We're thinking about how to clean up our database deployment story in general because right now we end up having 3 different database(s) in our helm chart and it has started causing some pain much like what you have highlighted here and also with upgrades. We decided to vendor in the charts to make it easier to start with, but it is making the ongoing maintenance of these things a little tricky.

PR #1668, opened 2026-08-25
feat(gpu-health-monitor): detect asserted hardware power brake

### LINKED ISSUE #1666: [Feature]: detect asserted hardware power brake; the POWER health watch does not, and its only code is documented as noise
### Prerequisites

### Code of Conduct

### Feature Summary

gpu-health-monitor has no detection for an asserted **hardware power brake**, which is the one genuinely actionable power condition. Meanwhile the POWER health watch's dominant error code, `DCGM_FR_CLOCK_THROTTLE_POWER`, is documented by NVSentinel itself as non-actionable noise to be suppressed.

Request: a field-based power watch reading `DCGM_FI_DEV_CLOCKS_EVENT_REASONS` and failing on the HW power brake bit (`0x80`), mirroring the existing `GpuThermalMarginWatch`.

### Problem/Use Case

On a 288-node GB200 NVL72 cluster, a datacenter power event left 9 nodes with a continuously asserted hardware power brake. Two GPUs on each affected node were clock-limited 100% of the time for over 36 hours. **NVSentinel never indicated it.** It was found by hand with `nvidia-smi -q | grep -i brake`.

#### The POWER health watch does not detect it

Measured over a 2 hour window, comparing `dcgm_health_active_events{event_type="GpuPowerWatch"}` against `DCGM_FI_DEV_CLOCKS_EVENT_REASONS` from dcgm-exporter on the same nodes and GPUs:

| Node | GPU | HW power brake (bit 0x80) | `GpuPowerWatch` active |
|---|---|---|---|
| A-09 | 2 | 100% | 23% |
| A-09 | 3 | 100% | 20% |
| A-09 | 0 | 0% | 5.6% |
| A-09 | 1 | 0% | 10.8% |
| A-05 | all | 0% | 98.5% |

`GpuPowerWatch` correlates with neither the hardware power brake nor the instantaneous SW power cap bit (`0x04`). On the node above it fires *less* on the two GPUs that are permanently braked than the 98.5% it reports on a node with no brake at all. Across the rack it fired on 69% of nodes while exactly 50% had a brake, and the ranking did not separate the two groups.

By contrast the raw field is exact. Selecting `DCGM_FI_DEV_CLOCKS_EVENT_REASONS >= 128` identified 9 of 9 affected nodes with zero false positives across all 288, matched a manual `nvidia-smi` audit exactly, and showed a clean step at onset. Observed values were `0`, `1` (GPU idle), `4` (SW power cap), `136` (HW slowdown + HW power brake) and `140` (adds SW power cap).

#### Why the noise makes it worse

`DCGM_FR_CLOCK_THROTTLE_POWER` maps to `NONE` in `dcgmerrorsmapping.csv`, and NVSentinel ships two separate mechanisms whose documented purpose is removing it, both naming this code as the example:

- `docs/configuration/gpu-health-monitor.md`, `suppressedErrorCodes`: "Useful for high-frequency, non-actionable flaps (e.g. normal power-cap boost-clock behavior)", with `DCGM_FR_CLOCK_THROTTLE_POWER` as the worked example.
- `docs/runbooks/node-conditions.md`: a CEL rule whose stated purpose is excluding `DCGM_FR_CLOCK_THROTTLE_POWER` from cordoning.

So the only power signal that reaches an operator is one the project already treats as noise, and it flaps on a 30 second poll. An operator who follows the documentation and suppresses it is left with no power signal whatsoever. An operator who does not is left with a signal that is on nearly always and carries no information. Either way a real power brake is invisible.

#### The pattern already exists

`GpuThermalMarginWatch` was added for exactly this class of problem: a hardware condition DCGM's health watches did not surface usefully, so gpu-health-monitor reads `DCGM_FI_DEV_GPU_TEMP_TLIMIT` (field 153) directly, compares against the hardware slowdown threshold, and raises its own error code mapped to `CONTACT_SUPPORT` (design 042, `docs/runbooks/gpu-thermal-margin.md`).

That field is currently the *only* DCGM field gpu-health-monitor watches directly. The proposal is a second one on the same pattern.

### Suggested shape

- New watch `GpuPowerBrakeWatch`, gated by its own values flag, defaulting off to preserve current behaviour.
- Reads `DCGM_FI_DEV_CLOCKS_EVENT_REASONS`; fails a GPU when bit `0x80` (HW power brake) is set.
- New error code, for example `GPU_HW_POWER_BRAKE_VIOLATION`, added to `dcgmerrorsmapping.csv`. `CONTACT_SUPPORT` matches the thermal precedent, since an asserted brake is a facility or power delivery fault rather than anything a node reboot fixes.
- A minimum assertion duration or consecutive-poll count would be worth having, so a brief brake during a load transient does not cordon a node. In our data the condition was continuous for 36+ hours, so a threshold of a few polls would lose nothing real.
- Reporting bit `0x40` (HW thermal slowdown) from the same field may be a natural companion, since it is the same register and complements the existing thermal margin watch.

Happy to contribute the implementation if the approach looks right.

lfriedman-netllama:
Closes #1666.

Adds `GpuPowerBrakeWatch`, a field monitor that fails a GPU whose clocks-event-reasons mask has the
hardware power brake bit (`0x80`) set, meaning the power delivery path is forcing clocks down.

## Summary

* Adds `DCGM_FIELDS_MONITORING["gpupowerbrakemonitoringenabled"]`, reading
  `DCGM_FI_DEV_CLOCKS_EVENT_REASONS` (falling back to `DCGM_FI_DEV_CLOCK_THROTTLE_REASONS`).
* Adds `_evaluate_gpu_power_brake`, mirroring `_evaluate_gpu_thermal_margin`.
* Adds `GPU_HW_POWER_BRAKE_VIOLATION,CONTACT_SUPPORT` to `dcgmerrorsmapping.csv`.
* New values: `gpuPowerBrakeMonitoringEnabled` (default **false**), `gpuPowerBrakeStoreOnly`
  (default **true**), `gpuPowerBrakeMinConsecutivePolls` (default 3).
* Config reference section plus a runbook, `docs/runbooks/gpu-power-brake.md`.
* 8 unit tests.

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [X] Documentation/CI
- [ ] Fault Management
- [X] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Why a new watch rather than using the POWER health watch

Measured on a 288-node GB200 cluster, `dcgm_health_active_events{event_type="GpuPowerWatch"}` was active
on 69% of nodes while exactly 50% had a brake asserted, and it tracked neither the brake bit nor the SW
power cap bit (`0x04`):

| Node | GPU | HW power brake (bit 0x80) | `GpuPowerWatch` active |
|---|---|---|---|
| A-09 | 2 | 100% | 23% |
| A-09 | 3 | 100% | 20% |
| A-09 | 0 | 0% | 5.6% |
| A-09 | 1 | 0% | 10.8% |
| A-05 | all | 0% | 98.5% |

It fires *less* on the two permanently braked GPUs than the 98.5% it reports on a node with no brake at
all. Selecting the raw field instead identified 9 of 9 affected nodes with zero false positives across all
288, and matched a manual `nvidia-smi -q | grep -i brake` audit exactly.

The noise side compounds it: `DCGM_FR_CLOCK_THROTTLE_POWER` maps to `NONE` and is the worked example for
`suppressedErrorCodes` in `docs/configuration/gpu-health-monitor.md`, so an operator who follows the
documentation is left with no power signal whatsoever, and one who does not is left with a signal that is
on nearly always.

## Design

Follows `GpuThermalMarginWatch` (design 042) exactly: read the field directly, compare, raise a dedicated
error code. That was already the established pattern for a hardware condition DCGM's health watches do not
surface usefully, and field 153 was until now the only field watched directly.

`CONTACT_SUPPORT` matches the thermal precedent. An asserted brake is a facility or power delivery fault,
so a node reboot does not resolve it.

**Defaults are deliberately conservative.** Off by default, so existing deployments see no behaviour
change, and store-only by default when enabled, so it is observable before it can cordon anything.

**Consecutive-poll threshold.** A brake asserted for a single poll can be a load transient. A clear resets
the streak, so a flapping brake never accumulates to a failure. A GPU with no usable sample is skipped and
keeps its counter, so a gap in DCGM data neither raises nor clears a finding. In the fleet data behind this
the condition was continuous for 36+ hours, so a threshold of 3 loses nothing real.

**Field group.** The single field group is now shared by both field monitors rather than being
thermal-specific, so one `WatchFields` call covers however many are enabled. Each evaluator reads back only
its own field id. All three pre-existing paths behave as before: thermal enabled with a metadata reader
watches field 153, thermal enabled without one logs the same warning and watches nothing, and neither
enabled creates no field group.

## Testing
- [X] Tests pass locally

8 unit tests cover: unsupported field self-disables, disabled watch publishes nothing, brake bit detected,
SW power cap alone does **not** fail, the consecutive-poll threshold, streak reset on clear, mixed GPUs, and
a data gap returning `None`.

One caveat worth stating plainly: `pydcgm` is neither a pip dependency nor vendored, and the Makefile gates
these tests to Linux, so I could not execute the suite on my macOS workstation and am relying on CI for it.
What I did verify locally was `black --check` (clean), that the chart renders the three new config keys, and
the decision logic itself against the five mask values actually observed on hardware, with ground truth from
`nvidia-smi`:

```
mask  decimal  expected  computed  meaning
0x00  0        False     False     no throttling
0x01  1        False     False     GPU idle
0x04  4        False     False     SW power cap
0x88  136      True      True      HW slowdown + HW power brake
0x8c  140      True      True      SW power cap + HW slowdown + brake
```

Happy to adjust the default threshold, the error code name, or the `CONTACT_SUPPORT` mapping if you would
rather they were different.

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added optional GPU hardware power-brake monitoring.
  * Supports store-only mode and configurable consecutive-poll thresholds before reporting a GPU failure.
  * Detects power-brake conditions across supported DCGM versions while distinguishing software power caps and thermal slowdowns.

* **Documentation**
  * Added configuration guidance and a troubleshooting runbook covering detection, verification, impact assessment, escalation, and resolution.
  * Linked the runbook from the Health Monitor issues list.

* **Tests**
  * Added coverage ensuring blank monitoring samples are ignored without affecting power-brake evaluation.


health-monitors/gpu-health-monitor/uv.lock
lalitadithya, line 1: if you don't mind, can we drop this uv lock file? the module uses poetry for package management for now

lfriedman-netllama: Dropped in 5b5ba35b, apologies. That was an accident on my side: I tried to run the test suite with `uv` because `pydcgm` is not pip-installable on macOS, and the lockfile got swept up by a `git add -A`. Removed from the index and the working tree, and poetry remains the only package manager touched.

docs/runbooks/gpu-power-brake.md
```diff
@@ -0,0 +1,105 @@
...
+
+```promql
+DCGM_FI_DEV_CLOCKS_EVENT_REASONS >= 128
+```
+
+Fraction of a window each GPU spent braked, which distinguishes a transient from a sustained assertion:
+
+```promql
+avg by (Hostname, gpu) (avg_over_time((DCGM_FI_DEV_CLOCKS_EVENT_REASONS >= bool 128)[1h:1m]))
+```
```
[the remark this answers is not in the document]
lfriedman-netllama: Good catch, and this one was a real defect rather than a doc nit. Fixed in 5b5ba35b.

`0x100` (`DISPLAY_CLOCKS`) is 256, and `256 >= 128` is true while `256 & 0x80` is false, so the bare comparison false-positives on `0x100` and `0x140`. Both queries now isolate the low byte, since PromQL has no bitwise AND:

```promql
(DCGM_FI_DEV_CLOCKS_EVENT_REASONS % 256) >= 128
```

Verified against the mask values I could enumerate, with the brake bit as ground truth:

```
mask  truth  >=128   %256>=128
64    False  False   False
128   True   True    True
136   True   True    True
140   True   True    True
192   True   True    True
256   False  True    False   <-- old form wrong
320   False  True    False   <-- old form wrong
384   True   True    True
```

Worth noting the Python path was never affected: `_evaluate_gpu_power_brake` uses a real `reasons_mask & HW_POWER_BRAKE_REASON_BIT`. This was PromQL-only. It did however mean the same mistake was live in a Grafana dashboard of ours, which I have also corrected.

docs/runbooks/gpu-power-brake.md
```diff
@@ -0,0 +1,105 @@
...
+`clocks_throttle_reasons.active` is the legacy alias, normalized to `clocks_event_reasons.active` in the CSV
+header.
+
+### 3. Confirm it in metrics, and check whether it is sustained
+
+If dcgm-exporter is deployed, the raw mask is the authoritative view. Bit `0x80` set means brake:
+
+```promql
+DCGM_FI_DEV_CLOCKS_EVENT_REASONS >= 128
+```
```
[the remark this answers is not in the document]
lfriedman-netllama: Added in 5b5ba35b. The verification section now gives the `DCGM_FI_DEV_CLOCK_THROTTLE_REASONS` form as well, and notes that dcgm-exporter names the series after whichever field it is configured to collect, so only the metric name changes: the bit layout is identical across the rename.

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
```diff
@@ -55,6 +55,21 @@ def _run_dcgm_server(port: int, bind_address: str) -> None:
...
+    DCGM renamed the clocks-throttle family to "clocks event" in 4.x. Looking up
+    both spellings keeps a field monitor working across that rename in either
+    direction. ``getattr`` chaining with ``or`` is avoided deliberately: a
+    legitimate value of 0 would be skipped.
+    """
+    for name in names:
+        value = getattr(module, name, None)
+        if value is not None:
+            return value
+    return None
```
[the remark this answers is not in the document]
lfriedman-netllama: Added in 5b5ba35b: `def _first_defined(module: object, *names: str) -> int | None:`.

Used `object` rather than `types.ModuleType` deliberately. This module does `from . import types, metrics`, so the local `types` package shadows the stdlib one and importing `ModuleType` here would either need an aliased import or read as if it came from the local module. `object` is sufficient for `getattr` and avoids that confusion.

health-monitors/gpu-health-monitor/gpu_health_monitor/tests/test_dcgm_watcher/test_dcgm.py
```diff
@@ -73,6 +73,23 @@ def _make_thermal_margin_watcher(self, metadata_reader: MetadataReader) -> dcgm.
...
+        watcher = dcgm.DCGMWatcher(
+            addr="localhost:5555",
+            poll_interval_seconds=10,
+            callbacks=[],
+            dcgm_k8s_service_enabled=False,
+            power_brake_enabled=True,
+            power_brake_min_consecutive_polls=min_consecutive_polls,
+        )
+        watcher._field_group = MagicMock()
+        return watcher
```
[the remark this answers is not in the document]
lfriedman-netllama: Added in 5b5ba35b: `-> None` on all 8 new test methods, and `monkeypatch: pytest.MonkeyPatch` where applicable. `pytest` was already imported in the file.

One observation for context rather than pushback: only 10 of the 67 test methods in this file carry `-> None` today, so the prevailing convention here is unannotated. I have followed the guideline for the new code and left the existing methods alone rather than retrofitting them in a feature PR. Happy to do that sweep separately if it would be useful.

docs/runbooks/gpu-power-brake.md
```diff
@@ -0,0 +1,117 @@
...
+
+Fraction of a window each GPU spent braked, which distinguishes a transient from a sustained assertion:
+
+```promql
+avg by (Hostname, gpu) (avg_over_time(((DCGM_FI_DEV_CLOCKS_EVENT_REASONS % 256) >= bool 128)[1h:1m]))
+```
+
+On a DCGM build that predates the clocks-throttle to clocks-event rename, gpu-health-monitor falls back to
+`DCGM_FI_DEV_CLOCK_THROTTLE_REASONS`, and dcgm-exporter names its series after whichever field it is
+configured to collect. Substitute that metric name in both queries if the one above returns nothing:
```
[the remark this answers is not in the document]
lfriedman-netllama: Applied in 64194ffb. Reads better, thanks.

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
```diff
@@ -603,6 +664,95 @@ def _evaluate_gpu_thermal_margin(
...
+                reasons_mask = int(raw_reasons)
+            except (ValueError, TypeError):
+                log.warning(
+                    "GPU %s clocks-event-reasons value %r is not a valid integer; skipping power brake evaluation",
+                    gpu_id,
+                    raw_reasons,
+                )
+                continue
+            evaluated = True
+
```
deesharma24, line 740:
```
            # DCGM encodes "no data" as int64 sentinels (DCGM_INT64_BLANK and
            # friends, 0x7ffffffffffffff0..f3) whose low byte has bit 0x80 set,
            # so an unchecked blank would count as an asserted brake. Treat it
            # like a missing sample: skip, keep the streak.
            if dcgmvalue.DCGM_INT64_IS_BLANK(reasons_mask):
                log.warning(
                    "GPU %s clocks-event-reasons value is blank (0x%x); skipping power brake evaluation",
                    gpu_id,
                    reasons_mask,
                )
                continue
            evaluated = True
```

can we add this here.So, that we don't have to unnecessary traverse when it is empty. Also, I feel _power_brake_streaks.pop should be done when we have a valid data.Thoughts?

deesharma24:
@lfriedman-netllama  overall LGTM.Just one minor comment and please fix the lint issues and rebase the code to to the latest and push

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
lfriedman-netllama: Great catch, and this was a real defect rather than an optimisation. Applied verbatim in f6d553c3.

You are right that it is not just an unnecessary traversal. Every sentinel has bit 0x80 set in its low byte, so an unchecked blank was indistinguishable from an asserted brake:

```
0x7ffffffffffffff0  low byte 0xf0 (240)  & 0x80 -> 0x80   DCGM_INT64_BLANK
0x7ffffffffffffff1  low byte 0xf1 (241)  & 0x80 -> 0x80   DCGM_INT64_NOT_FOUND
0x7ffffffffffffff2  low byte 0xf2 (242)  & 0x80 -> 0x80   DCGM_INT64_NOT_SUPPORTED
0x7ffffffffffffff3  low byte 0xf3 (243)  & 0x80 -> 0x80   DCGM_INT64_NOT_PERMISSIONED
```

And it compounded with the consecutive-poll threshold, which made it worse than a single bad reading. Simulating four blank polls at a threshold of 3:

```
without the check: ['pass', 'pass', 'FAIL', 'FAIL']
with the check   : ['skip', 'skip', 'skip', 'skip']
```

So a GPU whose field is merely unsupported or not-permissioned would have been reported as power-braked and, once store-only is turned off, cordoned with CONTACT_SUPPORT. That is exactly the failure mode this watch exists to avoid creating.

On your second point, agreed, and placing the check before `evaluated = True` handles both halves: the `pop` only runs in the `else` branch, which a blank now skips entirely, so the streak is preserved rather than cleared. It also means an all-blank poll returns `None` and publishes nothing, instead of a spurious PASS.

Added two tests: one that a blank alone publishes nothing, and one that a blank between two real assertions neither clears nor advances the streak.

One caveat worth stating: I could not execute this locally, since `pydcgm` and `dcgmvalue` are not pip-installable on macOS and the Makefile gates these tests to Linux. I used your `dcgmvalue.DCGM_INT64_IS_BLANK` call verbatim rather than guessing at the binding, and verified the arithmetic independently, but CI is the real check.

lfriedman-netllama: Correction on the SHA above: my first push was rejected as a non-fast-forward because main had been merged into the branch in the meantime, so `f6d553c3` never landed. I rebased onto `c15dc023` and the change is now on the branch as **`b9f73324`**. The content is identical; only the commit id changed. Apologies for the noise.

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
```diff
@@ -64,6 +79,31 @@ def _run_dcgm_server(port: int, bind_address: str) -> None:
...
+        violation_code="GPU_HW_POWER_BRAKE_VIOLATION",
+    )
+
+# An asserted external hardware power brake, i.e. the power delivery path telling
+# the GPU to drop clocks. Deliberately not bit 0x04 (SW power cap), which is normal
+# capping under load and carries no fault information, nor bit 0x40 (HW thermal
+# slowdown), which the thermal margin monitor already covers.
+HW_POWER_BRAKE_REASON_BIT = (
+    _first_defined(
+        dcgm_structs,
```
deesharma24: it should be dcgm_fields not dcgm_structs(Ref https://github.com/NVIDIA/DCGM/blob/master/testing/python3/dcgm_fields.py#L89)

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
```diff
@@ -603,6 +664,107 @@ def _evaluate_gpu_thermal_margin(
...
+                    raw_reasons,
+                )
+                continue
+
+            # DCGM encodes "no data" as int64 sentinels (DCGM_INT64_BLANK and
+            # friends, 0x7ffffffffffffff0..f3) whose low byte has bit 0x80 set,
+            # so an unchecked blank would count as an asserted brake. Treat it
+            # like a missing sample: skip, keep the streak.
+            if dcgmvalue.DCGM_INT64_IS_BLANK(reasons_mask):
+                log.warning(
```
deesharma24: it will flood the logs everytime if that happens. maybe we should use log.debug.

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
lfriedman-netllama: Fixed in 08c9b70e, and you were right that it matters beyond style. `_first_defined(dcgm_structs, ...)` found nothing, so the lookup silently fell through to the literal `0x80` and never used the real definition. Now reads from `dcgm_fields`, where `DCGM_CLOCKS_EVENT_REASON_HW_POWER_BRAKE = 0x0000000000000080` lives. Thanks for the pointer.

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
lfriedman-netllama: Agreed, fixed in 08c9b70e. A GPU whose field is unsupported returns a blank on every poll, so at warning level that is one line per GPU per poll forever.

Dropped both the blank and the missing-sample logs to `log.debug`. I also added a `gpu_power_brake_reasons_blank` counter, mirroring the existing `gpu_temp_limit_margin_blank`, because debug alone means an operator has no way to tell the watch has gone inert on a GPU. Happy to drop the counter if you would rather keep the diff smaller.

lfriedman-netllama:
@deesharma24 all three points addressed, rebased onto `39d747a0`, pushed as `08c9b70e`.

## The lint-test failure was my tests, and the cause was worth finding

`health-monitors-lint-test (gpu-health-monitor)` was failing on 7 of my 8 new tests with `KeyError: gpupowerbrakemonitoringenabled`. Black was clean; it was the test half of `lint-test`.

The cause: these tests run against the DCGM mocks in `gpu_health_monitor/tests/conftest.py`, not real bindings, and `MockDCGMFields` never defined `DCGM_FI_DEV_CLOCKS_EVENT_REASONS`. So the registry entry was never created and every test that touched it raised. The one test that passed was the one asserting `None`, for the wrong reason.

`MockDCGMValue` was an empty class as well, so the `DCGM_INT64_IS_BLANK` check from your earlier review would have failed the same way on the next run.

Added to conftest:

* `DCGM_FI_DEV_CLOCKS_EVENT_REASONS = 112`, deleting the generated `DCGM_FI_DEV_FIELD_112` so the `DCGM_FI_DEV_*` count stays at 320, which `test_get_available_fields` asserts. Same trick already used for `DCGM_FI_DEV_GPU_TEMP_TLIMIT`.
* The clocks-event reason bits.
* `DCGM_INT64_IS_BLANK` plus the four int64 sentinels.

## I can now run the suite, which closes the gap I flagged earlier

I said previously that I could not execute these tests because `pydcgm` is not pip-installable on macOS. That was wrong: `conftest.py` stubs every DCGM module, so the only real blockers were the Python deps. With those installed the whole suite runs on macOS:

```
138 passed, 5 subtests passed
```

That includes all 10 power-brake tests and `test_get_available_fields`. Black 26.5.1, matching poetry.lock, is clean across all 22 files. So I am no longer relying on CI to find out whether my own tests pass, which is how the `KeyError` slipped through in the first place.

Worth noting the module Makefile skips tests entirely on non-Linux (`ifneq ($(UNAME_S),Linux)`) and writes empty report files. Given the mocks, that gate looks stricter than it needs to be, and it is why this was invisible locally. Happy to raise that separately rather than widen this PR.

CI will need a fresh `/ok to test` for `08c9b70e` when you get a chance.

health-monitors/gpu-health-monitor/uv.lock
deesharma24: it came back again. Can you please  remove it.

deesharma24, line 1: Can you remove this file and take care of rabbit comments.

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
deesharma24: LGTM

health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
deesharma24: LGTM

health-monitors/gpu-health-monitor/uv.lock
lfriedman-netllama: Removed in befd35cb, and I have put a mechanical guard in place so it cannot recur. Apologies for making you ask twice.

Root cause: I run the test suite with `uv` because it is the quickest way to get the deps on my workstation, and `uv run` regenerates `uv.lock` from `pyproject.toml`. A blanket `git add -A` then swept it back in after I had already removed it once.

Three changes on my side:
- Tests now run with `uv run --no-project`, which never creates the file.
- A local `.git/info/exclude` entry for `uv.lock` and `.venv/`, so `git add -A` cannot stage them in my clone. It is per-clone and never committed, so nothing about it reaches this repo.
- Explicit paths instead of `git add -A` when committing here.

Verified rather than assumed: with the file force-created on disk, `git check-ignore` matches it and `git add -A` stages only the removal, never an add.

health-monitors/gpu-health-monitor/gpu_health_monitor/tests/conftest.py
```diff
@@ -179,12 +179,31 @@ class MockDCGMFields:
...
-    pass
+    # DCGM encodes "no data" as int64 sentinels starting at DCGM_INT64_BLANK.
+    DCGM_INT64_BLANK = 0x7FFFFFFFFFFFFFF0
+    DCGM_INT64_NOT_FOUND = 0x7FFFFFFFFFFFFFF1
+    DCGM_INT64_NOT_SUPPORTED = 0x7FFFFFFFFFFFFFF2
+    DCGM_INT64_NOT_PERMISSIONED = 0x7FFFFFFFFFFFFFF3
+
+    @staticmethod
+    def DCGM_INT64_IS_BLANK(val):
+        return val >= MockDCGMValue.DCGM_INT64_BLANK
```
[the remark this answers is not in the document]
lfriedman-netllama: Added in befd35cb: `def DCGM_INT64_IS_BLANK(val: int) -> bool:`.

lfriedman-netllama:
@deesharma24 both addressed in `befd35cb`.

- `uv.lock` removed, with a guard so it cannot come back. Details on the thread; short version is that `uv run` regenerates it and `git add -A` re-staged it, so I now use `uv run --no-project`, a local `.git/info/exclude`, and explicit paths. Sorry for the repeat.
- Type hints added to the mock `DCGM_INT64_IS_BLANK`.

Verified locally before pushing: **138 passed, 5 subtests passed**, and `black 26.5.1 --check` clean across all 22 files. Head is `befd35cb`, which will need a fresh `/ok to test`.

deesharma24 (approved):

lalitadithya (approved):

PR #1670, opened 2026-08-25
feat(platform-connector): gate health events for ext…

cbumb:
## Summary

Implements NKX-13877: prevents NVSentinel from acting on nodes that are externally owned (handed off via the External Remediation Request flow from ADR-040).

**Platform-connector (primary gate):**
- MetadataAugmentor now checks for the `nvsentinel.dgxc.nvidia.com/managed=false` label after enriching the event with node metadata. If the label matches, the event's `ProcessingStrategy` is set to `STORE_ONLY`, which prevents downstream remediation while preserving the audit trail in the database.
- Cache TTL reduced from 1 hour to 30 seconds so label changes take effect quickly.
- MetadataAugmentor enabled by default in all Helm values files.
- Deduplicator updated with an early return for `STORE_ONLY` events to prevent it from overwriting the strategy.
- Added `skipNodeLabelKey` / `skipNodeLabelValue` config fields with symmetric validation (both must be set or both empty).
- Added a startup warning if MetadataAugmentor is disabled but a skip label is configured.
- New Prometheus counters: `nvsentinel_platform_connector_metadata_skip_label_gated_total` and `nvsentinel_platform_connector_metadata_lookup_failures_total`.

**Docs:**
- Updated ADR-040 to reflect the platform-connector approach and the defense hierarchy.

## Live Cluster Validation

Tested on `nvs-dgxc-k8s-azr-scus-dev1` (50K+ nodes) with custom-built images deployed via Helm.

**Setup:**
- Platform-connector: MetadataAugmentor enabled with `cacheTTL=30s`, `skipNodeLabelKey=nvsentinel.dgxc.nvidia.com/managed`, `skipNodeLabelValue=false`
- Events triggered via KOM's `node-authtest-condition` policy (patching `NVSentinelAuthTest` condition on a GPU node)

**Test Results:**

| # | Scenario | Platform-Connector | Fault-Quarantine | Node Cordoned |
|---|----------|-------------------|------------------|---------------|
| 1 | `managed=false` label only | Gated to `STORE_ONLY` | N/A (event never reached FQ) | No |
| 2 | Release taint only (no label) | Passed through as `EXECUTE_REMEDIATION` | Skipped event (taint detected) | No |
| 3 | Both label and taint | Gated to `STORE_ONLY` (label caught first) | N/A (event never reached FQ) | No |

**Key findings:**
- MetadataAugmentor enriches events with node metadata before the skip-label check, so gated events still carry audit-useful metadata in the database
- `STORE_ONLY` events are correctly skipped by the Kubernetes connector (no node conditions written, no remediation triggered)
- When both signals are present, the primary gate (platform-connector) catches the event first, making the FQ backup unnecessary but still available as defense-in-depth

This is Part 1 of the external remediation gating work. Part 2 (monitor state synchronization for KOM/CSP after a node returns from external repair) will be addressed in a follow-up ADR.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: Platform Connectors, Commons

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Metadata augmentation is enabled by default with a 60-second metadata cache.
  * Configure node-label gating with a single `key=value` setting to store matching events without remediation or Kubernetes side effects.
  * Added metrics for label-gated events and metadata lookup failures.
  * Generic-provider reboot jobs can optionally write attribution messages to node syslog.

* **Bug Fixes**
  * Metadata lookup failures now fail open, preserving normal event processing.
  * Store-only events bypass duplicate filtering while retaining their processing strategy.
  * Invalid node-label configurations are rejected.

* **Documentation**
  * Updated cache timing and store-only behavior guidance.


cbumb:
Hey team, flagging a few known tradeoffs and open questions with this approach. Would love your thoughts:

1. **Cache TTL vs API cost**: We reduced MetadataAugmentor's cache TTL from 1 hour to 30 seconds so the managed=false label takes effect quickly. This means more frequent Kubernetes API calls for node lookups. Each call is lightweight (single GET from the local API server), but on large clusters this adds up. Are we comfortable with 30s, or should we tune it differently?

2. **MetadataAugmentor was disabled by default**: The gate only works when MetadataAugmentor is enabled. We've now enabled it by default in all values files, and added a startup warning if someone disables it while a skip label is configured. Wanted to confirm there's no reason it was previously disabled that we're missing.

3. **Fail-open on cache miss**: If the Kubernetes lookup fails (cold start, API server blip, etc.), we let the event through ungated rather than dropping it. The thinking here is that a false positive (briefly acting on an opted-out node) is recoverable, while a false negative (missing a real fault on a healthy node) is not. The FQ defense-in-depth check using the node informer acts as a backstop for this. Does this safety posture feel right, or should we be more conservative?

4. **Monitor state sync (Part 2)**: This PR does not address KOM/CSP internal state going out of sync when events are gated. That's deferred to a follow-up ADR since it needs a broader design. Just making sure everyone's aligned on that split.

lalitadithya:
> The gate only works when MetadataAugmentor is enabled. We've now enabled it by default in all values files, and added a startup warning if someone disables it while a skip label is configured. Wanted to confirm there's no reason it was previously disabled that we're missing.

The feature that we have here can be a different transformer and this will avoid this coupling. 

> The FQ defense-in-depth check using the node informer acts as a backstop for this. Does this safety posture feel right, or should we be more conservative?

Don't we already have a FQ check based on the label? Do we need an additional one?

cbumb:
> The feature that we have here can be a different transformer and this will avoid this coupling.

We actually explored the separate-transformer route early on, but @natherz97 pushed back on it. His feedback was that having two transformers doing node lookups with different caching mechanisms (one with an LRU cache doing GET calls, another with an informer or its own cache) would be hard to reason about, and that we should be consistent with the existing transformer instead of adding a new one. That's what led us to put the gate inside MetadataAugmentor, since it already has the node metadata cache infrastructure. The startup warning covers the case where someone disables MetadataAugmentor while a skip label is configured.

distros/kubernetes/nvsentinel/values-full.yaml
```diff
@@ -365,7 +365,14 @@ platformConnector:
       cacheSize: 50
       # Cache TTL in seconds (supports duration strings like "1h", "30m")
       # How long to cache node metadata before refreshing
-      cacheTTLSeconds: 3600
+      cacheTTLSeconds: 30
```
lalitadithya: can we make this 1 minute, i.e., 60 seconds?

distros/kubernetes/nvsentinel/values-full.yaml
```diff
@@ -365,7 +365,14 @@ platformConnector:
...
+      # Node label key/value to gate events as STORE_ONLY.
+      # When the target node carries this label with this value, the event
+      # is stored but does not trigger remediation or Kubernetes side-effects.
+      # Leave skipNodeLabelKey empty to disable the gate.
+      # Must match commons/pkg/managed.ManagedLabelKey and ManagedLabelValueFalse.
+      skipNodeLabelKey: "nvsentinel.dgxc.nvidia.com/managed"
+      skipNodeLabelValue: "false"
```
lalitadithya: can we make this simpler:

```suggestion
      skipNodeLabel: "nvsentinel.dgxc.nvidia.com/managed=false"
```

with equal being the only operator supported?

platform-connectors/pkg/transformers/dedup/transformer.go
```diff
@@ -69,8 +69,14 @@ func (d *Deduplicator) Name() string {
 }
 
 // Transform downgrades duplicate unhealthy events to STORE_ONLY so they are persisted
-// but do not create Kubernetes-side remediation effects.
+// but do not create Kubernetes-side remediation effects. Events already marked
+// STORE_ONLY (e.g. by the managed-label gate) are returned immediately without
+// entering the dedup tracker.
 func (d *Deduplicator) Transform(ctx context.Context, event *pb.HealthEvent) error {
+	if event.GetProcessingStrategy() == pb.ProcessingStrategy_STORE_ONLY {
```
natherz97, line 76: My understanding is that transformer run in the order they're defined, can you confirm that MetadataAugmentor will run before the dedup tracker? 

Are we not making a similar check in OverrideTransformer since it may still be valid to override event properties even if we already overrode the processing strategy STORE_ONLY?

platform-connectors/pkg/transformers/metadata/config.go
```diff
@@ -63,5 +67,13 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("cacheTTL must be positive")
 	}
 
+	if c.SkipNodeLabelKey == "" && c.SkipNodeLabelValue != "" {
```
natherz97: Nit: you could combine this into a single if check but this won't be relevant if you follow Lalit's suggestion to use a single setting that includes key + value:
```
if (c.SkipNodeLabelKey == "") != (c.SkipNodeLabelValue == "") {
    return fmt.Errorf("skipNodeLabelKey and skipNodeLabelValue must both be set or both be empty")
}
```

platform-connectors/pkg/transformers/metadata/transformer.go
```diff
@@ -100,6 +114,18 @@ func (a *Augmentor) Transform(ctx context.Context, event *pb.HealthEvent) error
 		}
 	}
 
+	if metadata.SkipMatched {
```
natherz97, line 116: Can you update the existing ERR test to cover this new functionality? https://github.com/NVIDIA/NVSentinel/blob/main/tests/external_remediation_test.go 

We could do something like send an unhealthy event that would normally result in cordoning and confirm that the node is not cordoned after ERR completion.

platform-connectors/pkg/transformers/dedup/transformer.go
cbumb: Yes, transformers run in the order they're listed in the pipeline config, so MetadataAugmentor always runs before Deduplicator. The ordering is defined in the values.yaml pipeline array and NewFromConfigs iterates it sequentially.

> Are we not making a similar check in OverrideTransformer since it may still be valid to override event properties even if we already overrode the processing strategy STORE_ONLY?

That was intentional. The reason we only added the guard in Deduplicator is that it's the only transformer that can overwrite ProcessingStrategy. Dedup can change a STORE_ONLY event back to STORE_AND_ANALYSE, which would break the gate. OverrideTransformer only touches isFatal, isHealthy, and recommendedAction, so it can't undo the gate. Letting it still run on STORE_ONLY events is actually useful since the stored audit record should have correct property values.

platform-connectors/pkg/transformers/metadata/transformer.go
```diff
@@ -155,5 +180,11 @@ func (a *Augmentor) fetchNodeMetadata(ctx context.Context, nodeName string) (*No
 		}
 	}
 
+	if a.config.skipLabelKey != "" {
+		if val, ok := node.Labels[a.config.skipLabelKey]; ok && val == a.config.skipLabelValue {
```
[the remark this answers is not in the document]
cbumb: This is the expected behavior and is documented in ADR-040 as "Cache staleness: Up to the cache TTL (default 60s)." The 60s window is safe in both directions: when a node becomes externally managed, the ExtRR reconciler also applies the release taint which blocks scheduling independently. When a node returns to NVSentinel management, there's at most a 60s delay before remediation resumes. Adding a watch/informer for cache invalidation would add significant complexity for a window that's already covered by other safety mechanisms.

natherz97 (approved):

PR #1676, opened 2026-08-27
feat(health-event-analyser): scope HEA rules to incoming event's node

### LINKED ISSUE #1593: [Feature]: Add nodename filter to HEA rule Stage 0 to eliminate COLLSCAN on every event
### Prerequisites

### Code of Conduct

### Feature Summary

Every Health Events Analyzer rule evaluation performs a full collection scan of HealthEvents, regardless of collection size. At 1M documents, processing one event takes 1,852 ms P50. The fix is a one-line change per rule: add the incoming event's node name to Stage 0.

### Problem/Use Case

Each of HEA's 22 rule pipelines starts with a Stage 0 that filters only on time window:

```js
{ "$match": { "generatedtimestamp": { "$gte": <window_start> } } }
```

No `nodename` filter. MongoDB scans the entire collection before Stage 2 narrows to the relevant node. This is a COLLSCAN — confirmed via `explain("executionStats")`.

Every rule scopes exclusively to the node that generated the incoming event. `RepeatedXID31OnDifferentGPU`, for example, looks for the same XID on different GPUs on *that same node* — not across nodes. The `nodename` filter belongs in Stage 0, and its absence means every rule evaluation pays the full collection scan cost.

The self-reinforcing consequence: unprocessed events accumulate in HealthEvents, making each subsequent scan slower. Measured throughput dropped from 4.6 events/s to 3.3 events/s across a short burst as the collection grew — not because the workload changed, but because Stage 0 got slower.

At 100k nodes with a realistic fault rate, HEA cannot sustain production load.

### Proposed Solution

Add `nodename` to Stage 0 in every rule pipeline, resolved at runtime from the incoming event:

```js
{ "$match": {
    "healthevent.nodename": "$$currentNodeName",
    "generatedtimestamp": { "$gte": "$$window_start" }
}}
```

`$$currentNodeName` is passed as a pipeline variable (via MongoDB's `let` clause) from the incoming event's node name. No hardcoding — fully dynamic per reconcile.

This lets the existing compound index `(nodename, entitytype, entityvalue, generatedtimestamp)` satisfy Stage 0, reducing it from a full collection scan to only that node's documents — a 170× reduction at 1M docs.

**Expected improvement**: Throughput from ≈5 events/s to ≈500 events/s. P50 per-event latency from 1,852 ms to ≈2 ms at 1M docs.

### Component

Health Monitor

XRFXLP:
### What this change is

Adds an enforced "same node" constraint as the mandatory first filter in every rule's pipeline, alongside the existing agent filter. A rule triggered by an event on node X now only queries events from node X — at the framework level, not left to individual rule authors.

### What rules benefit and why

Rules that detect repeated failures already embedded a "same node" check in their logic — those were unaffected. The rules that benefited are the XID 74 register-bit classifiers: rules that classify a single incoming event by its hardware register dump. Because they only inspect the incoming event's own data, the node constraint was never part of their query, and they silently scanned all events across all nodes at scale. This PR closes that gap — for them it's both a correctness fix and a large performance win.

### Benchmark results

1M events across 100 nodes (10k/node), 10 XID 74 trigger events, latency from `mongo_query_execution_duration_seconds`:

| Rule | v1.20.0 | This PR | Speedup |
|---|---|---|---|
| XID74Reg0SignalIntegrityError | 437 ms | 0.9 ms | ~485× |
| XID74Reg0SoloNVLinkError | 431 ms | 0.9 ms | ~465× |
| XID74Reg2Bit13Set | 432 ms | 15 ms | ~29× |
| XID74Reg3Bit18Set | 433 ms | 0.9 ms | ~495× |
| XID74Reg3UnexpectedError | 433 ms | 0.9 ms | ~476× |
| All other rules (17) | ~1 ms | ~1 ms | no change |

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Rule-based event aggregation now evaluates only events from the same node as the incoming event.
  * Prevents unrelated node activity from affecting aggregation results.
  * Preserves existing filtering for analyzer-generated events and configured processing strategies.
  * Improves the accuracy and consistency of event analysis across multiple nodes.
* **Tests**
  * Added regression coverage to verify node-specific filtering is applied consistently.


tanishagoyal2:
LGTM

lalitadithya (approved):

PR #1683, opened 2026-08-28
fix(store-client): stop on uncheckpointed events

Saibernard:
## Summary

`DefaultEventProcessor` currently logs handler and checkpoint errors, then continues consuming events. If a later event succeeds, its checkpoint can advance past the unresolved event.

This change stops the loop whenever the current event has not been safely checkpointed. It preserves the existing `MarkProcessedOnError=true` behavior when the failed event is successfully checkpointed.

## How this surfaced

I found this while tracing the Health Events Analyzer failure and restart path, not from a production incident. HEA sets `MarkProcessedOnError=false` and documents that failed events will be retried after restart.

I reproduced the failure with a deterministic watcher: event 1 returns a handler error, event 2 succeeds, and the processor checkpoints event 2 even though event 1 remains unresolved.

## Impact

For PostgreSQL, checkpointing event 2 marks changelog entries through that ID as processed. For MongoDB, it stores event 2's resume token. In either case, event 1 may not be retried after restart. In HEA, that means a transient rule-evaluation or publication failure can prevent the expected fatal event from reaching downstream processing.

Supersedes #1682, which GitHub closed when the fork branch was renamed to match the project naming convention.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring
- [ ] Build/CI

## Component Affected
- Core Services: store-client

## Testing
- [x] `go test -race ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run --config ../.golangci.yml`
- [x] No public API or configuration changes

## Checklist
- [x] Self-review completed
- [x] Documentation is not needed for this behavior fix
- [x] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Event processing now stops when an event cannot be safely checkpointed, preventing later events from being processed incorrectly.
  * Processing failures now preserve both the original error and any checkpoint failure.
  * Events with unavailable document IDs are reported with a consistent fallback value.
  * Configured error-handling behavior is now applied consistently when deciding whether to continue processing.

### DIFF
--- a/store-client/pkg/client/event_processor.go
+++ b/store-client/pkg/client/event_processor.go
@@ -16,6 +16,7 @@ package client
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"maps"
@@ -70,6 +71,24 @@ type DefaultEventProcessor struct {
 	config              EventProcessorConfig
 	eventHandler        EventHandler
 	stopCh              chan struct{}
+}
+
+// uncheckpointedEventError indicates that the processor must stop before
+// consuming another event, because the current event remains unresolved.
+type uncheckpointedEventError struct {
+	cause error
+}
+
+func (e *uncheckpointedEventError) Error() string {
+	return fmt.Sprintf("event was not checkpointed: %v", e.cause)
+}
+
+func (e *uncheckpointedEventError) Unwrap() error {
+	return e.cause
+}
+
+func newUncheckpointedEventError(cause error) error {
+	return &uncheckpointedEventError{cause: cause}
 }
 
 // NewEventProcessor creates a new unified event processor
@@ -145,6 +164,11 @@ func (p *DefaultEventProcessor) processEvents(ctx context.Context) error {
 
 			if err := p.handleSingleEvent(ctx, event); err != nil {
 				slog.Error("Failed to handle event", "eventID", eventID, "error", err)
+
+				var uncheckpointedErr *uncheckpointedEventError
+				if errors.As(err, &uncheckpointedErr) {
+					return fmt.Errorf("stopping at uncheckpointed event %q: %w", eventID, err)
+				}
 			}
 		}
 	}
@@ -194,7 +218,7 @@ func (p *DefaultEventProcessor) handleSingleEvent(ctx context.Context, event Eve
 	if markErr := p.markProcessed(ctx, token); markErr != nil {
 		p.updateMetrics("mark_processed_error", eventID, time.Since(startTime), false)
 
-		return fmt.Errorf("failed to mark event as processed: %w", markErr)
+		return newUncheckpointedEventError(fmt.Errorf("failed to mark event as processed: %w", markErr))
 	}
 
 	return nil
@@ -207,7 +231,7 @@ func (p *DefaultEventProcessor) handleProcessingError(
 		slog.Error("Event processing failed, NOT marking as processed - will retry on restart",
 			"eventID", eventID, "error", processErr)
 
-		return processErr
+		return newUncheckpointedEventError(processErr)
 	}
 
 	slog.Warn("Marking failed event as processed due to MarkProcessedOnError=true",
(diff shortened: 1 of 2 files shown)


Saibernard:
Addressed the test naming comment in 803e6ae6 by using the project's function, scenario, and expected-behavior format.

Saibernard:
Addressed both follow-up comments in 6b7e7e8f. The tests now assert the original event error and checkpoint error, and the test function has the prescribed comment.

XRFXLP:
hey @Saibernard, thanks for contributing!

Can you please add some context i.e how this issue manifested in your case and what was the impact?

Ideally, it should be GH issue but it can be in PR description as well if minor fix.

Saibernard:
Thanks @XRFXLP. I found this while tracing HEA's failure and restart behavior, not from a production incident.

HEA sets `MarkProcessedOnError=false` and expects failed events to be retried after restart. I reproduced a case where event 1 fails, processing continues, and event 2 advances the durable checkpoint. This can prevent event 1 from being retried.

I added the reproduction and impact details to the PR description.

store-client/pkg/client/event_processor_test.go
```diff
@@ -0,0 +1,236 @@
...
+		config        EventProcessorConfig
+		firstEvent    *eventProcessorTestEvent
+		handlerErrors map[string]error
+		markErrors    map[string]error
+		wantErrors    []error
+		wantHandled   []string
+		wantMarkCalls []string
+		wantMarked    []string
+	}{
+		{
```
KaivalyaMDabhadkar, line 55: Could you also add the plain happy-path case, where both events succeed and both get checkpointed in order?

store-client/pkg/client/event_processor.go
```diff
@@ -145,6 +164,11 @@ func (p *DefaultEventProcessor) processEvents(ctx context.Context) error {
 
 			if err := p.handleSingleEvent(ctx, event); err != nil {
 				slog.Error("Failed to handle event", "eventID", eventID, "error", err)
+
+				var uncheckpointedErr *uncheckpointedEventError
+				if errors.As(err, &uncheckpointedErr) {
+					return fmt.Errorf("stopping at uncheckpointed event %q: %w", eventID, err)
```
KaivalyaMDabhadkar, line 174: nit: in the document ID error path, eventID is empty here, so the message comes out as `stopping at uncheckpointed event ""`. Can we fall back to "unknown" so the log stays readable?

store-client/pkg/client/event_processor_test.go
Saibernard: Added in 33f78ab4. The table now includes a plain two-event success case and asserts handler order, checkpoint-call order, and successfully marked-token order as [1, 2].

store-client/pkg/client/event_processor.go
Saibernard: Fixed in 33f78ab4. processEvents now normalizes an empty or failed document-ID lookup to "unknown" before logging or returning the uncheckpointed-event error, and the regression case asserts that readable message.

Saibernard:
Both requested review changes are addressed at `33f78ab453eb5fec9c60cc0f0fcbb3cfc5e5e637`, and both threads are resolved. Could a PR vetter please authorize this exact head with `/ok to test 33f78ab453eb5fec9c60cc0f0fcbb3cfc5e5e637` so the refreshed CI and E2E matrices can run?

Saibernard:
The two red Container Build Validation checks are one underlying failure plus its aggregate summary: `preflight-dcgm-diag` hit a BuildKit QEMU ARM64 emulator SIGSEGV while running the Ubuntu `libc-bin` post-install script (`exit 139`), and `container-build-summary` reflects that child failure. The PR changes only store-client files. I reran the exact multi-platform target locally from this head (`linux/arm64,linux/amd64`), including the same `libc-bin` stage, and it completed successfully. Could a maintainer please rerun the failed job in [run 33389524508](https://github.com/NVIDIA/NVSentinel/actions/runs/33389524508)?

KaivalyaMDabhadkar (approved):

Saibernard:
thanks @KaivalyaMDabhadkar :)

Issue #126, opened 2025-10-23
[Feature]: Support for GPU reset


lalitadithya:
### Prerequisites

- [x] I searched existing issues

### Feature Summary

Not all GPU failures need to be remediated with a reboot; most can be remediated with a GPU reset. Some inference workloads do not use all the GPUs available on the node and therefore reboot a node will require all the workloads running on the node to be evicted. In some case reboots of the bare metal can take close to 30 minutes which increases the mean time to repair.

### Problem/Use Case

Reduce the mean time to repair by using GPU reset instead of node reboot for supported faults

### Proposed Solution

Introduce a new GPU reset CRD, controller and update the fault remediation module to create the appropriate CR based on the nature of the fault.

### Component

Fault Management

natherz97:
@ivelichkovich Thanks for the context! We want to add support for GPU reset to prevent needing to drain pods which are using healthy GPUs on the same node. This prevents us from force removing all processes leveraging the driver on that node. Our plan is to drain only GPU pods leveraging the GPU that needs reset and then temporarily disabling all gpu-operator pods except the driver pod to prevent impact to other pods on healthy GPUs (along with also disabling persistent mode on that GPU which is a process started the driver pod). This will allow us to reset only unhealthy GPUs without impact to existing workloads.

ivelichkovich:
> [@ivelichkovich](https://github.com/ivelichkovich) Thanks for the context! We want to add support for GPU reset to prevent needing to drain pods which are using healthy GPUs on the same node. This prevents us from force removing all processes leveraging the driver on that node. Our plan is to drain only GPU pods leveraging the GPU that needs reset and then temporarily disabling all gpu-operator pods except the driver pod to prevent impact to other pods on healthy GPUs (along with also disabling persistent mode on that GPU which is a process started the driver pod). This will allow us to reset only unhealthy GPUs without impact to existing workloads.

Awesome! That's even better 😄

PR #1692, opened 2026-08-28
fix(gpu-reset): disable persistence mode before resetting a GPU

### LINKED ISSUE #1690: [Bug]: GPU reset fails with "In use by another client" because gpu_reset.sh never disables persistence mode
### Prerequisites

### Code of Conduct

### Bug Description

`gpu_reset.sh` invokes `nvidia-smi --gpu-reset` without first disabling persistence
mode on the target GPU. Where persistence mode is enabled, `nvidia-persistenced`
holds an attachment to the device and the reset is refused:

```
The following GPUs could not be reset:
  GPU 00000000:1A:00.0: In use by another client

1 device is currently being used by one or more other processes (e.g. Fabric
Manager, CUDA application, graphics application such as an X server, or a
monitoring application such as another instance of nvidia-smi). Please first kill
all processes using this device and all compute applications running in the
system.
exit=255
```

Disabling persistence mode is sufficient to make the same reset succeed on the
same GPU seconds later, with no other change:

```
$ nvidia-smi -pm 0 -i <uuid>
All done.
$ nvidia-smi -r -i <uuid>
GPU 00000000:1A:00.0 was successfully reset.
exit=0
```

This appears to be a known-but-unimplemented step. #126 describes the intended
design as "temporarily disabling all gpu-operator pods except the driver pod ...
along with also disabling persistent mode on that GPU which is a process started
the driver pod". The pod teardown half is implemented; the persistence-mode half
is not, in any released version I can find.

Worth noting that an open file descriptor on `/dev/nvidia*` is *not* what blocks
the reset. During the successful run above, `nvidia-persistenced` still held the
device open according to `/proc/<pid>/fd`. It is persistence *mode* that
constitutes the attachment, which is why the fix is `-pm 0` rather than stopping
the driver pod.

### Component

Fault Management

### Steps to Reproduce

1. On a node where persistence mode is enabled (`nvidia-smi -q | grep -i
   "Persistence Mode"` reports `Enabled`), ensure no compute processes hold the
   target GPU.
2. Create a `GPUReset` CR selecting that GPU, or run the reset job's command
   directly: `nvidia-smi --gpu-reset -i <uuid>`.
3. The reset fails with `In use by another client`, exit 255.
4. Run `nvidia-smi -pm 0 -i <uuid>`, then repeat step 2. It succeeds.

### Environment

- NVSentinel version: reproduced against v1.16.0; `gpu_reset.sh` is byte-identical in v1.20.0
- GPU: NVIDIA B300 SXM6, 8 per node
- NVIDIA driver: 580.95.05
- Platform: bare metal (`systemd-detect-virt` reports `none`)
- Deployment method: Helm chart from `oci://ghcr.io/nvidia/nvsentinel`

### Logs/Output

```
The following GPUs could not be reset:
  GPU 00000000:1A:00.0: In use by another client
exit=255
```

### Suggested fix

In `gpu_reset.sh`, around the existing reset call: record the current persistence
mode, set it to `0` for the target UUIDs before `--gpu-reset`, and restore it
afterwards regardless of the reset outcome. Happy to send a PR if the approach
looks right.

gopuman:
Fixes #1690.

## What

Disables persistence mode on the target GPUs before `nvidia-smi --gpu-reset`, and
restores it afterwards.

## Why

Persistence mode counts as an attachment to the device, so the reset is refused
while it is enabled:

```
The following GPUs could not be reset:
  GPU 00000000:1A:00.0: In use by another client
exit=255
```

Disabling it is sufficient to make the same reset succeed seconds later on the
same GPU, with nothing else changed:

```
$ nvidia-smi -pm 0 -i <uuid>
All done.
$ nvidia-smi -r -i <uuid>
GPU 00000000:1A:00.0 was successfully reset.
exit=0
```

#126 describes this as part of the intended design — "along with also disabling
persistent mode on that GPU which is a process started the driver pod". The pod
teardown half is implemented; this adds the other half.

## Notes on the implementation

**Prior state is recorded per GPU and restored from the existing `EXIT` trap.** A
GPU whose persistence mode was already disabled is left untouched, and a failed
reset does not leave the setting changed — the trap runs on the error path too.

**The query is `|| true`-guarded.** The script runs under `set -eou pipefail`, so
an unqueryable GPU would otherwise abort the whole reset via the failed command
substitution. That case is real rather than theoretical: a GPU that has dropped
off the PCIe bus is exactly the kind of fault a reset gets requested for, and one
bad device must not prevent the others from being reset.

**No behaviour change when persistence mode is already off**, which is the default
after boot, so this is inert on nodes that never enable it.

## Testing

Verified on an 8-GPU bare-metal node (B300 SXM6, driver 580.95.05):

- with persistence mode enabled, `--gpu-reset` returns exit 255 `In use by another client`
- after `-pm 0`, the same reset returns exit 0 and the node returns to its full GPU count

Also exercised the `set -e` guard with a stub that fails the query, confirming the
loop continues rather than exiting.

One incidental finding worth recording: an open file descriptor on `/dev/nvidia*`
is *not* what blocks the reset. `nvidia-persistenced` still held the device open
during the successful run above. It is persistence *mode* that constitutes the
attachment, which is why this is the right fix rather than stopping the driver pod.

## Summary by CodeRabbit

- **Bug Fixes**
  - GPU resets now safely handle GPUs with persistence mode enabled.
  - Persistence mode is restored after resets, including when failures occur.
  - Unavailable GPU status information and restoration failures no longer interrupt the reset workflow.
  - Reset operations now log persistence-mode changes and related failures clearly.

### DIFF
--- a/gpu-reset/gpu_reset.sh
+++ b/gpu-reset/gpu_reset.sh
@@ -51,6 +51,42 @@ nvidia_smi_helper() {
   chroot "$DRIVER_ROOT" nvidia-smi "$@"
 }
 
+# Persistence mode counts as an attachment to the device, so nvidia-smi --gpu-reset
+# is refused with "In use by another client" while it is enabled. Record the
+# per-GPU state so it can be restored exactly, including on the failure path.
+PERSISTENCE_RESTORE=""
+
+disable_persistence_mode() {
+  local uuid
+  local mode
+  for uuid in $(echo "$1" | tr ',' ' '); do
+    # Tolerate an unqueryable GPU: the script runs under `set -e`, and a device
+    # that has dropped off the bus must not abort the whole reset here.
+    mode=$(nvidia_smi_helper --query-gpu=persistence_mode --format=csv,noheader -i "$uuid" 2>/dev/null | tr -d '[:space:]' || true)
+    if [ "$mode" != "Enabled" ]; then
+      continue
+    fi
+    if nvidia_smi_helper -pm 0 -i "$uuid" >/dev/null 2>&1; then
+      PERSISTENCE_RESTORE="${PERSISTENCE_RESTORE}${uuid} "
+      log "INFO: Disabled persistence mode on ${uuid} for the duration of the reset."
+    else
+      log "WARN: Failed to disable persistence mode on ${uuid}; the reset may be refused."
+    fi
+  done
+}
+
+restore_persistence_mode() {
+  local uuid
+  for uuid in $PERSISTENCE_RESTORE; do
+    if nvidia_smi_helper -pm 1 -i "$uuid" >/dev/null 2>&1; then
+      log "INFO: Restored persistence mode on ${uuid}."
+    else
+      log "WARN: Failed to restore persistence mode on ${uuid}."
+    fi
+  done
+  PERSISTENCE_RESTORE=""
+}
+
 prepend_driver_root() {
   if [ "$DRIVER_ROOT" = "/" ]; then
     printf "%s\n" "$1"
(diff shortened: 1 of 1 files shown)


lalitadithya:
hi @gopuman , can you please complete the DCO on this PR? https://github.com/NVIDIA/NVSentinel/pull/1692/checks?check_run_id=98913196038

gopuman:
Closing as unnecessary based on the controlled B200/B300 A/B recorded in #1690: with the persistenced socket visible in Janitor's supported driver-root layout, unmodified v1.16.0 handles persistence automatically and restores it after a successful reset.

PR #1693, opened 2026-08-28
feat(fault-quarantine): recover unresolved events on startup

### LINKED ISSUE #1674: [Bug]: not recovering from loss-of-mongodb queue
### Prerequisites

### Code of Conduct

### Bug Description

https://github.com/NVIDIA/NVSentinel/blob/83d88554fb1ee224af5b746a485a8105e4f1c8e6/health-monitors/gpu-health-monitor/gpu_health_monitor/platform_connector/platform_connector.py#L680 seems to not emit an event if the event has already been emitted.

I think that the thinking is that if the event has been emitted, it is somewhere in the mongodb queue, and will eventually be processed.

Where it fails: if we lose mongodb state (e.g. #1594 + long Circuit Breaker condition, but I suppose that this failure mode might occur under different conditions), then I think that the condition may never be processed.

Is the idea that in this case, https://github.com/NVIDIA/NVSentinel/blob/main/docs/runbooks/stale-events.md should be run, and that nvsentinel does not try to self-heal? If that is the case it's not ideal but I could respect this software choice ;-)

At the same time, one could wonder if, at startup, health monitors should try to re-emit events, at least once, to improve the chance of recovery.

Downside of this suggestion: a crash-looping Monitor might spam the queue and cause a cascading failure. I wonder if there might be a way to mitigate this a bit with some persistence as a PVC for monitors perhaps, and re-emit if last emitted ... 1h ago?

Upside is a higher chance of these complex systems converging and self-healing :-)

### Component

Health Monitor

### Steps to Reproduce

1. Activate Circuit Breaker
2. Have some nodes with gpu-health-monitor failures
3. Losing mongodb queue
4. <never treat the original failures AFAICT>

### Environment

- NVSentinel version: 1.20
- Kubernetes version: 1.35

### Logs/Output

_No response_

Saibernard:
## Summary

Fault-quarantine relies on its datastore resume position to continue processing health events. If that position is lost or expires, events already stored in MongoDB or PostgreSQL can remain unresolved.

This change adds startup recovery for the fault-quarantine side of #1674:

- Opens the live watcher before recovery so new events are not missed.
- Replays unresolved events in creation order through the existing rule and circuit-breaker paths.
- Suppresses obsolete events when newer events have changed the same state.
- Marks skipped, invalid, and superseded events as complete so later startups converge.
- Preserves the `CREATE` reset boundary so deliberately abandoned backlogs are not revived.
- Adds bounded datastore iteration, supporting indexes, metrics, and documentation.

During recovery, node-dependent rules read current API-server state instead of relying on the informer cache. Transient failures stop recovery and are retried on restart, while permanent event-specific failures are recorded and skipped.

### Scope

This recovers unresolved events that are still present in the datastore. It does not recreate events that are no longer stored or change health-monitor re-emission behavior. For that reason, this PR references #1674 rather than closing it.

Supersession is document-level. If a newer event changes state represented by a compound event, the older document is skipped as a unit because partially replaying it would still expose the original complete document downstream.

### Deployment note

The required indexes are included in the MongoDB initialization templates and PostgreSQL schema paths. Existing or custom deployments that do not rerun those initialization paths must apply the corresponding indexes during upgrade.

Refs #1674

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Refactoring
- [ ] Build/CI

## Component(s) Affected

- [x] Core Services
- [x] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other

## Testing

- [x] Tests pass locally
- [x] No breaking changes

### Verification

Key regression coverage:

- Missed fatal events are recovered end-to-end: `TestE2E_ColdStartRecoversMissedFatalEvent`
- Obsolete failures do not quarantine recovered nodes: `TestE2E_ColdStartSkipsFailureSupersededByRecovery`
- Deleted nodes cannot cause a startup crash loop: `TestE2E_ColdStartCompletesEventForDeletedNode`
- Live events arriving during recovery are preserved and deduplicated: `TestProcessStoredEventUsesLiveProcessingPathAndDeduplicatesReplay`
- PostgreSQL iteration remains stable for mutating and OR-rooted queries: `TestFindHealthEventsByQueryBatchedUsesStableKeysetPagination`, `TestFindHealthEventsByQueryBatchedParenthesizesORFilterBeforeCursor`
- Operator-requested `CREATE` resets remain crash-safe: `TestHandleCircuitBreakerCreatePersistsCutoffBeforeDeletingToken`
- Missing and explicit-null status documents complete successfully: `TestCompleteStoredEventReplacesExplicitNullStatus`

Validation completed at `656ab67e`:

- `go test -race ./... -count=1` in `store-client`
- `go test -race ./... -count=1` in `fault-quarantine`, including envtest
- `go vet ./...` in both modules
- `golangci-lint run ./...` in both modules
- Manual reproduction against MongoDB 8.0

## Checklist

- [x] Self-review completed
- [x] Documentation updated
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Fault Quarantine now recovers unresolved health events at startup before processing new events.
  * Added supersession handling, batched completion updates, and startup recovery metrics.
  * Completion-only updates can be skipped without unnecessary processing.

* **Bug Fixes**
  * Invalid health events without status data now return clear errors.
  * Recovery more reliably handles deleted nodes, failures, ordering, retries, and partial updates.
  * Skipped and recovered events are durably completed before checkpointing.

* **Performance**
  * Improved health-event lookup with stable indexes and keyset pagination.

* **Documentation**
  * Documented startup recovery behavior, metrics, and cutoff configuration.


fault-quarantine/pkg/evaluator/rule_evaluator.go
```diff
@@ -143,26 +163,33 @@ func NewNodeRuleEvaluator(expression string, nodeLister corelisters.NodeLister)
...
+	ctx context.Context,
+	event *protos.HealthEvent,
+) (common.RuleEvaluationResult, error) {
 	slog.Info("Evaluating NodeRuleEvaluator for node", "node", event.NodeName)
 
-	nodeInfo, err := nm.getNode(event.NodeName)
+	nodeInfo, err := nm.getNode(ctx, event.NodeName)
 	if err != nil {
 		return common.RuleEvaluationFailed, fmt.Errorf("failed to get node metadata: %w", err)
 	}
```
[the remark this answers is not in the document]
Saibernard: Fixed in dd71ff7e. Recovery-time node-rule reads now classify Kubernetes NotFound errors as permanent, so a deleted node cannot block startup recovery. TestNodeRuleEvaluator_DeletedNodeDuringRecovery_ReturnsPermanentError covers this path.

store-client/pkg/datastore/providers/postgresql/health_events.go
```diff
@@ -869,24 +870,34 @@ func (p *PostgreSQLHealthEventStore) FindLatestHealthEventByQuery(ctx context.Co
 
 	//nolint:gosec // G202 false positive - using parameterized query with placeholders
 	query := `
-		SELECT document FROM health_events
+		SELECT created_at, document FROM health_events
 		WHERE ` + whereClause + `
 		ORDER BY created_at DESC
 		LIMIT 1
 	`
```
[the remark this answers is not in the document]
Saibernard: Fixed in c8d7ec22. PostgreSQL now orders equal-timestamp events by id DESC, with a focused regression test. The MongoDB path uses the equivalent createdAt and _id ordering for provider consistency.

lalitadithya:
@KaivalyaMDabhadkar , can you review please?

fault-quarantine/pkg/coldstart/supersession.go
```diff
@@ -0,0 +1,221 @@
...
+	for i := range timeline {
+		if !eventAfter(timeline[i], record.CreatedAt, candidateID) {
+			continue
+		}
+
+		if candidate.HealthEvent.GetIsHealthy() && len(candidate.HealthEvent.GetEntitiesImpacted()) == 0 {
+			return true, nil
+		}
+
+		if affected.RemoveEvent(timeline[i].event.HealthEvent) > 0 {
```
KaivalyaMDabhadkar: This skip logic goes wrong in both directions with real events:

Too little skipping: an old healthy event (which carries no error codes) is not recognized as outdated by a newer failure that does carry error codes, because the matching only works one way. When that old healthy event is replayed, it wipes the newer failure from the node and uncordons a node that is still broken.

Too much skipping: an event covering several GPUs is dropped entirely if a newer event touched just one of them, so the GPUs that never recovered are ignored and the node is never cordoned. The same happens to check-level "all clear" events like the NIC baseline clear: the special case above skips them whenever any newer event exists, which leaves stale entries behind so the node can never be uncordoned.

my suggestion is to skip an old event only when the newer events cover everything it would do, and make the matching work in both directions. The special case for healthy events without entities can then go away. Could you also add tests where the newer failure carries error codes? The current test data never sets them, which is why this passes today

fault-quarantine/pkg/coldstart/coldstart.go
```diff
@@ -0,0 +1,298 @@
...
+		query.Eq("healtheventstatus.nodequarantined", ""),
+		query.Eq("healtheventstatus.nodequarantined", string(model.StatusNotStarted)),
+	)
+	recoveryIncomplete := query.Or(
+		query.Eq(RecoveryCompletionStatusPath, nil),
+		query.Eq(RecoveryCompletionStatusPath, ""),
+	)
+	condition := query.And(unresolved, recoveryIncomplete, processableCondition())
+
+	if !coldStartAfter.IsZero() {
```
KaivalyaMDabhadkar, line 452: On the very first startup after upgrading, no cutoff exists yet, so this scan has no starting point and walks the entire stored history. Every event the service ever chose not to act on (healthy events, events matching no rule) still counts as unresolved, so that can be weeks of data, with a database write for each one, all before a single live event is handled. And one temporary database or API error exits the pod and the whole sweep starts over. Could we default the cutoff to now when none exists, and keep going past a temporary failure on one event?

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1828,11 +1886,31 @@ func (r *Reconciler) getNodeQuarantineAnnotations(ctx context.Context, nodeName
...
+	slog.DebugContext(ctx, "Retrieved quarantine annotations for node", "node", nodeName)
 
 	return quarantineAnnotations, nil
 }
 
+// getNode bypasses the informer during recovery so reconciliation uses current node state.
+// A node deleted before replay is a permanent event error.
+func (r *Reconciler) getNode(ctx context.Context, nodeName string) (*corev1.Node, error) {
+	if coldstart.IsRecoveryContext(ctx) {
+		node, err := r.k8sClient.Clientset.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
```
KaivalyaMDabhadkar: I think the reason for going straight to the api server here is that replay happens back to back, so the informer cache may not have caught up with our own writes from the previous event yet. That is a fair concern, but the only freshness we really need is our own writes, and right now every read is a fresh api call, so one replayed event can fetch the same node up to 6 times through a client limited to 5 requests per second. A big backlog then spends an hour or more just waiting on the rate limiter while live events queue, and these direct reads can also fail transiently in ways the informer never did.
Could we fetch the node once at the start of each replayed event, keep it in the recovery context, and reuse it for every read within that event, including the rule evaluator's read in rule_evaluator.go? One fresh read per event still sees everything the previous events wrote, so the guarantee is kept and the api load drops several fold

fault-quarantine/pkg/eventwatcher/event_watcher.go
```diff
@@ -178,6 +240,150 @@ func (w *EventWatcher) processEvent(ctx context.Context, event client.Event) err
...
+	}
+
+	fields := map[string]any{coldstart.RecoveryCompletionStatusPath: string(result)}
+	if status, exists := event.RawEvent["healtheventstatus"]; exists && status == nil {
+		fields = map[string]any{"healtheventstatus": map[string]any{
+			"faultquarantinerecovery": string(result),
+		}}
+	}
+
+	if err := w.databaseClient.UpdateDocumentStatusFields(ctx, recordUUID, fields); err != nil {
```
KaivalyaMDabhadkar: On postgres this write lands in the changelog, and health event analyzer listens to updates there without checking whether it already handled the document. So every historical event we mark here gets delivered to the analyzer again, its rules match again (they look at the event's own timestamp), and it publishes new fatal events that can quarantine nodes that recovered long ago. The first post-upgrade run would do this for thousands of events at once. Could the analyzer ignore updates that only set the recovery marker?

lalitadithya:
hey @Saibernard , can you please check the failing end to end tests?

store-client/pkg/datastore/providers/mongodb/health_store.go
```diff
@@ -388,15 +395,20 @@ func decodeRawDocToHealthEvent(rawDoc map[string]any) (datastore.HealthEventWith
...
 	filter := builder.ToMongo()
+	findOptions := &client.FindOptions{
+		Sort: bson.D{
+			{Key: "createdAt", Value: 1},
+			{Key: "_id", Value: 1},
+		},
+	}
 
-	cursor, err := h.databaseClient.Find(ctx, filter, nil)
+	cursor, err := h.databaseClient.Find(ctx, filter, findOptions)
```
KaivalyaMDabhadkar, line 479: This holds one cursor open for the whole recovery, and all the slow per-event work happens between reads. mongo kills a cursor that sits idle for 10 minutes, so one slow batch makes the next read fail with CursorNotFound and the whole recovery restarts. A batch size cannot fix this, since the idle time is the processing time. My suggestion is to page the query the way the postgres side already does, one bounded query per batch resuming from the last (createdAt, id). This stays inside this one function, FindOptions already supports Limit, and the createdAt plus id index added in this PR is exactly what the paging needs.

fault-quarantine/pkg/coldstart/coldstart.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil
+	}
+
+	result, err := processor.ProcessStoredEvent(ctx, event)
+	if err != nil {
+		return recordRecoveryFailure(fmt.Errorf("failed to recover stored event: %w", err))
+	}
+
+	if result == ProcessResultSkipped || result == ProcessResultInvalid {
+		if err := completeStoredEvent(ctx, processor, event, result); err != nil {
```
KaivalyaMDabhadkar: This writes one status update per skipped event, so the first big sweep pays one database round trip for every historical event it decides to leave alone. Both database clients already have bulk update support (UpdateManyDocuments). Could we collect the IDs within each scan batch and mark them with one bulk write per batch instead?

fault-quarantine/pkg/coldstart/supersession.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	records := make([]datastore.HealthEventWithStatus, 0)
+
+	err := r.store.FindHealthEventsByQueryBatched(
+		ctx,
+		query.New().Build(condition),
+		batchSize,
+		func(batch []datastore.HealthEventWithStatus) error {
+			records = append(records, batch...)
```
KaivalyaMDabhadkar: This loads the full history of newer events for a check into memory, and the cache keeps it for the whole recovery. For a check that flaps often, that can be tens of MB. It could stop reading as soon as it has the answer instead of loading everything first

store-client/pkg/client/postgresql_client.go
```diff
@@ -338,6 +338,8 @@ func (c *PostgreSQLClient) UpdateDocumentStatusFields(
 		parts := strings.Split(path, ".")
 		jsonbPath := "{" + strings.Join(parts, ",") + "}"
 
+		setExpression = initializeJSONBParents(setExpression, parts)
```
KaivalyaMDabhadkar: This rebuilds the parent-initialization SQL for every field, and each nesting level triples the size of the statement built so far. The normal quarantine status update (3 fields) already produces a 9KB SQL statement, and one more nested field would multiply that again. The update builder in store-client/pkg/query/update.go already solves this by remembering which parents it initialized. Could both use one shared helper?

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -637,6 +671,8 @@ func (r *Reconciler) hasExistingQuarantine(ctx context.Context, nodeName string)
 	annotations, err := r.getNodeQuarantineAnnotations(ctx, nodeName)
 	if err != nil {
 		slog.ErrorContext(ctx, "Failed to fetch annotations for node", "node", nodeName, "error", err)
+		coldstart.RecordError(ctx, err)
+
 		return make(map[string]string), false
```
KaivalyaMDabhadkar: If this lookup fails for a moment during recovery (it is now a direct api server call, so that can happen), the node is treated as not quarantined and gets quarantined again. In doing so the code sees the node is already cordoned and records that the cordon existed before us, and that flag later blocks the automatic uncordon forever. It is better to return the error here instead of returing "not quarantined"

fault-quarantine/pkg/eventwatcher/event_watcher.go
```diff
@@ -87,7 +110,21 @@ func (w *EventWatcher) Start(ctx context.Context) error {
...
+	}
+
+	if ctx.Err() != nil {
+		w.closeAfterColdStart(ctx)
+
+		return nil //nolint:nilerr // Cancellation is the expected shutdown path.
+	}
+
+	w.armRecoveredEventExpiry(time.Now())
 	go w.updateUnprocessedEventsMetric(ctx)
```
KaivalyaMDabhadkar: this used to start right away, but now it only starts after recovery finishes, so the backlog metric shows nothing during recovery, which is exactly when the backlog is biggest. The fix is just to move this line above the cold start call so the goroutine only reads counts, so it is safe to run while recovery is going

fault-quarantine/pkg/eventwatcher/event_watcher.go
```diff
@@ -178,6 +240,150 @@ func (w *EventWatcher) processEvent(ctx context.Context, event client.Event) err
...
+	return nil
+}
+
+// ProcessStoredEvent sends a durable health-event document through the same
+// callback and status update path used for live change-stream events.
+func (w *EventWatcher) ProcessStoredEvent(
+	ctx context.Context,
+	event datastore.HealthEventWithStatus,
+) (coldstart.ProcessResult, error) {
+	healthEventWithStatus, err := eventutil.ParseHealthEventFromEvent(event.RawEvent)
```
KaivalyaMDabhadkar: By the time an event is fully handled it has been JSON-parsed up to four or five times: once in the skip check, once here, and again for each appearance in a history lookup, and each of those parses is itself two full serialize and deserialize passes. My suggestion: parse the document once in recoverStoredEvent and pass the parsed event (plus its document ID) into both the skip check and ProcessStoredEvent, instead of each of them re-parsing the raw document. That is a small signature change inside the coldstart package and saves a lot of CPU during a large recovery

store-client/pkg/datastore/providers/postgresql/datastore.go
```diff
@@ -399,6 +399,16 @@ func createTables(ctx context.Context, db *sql.DB) error {
 		`CREATE INDEX IF NOT EXISTS idx_health_events_quarantined ON health_events(node_quarantined) ` +
 			`WHERE node_quarantined IS NOT NULL`,
 		`CREATE INDEX IF NOT EXISTS idx_health_events_created_desc ON health_events(created_at DESC)`,
+		`CREATE INDEX IF NOT EXISTS idx_health_events_created_id ON health_events(created_at, id)`,
+		`CREATE INDEX IF NOT EXISTS idx_health_events_recovery_identity ON health_events (` +
```
KaivalyaMDabhadkar: These two indexes are maintained on every health event insert forever, but they are only read during cold start, and the scan's actual filter (the unresolved status check) has no index, so every restart still tests each row in the window one by one. Suggestion: add a WHERE clause to the identity index matching the unresolved condition the cold start query uses (nodequarantined missing, empty, or NotStarted, written with the same COALESCE form the query builder generates so the planner can match it). Then the index serves the scan directly, stays tiny because resolved rows drop out of it, and costs less on every insert

KaivalyaMDabhadkar:
Thanks for putting this together. This closes a real gap, and a lot of the hard parts hold up well: opening the live watcher before recovery starts, replaying stored events through the same code path as live events, the postgres pagination, and the CREATE cutoff handling all look solid.

I have three asks before this merges.

1. The logic that decides which old events to skip needs a change. Right now it can skip too much: when a newer event only partially covers an older one, the older event is dropped entirely, so faults that never recovered are silently ignored and the node is never cordoned. It can also skip too little: an old healthy event can sneak past a newer failure and, when replayed, uncordon a node that is still broken. Both problems live in the same loop. The rule we want is simple to state: during recovery, only act on the latest known state for each check and entity, and skip an old event only when everything it would do is already outdated. Details in the line comments.

2. Whether a replayed event counts as "failed, retry next time" or "intentionally skipped, never look again" depends on many scattered coldstart.RecordError calls all remembering to fire. If one is missed, an event is written off forever. Having the processing callback return an error would let the compiler enforce this, and would actually remove code overall.

3. The very first startup after upgrading needs guardrails. There is no cutoff recorded yet, so recovery scans the entire stored history, every node lookup goes straight to the api server through a rate-limited client, and a single temporary error crashes the pod and starts the whole thing over. Details in the line comments.

The overall structure is right, It is the skip logic and the error handling contract I would like reworked

Saibernard:
@KaivalyaMDabhadkar Thank you for the detailed review. I addressed all review comments in ac0b8e4a and integrated the current main branch, including the processing-error callback contract and the merged checkpoint-ordering behavior. The affected modules pass local race tests, vet, tidy checks, lint, schema validation, and amd64/arm64 image builds. Could you please take another look when you have a chance?

fault-quarantine/pkg/coldstart/coldstart.go
KaivalyaMDabhadkar: @Saibernard The failure handling and batched completions look good, but I think the cutoff seeding now turns recovery off completely for any install that never used the CREATE flow. Nothing persists a cutoff on the normal path (SetColdStartCutoff has no callers), so it is zero on every startup and Handle re-seeds it to time.Now. Since the upper bound was computed earlier in the reconciler, the window becomes "created after a time that is already past the upper bound", which matches nothing, on every restart. So in the main scenario this feature exists for (resume token lost, events missed while we were down), recovery now finds nothing. Please persist the seed as a real watermark, for example through the resume-control store the first time it is seeded or when a sweep completes. A test with a nonzero UntilTime and a lost token would catch this; the current seed test passes a zero UntilTime.

fault-quarantine/pkg/coldstart/supersession.go
KaivalyaMDabhadkar: The rewrite looks right and the new tests cover the main cases. Two small edges in the new matching: a healthy event that carries error codes can mark an uncoded failure as covered, even though replaying it live would not have cleared that failure, so we can miss a cordon when a check mixes coded and uncoded events. And check-wide coverage ignores error codes entirely, so a check-wide clear for code 79 counts as covered by a clear for code 48

fault-quarantine/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: The per-event node cache looks good and closes the rate limiter problem. One small thing: the cache is always seeded by the reconciler's loader, which does not strip node.Status, so custom rules that reference node.status see a different node shape during replay than live (the informer strips status). Stripping Status inside GetRecoveryNode or in this loader would keep the shapes identical

KaivalyaMDabhadkar:
@Saibernard I have re-opened a few resolved comments with a few more observations, could you please address them as well as the code rabbit reviews too?

Saibernard:
done, addressing it now, thanks @KaivalyaMDabhadkar

store-client/pkg/query/update.go
```diff
@@ -95,6 +96,70 @@ func (u *UpdateBuilder) ToMongo() map[string]any {
...
+// parent field is missing or explicitly null.
+func (u *UpdateBuilder) ToMongoPipeline() []any {
+	if u == nil || len(u.operations) == 0 {
+		return nil
+	}
+
+	pipeline := make([]any, 0, len(u.operations))
+
+	for _, op := range u.operations {
+		setDoc, _ := op.ToMongo()[opSet].(map[string]any)
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: @Saibernard Can you please address this as well?

KaivalyaMDabhadkar:
@Saibernard  I dug into the failing AMD64 e2e jobs and I believe this is a real bug, not a flaky test (the scale test only runs on AMD64, so green ARM64 does not cover it).

What happens: an earlier test sends an event that fault-quarantine correctly ignores under that test's rules. Ignored events get no status written, so they stay "unresolved" in the datastore. When a later test restarts fault-quarantine with the default rules, the cold-start sweep replays that old event under the new rules and cordons the node. The scale test then finds one of its nodes already cordoned, that node never feeds the circuit breaker, and the breaker ends up exactly one event short of tripping, so the test times out. The bigger concern is that every restart re-decides old skipped events under whatever rules are loaded now, so a routine config change plus rollout could cordon nodes for events that were correctly ignored when they arrived.

Suggested fix: advance the persisted watermark from live processing too, so restarts only rescan the recent gap and old skipped events are never re-decided. Marking each skipped event as done would also work but costs a write per event. Smaller follow-ups: count already-cordoned nodes in the circuit breaker, and have the scale test check for clean node state in setup.

Could you also add a regression test for this exact sequence: process an event live under rules that do not match it, then run a fresh cold start under rules that do match it, and assert the node is not cordoned. The existing cold-start tests all replay events fault-quarantine never saw, so this case (saw it, skipped it, rules changed) is currently untested, which is why CI was the first thing to catch it.

Saibernard:
@KaivalyaMDabhadkar Thanks for catching these. I’ve addressed the latest comments in `c0d13f10`, and local validation passed as well. Could you please take another look when you get a chance?

KaivalyaMDabhadkar:
> @KaivalyaMDabhadkar Thanks for catching these. I’ve addressed the latest comments in `c0d13f10`, and local validation passed as well. Could you please take another look when you get a chance?

@Saibernard DCO check is failing because the newest commit (c0d13f10) is missing its Signed-off-by line. Since it is the head commit, a quick `git commit --amend --no-edit --signoff` and a force push will fix it. Can you please push it?

KaivalyaMDabhadkar:
@Saibernard  I looked into the two PostgreSQL e2e failures and I think they both come from a small ordering change in 7474a864.
Every document already stored in Postgres contains a "RawEvent": null entry, because the struct field has no json tag, so it gets written into the document on insert. After the reorder, the decode function fills the raw map first and runs the typed unmarshal second, so that null value quietly resets RawEvent back to nil without any error. The batch reader then writes RawEvent["id"] into it, which panics on a nil map. That makes both fault-quarantine and node-drainer crash on the first row their cold start reads, so they crash-loop on Postgres, which is exactly the two rollout timeouts we see in CI.
The fix should be small: set RawEvent after the typed unmarshal runs (or set it again once the unmarshal succeeds). One thing to watch out for is that adding a json tag to the field will not be enough on its own, because all the existing rows already contain the null. Could you also add a decode test using a document that includes "RawEvent": null? Since every legacy row has that shape, it would keep this from coming back

KaivalyaMDabhadkar:
> @Saibernard I looked into the two PostgreSQL e2e failures and I think they both come from a small ordering change in [7474a86](https://github.com/NVIDIA/NVSentinel/commit/7474a8649e0918b379a2f02673c4b4501b2235bb). Every document already stored in Postgres contains a "RawEvent": null entry, because the struct field has no json tag, so it gets written into the document on insert. After the reorder, the decode function fills the raw map first and runs the typed unmarshal second, so that null value quietly resets RawEvent back to nil without any error. The batch reader then writes RawEvent["id"] into it, which panics on a nil map. That makes both fault-quarantine and node-drainer crash on the first row their cold start reads, so they crash-loop on Postgres, which is exactly the two rollout timeouts we see in CI. The fix should be small: set RawEvent after the typed unmarshal runs (or set it again once the unmarshal succeeds). One thing to watch out for is that adding a json tag to the field will not be enough on its own, because all the existing rows already contain the null. Could you also add a decode test using a document that includes "RawEvent": null? Since every legacy row has that shape, it would keep this from coming back

@Saibernard Friendly reminder regarding the above comment

Saibernard:
@KaivalyaMDabhadkar yup working on it, validating the entire pipeline now, will push by today and hopefully this fix should fix everything :)

Saibernard:
@KaivalyaMDabhadkar can you give an ok to test?

KaivalyaMDabhadkar (approved):

PR #1694, opened 2026-08-28
feat: validation-controller implementation

natherz97:
## Summary
This PR adds the core reconciliation logic to the validation-controller. See the corresponding ADR for more information: https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/049-node-validation.md. For reviewers, please start with the primary validationrequest_controller.go file which has comments outlining the high-level implementation. 

#### How reconciling is triggered

Outside of controller start-up or a SyncPeriod, we expect the Reconcile function to be triggered by edge-based signals for ValidationRequest, nodes, or test provider resources. This is configured above in SetupWithManager:
- ValidationRequest CREATE, UPDATE, and DELETE events will fire the reconciler.
- Node CREATE, UPDATE, and DELETE events where the node has the validation-session annotation will fire the reconciler for each ValidationRequest listed in the session annotation. In practice, CREATE events will never fire due to the annotation not existing. DELETE events still fire because the last-known state of the object retains the annotation and this is required to drive reconciling a ValidationRequest after one of its nodes is deleted.
- TestProvider resource CREATE, UPDATE, and DELETE events which have an OwnerReference for a ValidationRequest.

The only place where we rely on a level-based signal to trigger reconciling is to detect test provider timeouts where we specify an explicit RequeueAfter time that is after the configured test provider timeout.

#### Phase transitions for ValidationRequests and TestGroups

ValidationRequest phase transitions:
1. Init -> Pending -> Succeeded/Failed
2. Init -> Pending -> Running -> Succeeded/Failed

TestGroup transitions within a ValidationRequest:
1. Without any retries: Pending -> Running -> Succeeded/Failed
2. With 1 retry: Pending -> Running -> Pending -> Running -> Succeeded/Failed
3. Built directly as Failed with no attempts, if it cannot meet its batch minimum when initial TestGroups are built.
4. Pending -> Failed/Succeeded with no attempts, if a deleted node drops it below its batch minimum before it starts.

#### Full list of validation-controller test cases

- group construction
  - request for one node and one test
  - batches multiple nodes into one group when node batching is true
  - produces one combined group when both node batching and test batching are enabled
  - produces four groups when both node batching and test batching are disabled
  - produces one group per node when test batching is enabled but node batching is disabled
  - produces one group per test when node batching is enabled but test batching is disabled

- pending phase
  - succeeds immediately when all nodes are deleted
  - skips the deleted node and continues with the remaining node
  - stays pending until node readiness passes then transitions to running
  - fails immediately when the batch minimum is not met and BatchFailurePolicyFail is fail
  - skips the test and succeeds when the batch minimum is not met and BatchFailurePolicyIgnore is ignore
  - skips one test and proceeds to running when only one test's batch minimum is met
  - writes the session annotation while blocked on readiness
  - adds the session annotation to every node even when an earlier node isn't ready

- running phase
  - runs groups sequentially when they share a node
  - runs groups sequentially when MaxConcurrentGroups is 1
  - does not start a pending group with a node readiness failure
  - runs groups concurrently when MaxConcurrentGroups is greater than 1 and groups have no node overlap
  - does not start a pending group after another group fails
  - allows a currently running group to finish when another group fails
  - fails a pending group without starting it when its nodes are deleted below the batch minimum with BatchFailurePolicyFail is set to fail
  - succeeds a pending group without starting it when a deleted node violates the batch minimum and BatchFailurePolicyIgnore is set to ignore
  - starts a pending group without the deleted node when the batch minimum is still met
  - does not mutate a VR that is already terminal

- group failures failures
  - marks the request as failed when no retries configured
  - retries and ultimately fails when all retries are exhausted
  - succeeds after one retry when the first attempt fails and the second succeeds
  - deletes the failed attempt's provider resource before starting the retry
  - fails the attempt with TestTimeout
  - treats an externally deleted provider resource as a test failure
  - fails the attempt and retries when a node readiness violation occurs while running and then succeeds after readiness recovers

- node deletion while running
  - fails the attempt with NodeDeleted and retries when a node is deleted but the batch minimum is still met
  - drops one test and retries with the other when a node deletion violates only one test's batch minimum in a shared group
  - fails the whole group when a node deletion violates one test's batch minimum under BatchFailurePolicyFail in a shared group
  - fails the group even with retries configured when a deleted node violates the batch minimum (non-retryable)
  - skips the test and marks the group successful when a deleted node violates the batch minimum and BatchFailurePolicyIgnore is set
  - marks the request as successful when all nodes are deleted during running despite a failed group

- multiple requests
  - runs two requests concurrently when there's no node overlap
  - blocks a second request in pending from node overlap, unblocks when the first request succeeds
  - blocks a second request in pending from node overlap, unblocks when the first request fails
  - supersedes failed request session entry when a new request succeeds with the same tests
  - leaves the failed request session entry when the succeeding request targets different tests
  - leaves the failed request session entry when the succeeding request targets a subset of the failed tests
  - adds concurrent requests to the session annotation

- request deletion
  - releases node annotations and removes the finalizer
  - removes failed session entry when a failed request is deleted

- scheduling gate
  - releases the cordon and removes configured taints after success
  - leaves the cordon and taints after failure
  - leaves taints non-matching taints
  - leaves a taint in place when remove is set to false
  - renders all template context fields into the provider resource
  - does not release the scheduling gate on a node with no session entries for this request

#### Example ValidationRequest

Example ValidationRequest which targets 2 nodes and 3 tests. In this example, the nccl-provider supports test and node batching which results in a single test object which targets both nodes and the dcgm-level4 and nccl-loopback tests. Additionally, the k8s-job-provider does not support node batching which results in 2 separate test objects for each node which runs the cuda-smoke-test
```
apiVersion: nvsentinel.nvidia.com/v1alpha1
kind: ValidationRequest
metadata:
  name: gpu-node-validation
spec:
  nodes:
    - name: node-a
    - name: node-b
  tests:
    - dcgm-level4       
    - nccl-loopback    
    - cuda-smoke-test 
status:
  phase: Succeeded
  testGroups:
    - name: dcgm-nccl-group-1
      provider: nccl-provider
      tests: [dcgm-level4, nccl-loopback]
      nodes: [node-a, node-b]
      phase: Succeeded
      attempts:
        - objectName: gpu-node-validation-dcgm-nccl-group-1-1
          phase: Failed
          failureReason: TestFailed
          startTime: "2026-08-27T10:00:00Z"
          endTime: "2026-08-27T10:15:00Z"
        - objectName: gpu-node-validation-dcgm-nccl-group-1-2   
          phase: Succeeded
          startTime: "2026-08-27T10:15:05Z"
          endTime: "2026-08-27T10:30:00Z"
    - name: cuda-smoke-test-group-1
      provider: k8s-job-provider
      tests: [cuda-smoke-test]
      nodes: [node-a]
      phase: Succeeded
      attempts:
        - objectName: gpu-node-validation-cuda-smoke-test-group-1-1
          phase: Succeeded
          startTime: "2026-08-27T10:00:00Z"
          endTime: "2026-08-27T10:05:00Z"
    - name: cuda-smoke-test-group-2
      provider: k8s-job-provider
      tests: [cuda-smoke-test]
      nodes: [node-b]
      phase: Succeeded
      attempts:
        - objectName: gpu-node-validation-cuda-smoke-test-group-2-1
          phase: Succeeded
          startTime: "2026-08-27T10:00:00Z"
          endTime: "2026-08-27T10:05:00Z"
```

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [X] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [NA] Documentation updated: will include documentation update as part of fault-quarantine integration
- [X] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added lifecycle management for validation requests, including batching, retries, concurrency limits, readiness checks, cleanup, and node-deletion handling.
  - Added configurable test environment variables, commands, tolerations, provider timeouts, and provider resources.
  - Added readiness validation for GPU resources, labels, cordons, and quarantine annotations.
  - Added validation status details for participating nodes and tests.
  - Added network access for lifecycle-manager webhook and metrics endpoints.
  - Added end-to-end validation workflow coverage.

- **Bug Fixes**
  - Improved handling of missing GPU data and annotations.
  - Prevented naming collisions and ensured Kubernetes-compatible generated names.
  - Improved cleanup and status reporting across failed validation attempts.


Makefile
```diff
@@ -859,6 +859,9 @@ cluster-create: ## Create local ctlptl-managed Kind cluster with registry
 	ctlptl apply -f $(CTLPTL_CONFIG_FILE)
 	@echo "Waiting for all nodes to be ready..."
 	@kubectl wait --for=condition=ready nodes --all --timeout=300s
+	@echo "Raising kindnet memory limit"
+	@kubectl -n kube-system patch daemonset kindnet -p '{"spec":{"template":{"spec":{"containers":[{"name":"kindnet-cni","resources":{"limits":{"memory":"250Mi"}}}]}}}}'
```
lalitadithya, line 863: out of curiosity why we do need this now? I doubt this has anything to do with the validation controller, was this something that was always a problem but we just didn't notice?

tilt/Tiltfile
```diff
@@ -259,6 +260,7 @@ else:
 # Pulled out of auto-grouped workload resources so they get applied after
 # CRDs exist (otherwise Tilt fails to apply the whole group).
 preflight_enabled = effective_values.get('global', {}).get('preflight', {}).get('enabled', False)
+lifecycle_manager_enabled = effective_values.get('global', {}).get('lifecycleManager', {}).get('enabled', False)
```
lalitadithya: let's keep this always enabled in CI so that we can catch flakes and issues in CI first. It also gives us a good signal for our default enablement strategy

lifecycle-manager/internal/controller/validationrequest_controller.go
```diff
@@ -82,24 +71,587 @@ func (r *ValidationRequestReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+
+	validationRequest.Status.Skipped = buildSkippedStatus(validationRequest.Status.Skipped, skippedNodes, skippedTests)
+
+	now := metav1.Now()
+	validationRequest.Status.Phase = phase
+	validationRequest.Status.CompletionTime = &now
+
+	if err := r.Status().Update(ctx, validationRequest); err != nil {
+		return ctrl.Result{}, err
```
lalitadithya: can wrap errors so that we can trace where they are coming from? We should at least wrap the lowest set of errors

lifecycle-manager/internal/controller/validationrequest_provider.go
```diff
@@ -0,0 +1,262 @@
...
+	sigsyaml "sigs.k8s.io/yaml"
+
+	"github.com/nvidia/nvsentinel/lifecycle-manager/api/v1alpha1"
+)
+
+type TemplateContext struct {
+	ValidationRequestName string
+	TestGroupName         string
+	Namespace             string
+	Timeout               int64 // seconds
```
lalitadithya: can we just call it 


```suggestion
	TimeoutSeconds               int64 // seconds
```

lifecycle-manager/internal/controller/validationrequest_controller.go
```diff
@@ -82,24 +71,587 @@ func (r *ValidationRequestReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	case v1alpha1.PhaseSucceeded, v1alpha1.PhaseFailed:
+		return ctrl.Result{}, nil
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ValidationRequestReconciler) reconcileInit(ctx context.Context,
+	validationRequest *v1alpha1.ValidationRequest) (ctrl.Result, error) {
+	controllerutil.AddFinalizer(validationRequest, finalizerName)
```
lalitadithya: IIRC this function call should return a bool saying if the finalizer was updated, we should guard the update based on this flag

lifecycle-manager/internal/controller/validationrequest_nodes.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	var entries []sessionEntry
+
+	_ = json.Unmarshal([]byte(sessionAnnotation), &entries)
+
+	return entries
+}
+
+func (r *ValidationRequestReconciler) checkPendingNode(ctx context.Context, ns v1alpha1.NodeSpec,
```
lalitadithya: nit: we might need a better name for this, since the function has a side effect of patching the node annotation which isn't evident from the function name

lifecycle-manager/internal/controller/validationrequest_controller.go
```diff
@@ -82,24 +71,587 @@ func (r *ValidationRequestReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				return ctrl.Result{}, nil
+			}
+
+			existingNodes[ns.Name] = true
+		}
+	}
+
+	// If every node has been deleted, mark the request as successful
+	if len(existingNodes) == 0 {
+		return r.completeValidationRequest(ctx, validationRequest, v1alpha1.PhaseSucceeded, deletedNodes, nil)
```
lalitadithya, line 269: I feel like this should be a terminal error state since we didn't do anything as in the user asked for an impossible request which we cannot fulfill.

lifecycle-manager/internal/controller/validationrequest_groups.go
```diff
@@ -0,0 +1,313 @@
...
+	for _, s := range items {
+		if !remove[s] {
+			result = append(result, s)
+		}
+	}
+
+	return result
+}
+
+func groupName(tests []string, index int) string {
```
lalitadithya, line 295: I think a comment with an example of what the input and the resulting name would look like would be helpful here. I presume this will always result in a valid object name?

Makefile
natherz97: Yes, this looks like it's been a persistent issue where kindnet was continually getting OOMKilled. It was uncovered with lifecycle-manager webhook unavailability during the e2e test around 40% of runs:
```
  === Failed
  === FAIL: . TestValidationController/TestValidationController/ValidationRequest_is_created_and_reaches_Running (10.02s)
      validation_request_test.go:68:
              Error Trace:    /home/runner/_work/NVSentinel/NVSentinel/tests/validation_request_test.go:68
                                          /home/runner/go/pkg/mod/sigs.k8s.io/e2e-framework@v0.7.0/pkg/env/env.go:479
                                          /home/runner/go/pkg/mod/sigs.k8s.io/e2e-framework@v0.7.0/pkg/env/env.go:523
              Error:          Received unexpected error:
                              failed to create ValidationRequest e2e-vr-1787958988153283149: Internal error occurred: failed calling webhook "vvalidationrequest-v1alpha1.kb.io": failed to call webhook: Post "https://lifecycle-manager.nvsentinel.svc:9443/validate-nvsentinel-nvidia-com-v1alpha1-validationrequest?timeout=10s": context deadline exceeded
              Test:           TestValidationController/TestValidationController/ValidationRequest_is_created_and_reaches_Running
              Messages:       ValidationRequest should be created successfully

  === FAIL: . TestValidationController/TestValidationController (10.06s)
      validation_request_test.go:47: Selected node for validation-controller test: nvsentinel-worker
      validation_request_test.go:55: Node nvsentinel-worker cordoned
```
During these times, the backing webhook pods were healthy but the kindnet pods were continually restarting. My best theory is that we didn't run any e2e tests that required webhook connectivity after kindnet starting OOM'ing until the new validation-controller test was added. Have no observed the webhook connectivity issues after the memory limit was increased.

tilt/Tiltfile
natherz97: Sure, it's currently enabled in values-tilt.yaml but let me update the Tiltfile to be consistent with Janitor and always deploy resources.

lifecycle-manager/internal/controller/validationrequest_provider.go
natherz97: I updated this in both the provider resource template and in the ValidationConfiguration. An operator can optionally pass the TimeoutSeconds to their yaml template so I also updated the ValidationConfiguration provider section to accept an int for TimeoutSeconds. This prevents an operator from needing to reason about passing a duration type in the ValidationConfiguration but having to accept a number of seconds in their template.

lifecycle-manager/internal/controller/validationrequest_controller.go
natherz97: Good callout, wrapped all direct K8s errors in validationrequest_controller.go to include context about which call was made. K8s errors from other files are already wrapped with context related to the error so I didn't make changes to other files.

lifecycle-manager/internal/controller/validationrequest_controller.go
natherz97: Nice catch, updated the finalizer update to be conditional:
```
	if controllerutil.AddFinalizer(validationRequest, finalizerName) {
		if err := r.Update(ctx, validationRequest); err != nil {
			return ctrl.Result{}, fmt.Errorf("add finalizer to ValidationRequest %q: %w", validationRequest.Name, err)
		}
	}
```
reconcileInit would've only ever been called if the status wasn't updated to set the phase as pending so there's no need for a similar check prior to updating the status.

lifecycle-manager/internal/controller/validationrequest_nodes.go
natherz97: Agreed, updated to addToSessionAndCheckEligibility to make it clear we always add the current request to the node's validation-session if it exists even if it is not eligible from being targeted from another request or doesn't meet the readiness criteria.

lifecycle-manager/internal/controller/validationrequest_controller.go
natherz97: The webhook ensures that all ValidationRequests target >= 1 node and >= 1 test. After webhook admission, the current behavior we have for deleted nodes is that we skip deleted nodes and try re-running any tests against existing nodes (and if deleted nodes cause us to drop below the batch node minimum for a given test, we'll defer to the batch policy to determine whether we skip the test or fail the test). 

I was being consistent with this approach where we skip deleted nodes and implicitly consider them as successful so if all nodes or all tests were deleted, we'd consider the request as successful. I think it would be difficult to reason about if we considered a test that targeted 99 deleted nodes and 1 existing node as successful but then considered 100 deleted nodes as failed. It doesn't matter too much but this does allow us to accurately alert on the ValidationRequest status without needing to worry about an operator scaling down node pools while they are running validation. Let me know if you disagree.

lifecycle-manager/internal/controller/validationrequest_groups.go
natherz97: Good callout, I included a comment above the attemptObjectName since this is the function which derives the object name using the GroupName, ValidationRequest, and attempt number. This attemptObjectName function ensures that we generate a valid DNS1123 name (specifically that we conform to its max length restriction and that our name does not start or end with a hyphen). Examples:

1. No hash needed:
```
Input: tests: ["dcgm-level4", "nccl-loopback"], vrName: "vr-8f3a2b", attempt 1
groupName  = dcgm-level4-nccl-loopback-group-1
attemptName = vr-8f3a2b-dcgm-level4-nccl-loopback-group-1-1
```

2. Hash required for long test names:
```
Input: tests: ["multi-node-nccl-all-reduce-benchmark", "nemotron4-15b-goodput-check"], vrName: "vr-8f3a2b", attempt 1
groupName  = multi-node-nccl-all-reduce-benchmark-nemotr-4c26a2c1-group-2
attemptName = multi-node-nccl-all-reduce-benchmark-nemotr-4c26a2c1-group-2-1
```

distros/kubernetes/nvsentinel/charts/lifecycle-manager/values.yaml
```diff
@@ -60,9 +60,13 @@ config:
...
     - name: gpu-allocatable
-      expression: 'node.status.allocatable["nvidia.com/gpu"] > 0'
+      expression: >-
+        has(node.status.allocatable) && "nvidia.com/gpu" in node.status.allocatable &&
+        quantity(node.status.allocatable["nvidia.com/gpu"]) > 0
     - name: not-under-quarantine
-      expression: '!("nvsentinel.nvidia.com/quarantineHealthEvent" in node.metadata.annotations)'
+      expression: >-
+        !(has(node.metadata.annotations) &&
+        "nvsentinel.nvidia.com/quarantineHealthEvent" in node.metadata.annotations)
```
XRFXLP: we don't put `nvsentinel.nvidia.com` prefix in the annotation key, right: https://github.com/NVIDIA/NVSentinel/blob/main/fault-quarantine/pkg/common/common.go#L27?

tests/validation_request_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+		nodeName, err := helpers.GetRealNodeName(ctx, client)
+		require.NoError(t, err, "failed to get real node")
+		t.Logf("Selected node for validation-controller test: %s", nodeName)
+
+		node, err := helpers.GetNodeByName(ctx, client, nodeName)
+		require.NoError(t, err, "failed to get node")
+
+		node.Spec.Unschedulable = true
+		err = client.Resources().Update(ctx, node)
```
XRFXLP: Should we do retryOnConflict or something similar to avoid flakes?

lifecycle-manager/internal/controller/validationrequest_scheduling.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	if err := r.Patch(ctx, &node, patch); err != nil {
+		return fmt.Errorf("release scheduling gate on node %q: %w", nodeName, err)
+	}
+
+	return nil
+}
+
+func removeConfiguredTaints(taints []corev1.Taint, configuredTaints []v1alpha1.TaintConfig) ([]corev1.Taint, bool) {
```
XRFXLP: nit: we can use more readable slices.DeleteFunc here:
```golang
func removeConfiguredTaints(taints []corev1.Taint, configuredTaints []v1alpha1.TaintConfig) ([]corev1.Taint, bool) {
      remaining := slices.DeleteFunc(slices.Clone(taints), func(t corev1.Taint) bool {
              return slices.ContainsFunc(configuredTaints, func(cfg v1alpha1.TaintConfig) bool {
                      return cfg.Remove && taintMatches(t, cfg)
              })
      })
      return remaining, len(remaining) != len(taints)
}
```

lifecycle-manager/internal/controller/validationrequest_groups.go
```diff
@@ -0,0 +1,321 @@
...
+		if !existing[n] {
+			skipped.Nodes = append(skipped.Nodes, n)
+			existing[n] = true
+		}
+	}
+
+	return skipped
+}
+
+func appendSkippedTests(skipped *v1alpha1.SkippedStatus, tests []string) *v1alpha1.SkippedStatus {
```
XRFXLP: this looks like a duplicate of `appendSkippedNodes` defined just above, why can't we use the same function for both?

lifecycle-manager/internal/controller/validationrequest_readiness.go
```diff
@@ -0,0 +1,109 @@
...
+
+		if !ok {
+			return c.Name, nil
+		}
+	}
+
+	return "", nil
+}
+
+func evalReadinessCriterion(expr string, nodeMap map[string]any) (bool, error) {
```
XRFXLP: should we cache the compiled CEL program instead of `Parse+Check+Program()` every time?

lifecycle-manager/internal/controller/validationrequest_nodes.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+func parseSessionAnnotation(sessionAnnotation string) []sessionEntry {
+	if len(sessionAnnotation) == 0 {
+		return nil
+	}
+
+	var entries []sessionEntry
+
+	_ = json.Unmarshal([]byte(sessionAnnotation), &entries)
```
XRFXLP: should we at least log the error for unmarshalling?

XRFXLP (changes requested):

lifecycle-manager/internal/controller/validationrequest_controller.go
```diff
@@ -82,24 +71,602 @@ func (r *ValidationRequestReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if !exists {
+			// Deleted nodes will be marked as skipped
+			deletedNodes = append(deletedNodes, ns.Name)
+		} else {
+			// Any node with a NodeReadinessViolation will keep the request in Pending
+			if !ready {
```
XRFXLP: the early return inside the node loop skips session-registration for later nodes, so two VRs could end up running on the same node at once. Can we loop over all nodes first, then return pending if any weren't ready?

lifecycle-manager/internal/controller/validationrequest_controller.go
lalitadithya: ah, that makes sense

lifecycle-manager/internal/controller/validationrequest_provider.go
```diff
@@ -0,0 +1,282 @@
...
+func (r *ValidationRequestReconciler) deleteTestGroupObject(ctx context.Context, group *v1alpha1.TestGroupStatus,
+	objectName string) error {
+	providerCfg := r.Config.Validation.Spec.Providers[group.Provider]
+
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(providerGVK(providerCfg))
+	u.SetName(objectName)
+	u.SetNamespace(r.Namespace)
+
+	if err := r.Delete(ctx, u); client.IgnoreNotFound(err) != nil {
```
KaivalyaMDabhadkar: I think this delete might orphan the test pods, batch/v1 jobs default to OrphanDependents when no propagation policy is set, so a still-running pod (for example after a readiness violation or when a running request is deleted) survives the job deletion and can overlap the next attempt's pod on the same node. Would you consider passing `client.PropagationPolicy(metav1.DeletePropagationBackground)` here?

lifecycle-manager/internal/controller/validationrequest_scheduling.go
```diff
@@ -0,0 +1,90 @@
...
+	patch := client.MergeFrom(node.DeepCopy())
+	changed := false
+
+	if schedulingGate.Cordon.Remove && node.Spec.Unschedulable {
+		node.Spec.Unschedulable = false
+		changed = true
+	}
+
+	remainingTaints, taintsChanged := removeConfiguredTaints(node.Spec.Taints, schedulingGate.Taints)
+	node.Spec.Taints = remainingTaints
```
KaivalyaMDabhadkar: This replaces the whole taints array based on a possibly stale cached node, without a resourceVersion precondition, so I think it could erase a taint that FQ adds at the same moment. It looks like we already use commons/pkg/kubeclient/nodepatch.go (NodeMergePatch) to guard exactly against this in FQ, ND. Would it make sense to reuse it here?

lifecycle-manager/internal/controller/validationrequest_nodes.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+
+		if supersedeTests != nil && e.Failed && sameTests(e.Tests, supersedeTests) {
+			continue
+		}
+
+		remaining = append(remaining, e)
+	}
+
+	sessionEmpty := len(remaining) == 0
```
KaivalyaMDabhadkar: If the node has no session annotation at all, this returns sessionEmpty=true, so deleting an old, already completed request would uncordon a node that may have been cordoned later for unrelated reasons. Should we only uncordon and remove taints when we actually removed our own entry from the session?

lifecycle-manager/internal/controller/validationrequest_controller.go
```diff
@@ -82,24 +71,602 @@ func (r *ValidationRequestReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	// reason is not possible for any TestGroup after the ValidationRequest transitions to running and this reason will
+	// only be reflected during buildInitialTestGroups. As a result, any running or pending TestGroup that encounters a
+	// deleted node will show the NodeDeleted reason regardless of the batch failure policy.
+	switch {
+	case len(deletedNodes) > 0:
+		failedNodes = deletedNodes
+		newGroupPhase = removeDeletedNodesFromTestGroup(validationRequest, currentTestGroup, deletedNodes, r.Config)
+		attemptPhase = v1alpha1.PhaseFailed
+		attemptFailureReason = v1alpha1.FailureReasonNodeDeleted
+	case timedOut:
```
KaivalyaMDabhadkar: Because timedOut is checked before testGroupObjectSucceeded, a job that completed within the deadline can still be recorded as TestTimeout if the reconcile runs after the deadline. Since the RequeueAfter fires right at the deadline, a missed watch event is enough to hit this, no controller downtime needed. Should the success check win when the object already reports the success condition?

lifecycle-manager/internal/controller/validationrequest_controller.go
```diff
@@ -82,24 +71,602 @@ func (r *ValidationRequestReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			Kind:    provider.Kind,
 		})
 
-		b = b.Owns(u)
+		controllerManager = controllerManager.Owns(u)
```
KaivalyaMDabhadkar, line 100: Two small concerns about these watches, 1) they are cluster wide, so with the job provider we end up caching every job in the cluster 2) if a configured provider's CRD is not installed yet, the cache never syncs and the whole pod, including the webhook, crash loops roughly every two minutes. Would it be better scoping the cache to our namespace and handling a missing CRD more gracefully?

distros/kubernetes/nvsentinel/charts/lifecycle-manager/values.yaml
```diff
@@ -99,16 +111,30 @@ config:
         name: {{.ValidationRequestName}}-{{.TestGroupName}}
         namespace: {{.Namespace}}
       spec:
-        activeDeadlineSeconds: {{.Timeout}}
+        activeDeadlineSeconds: {{.TimeoutSeconds}}
```
KaivalyaMDabhadkar, line 114: Should we set backoffLimit: 0 here so the job's default pod retries don't stack on top of the provider level retries?

distros/kubernetes/nvsentinel/charts/lifecycle-manager/values.yaml
natherz97: Great catch, fixed in latest revision.

tests/validation_request_test.go
natherz97: Wrapped the 2 updates to retryOnConflict for cordoning the test node in setup and for uncordoning the node in teardown.

lifecycle-manager/internal/controller/validationrequest_scheduling.go
natherz97: Nice, I was not aware of slices.DeleteFunc. I made this update.

lifecycle-manager/internal/controller/validationrequest_groups.go
natherz97: Thanks for catching this, this was a side-effect of initially implementing the skip behavior for nodes prior to the allowing tests to be skipped. I refactored this into a single helper function which takes a bool to indicate whether the Nodes or Tests field should be set on SkippedStatus.

lifecycle-manager/internal/controller/validationrequest_nodes.go
natherz97: Updated parseSessionAnnotation and marshalSessionAnnotation to both return errors which result in errors returned from reconciler function. The only place where we can't return an error is within the nodeToValidationRequest function which is used to enqueue all ValidationRequests for a given node when the node is modified and in this case we won't enqueue anything and log the error.

lifecycle-manager/internal/controller/validationrequest_controller.go
natherz97: With the current return-early, there's no risk of two VRs running against the same node since we're checking if the node already has the active-validation-request annotation present prior to transitioning a new VR which targets this node from pending to running.

What could happen with the return-early pattern is that only a subset of nodes targeted by the VR are added to the validation-session annotation since we stop processing nodes as soon as we find any node that already has an in-progress VR or if it's failing its readiness criteria. In the worst case, these nodes that miss getting the VR added to their session annotation could be uncordoned even though a pending VR targets it until the blocking node becomes eligible.

I updated this logic to loop over all nodes so that we update the validation-session for all nodes and check after iterating if any node isn't eligible in the current VR.

natherz97: Here's the new logic:
```
	for _, ns := range validationRequest.Spec.Nodes {
		exists, ready, err := r.addToSessionAndCheckEligibility(ctx, ns, validationRequest, criteria, resolvedTests)
		if err != nil {
			return ctrl.Result{}, err
		}

		if !exists {
			deletedNodes = append(deletedNodes, ns.Name)
		} else if !ready {
			allReady = false
		} else {
			existingNodes[ns.Name] = true
		}
	}

	if !allReady {
		return ctrl.Result{}, nil
	}
```

lifecycle-manager/internal/controller/validationrequest_readiness.go
natherz97: Great point, added a readinessPrograms struct member to the controller that caches the compiled CEL program at construction time.

lifecycle-manager/internal/controller/validationrequest_provider.go
natherz97: That is a great catch, I was not aware that it's possible for objects to have a different default garbage collection policy than background cascading deletion (this section in documentation mentions it: https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs). I went with your suggestion and am always passing DeletePropagationBackground.

distros/kubernetes/nvsentinel/charts/lifecycle-manager/values.yaml
natherz97: For our default job provider that sounds good to me. If someone would like to set a custom backoffLimit later, we could require them to either use a custom provider template or we could choose to expose this as a test provider setting.

lifecycle-manager/internal/controller/validationrequest_nodes.go
natherz97: In normal operation, there should not be a case where a successful ValidationRequest does not result in an update to the validation-session of a node that it targets. It is possible that we remove the current VR from the annotation and successfully patch it but then our call to uncordon the node fails. The current implementation ensures that we would still proceed to the retrying the second uncordon request after a re-queue (even if we detect that the request did not need removed from the annotation).

To address the callout you have and keep the invariant that we only allow cordon or taint removal if there was a successful ValidationRequest and that request resulted in an update to the validation-session annotation, I think we can make a single patch request which updates the annotation + updates the cordon/taint if required. This keeps the invariant while also ensuring we can continually retry the patch request.

lifecycle-manager/internal/controller/validationrequest_controller.go
natherz97: Great point, there's no reason for us to throwaway a successful result in favor of a timeout since our RequeueAfter period to catch a TestTimeout has no guarantee it is processed by the controller at that time. Updated the check ordering so that we check for success prior to timeout.

lifecycle-manager/internal/controller/validationrequest_scheduling.go
natherz97: Thanks for sharing those learnings. I updated this function to use a consistent get using r.APIReader.Get and building the patch via kubeclient.NodeMergePatch.

lifecycle-manager/internal/controller/validationrequest_controller.go
natherz97: Good callouts, for 1, I updated overall controller-runtime manager to set the namespace to only target the same namespace where the lifecycle-manager is running which is also the namespace for all provider resources created. 

For 2, if someone starts the validation-controller with a provider that targets a CRD which isn't available, I think it is preferable to crash the controller like we do for other invalid ValidationConfiguration settings. We could explicitly check for this existence at controller start-up when we read the configuration for a better error message, but I do think it is preferable to crash the controller rather than allow continuing to run with a subset of the available providers since an operator may not be aware. Let me know if you disagree but I'll plan to leave the current behavior and we can re-evaluate that when we add the first CRD to the test provider since currently we're only using built-in types.

distros/kubernetes/nvsentinel/charts/lifecycle-manager/values.yaml
XRFXLP: Looks like you forgot to push the commit?

lifecycle-manager/api/v1alpha1/validationconfiguration_types.go
```diff
@@ -142,6 +142,16 @@ type TaintConfig struct {
 	Remove bool `json:"remove,omitempty"`
 }
 
+// EnvVarConfig defines a single environment variable for a k8s-job-provider test's container.
+type EnvVarConfig struct {
```
XRFXLP, line 146: Does it make sense to add ValueFrom as well? Like https://pkg.go.dev/k8s.io/api/core/v1#EnvVar?

In general, there are lot fields that could directly inherit from upstream corev1, like
```
Effect string `json:"effect,omitempty"
```
from `corev1.TaintEffect` in upstream, same for ConditionMatch.Status, which mirrors metav1.ConditionStatus.

tests/validation_request_test.go
```diff
@@ -0,0 +1,162 @@
...
+		return ctx
+	})
+
+	feature.Teardown(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		client, err := c.NewClient()
+		require.NoError(t, err, "failed to create kubernetes client")
+
+		nodeName := ctx.Value(keyNodeName).(string)
+
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
XRFXLP, line 135: nit: can we extract this out in to helper for both cordoning and uncordoning? There are lot of occurences of this exact logic in tests/*. Perhaps we can do a claude pass to find out all of them and replace?

XRFXLP (commented):
Only nits, nothing blocking. Can do the cleanup in seperate PR

lifecycle-manager/api/v1alpha1/validationconfiguration_types.go
natherz97: Good callout, I will address in a follow-up PR.

tests/validation_request_test.go
natherz97: Sounds good, will address in a follow-up.

XRFXLP (approved):

PR #1700, opened 2026-08-30
feat(janitor-provider): write attribution to syslog before generic reboot

### LINKED ISSUE #1688: [Feature]: generic provider reboot leaves no attribution on the node; write the reason to syslog before rebooting
### Prerequisites

- [x] I searched existing issues. #1402 and #1418 added the sysrq reboot option, and #964 added the generic reboot ability, but neither covers recording *why* the reboot happened on the node itself. The syslog-health-monitor issues (#1417, #1532 and similar) concern reading the journal, not writing to it.

### Code of Conduct

### Feature Summary

When the generic bare-metal provider reboots a node, nothing on the node records that NVSentinel did it or why. `last reboot` shows a reboot with no attribution, so anyone triaging from the host has no link back to the health event that caused it.

Request: have the reboot Job write a short attribution line to the node's syslog before rebooting.

### Problem/Use Case

The generic provider runs a Job on the target node with either:

```
chroot /host reboot                        # default
echo b > /host-proc/sysrq-trigger          # useSysrqReboot: true
```

The default path produces a normal systemd shutdown, so wtmp and the journal record *that* a reboot occurred, which is useful. Neither path records *why*, and the sysrq path leaves nothing at all because it is an immediate kernel reset.

The consequence is a broken triage path. Someone looking at a node after the fact, often not the person who enabled NVSentinel, sees an unexplained reboot. The explanation exists, but only in the cluster: the `RebootNode` CR, janitor's logs, `janitor_actions_count`, and the health event's `faultremediated` and `lastremediationtimestamp` fields. On bare metal, where nodes are long-lived and frequently examined directly by hardware and datacenter teams who may have no cluster access, that separation is awkward. It also makes correlating against vendor logs or a BMC event log harder than it needs to be.

There is already a precedent for solving this the other way round. `gpu-reset/gpu_reset.sh` writes its outcome to the node's syslog:

```sh
logger -t nvsentinel-gpu-reset -p daemon.err "..."
```

So the pattern of "the thing that acts on the node also annotates the node" is established in this codebase. The reboot path is the one place where the most disruptive action leaves the least local evidence.

### Suggested approach

Prepend a `logger` call to the reboot Job's command, before the reboot, carrying enough to correlate:

```sh
chroot /host logger -t nvsentinel-reboot -p daemon.notice \
  "NVSentinel reboot: node=$NODE_NAME cr=$CR_NAME event=$HEALTH_EVENT_ID action=$ACTION code=$ERROR_CODE"
chroot /host reboot
```

Notes on the details, which are yours to decide:

- The same line matters *more* on the sysrq path, since that leaves no other trace. It would need to be written and flushed before triggering the reset, and `logger` is synchronous to the journal socket so ordering should hold, though on the sysrq path there is a real risk the journal has not yet been persisted to disk when the kernel resets. A short `sync` beforehand, or accepting best-effort on that path, would both be reasonable.
- Worth making it opt-out rather than mandatory, for operators who do not want NVSentinel writing to host logs.
- The fields available depend on what the provider is given today; if the health event ID is not currently plumbed through to `SendRebootSignal`, node name plus CR name alone would still be a large improvement over nothing.

We are adopting NVSentinel on a bare-metal GB200 cluster and expect to use the generic provider, so this would materially improve our post-incident triage. Not blocking, unlike the detection-side gaps we have raised separately.

Nikhil-cp1905:
<html><head></head><body><h2>Summary</h2>
<p>Fixes #1688</p>
<p>When the generic bare-metal reboot provider reboots a node (via default <code>chroot /host reboot</code> or <code>echo b &gt; /proc/sysrq-trigger</code>), no attribution was recorded on the target node explaining that NVSentinel initiated the reboot. This created triage gaps for hardware, host, and datacenter operators inspecting post-reboot logs locally without Kubernetes cluster access.</p>
<p>This PR adds <strong>optional syslog attribution</strong> to the reboot Job executed on the target node:</p>
<ul>
<li>Invokes <code>chroot /host logger -t nvsentinel-reboot -p daemon.notice "NVSentinel reboot: node=&lt;nodeName&gt;" || true</code> prior to executing the reboot command.</li>
<li>For the SysRq reboot path (<code>useSysrqReboot: true</code>), mounts <code>/</code> at <code>/host</code> when <code>writeSyslog</code> is enabled, logs attribution via <code>logger</code>, and invokes <code>sync || true</code> before writing to <code>/host-proc/sysrq-trigger</code> to flush dirty disk buffers.</li>
<li>For the SysRq path without syslog (<code>useSysrqReboot: true</code>, <code>writeSyslog: false</code>), ensures <code>sync || true</code> is executed before writing to <code>/host-proc/sysrq-trigger</code>.</li>
<li>Appends <code>|| true</code> to prevent logging failures (such as a missing <code>logger</code> binary or unresponsive syslog daemon) from blocking node remediation.</li>
<li>Adds an opt-in configuration flag <code>writeSyslog</code> (default: <code>false</code>, env <code>GENERIC_REBOOT_WRITE_SYSLOG</code>) in Helm values and provider configuration.</li>
<li>Updates documentation accordingly.</li>
</ul>
<hr>
<h2>Type of Change</h2>

Type | Applies
-- | --
🐛 Bug fix | ☐
✨ New feature | ☑
💥 Breaking change | ☐
📚 Documentation | ☑
🔧 Refactoring | ☐
🔨 Build/CI | ☐

<hr>
<h3>2. Local Simulation &amp; Failure-Mode Testing (Without Rebooting Local Machine)</h3>
<p>Since CI cannot reboot nodes, the containerized execution environment and failure handling were verified locally via Docker.</p>
<h4>A. Positive Path — Host Logging Verification</h4>
<p>Simulated the reboot container mounting the host root filesystem at <code>/host</code> and writing to host syslog/journal:</p>
<pre><code class="language-bash">docker run --rm --privileged -v /:/host busybox:1.37 sh -c \
  "chroot /host logger -t nvsentinel-reboot -p daemon.notice \"NVSentinel reboot: node=local-test-node\" || true"
</code></pre>
<p><strong>Verification command:</strong></p>
<pre><code class="language-bash">journalctl -t nvsentinel-reboot -n 1 --no-pager
</code></pre>
<p><strong>Output received:</strong></p>
<pre><code>Aug 31 19:38:00 my-host nvsentinel-reboot: NVSentinel reboot: node=local-test-node
</code></pre>
<h4>B. Negative / Fault Injection Path — Resilience to Missing <code>logger</code></h4>
<p>Simulated an environment where the <code>logger</code> binary does not exist on the target node OS:</p>
<pre><code class="language-bash">docker run --rm --privileged -v /:/host busybox:1.37 sh -c \
  "chroot /host non_existent_logger -t nvsentinel-reboot \"test\" || true; echo 'Reboot would proceed successfully'"
</code></pre>
<p><strong>Output received:</strong></p>
<pre><code>chroot: can't execute 'non_existent_logger': No such file or directory
Reboot would proceed successfully
</code></pre>
<p><strong>Result:</strong> Confirmed that logging failures cleanly fall through <code>|| true</code> and never block the reboot command.</p>
<hr>
<h3>3. Real Cluster End-to-End Testing Procedure</h3>
<p>For staging / bare-metal clusters, the flow is tested through the following steps:</p>
<ol>
<li>
<p><strong>Deploy Helm chart with the feature enabled:</strong></p>
<pre><code class="language-yaml">csp:
  provider: generic
  generic:
    writeSyslog: true
    useSysrqReboot: false  # or true to test SysRq
</code></pre>
</li>
<li>
<p><strong>Cordon the target worker node:</strong></p>
<pre><code class="language-bash">kubectl cordon &lt;target-node-name&gt;
</code></pre>
</li>
<li>
<p><strong>Record the initial node boot ID:</strong></p>
<pre><code class="language-bash">kubectl get node &lt;target-node-name&gt; -o jsonpath='{.status.nodeInfo.bootID}'
</code></pre>
</li>
<li>
<p><strong>Trigger reboot remediation:</strong></p>
<p>Trigger Janitor remediation, or trigger <code>SendRebootSignal</code> against <code>&lt;target-node-name&gt;</code>.</p>
</li>
<li>
<p><strong>Verify Job execution:</strong></p>
<pre><code class="language-bash">kubectl get jobs -n nvsentinel -l app.kubernetes.io/managed-by=nvsentinel-janitor-provider
</code></pre>
<p>Verify the Pod is scheduled specifically onto <code>&lt;target-node-name&gt;</code>, runs privileged with <code>/host</code> mounted, and runs the reboot command.</p>
</li>
<li>
<p><strong>Inspect host journal post-reboot:</strong></p>
<p>Once the node finishes rebooting, SSH into the host and query the journal:</p>
<pre><code class="language-bash">journalctl -t nvsentinel-reboot -n 5 --no-pager
# or: grep nvsentinel-reboot /var/log/syslog
</code></pre>
<p>Verify the attribution message appears with the timestamp right before the system shutdown/restart.</p>
</li>
<li>
<p><strong>Verify node recovery in Kubernetes:</strong></p>
<p>Verify that the node has a new boot ID and returns to <code>Ready</code> status:</p>
<pre><code class="language-bash">kubectl get node &lt;target-node-name&gt; -o jsonpath='{.status.nodeInfo.bootID}'
kubectl get node &lt;target-node-name&gt;
</code></pre>
</li>
</ol>
<hr>
<h2>Checklist</h2>
<ul>
<li>[x] Self-review completed</li>
<li>[x] Documentation updated</li>
<li>[x] Ready for review</li>
</ul></body></html>

## Summary by CodeRabbit

- **New Features**
  - Generic-provider reboot syslog messages can include the originating reboot request name when provided.
  - Syslog attribution remains optional and supports standard and SysRq reboots.

- **Bug Fixes**
  - Syslog attribution is now disabled by default, reducing unexpected node log messages.
  - Invalid syslog configuration safely falls back to disabled behavior.

- **Documentation**
  - Updated configuration defaults and guidance to reflect that syslog attribution is disabled by default.


distros/kubernetes/nvsentinel/charts/janitor-provider/values.yaml
```diff
@@ -128,6 +128,8 @@ csp:
     rebootJobTTLSeconds: 3600
     # Comma-separated image pull secret names for the reboot Job (optional)
     imagePullSecrets: ""
+    # Write a reboot attribution message to the node's syslog before rebooting
+    writeSyslog: true
```
lalitadithya: can we start this feature off as default disabled?

docs/designs/028-generic-baremetal-reboot-provider.md
lalitadithya, line 1: we typically don't update older ADRs since they contain decisions that were made at that point and if we keep updating them then it will no longer reflect the original decision.

lalitadithya:
also, can you please add details to the PR description on how you tested this flow in a real cluster since we don't have the option to reboot nodes in CI?

Nikhil-cp1905:
Hey @lalitadithya I have made the changes for the comments by you and code rabbit , i have a RTX 3050 laptop , so i could test it out locally and check if im receiving a desired output , i have also updated the description for how i tested locally and how it can be done E2E , please let me know if the PR desc should be simplified or changed , and any other code changes are required , ill review it and make the changes accordingly , thankyou very much for reviewing my PR

janitor-provider/pkg/csp/generic/generic.go
```diff
@@ -235,9 +256,26 @@ func (c *Client) buildRebootJob(nodeName string) *batchv1.Job {
...
-		return []string{"sh", "-c", fmt.Sprintf("echo b > %s/sysrq-trigger", hostProcMountPath)}
+		if c.WriteSyslog {
+			return []string{
+				"sh", "-c",
+				fmt.Sprintf("chroot %s logger -t nvsentinel-reboot -p daemon.notice %q || true; sync || true; echo b > %s/sysrq-trigger",
+					hostMountPath, fmt.Sprintf("NVSentinel reboot: node=%s", nodeName), hostProcMountPath),
+			}
+		}
+
+		return []string{"sh", "-c", fmt.Sprintf("sync || true; echo b > %s/sysrq-trigger", hostProcMountPath)}
```
lalitadithya: we don't need the `sync || true` here since we are not writing to the journal

janitor-provider/pkg/csp/generic/generic.go
```diff
@@ -235,9 +256,26 @@ func (c *Client) buildRebootJob(nodeName string) *batchv1.Job {
...
-func (c Config) rebootCommand() []string {
+// rebootCommand generates the container command slice based on sysrq and syslog settings.
+func (c Config) rebootCommand(nodeName string) []string {
 	if c.UseSysrqReboot {
-		return []string{"sh", "-c", fmt.Sprintf("echo b > %s/sysrq-trigger", hostProcMountPath)}
+		if c.WriteSyslog {
+			return []string{
+				"sh", "-c",
+				fmt.Sprintf("chroot %s logger -t nvsentinel-reboot -p daemon.notice %q || true; sync || true; echo b > %s/sysrq-trigger",
+					hostMountPath, fmt.Sprintf("NVSentinel reboot: node=%s", nodeName), hostProcMountPath),
```
lalitadithya: I believe the RFE also mentioned we should include the CR name which is something we need to look into as well. I suspect the janitor provider doesn't have it so we will likely have to plumb that all the way to it to this point

lalitadithya:
ah, thank you for making the changes. Added some final feedback which I missed in the previous pass, please take a look when you have a min

janitor-provider/pkg/csp/generic/generic.go
Nikhil-cp1905: i have made the changes

janitor-provider/pkg/csp/generic/generic.go
Nikhil-cp1905: Hey i have  added cr_name to SendRebootSignalRequest in provider.proto and regenerated proto Go stubs
so plumbed cr_name all the way from Janitor to Janitor Provider
-also Updated janitor/pkg/controller/rebootnode_controller.go to forward rebootNode.Name
-Updated janitor-provider's main.go, model.CSPClient, and all CSP provider method signatures
-Updated the Generic provider so syslog attribution outputs NVSentinel reboot: node=<node> cr=<cr> (with fallback to node=<node> if CR is unset).

Nikhil-cp1905:
Hey lalith , ill fix the lint in some time , i have setup an agent to run these test locally for the desired output everytime there is a code change ,i have defined my agent to follow few acceptance criteria so there is no defects in workflow , im mostly sure these test i mock localy would also be good to analyse how the workflow is if its run E2E , please let me know if there are any changes required

lalitadithya:
> Hey lalith , ill fix the lint in some time , i have setup an agent to run these test locally for the desired output everytime there is a code change ,i have defined my agent to follow few acceptance criteria so there is no defects in workflow , im mostly sure these test i mock localy would also be good to analyse how the workflow is if its run E2E , please let me know if there are any changes required

the changes look good, thank you for working through the reviews! We just need to get the lint and tests to pass

Nikhil-cp1905:
> > Hey lalith , ill fix the lint in some time , i have setup an agent to run these test locally for the desired output everytime there is a code change ,i have defined my agent to follow few acceptance criteria so there is no defects in workflow , im mostly sure these test i mock localy would also be good to analyse how the workflow is if its run E2E , please let me know if there are any changes required
> 
> the changes look good, thank you for working through the reviews! We just need to get the lint and tests to pass

i have made the lint changes

Nikhil-cp1905:
I have fixed few more lint errors

lalitadithya (approved):

Nikhil-cp1905:
Thankyou adithya

PR #1706, opened 2026-08-31
feat(health-events-analyzer): recover derived conditions

### LINKED ISSUE #1553: [Feature]: Support recovery events for health-events-analyzer derived conditions
### Prerequisites

### Code of Conduct

### Feature Summary

  `health-events-analyzer` can generate derived unhealthy events from aggregated
  source events, but it cannot generate the corresponding healthy event when the
  source condition recovers.

  For example, a rule may aggregate repeated `SysLogsXIDError` events into
  `RepeatedXID94OnSameGPU`. After a node reboot, `SysLogsXIDError` becomes healthy,
  but the derived Node Condition and quarantine state remain active and require
  manual cleanup.

  It would be useful for analyzer rules to support a recovery mapping that:

  - Maps a source healthy event to the matching derived healthy event.
  - Recovers by node or entity, such as `GPU_UUID`.
  - Clears or truncates aggregation history at the recovery time.
  - Publishes the derived event with `isHealthy=true`,
    `isFatal=false`, and `recommendedAction=NONE`.
  - Keeps uncordon decisions in `fault-quarantine`, so the node is released only
    when no other active faults remain.

  Related PRs #160 and #248 already document that analyzer-derived conditions
  currently require manual removal because the analyzer does not publish healthy
  events.

### Problem/Use Case

Analyzer rules can create derived unhealthy events, but they cannot create the
  matching healthy event when the source condition recovers.

  For example, repeated `SysLogsXIDError` events may produce
  `RepeatedXID94OnSameGPU`. After a reboot, the source event becomes healthy, but
  the derived Node Condition and quarantine state remain active and require manual
  cleanup.

### Proposed Solution

Add an optional recovery mapping to analyzer rules.

  When a matching source healthy event is received, the analyzer should:

  - Publish a healthy event with the same derived `agent` and `checkName`.
  - Set `isHealthy=true`, `isFatal=false`, and `recommendedAction=NONE`.
  - Recover the matching node or entity, such as a `GPU_UUID`.
  - Clear or truncate aggregation history at the recovery time.
  - Avoid publishing a recovery event when no active derived fault exists.

  `fault-quarantine` should continue deciding whether the node can be uncordoned
  based on all remaining active faults.

### Component

Health Monitor

Saibernard:
## Summary

Closes #1553.

Add opt-in recovery mappings for health-events-analyzer rules so derived conditions can recover automatically when their source condition becomes healthy.

- Recover derived conditions by node or configured entity identity.
- Publish matching healthy transitions without bypassing fault-quarantine final uncordon decision.
- Persist recovery boundaries so pre-recovery history and delayed events cannot immediately recreate a cleared condition.
- Wait for derived fault and recovery transitions to become visible in the store before acknowledging their source event.
- Preserve equivalent recovery filtering and deterministic-failure behavior across MongoDB and PostgreSQL.
- Expose recovery publication, persistence timeout, and stored-document diagnostics.
- Document the design, tradeoffs, failure semantics, and alternatives in [ADR-056](https://github.com/Saibernard/NVSentinel/blob/feat/health-events-analyzer-recovery/docs/designs/056-derived-condition-recovery.md).

The adjacent Kubernetes delivery retry fix requested as a separate change is in #1747.

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected

- [x] Core Services
- [x] Documentation/CI
- [x] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing

- Uncached Go 1.27 race suites pass for `commons`, `store-client`, and `health-events-analyzer` on native macOS ARM64.
- Real MongoDB replica-set and PostgreSQL recovery integration tests pass, including persistence-before-acknowledgement through the production processor configuration.
- Build, vet, `go mod tidy -diff`, and golangci-lint v2.13.1 pass for the affected modules.
- Repository-wide `gomod-lint` and license checks pass. The analyzer no longer imports the MongoDB driver directly.
- Helm unit tests/lint and PostgreSQL schema parity pass.
- Both documentation mappings load with the production config loader. The publisher example was compiled and exercised against a local capture sink.
- The new ordering regression fails when the recovery checkpoint guard is removed through a Go overlay.
- `make lint-test-all` was attempted in an isolated copy and stops during protobuf generation because `protoc` is not installed. Full repository validation and GitHub E2E remain pending.

## Checklist

- [x] Self-review completed
- [x] Ready for review


lfriedman-netllama:
@Saibernard I came at this from #1553 and had a look at what is actually blocking here. It is in better shape than the `CONFLICTING` label suggests, so offering to help if useful.

**The conflicts are small.** A dry-run merge against current `main` gives 4 files, 12 hunks, and 3 of the 4 are test files:

| File | Hunks |
|---|---|
| `health-events-analyzer/pkg/publisher/publisher_test.go` | 2 (add/add) |
| `store-client/pkg/client/event_processor_test.go` | 8 |
| `store-client/pkg/client/postgresql_client_test.go` | 1 |
| `store-client/pkg/datastore/providers/postgresql/datastore.go` | 1 |

The only production conflict is a 6-line one where `main` added `idx_health_events_created_id` and you added `idx_health_events_analyzer_lookup`. They serve different queries, so keeping both is almost certainly the resolution.

I am happy to do that resolution, since the `publisher.go` / `publisher_test.go` and `reconciler.go` side of the conflicts is mostly my own recently merged work (#1711), so I know what the `HEAD` side intends. Either add me as a collaborator on your fork, or say the word and I will post a patch you can apply.

**Two other things standing in the way, neither about the code:** CI has never run on this PR (3 check rows, so it needs an `/ok to test` from a maintainer), and it has no review yet. On the second I can help directly if you want a detailed pass.

For what it is worth, the branch builds cleanly and `pkg/reconciler` tests pass as-is, and the design reads correctly to me: recovery keyed on `IsHealthy && CheckName == mapping.SourceCheckName`, entity-scoped with configurable types, and `EntitiesImpacted` carried through so the Kubernetes connector's existing `removeImpactedEntitiesMessagesScoped` can do the scoped clear.

One thing worth knowing that is outside this PR: the Kubernetes connector drops an event when the node-status write fails, while `store` and `grpcsink` requeue (#1743). Today that is harmless because no healthy derived events exist. Once this lands, every clear depends on one write surviving, so a condition could still latch through no fault of your logic. Not a blocker here, just adjacent.

Saibernard:
@lfriedman-netllama  Thanks for taking the time to investigate this. I had already been working through the branch sync and validation for a while (as you see can see from the draft branch), and your breakdown is consistent with what we’re seeing. I am resolving the latest main conflicts now and have also caught a couple of semantic integration issues around derived-event timestamps and recovery persistence.

I was already going to address #1743 in this update as well, since reliable delivery of the healthy transition is necessary to prevent derived conditions from remaining latched. I am keeping that fix narrowly scoped and validating the complete path on both ARM64 and AMD64 before updating the PR. I’ll post the final details once the full matrix is clean. Thanks again for the careful review and offer to help, but I'll push it today for review :)

store-client/pkg/datastore/providers/postgresql/datastore.go
```diff
@@ -92,6 +93,11 @@ func NewPostgreSQLStore(ctx context.Context, config datastore.DataStoreConfig) (
 		return nil, fmt.Errorf("failed to create tables: %w", err)
 	}
 
+	// Runtime upgrades use statements such as CREATE INDEX CONCURRENTLY that
+	// PostgreSQL forbids inside a transaction. Keep them separate from schema
+	// setup and non-fatal so a performance migration cannot prevent startup.
+	runRuntimeUpgrades(ctx, db)
```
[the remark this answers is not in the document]
Saibernard: Fixed in 5fd246f2. Runtime upgrades now run in a datastore-owned background task, and Close cancels and waits for that task before closing the pool. Interrupted builds remain retryable because the next startup still detects missing or invalid indexes. The new lifecycle test pins non-blocking startup and shutdown cancellation, and I also verified the cancel-and-retry path against a disposable PostgreSQL instance with an old snapshot held open.

store-client/pkg/datastore/providers/postgresql/datastore_test.go
```diff
@@ -254,6 +256,101 @@ func TestRecoveryIndexesIncludePartialPendingEventCursor(t *testing.T) {
...
+		WithArgs(index.name).
+		WillReturnRows(sqlmock.NewRows([]string{"indisvalid"}).AddRow(true))
+	// sqlmock has no negative expectations. Queue both destructive statements
+	// and require them to remain unmet so removing the valid-index guard fails
+	// this test instead of being hidden by the helper's best-effort error path.
+	mock.ExpectExec(regexp.QuoteMeta(index.dropStatement)).WillReturnResult(sqlmock.NewResult(0, 0))
+	mock.ExpectExec(regexp.QuoteMeta(index.createStatement)).WillReturnResult(sqlmock.NewResult(0, 0))
+
+	runRuntimeUpgrades(context.Background(), db)
+	require.Error(t, mock.ExpectationsWereMet())
```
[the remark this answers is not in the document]
Saibernard: Fixed in c5476d1c. The test now requires the remaining expectation error to identify ExpectedExec, which proves the validity query was consumed while the destructive statements were not executed.

lalitadithya:
hey @Saibernard , thank you very much for your work on this! I have a couple of small requests:
1. can we keep this PR focused on only the health event recovery and we can fix the retry bug in a different PR as this will allow us to merge the bug and feature seperately 
2. would it be possible to get an ADR written out for this feature first since this is a new feature so that we know what the pros/cons and what alternatives we have considered (if any)

Saibernard:
Thanks @lalitadithya — both requests are addressed. I split the ordered Kubernetes write retry into #1747, so this PR’s final diff is focused on derived health-event recovery and its supporting datastore/schema work. I also added [ADR-053](https://github.com/NVIDIA/NVSentinel/blob/9677f5315c604550e87532a8e82b884bc8b4af49/docs/designs/053-derived-condition-recovery.md), covering the decision, scope, persistence and replay semantics, tradeoffs, failure modes, and alternatives considered.

The branch is synced with current `main`. Local validation passed on Linux ARM64 and AMD64, including the full affected race suites and real PostgreSQL/MongoDB recovery integration tests. Both PR heads have green DCO and are ready for review.

lfriedman-netllama:
@Saibernard heads up on a numbering collision, already resolved on my side so no action needed from you.

Your `docs/designs/053-derived-condition-recovery.md` and my #1738 both landed on 053 within eight minutes of each other. I have moved mine to **054**, so **053 is yours** and you do not need to touch it.

Also noting for the record that #1747 closes #1743, which I filed. Thanks for picking it up, and the "preserving fault/recovery ordering" framing in its description addresses the specific concern I raised there about requeueing reordering a fault and its recovery for the same condition.

Saibernard:
Thanks @lfriedman-netllama for sorting out the ADR numbering and flagging the retry ordering concern. Appreciate it!

docs/designs/053-derived-condition-recovery.md
lalitadithya, line 1: It is still unclear to me how an operator would request for recovery? The doc covers that we are going to clear the condition on a health event, but doesn't specify how the health event will be generated. Can you please provide some info on this?

docs/designs/056-derived-condition-recovery.md
Saibernard: @lalitadithya I’ve clarified where the recovery event comes from. For GPU resets, the existing reset workflow writes its result to syslog. On success, syslog-health-monitor publishes a healthy `SysLogsXIDError` event for the affected node and GPU. The analyzer matches that against the rule’s recovery mapping and publishes the corresponding derived healthy event.

For manual repairs where no monitor emits a suitable healthy event, the operator would need to verify recovery and publish that event through `PlatformConnector.HealthEventOccurredV1`. That publisher is operator-provided; this PR doesn’t add a built-in manual recovery command.

I’ve documented both paths and included the recovery configuration and publishing code so that requirement is explicit.

Saibernard:
@lalitadithya the review feedback and merge conflicts are addressed, and the updates are pushed. The affected local tests and MongoDB/PostgreSQL recovery tests pass. Could you please run `/ok to test` so we can get the full CI results?

docs/designs/056-derived-condition-recovery.md
lalitadithya, line 1: nit: can we format this document consistently with respect to word wrapping?

docs/designs/056-derived-condition-recovery.md
```diff
@@ -0,0 +1,255 @@
...
+from input, so accepting it as a recovery source would create an unreachable and
+misleading configuration.
+
+### How an operator requests recovery
+
+The operator requests recovery by completing a verified recovery workflow that publishes a healthy source event. The analyzer does not provide a new recovery command, REST endpoint, or Kubernetes resource.
+
+1. Configure the affected rule's `[rules.recovery]` mapping before starting the workflow.
+2. Repair the affected node or device and verify its health.
+3. Have the configured monitor or operator automation publish the matching healthy event through `PlatformConnector.HealthEventOccurredV1`.
```
lalitadithya, line 81: if another monitor is publishing the health event, it is still unclear to me as to why we need the health event analyzer in the loop. The other monitor should be able to "impersonate" the health event analyzer using the same agent and everything should work as expected, right?

docs/designs/056-derived-condition-recovery.md
Saibernard: Done, the ADR now uses one paragraph or list item per line throughout. I also moved it to ADR-059 because main now uses 056 for concurrent event processing.

docs/designs/056-derived-condition-recovery.md
Saibernard: Yes, that can clear the downstream condition if it uses the same agent, rule check name, node, entity scope, and processing strategy. The part it does not handle by itself is the history that triggered the rule. Analyzer events are excluded from rule input, so a direct clear does not create a recovery boundary. A later evaluation can still match the old XID history and raise the condition again.

I added this alternative and the tradeoff to ADR-059. The mapping keeps the history boundary and persisted-state/replay handling alongside the rule, while the other monitor reports that the source condition has recovered.

Saibernard:
@lalitadithya  I’ve resolved the merge conflicts and addressed the two open comments. Recovery tests pass with MongoDB and PostgreSQL, including concurrent workers. Could you run `/ok to test` and take another look?

lalitadithya:
thank for taking a look!

I'm not too big of a fan with the two level publishing of the healthy events we have in 

> 1. Configure the affected rule's `[rules.recovery]` mapping before starting the workflow.
> 2. Repair the affected node or device and verify its health.
> 3. Have the configured monitor or operator automation publish the matching healthy event through `PlatformConnector.HealthEventOccurredV1`.
> 4. The platform connector stores the source event. The analyzer matches its rule, node, and entity identity.
> 5. The analyzer publishes the derived healthy event and waits for storage confirmation. Downstream components apply their normal recovery policies.

I'm wondering can we simplify this into a label contract? If the recovery is deemed to be completed, the operator applies a label/annotation to the node and the would be used by the health event analyzer to clear the derived condition? This would be a simpler lift for the operators as well since they don't have to publish to an API either

Saibernard:
@lalitadithya Agreed, the annotation approach is simpler for operators. I've added it: after verifying recovery, the operator annotates the node with the verification time, optionally scoped to a GPU. The analyzer publishes the derived healthy event directly and removes the annotation once it is stored. No custom publisher is needed.

The docs now include the configuration and `kubectl annotate` commands. I also resolved the merge conflict and tested recovery, retries, and restart behavior against a real Kubernetes API server and both databases. Could you run `/ok to test` and take another look?

docs/designs/059-derived-condition-recovery.md
```diff
@@ -0,0 +1,182 @@
...
+
+Add opt-in, rule-specific recovery triggers to the health-events-analyzer. A rule selects a node annotation for operator recovery or a trusted healthy source event for automatic recovery, plus node or entity scope. The analyzer publishes a derived healthy transition only when persisted state shows that the same rule and recovery identity are currently unhealthy.
+
+Recovery is not inferred from an aggregation pipeline. Rules without a recovery mapping retain manual-recovery behavior.
+
+## Implementation
+
+### Recovery configuration
+
+An enabled rule may define a `[rules.recovery]` block with:
```
lalitadithya, line 44: can we start with just one of the approaches and then we can iterate on it? So, in this case we can start with just the annotation recovery path and based on the feedback and usage we can add additional options.

docs/designs/059-derived-condition-recovery.md
```diff
@@ -0,0 +1,182 @@
...
+- `scope = "node"` or `scope = "entity"`; and
+- `entity_types` for entity scope.
+
+Configuration validation rejects empty or duplicate values, entity types on a node-scoped mapping, an entity-scoped mapping without entity types, and `source_agent = "health-events-analyzer"`. The analyzer's own output is excluded from input, so accepting it as a recovery source would create an unreachable and misleading configuration.
+
+### How an operator requests recovery
+
+For Kubernetes deployments, configure `annotation_key` on the affected rule's recovery block, alongside its node or entity scope. This mode does not require `source_check_name`, a monitor, or an operator-provided publisher. It is mutually exclusive with source-event matching on the same rule.
+
+After repairing and verifying the node, the operator sets the configured node annotation to the verification time in RFC 3339 format. The analyzer watches nodes with an informer, reads the active derived state, and directly publishes the corresponding derived healthy event through its existing platform-connector connection. No intermediate source health event is published.
```
tanishagoyal2, line 57: I am confused on the working of verification time. Can you add more details on it? like who is adding this timestamp, why do we need it and will it be cleared when healthy event is published?

docs/designs/059-derived-condition-recovery.md
```diff
@@ -0,0 +1,182 @@
...
+- either `annotation_key` for operator requests or `source_check_name` for source-event recovery;
+- optional `source_agent` and `source_error_codes` constraints in source-event mode;
+- `scope = "node"` or `scope = "entity"`; and
+- `entity_types` for entity scope.
+
+Configuration validation rejects empty or duplicate values, entity types on a node-scoped mapping, an entity-scoped mapping without entity types, and `source_agent = "health-events-analyzer"`. The analyzer's own output is excluded from input, so accepting it as a recovery source would create an unreachable and misleading configuration.
+
+### How an operator requests recovery
+
+For Kubernetes deployments, configure `annotation_key` on the affected rule's recovery block, alongside its node or entity scope. This mode does not require `source_check_name`, a monitor, or an operator-provided publisher. It is mutually exclusive with source-event matching on the same rule.
```
tanishagoyal2, line 55: Is there any specific format that user has to follow to apply annotation_key? Can you add example of it in doc?

PR #1712, opened 2026-09-01
fix(health-events-analyzer): exclude XID 45 from RepeatedXIDErrorOnSameGPU

### LINKED ISSUE #1710: [Bug]: RepeatedXIDErrorOnSameGPU escalates XID 45 to CONTACT_SUPPORT, but the shipped Xid catalogue says a non-solo XID 45 should be ignored
### Prerequisites

### Code of Conduct

### Describe the bug

`RepeatedXIDErrorOnSameGPU` escalates repeated **XID 45** to `CONTACT_SUPPORT` with
`isFatal: true`, but the shipped `Xid-Catalog.xlsx` says a non-solo XID 45 should be **ignored**.
On a 288 node GB200 fleet this produced **672 fatal events across 3 nodes in 3 hours**, driven
entirely by normal application churn.

### The contradiction with the shipped catalogue

From `health-monitors/syslog-health-monitor/pkg/common/Xid-Catalog.xlsx`, the XID 45 row:

| Field | Value |
| --- | --- |
| Mnemonic | `ROBUST_CHANNEL_PREEMPTIVE_REMOVAL` |
| Applies to GB200 | YES |
| Resolution Bucket (Immediate Action) | `WORKFLOW_XID_45` |
| Resolution Bucket (Investigatory Action) | **`Solo: RESTART_FM` / `Not Solo: IGNORE (follow other Xid)`** |
| Trigger Conditions | "logged when the user application aborts and the kernel driver tears down the GPU appli..." |

"Repeated 5 times within 24 hours" is by definition **not solo**, so the rule fires in exactly
the case the catalogue says to ignore. The catalogue's own trigger description explains why the
volume is high: XID 45 is emitted when a user application aborts, so a tenant starting and
killing CUDA processes generates it continuously. It is a breadcrumb that follows another XID,
not a fault in itself.

This is also consistent with the rest of the project: the detector maps XID 45 to
`recommendedAction: NONE` with `isFatal: false`, which as of v1.21.0 is a deliberate
reclassification. The analyzer then re-escalates it to `CONTACT_SUPPORT`.

### Measured on a live fleet

Over 3 hours, events written by `health-events-analyzer`:

| checkName | recommendedAction | isFatal | errorCode | events | nodes |
| --- | --- | --- | --- | --- | --- |
| `RepeatedXIDErrorOnSameGPU` | `CONTACT_SUPPORT` (5) | **true** | **45** | **672** | 3 |
| `RepeatedXID31OnSameGPU` | `RUN_DCGMEUD` (26) | true | 31 | 11 | 1 |
| `RepeatedXID31OnDifferentGPU` | `NONE` (0) | false | 31 | 11 | 1 |
| `XIDErrorSoloNoBurst` | `NONE` (0) | false | 31 | 6 | 1 |

The underlying detector events on the worst node in the same window:

```
syslog-health-monitor   XID 45                        n=1232
gpu-health-monitor      DCGM_FR_CLOCK_THROTTLE_POWER  n=45
```

So 1,232 application-abort breadcrumbs on one node became 273 `CONTACT_SUPPORT` escalations.
With fault-quarantine enabled, `isFatal: true` drives quarantine, so those three nodes would be
cordoned and drained continuously while nothing is wrong with them.

### Why the rule matches it

`RepeatedXIDErrorOnSameGPU` is described as "Detect occurrence of **fatal XIDs** 5 times within
24 hours", but the rule body contains **no reference to `isfatal` or `recommendedaction`** at
all. Its only code predicate is an exclusion of XID 31:

```json
{"$ne": [{"$arrayElemAt": ["$healthevent.errorcode", 0]}, "31"]}
```

XID 31 is excluded because `RepeatedXID31OnSameGPU` handles it. The
`["74","79","95","109","119"]` list elsewhere in the rule is the sticky-XID window helper, not an
in-scope allowlist. So every XID except 31 is in scope, including ones the catalogue says to
ignore.

### Possible fixes

I do not want to assert which is right, since it depends on what "fatal XIDs" was intended to
mean. Two candidates:

1. **Exclude XID 45 the way XID 31 already is.** Narrow, immediately matches the catalogue's
   "Not Solo: IGNORE" guidance, and the existing `XIDErrorSoloNoBurst` rule already covers the
   solo case the catalogue treats differently. Lowest risk.
2. **Add an `isfatal: true` predicate to the searched event set**, so the rule matches its own
   description. This would exclude XID 45 as a side effect, since the detector emits it
   non-fatal. Broader, and it would change behaviour for any other non-fatal code currently
   being escalated, which may or may not be wanted.

Option 1 treats the specific miscategorisation; option 2 treats the class. Happy to implement
either, or both, once you have a view.

### A related naming observation

While checking whether option 2 was even possible: `BuildProcessableNonFatalUnhealthyInsertsPipeline`
(`store-client/pkg/client/mongodb_pipeline_builder.go:119`) filters on
`fullDocument.healthevent.ishealthy = false` but **not** on `isfatal`, despite "NonFatal" in the
name. So fatal events do reach the analyzer. Not a bug in itself, but the name implies a filter
that is not there, and it briefly led me to the wrong conclusion about what a fix could look
like. Might be worth a rename or a comment.

### Expected behavior

A repeated XID whose catalogue entry says to ignore it when not solo should not be escalated to
`CONTACT_SUPPORT`, and should not produce `isFatal: true` events that drive quarantine.

### Component

Health Events Analyzer

lfriedman-netllama:
## Summary

Fixes #1710, taking option 1 as agreed: exclude XID 45 from `RepeatedXIDErrorOnSameGPU` the
same way XID 31 already is, rather than adding a broad `isfatal` predicate.

## Why

The rule is described as detecting fatal XIDs repeating on a GPU and carries
`CONTACT_SUPPORT`. Its only code exclusion was XID 31, so XID 45 was in scope and repeats were
escalated to `CONTACT_SUPPORT` with `isFatal: true`.

The shipped `Xid-Catalog.xlsx` resolves XID 45 (`ROBUST_CHANNEL_PREEMPTIVE_REMOVAL`) as:

> **Solo: RESTART_FM / Not Solo: IGNORE (follow other Xid)**

with trigger condition "logged when the user application aborts and the kernel driver tears
down the GPU appli...". A repeat is by definition **not solo**, so the rule fired exactly where
the catalogue says to ignore. It is also inconsistent with the detector, which maps XID 45 to
`NONE` with `isFatal: false`.

Measured on a 288 node GB200 fleet: **672** `CONTACT_SUPPORT` events with `isFatal: true`
across 3 nodes in 3 hours, from 1,232 XID 45 detector events on the worst node. Over 2 days the
collection holds **168,924** XID 45 events against **10** of every other XID combined. With
fault-quarantine enabled those nodes would be cordoned and drained continuously.

## Change

One predicate, in both files that define the rule:

```diff
-{"$ne": [{"$arrayElemAt": ["$healthevent.errorcode", 0]}, "31"]},
+{"$not": [{"$in": [{"$arrayElemAt": ["$healthevent.errorcode", 0]}, ["31", "45"]]}]},
```

A `$not`/`$in` list rather than a second `$ne`, so the next per-XID decision is one list entry,
which is how you framed it on the issue. `values-tilt.yaml` gets the same change so Tilt and
e2e behaviour match production.

## Verification

- **The predicate was run against a live MongoDB**, not just reasoned about. Same match stage,
  2 days of real data: the control returns `45`, `44`, `145.RLW_SRC_TRACK`,
  `149.NETIR_BER_EVENT`; with the predicate, `45` is dropped and the rest retained. Removed
  exactly `45`. No aggregation error, so the `$not`/`$in` form is valid on the server.
- YAML parses and the edited stage is valid JSON in both files.
- Confirmed these are the only two files defining the rule and no `$ne ... "31"` copy remains.
- Confirmed `extractXidDetectorConfig` keys off a rule named `RepeatedXidError`, which does not
  exist in the shipped config, so the PostgreSQL burst-detector path is unaffected.
- Confirmed the existing e2e tests never use XID 45, so nothing there changes behaviour. The
  test at `tests/health_events_analyzer_test.go:379` already documents the XID 31 exclusion as
  precedent for this shape.

## On an e2e test

I did not add one. The natural case is sending 5 XID 45 events and asserting
`EnsureNodeConditionNotPresent`, mirroring the XID 31 note above, but `e2e-test-ci` depends on
`tilt-ci` and I have no way to run it here. I would rather not ship an e2e test I could not
execute. Happy to add it if you want it, or if CI will exercise it on this branch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved repeated XID event detection by excluding XID 31 and XID 45 from candidate events.
  - Reduced false positives from preemptive channel cleanup and separately handled events being flagged as repeated errors.
  - Applied consistent filtering across supported deployment configurations.

- **Documentation**
  - Updated configuration guidance to explain how XID 31 and XID 45 are handled.

### DIFF
--- a/distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
+++ b/distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
@@ -115,7 +115,7 @@ config: |
 
   [[rules]]
   name = "RepeatedXIDErrorOnSameGPU"
-  description = "Detect occurrence of XID errors (any XID except 31) on the same GPU 5 times within 24 hours where the burst window is 3 minutes and sticky XIDs window is 3 hours"
+  description = "Detect occurrence of XID errors (any XID except 31 and 45) on the same GPU 5 times within 24 hours where the burst window is 3 minutes and sticky XIDs window is 3 hours"
   recommended_action = "CONTACT_SUPPORT"
   evaluate_rule = {{ .Values.enableRepeatedXIDErrorOnSameGPURule }}
   stage = [
--- a/distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
+++ b/distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
@@ -163,6 +163,10 @@ config: |
     ''',
     # Candidate docs must also be XID events so non-XID unhealthy events on
     # the same node are not merged into GPU burst counts.
+    #
+    # Excluded codes: 31 has its own dedicated rules; 45 is preemptive channel cleanup that
+    # the Xid catalogue resolves as "Not Solo: IGNORE (follow other Xid)", and a repeat is
+    # never solo.
     '''
     {
       "$match": {
@@ -171,7 +175,7 @@ config: |
         "healthevent.nodename": "this.healthevent.nodename",
         "$expr": {
           "$and": [
-            {"$ne": [{"$arrayElemAt": ["$healthevent.errorcode", 0]}, "31"]},
+            {"$not": [{"$in": [{"$arrayElemAt": ["$healthevent.errorcode", 0]}, ["31", "45"]]}]},
             {
               "$gt": [
                 {
--- a/distros/kubernetes/nvsentinel/values-tilt.yaml
+++ b/distros/kubernetes/nvsentinel/values-tilt.yaml
@@ -435,7 +435,7 @@ health-events-analyzer:
 
     [[rules]]
     name = "RepeatedXIDErrorOnSameGPU"
-    description = "Detect occurrence of XID errors (any XID except 31) on the same GPU 2 times within 2 minutes where the burst window is 15 seconds and sticky XIDs window is 30 seconds (test config)"
+    description = "Detect occurrence of XID errors (any XID except 31 and 45) on the same GPU 2 times within 2 minutes where the burst window is 15 seconds and sticky XIDs window is 30 seconds (test config)"
     recommended_action = "CONTACT_SUPPORT"
     evaluate_rule = true
     stage = [
--- a/distros/kubernetes/nvsentinel/values-tilt.yaml
+++ b/distros/kubernetes/nvsentinel/values-tilt.yaml
@@ -483,6 +483,10 @@ health-events-analyzer:
       ''',
       # Candidate docs must also be XID events so non-XID unhealthy events on
       # the same node are not merged into GPU burst counts.
+      #
+      # Excluded codes: 31 has its own dedicated rules; 45 is preemptive channel cleanup that
+      # the Xid catalogue resolves as "Not Solo: IGNORE (follow other Xid)", and a repeat is
+      # never solo.
       '''
       {
         "$match": {
@@ -491,7 +495,7 @@ health-events-analyzer:
           "healthevent.nodename": "this.healthevent.nodename",
           "$expr": {
           "$and": [
-            {"$ne": [{"$arrayElemAt": ["$healthevent.errorcode", 0]}, "31"]},
+            {"$not": [{"$in": [{"$arrayElemAt": ["$healthevent.errorcode", 0]}, ["31", "45"]]}]},
             {
               "$gt": [
                 {
--- a/docs/configuration/health-events-analyzer.md
+++ b/docs/configuration/health-events-analyzer.md
@@ -113,7 +113,7 @@ The following table summarises every flag, the XID or event type it covers, and 
 | Flag | Rule | Recommended Action | Description |
 |------|------|--------------------|-------------|
 | `enableMultipleRemediationsRule` | MultipleRemediations | `CONTACT_SUPPORT` | 5 or more remediations on the same node within 7 days |
-| `enableRepeatedXIDErrorOnSameGPURule` | RepeatedXIDErrorOnSameGPU | `CONTACT_SUPPORT` | Any XID except 31 five or more times within 24 hours on the same GPU (burst window 3 min, sticky XID window 3 h) |
+| `enableRepeatedXIDErrorOnSameGPURule` | RepeatedXIDErrorOnSameGPU | `CONTACT_SUPPORT` | Any XID except 31 and 45 five or more times within 24 hours on the same GPU (burst window 3 min, sticky XID window 3 h) |
 | `enableRepeatedXID31OnSameGPURule` | RepeatedXID31OnSameGPU | `RUN_DCGMEUD` | XID 31 two or more times on the same GPU within 24 hours |
 | `enableRepeatedXID31OnDifferentGPURule` | RepeatedXID31OnDifferentGPU | `NONE` | XID 31 on two or more different GPUs within 24 hours |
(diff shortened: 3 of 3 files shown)


distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
```diff
@@ -171,7 +175,7 @@ config: |
         "healthevent.nodename": "this.healthevent.nodename",
         "$expr": {
           "$and": [
-            {"$ne": [{"$arrayElemAt": ["$healthevent.errorcode", 0]}, "31"]},
+            {"$not": [{"$in": [{"$arrayElemAt": ["$healthevent.errorcode", 0]}, ["31", "45"]]}]},
```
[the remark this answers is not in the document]
lfriedman-netllama: @coderabbitai Valid, fixed in 31ced550. The predicate excluded 31 and 45 while the descriptions still said "any XID except 31".

Updated both rule descriptions:

- `charts/health-events-analyzer/values.yaml:118`
- `values-tilt.yaml:438`

One addition beyond the finding: grepping the stale phrasing across the repo turned up a third occurrence you did not flag, `docs/configuration/health-events-analyzer.md:116`, in the configuration reference table. That is the user-facing one, so it is updated too. No stale "except 31" phrasing remains anywhere.

Minor note for accuracy: the finding anchored at line 178 in each file, which is the predicate line rather than the description. The substance was right, the line reference was off by the length of the stage block.

distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
```diff
@@ -163,6 +163,10 @@ config: |
     ''',
     # Candidate docs must also be XID events so non-XID unhealthy events on
     # the same node are not merged into GPU burst counts.
+    #
+    # Excluded codes: 31 has its own dedicated rules; 45 is preemptive channel cleanup that
+    # the Xid catalogue resolves as "Not Solo: IGNORE (follow other Xid)", and a repeat is
+    # never solo.
```
tanishagoyal2, line 169: I think there is a confusion. Here solo doesn't mean single occurrence of XID 45. Here it means XID 45 should not be followed by any other XID. If any other XID is present then we can perform recovery action for that XID. 
XID 45 detection in RepeatedXIDErrorOnSameGPU seems to be correct as it finds any XID (fatal or non-fatal) repeating over a period of time.

tanishagoyal2 (changes requested):

distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
lalitadithya: correct, but since this is a burst analysis, there will always be some other XID which will follow the 45, right? So, if not other XIDs other than 45 are present it isn't really actionable

tanishagoyal2: Yeah, lets exclude XID 45 from rule.

tanishagoyal2 (approved):

lalitadithya (approved):

PR #1713, opened 2026-09-01
feat(fault-quarantine): add an absolute max-nodes bound to the circuit breaker

### LINKED ISSUE #1686: [Feature]: circuit breaker supports only a percentage; add an absolute max-nodes bound
### Prerequisites

- [x] I searched existing issues. Closest prior art is #1228 and #1482, which settled the circuit breaker *denominator* (GPU nodes rather than all nodes). This request is about the *threshold expression*, which is a separate question.

### Code of Conduct

### Feature Summary

The fault-quarantine circuit breaker can only be expressed as a percentage. On a large cluster the smallest expressible limit is coarser than an operator would want, and the effective node count silently changes as the fleet grows or shrinks.

```go
type CircuitBreaker struct {
    Percentage int    `toml:"percentage"`
    Duration   string `toml:"duration"`
}
```

Request: allow an absolute node count as an alternative or additional bound.

### Problem/Use Case

On a 288-GPU-node cluster, `Percentage` being an `int` means 1% is the floor, which is roughly 3 nodes, and each further step is another 3. An operator who wants "never quarantine more than 5 nodes in a window" cannot express it.

Worse, a percentage is a moving target. A limit set at 5% means 14 nodes today and a different number after the next expansion, silently. For a safety control whose entire job is bounding blast radius, an operator generally wants to reason in absolute terms: "no more than N machines out of service before a human looks".

This matters most during initial adoption, which is exactly when the breaker is load-bearing. We are working toward enabling remediation on this cluster and would want a hard cap of a handful of nodes for the first weeks, independent of fleet size.

`Duration` is already a string, so long windows such as `"24h"` work today. Only the threshold is the limitation.

### Suggested approach

Add an optional absolute bound alongside the existing one:

```go
type CircuitBreaker struct {
    Percentage int    `toml:"percentage"`
    MaxNodes   int    `toml:"maxNodes"`
    Duration   string `toml:"duration"`
}
```

Semantics I would propose, subject to your preference:

- Both optional, but at least one must be set when `enabled: true`; validate at config load.
- When both are set, **whichever trips first wins**. That is the safe reading, and it lets an operator express "5% but never more than 10 nodes".
- `fault_quarantine_breaker_utilization` currently reports one utilisation figure. It would need to report against whichever bound is closer, or gain a label distinguishing the two, so a dashboard can show which constraint is binding.

Estimated as a contained change: the config struct, validation, the breaker evaluation, the utilisation metric, and tests.

Happy to contribute the implementation. Please confirm the semantics you would accept, particularly the precedence rule and whether you would prefer the metric to gain a label or report the binding constraint, before I open a PR.

lfriedman-netllama:
## Summary

Fixes #1686. The circuit breaker threshold could only be expressed as a percentage of GPU
nodes. Adds an optional absolute `maxNodes` bound alongside it.

## Why

On a 288 node cluster, `Percentage` being an `int` means 1% is the floor, roughly 3 nodes, and
each further step is another 3, so "never quarantine more than 5 nodes in a window" is not
expressible. A percentage is also a moving target: 5% is 14 nodes today and a different number
after the next expansion, silently. For a control whose job is bounding blast radius, an
operator generally wants to reason in absolute terms.

## Semantics

- Both bounds optional; `maxNodes: 0` disables the new one, so existing configs are unchanged.
- **When both are set the lower one binds**, so `percentage: 5, maxNodes: 10` reads as "5% but
  never more than 10 nodes", and growing the fleet cannot silently raise the effective limit.
- The initializer now **rejects a config with neither bound set**. Previously that computed a
  threshold of 0 and tripped on the first cordon, which is a silent halt rather than a safety
  control.

## Observability

You asked on the issue whether the utilisation metric should gain a label or report the binding
constraint. I went with neither, to avoid changing an existing metric's meaning:

- `fault_quarantine_breaker_utilization` keeps its current definition, recent cordons over GPU
  nodes, so existing dashboards and alerts are untouched.
- A new gauge `fault_quarantine_breaker_threshold_nodes{bound="percentage|maxNodes"}` reports
  the effective threshold in nodes, labelled by the bound that produced it.

That gives strictly more information than a label on the old metric: utilisation against the
actual limit is now derivable as `fault_quarantine_breaker_utilization * <gpu nodes> /
fault_quarantine_breaker_threshold_nodes`, and the label shows which constraint is active. The
setter resets before setting so a change of binding bound leaves no stale series. Happy to
change this if you would rather it were a label on the existing metric.

## Documentation

Per your note about multi-cluster operators, `values.yaml` states that an absolute bound is per
cluster by nature, so an operator running many clusters of differing size will need to tune it
per cluster or leave it unset and rely on percentage alone.

## Verification

- New table-driven `TestTripThreshold` covers 8 cases: percentage-only (unchanged behaviour),
  the ceiling rounding, maxNodes-only, both-set in each direction, both-equal, the
  fleet-growth case that motivated the issue, and neither-set returning no threshold so the
  reason for the initializer guard stays visible.
- Full `fault-quarantine` module test suite passes, including the existing breaker,
  reconciler, evaluator and informer suites.
- `golangci-lint` v2.13.1 with the repo config: **0 issues**.
- Helm renders correctly through the umbrella chart, both at the default and with an override:

  ```
  [circuitBreaker]
  percentage = 50
  maxNodes = 10
  duration = "5m"
  ```

## Unrelated tooling note

`make install-golangci-lint` fails on macOS arm64 with a checksum error. It is not a corrupt
download. The Makefile pipes `install.sh` from golangci-lint `master`, and that script's
checksum lookup matches `golangci-lint-<v>-darwin-arm64.tar.gz.sbom.json` as well as
`...tar.gz`, then compares the SBOM's hash against the tarball. Verifying the tarball against
the release's own `checksums.txt` with an exact filename match passes. Mentioning it in case it
bites others; happy to raise it separately if useful.

## Summary by CodeRabbit

* **New Features**
  * Added an optional absolute node limit for the fault-quarantine circuit breaker.
  * When percentage and node limits are configured, the lower applicable threshold is used, with percentage values rounded up.
  * Added metrics reporting the effective threshold and the limit that determined it.

* **Bug Fixes**
  * Invalid threshold configurations are now rejected, and thresholds are capped at the fleet size.

* **Documentation**
  * Updated configuration and metrics documentation with new options, behavior, validation rules, and examples.

### DIFF
--- a/distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
+++ b/distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
@@ -65,6 +65,15 @@ circuitBreaker:
   # before the circuit breaker trips.
   # Example: 50 means if 50% or more GPU nodes are cordoned, no new quarantines will be allowed
   percentage: 50
+  # Optional absolute cap on cordoned nodes within the window, independent of fleet size.
+  # 0 disables it. When both are set the lower bound binds, so "percentage: 5, maxNodes: 10"
+  # means 5% but never more than 10 nodes.
+  #
+  # Prefer this when the blast radius you care about is a number of machines rather than a
+  # share of the fleet, since a percentage silently changes meaning as the fleet grows. Note
+  # it is per cluster by nature: an operator running many clusters of differing size will
+  # need to tune it per cluster, or leave it 0 and rely on percentage alone.
+  maxNodes: 0
   # Time window for the sliding window circuit breaker
   # The circuit breaker counts cordon events within this time window
   # Example: "5m" means if 50% of GPU nodes are cordoned within any 5-minute window, the circuit breaker trips
--- a/docs/configuration/fault-quarantine.md
+++ b/docs/configuration/fault-quarantine.md
@@ -101,6 +101,7 @@ fault-quarantine:
   circuitBreaker:
     enabled: true
     percentage: 50
+    maxNodes: 0
     duration: "5m"
 ```
 
@@ -110,10 +111,21 @@ fault-quarantine:
 Enables or disables circuit breaker protection. When disabled, unlimited nodes can be quarantined.
 
 #### percentage
-Maximum percentage of total cluster nodes that can be quarantined within the time window. When exceeded, the circuit breaker trips and blocks all new quarantine actions.
+Maximum percentage of GPU nodes that can be quarantined within the time window. When reached, the circuit breaker trips and blocks all new quarantine actions. Rounds up, so `1` on a 288-node cluster is 3 nodes, not 2. Set to `0` to bound by `maxNodes` alone.
+
+#### maxNodes
+Maximum absolute number of nodes that can be quarantined within the time window, independent of cluster size. Defaults to `0`, which disables this bound and preserves the percentage-only behaviour.
+
+When both bounds are set, **the lower one binds**. This is the point of the setting: a percentage silently tracks fleet growth, so a limit chosen for a 100 node cluster becomes three times as permissive at 300 nodes, while an absolute bound does not move. Fleet-wide operators can therefore express "no more than X% *and* never more than N nodes" in one config that stays correct as clusters grow.
+
+At least one of `percentage` and `maxNodes` must be positive, and a negative value for either is rejected rather than ignored. The breaker refuses to start otherwise, so it cannot be silently reduced to a no-op while `enabled: true`. For the same reason, a threshold above the fleet size is clamped to the fleet size rather than being left unreachable, and the `bound` label described below reports `fleetSize` when that happens. To turn the breaker off, use `enabled: false` rather than an unreachable threshold.
 
 #### duration
 Time window for tracking cordon events. The circuit breaker counts unique node cordons within this sliding window.
+
+### Observability
+
+`fault_quarantine_breaker_threshold_nodes{bound}` reports the effective threshold in nodes, with `bound` naming what produced it: `percentage`, `maxNodes`, or `fleetSize` when a configured bound exceeded the cluster size and was clamped. `fault_quarantine_breaker_utilization` keeps its existing meaning, the recent cordon count as a fraction of GPU nodes, so existing dashboards and alerts are unaffected.
 
 ### Configuration Examples
 
@@ -131,6 +143,24 @@ circuitBreaker:
   enabled: true
   percentage: 75
   duration: "3m"
+```
+
+Percentage with an absolute ceiling, for a fleet whose clusters vary in size:
+```yaml
+circuitBreaker:
+  enabled: true
+  percentage: 10
+  maxNodes: 20
+  duration: "5m"
+```
+
+Absolute bound only, for "never cordon more than 2 nodes regardless of cluster size":
+```yaml
+circuitBreaker:
+  enabled: true
+  percentage: 0
+  maxNodes: 2
+  duration: "5m"
 ```
 
 Disabled:
--- a/fault-quarantine/pkg/breaker/breaker.go
+++ b/fault-quarantine/pkg/breaker/breaker.go
@@ -226,8 +226,8 @@ func (b *slidingWindowBreaker) tripThreshold(totalNodes int) (int, string) {
 // It returns true if:
 // 1. The breaker is already in TRIPPED state, OR
 // 2. Recent cordon events reach the configured threshold
-// The threshold is the lower of TripPercentage * GPU nodes and TripMaxNodes, whichever
-// bounds are configured. The method automatically trips the breaker if it is exceeded.
+// The threshold is the lower of the configured percentage of GPU nodes and TripMaxNodes,
+// clamped to the fleet size. The breaker trips when the recent-cordon count reaches it.
 func (b *slidingWindowBreaker) IsTripped(ctx context.Context) (bool, error) {
 	b.mu.RLock()
 
--- a/fault-quarantine/pkg/breaker/breaker.go
+++ b/fault-quarantine/pkg/breaker/breaker.go
@@ -45,6 +45,10 @@ var (
 // It prevents cordoning more than a specified percentage of nodes within a time window.
 // The breaker uses a ring buffer with 1-second granularity to track unique cordoned nodes.
 func NewSlidingWindowBreaker(ctx context.Context, cfg Config) (CircuitBreaker, error) {
+	if err := cfg.validateBounds(); err != nil {
+		return nil, fmt.Errorf("validate circuit breaker bounds: %w", err)
+	}
+
 	numBuckets := int((cfg.Window + time.Second - 1) / time.Second)
 	b := &slidingWindowBreaker{
 		cfg:          cfg,
@@ -176,11 +180,54 @@ func (b *slidingWindowBreaker) sumBuckets() int {
 	return sum
 }
 
+// Names for the configured bounds, used in logs and the binding-bound metric.
+const (
+	boundPercentage = "percentage"
+	boundMaxNodes   = "maxNodes"
+	// Reported when the configured bound exceeds the fleet size and was clamped to it.
+	boundFleetSize = "fleetSize"
+)
+
+// tripThreshold returns the recent-cordon count that trips the breaker for the given GPU
+// node count, along with the bound that produced it. When both bounds are configured the
+// lower one binds, so growing the fleet cannot silently raise the effective limit.
+func (b *slidingWindowBreaker) tripThreshold(totalNodes int) (int, string) {
+	threshold, bound := 0, ""
+
+	if b.cfg.TripPercentage > 0 {
+		// Compared as a float before converting: a percentage large enough to exceed
+		// math.MaxInt makes the int conversion implementation-defined per the Go spec, and
+		// a negative result would slip past the clamp below and trip on every evaluation.
+		scaled := math.Ceil(float64(totalNodes) * b.cfg.TripPercentage / 100)
+		if scaled > float64(totalNodes) {
+			threshold, bound = totalNodes, boundFleetSize
+		} else {
+			threshold, bound = int(scaled), boundPercentage
+		}
+	}
+
+	if b.cfg.TripMaxNodes > 0 && (threshold == 0 || b.cfg.TripMaxNodes < threshold) {
+		threshold = b.cfg.TripMaxNodes
+		bound = boundMaxNodes
+	}
+
+	// A threshold above the fleet size can never be reached, which would turn either
+	// bound into an off switch for the breaker. Clamp so a too-large value degrades to
+	// "every node" instead of "never trip"; the bound label reports when this happens.
+	if threshold > totalNodes {
+		threshold = totalNodes
+		bound = boundFleetSize
+	}
+
+	return threshold, bound
+}
+
 // IsTripped checks if the circuit breaker should prevent further node cordoning.
 // It returns true if:
 // 1. The breaker is already in TRIPPED state, OR
-// 2. Recent cordon events reach the configured threshold (TripPercentage * GPU nodes)
-// The method automatically trips the breaker if the threshold is exceeded.
+// 2. Recent cordon events reach the configured threshold
+// The threshold is the lower of TripPercentage * GPU nodes and TripMaxNodes, whichever
+// bounds are configured. The method automatically trips the breaker if it is exceeded.
 func (b *slidingWindowBreaker) IsTripped(ctx context.Context) (bool, error) {
 	b.mu.RLock()
 
@@ -210,7 +257,7 @@ func (b *slidingWindowBreaker) IsTripped(ctx context.Context) (bool, error) {
 
 	b.slideWindow(now)
 	recentCordonedNodes := b.sumBuckets()
-	threshold := int(math.Ceil(float64(totalNodes) * b.cfg.TripPercentage / 100))
+	threshold, bindingBound := b.tripThreshold(totalNodes)
 	shouldTrip := recentCordonedNodes >= threshold
 
 	b.mu.Unlock()
@@ -218,9 +265,13 @@ func (b *slidingWindowBreaker) IsTripped(ctx context.Context) (bool, error) {
 	slog.DebugContext(ctx, "Recent cordoned nodes status",
 		"recentCordonedNodes", recentCordonedNodes,
 		"totalNodes", totalNodes,
-		"tripPercentage", b.cfg.TripPercentage)
+		"tripPercentage", b.cfg.TripPercentage,
+		"tripMaxNodes", b.cfg.TripMaxNodes,
+		"threshold", threshold,
+		"bindingBound", bindingBound)
 
 	metrics.SetFaultQuarantineBreakerUtilization(float64(recentCordonedNodes) / float64(totalNodes))
+	metrics.SetFaultQuarantineBreakerThresholdNodes(float64(threshold), bindingBound)
 
 	if shouldTrip {
 		err := b.ForceState(ctx, StateTripped)
--- a/fault-quarantine/pkg/breaker/breaker.go
+++ b/fault-quarantine/pkg/breaker/breaker.go
@@ -180,19 +180,25 @@ func (b *slidingWindowBreaker) sumBuckets() int {
 	return sum
 }
 
-// Names for the configured bounds, used in logs and the binding-bound metric.
+// Bound names the configured limit that produced the effective threshold. It appears in
+// logs and as the label on fault_quarantine_breaker_threshold_nodes.
+type Bound string
+
 const (
-	boundPercentage = "percentage"
-	boundMaxNodes   = "maxNodes"
-	// Reported when the configured bound exceeds the fleet size and was clamped to it.
-	boundFleetSize = "fleetSize"
+	// boundNone is the zero value, reported when no bound is configured.
+	boundNone       Bound = ""
+	boundPercentage Bound = "percentage"
+	boundMaxNodes   Bound = "maxNodes"
+	// boundFleetSize is reported when a configured bound exceeded the fleet size and was
+	// clamped to it.
+	boundFleetSize Bound = "fleetSize"
 )
 
 // tripThreshold returns the recent-cordon count that trips the breaker for the given GPU
 // node count, along with the bound that produced it. When both bounds are configured the
 // lower one binds, so growing the fleet cannot silently raise the effective limit.
-func (b *slidingWindowBreaker) tripThreshold(totalNodes int) (int, string) {
-	threshold, bound := 0, ""
+func (b *slidingWindowBreaker) tripThreshold(totalNodes int) (int, Bound) {
+	threshold, bound := 0, boundNone
 
 	if b.cfg.TripPercentage > 0 {
 		// Compared as a float before converting: a percentage large enough to exceed
@@ -271,7 +277,7 @@ func (b *slidingWindowBreaker) IsTripped(ctx context.Context) (bool, error) {
 		"bindingBound", bindingBound)
 
 	metrics.SetFaultQuarantineBreakerUtilization(float64(recentCordonedNodes) / float64(totalNodes))
-	metrics.SetFaultQuarantineBreakerThresholdNodes(float64(threshold), bindingBound)
+	metrics.SetFaultQuarantineBreakerThresholdNodes(float64(threshold), string(bindingBound))
 
 	if shouldTrip {
 		err := b.ForceState(ctx, StateTripped)
--- a/fault-quarantine/pkg/initializer/init.go
+++ b/fault-quarantine/pkg/initializer/init.go
@@ -184,15 +184,20 @@ func initializeCircuitBreaker(
 		return nil, fmt.Errorf("invalid circuit breaker duration %q: %w", cbConfig.Duration, err)
 	}
 
+	// Bounds are validated by NewSlidingWindowBreaker below, so a caller that builds the
+	// config directly cannot bypass the check.
+
 	slog.InfoContext(ctx, "Initializing circuit breaker",
 		"configMap", circuitBreakerName,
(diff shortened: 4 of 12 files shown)


lfriedman-netllama:
Applied the rename in `13bea805` `TestTripThreshold` is now `TestTripThreshold_ConfiguredBounds_ReturnsEffectiveThreshold`, matching the `TestFunctionName_Scenario_ExpectedBehavior` convention in `.github/copilot-instructions.md`. Breaker suite still green.

lalitadithya:
can we also update the documentation in the docs folder please?

XRFXLP:
Why not allow percentage to be a floating point? That should solve the problem that is stated in the description.

lalitadithya:
> Why not allow percentage to be a floating point? That should solve the problem that is stated in the description.

that will become complex if the use case is no more than 2 nodes can be cordoned irrespective of cluster. Based on the use case the operator can either use percentage to say no more X% or number to say no more than X, it makes it simpler for fleet wide operations. I feel it is cleaner this way

fault-quarantine/pkg/breaker/threshold_test.go
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
XRFXLP: we already had breaker_test.go, can we move this test over there?

fault-quarantine/pkg/breaker/threshold_test.go
```diff
@@ -0,0 +1,109 @@
...
+		{
+			name:          "percentage rounds up, so 1 percent of 288 is 3 not 2",
+			percentage:    1,
+			totalNodes:    288,
+			wantThreshold: 3,
+			wantBound:     boundPercentage,
+		},
+		{
+			name:          "maxNodes only, independent of fleet size",
+			maxNodes:      5,
```
XRFXLP: So technically one can make this number higher than totalNodes hence effectively disable circuit breaker in another way?

fault-quarantine/pkg/breaker/threshold_test.go
lfriedman-netllama: Done in `3605ebcb` — `threshold_test.go` is deleted and `TestTripThreshold_ConfiguredBounds_ReturnsEffectiveThreshold` now lives at the end of `breaker_test.go`. Thanks, that is the better home; `breaker_test.go` already has the `TestMain` envtest setup so there was no reason for a second file.

fault-quarantine/pkg/breaker/threshold_test.go
lfriedman-netllama: Good catch, you were right. `shouldTrip := recentCordonedNodes >= threshold` and `recentCordonedNodes` can never exceed `totalNodes`, so `maxNodes: 1000` on a 288 node cluster made the threshold unreachable and the breaker could never trip. Note it was reachable through `percentage` too, since nothing validated it at or below 100.

Worth separating from `percentage: 100`, which is *not* the same thing: that yields `threshold == totalNodes`, which still trips once every node is cordoned. A threshold above the fleet size trips never, so this was genuinely a new off switch rather than just a very permissive setting.

Fixed in `3605ebcb` by clamping in `tripThreshold`:

```go
// A threshold above the fleet size can never be reached, which would turn either
// bound into an off switch for the breaker. Clamp so a too-large value degrades to
// "every node" instead of "never trip"; the bound label reports when this happens.
if threshold > totalNodes {
    threshold = totalNodes
    bound = boundFleetSize
}
```

So a too-large bound now degrades to "every node" rather than "never", and `fault_quarantine_breaker_threshold_nodes{bound="fleetSize"}` makes the clamp visible instead of silent. Safe to clamp at that point because `IsTripped` already returns an error when `totalNodes == 0`, so this cannot produce a `threshold` of 0 that would trip on everything.

Added four cases: `maxNodes` above the fleet size, `percentage` above 100, and the two boundaries that must *not* clamp (`maxNodes == totalNodes` still reports `maxNodes`, `percentage: 100` still reports `percentage`). Also documented that `enabled: false` is the way to disable the breaker.

Let me know if you would rather reject this at config time instead; I went with the clamp because the fleet size is not known at initialization.

lfriedman-netllama:
@lalitadithya docs added in `3605ebcb`, three files:

- **`docs/configuration/fault-quarantine.md`** — a `maxNodes` parameter section covering the lower-bound-binds rule and why an absolute bound exists (a percentage tracks fleet growth, so a limit chosen at 100 nodes is three times as permissive at 300), the at-least-one-positive requirement, plus two new examples: percentage with an absolute ceiling, and absolute-only. Also noted that `percentage` rounds up, since `1` on 288 nodes is 3 rather than 2.
- **`docs/circuit-breaker.md`** — `maxNodes` in the values example and a worked case of the two bounds interacting on different cluster sizes.
- **`docs/METRICS.md`** — `fault_quarantine_breaker_threshold_nodes{bound}`.

I also answered @XRFXLP's floating-point question implicitly in the docs by documenting the two bounds as expressing different intents, matching your reasoning above, rather than presenting `maxNodes` as a workaround for percentage granularity.

Two review items from @XRFXLP are addressed in the same commit: the test moved into `breaker_test.go`, and a real bug they found where a bound above the fleet size made the threshold unreachable and silently disabled the breaker. Details in those threads. Since that last one changed `tripThreshold`'s behaviour, this needs a fresh `/ok to test` when you get a chance.

fault-quarantine/pkg/initializer/init.go
```diff
@@ -184,15 +184,23 @@ func initializeCircuitBreaker(
 		return nil, fmt.Errorf("invalid circuit breaker duration %q: %w", cbConfig.Duration, err)
 	}
 
+	// Without at least one bound the threshold would be 0 and the breaker would trip on the
+	// first cordon, which is a silent halt rather than a safety control.
+	if cbConfig.Percentage <= 0 && cbConfig.MaxNodes <= 0 {
```
KaivalyaMDabhadkar: Could this check move into NewSlidingWindowBreaker? Right now anything that builds the config directly (like the e2e tests) skips this check, and with both values at zero the breaker trips on the very first check, even with no cordons. Also a negative value (e.g maxNodes: -10) gets past this check and is just silently ignored, so it's probably better to reject negatives with an error too

lfriedman-netllama: @KaivalyaMDabhadkar all three points are right, and all three are fixed in `041c3425`.

**1. Moved into the breaker.** Validation is now `Config.validateBounds()` in `types.go`, called at the top of `NewSlidingWindowBreaker` before any Kubernetes work, and the duplicate check is gone from `init.go`. So `reconciler_e2e_test.go:289`, which builds `breaker.Config` directly, is covered now.

**2. The both-zero case really does trip immediately**, which is worth spelling out since it is worse than it sounds: `tripThreshold` returns 0, `IsTripped` compares `recentCordonedNodes >= threshold`, and `0 >= 0` is true, so the breaker trips on its **first evaluation with no cordons at all**. That is now rejected wherever the config is built.

**3. Negatives rejected.** `percentage: -10` or `maxNodes: -10` now error instead of being silently ignored by the `> 0` guards in `tripThreshold`. A negative is rejected even when the other bound is valid, since it is a config error rather than a typo worth quietly dropping.

```go
func (c Config) validateBounds() error {
	if c.TripPercentage < 0 {
		return fmt.Errorf("circuit breaker percentage must not be negative, got %v", c.TripPercentage)
	}

	if c.TripMaxNodes < 0 {
		return fmt.Errorf("circuit breaker maxNodes must not be negative, got %d", c.TripMaxNodes)
	}

	if c.TripPercentage == 0 && c.TripMaxNodes == 0 {
		return errors.New("circuit breaker requires percentage or maxNodes to be set to a positive value")
	}

	return nil
}
```

Covered by `TestValidateBounds_RejectsConfigsThatCannotLimit` (7 cases: the three valid shapes, both-zero, each negative alone, and a negative alongside a valid percentage) plus `TestNewSlidingWindowBreaker_NoBoundsConfigured_ReturnsError` for the constructor path. Full `fault-quarantine` suite green including the 47s reconciler e2e suite, `golangci-lint` 0 issues. Docs updated to say negatives are rejected too.

docs/configuration/fault-quarantine.md
```diff
@@ -110,11 +111,22 @@ fault-quarantine:
 Enables or disables circuit breaker protection. When disabled, unlimited nodes can be quarantined.
 
 #### percentage
-Maximum percentage of total cluster nodes that can be quarantined within the time window. When exceeded, the circuit breaker trips and blocks all new quarantine actions.
+Maximum percentage of total cluster nodes that can be quarantined within the time window. When exceeded, the circuit breaker trips and blocks all new quarantine actions. Rounds up, so `1` on a 288 node cluster is 3 nodes, not 2. Set to `0` to bound by `maxNodes` alone.
```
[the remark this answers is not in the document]
lfriedman-netllama: Both corrections are right and applied. Verified rather than taken on faith:

- **GPU nodes, not cluster nodes.** `informer/node_informer.go:39` defines `GPUNodeLabel = "nvidia.com/gpu.present"`, and the initializer passes the GPU node label into the client that `GetTotalNodes` uses.
- **Reached, not exceeded.** `IsTripped` computes `shouldTrip := recentCordonedNodes >= threshold`.

Both inaccuracies predate this PR, but the line was in my diff so they are fixed here. Also took `288-node`.

fault-quarantine/pkg/breaker/breaker.go
```diff
@@ -176,11 +180,47 @@ func (b *slidingWindowBreaker) sumBuckets() int {
...
+)
+
+// tripThreshold returns the recent-cordon count that trips the breaker for the given GPU
+// node count, along with the bound that produced it. When both bounds are configured the
+// lower one binds, so growing the fleet cannot silently raise the effective limit.
+func (b *slidingWindowBreaker) tripThreshold(totalNodes int) (int, string) {
+	threshold, bound := 0, ""
+
+	if b.cfg.TripPercentage > 0 {
+		threshold = int(math.Ceil(float64(totalNodes) * b.cfg.TripPercentage / 100))
```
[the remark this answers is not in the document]
lfriedman-netllama: Valid, and applied, though the practical severity is lower than Major and worth recording accurately.

I measured it on this platform (arm64, Go 1.27) rather than assuming:

```
pct=150      raw=432        int()=432
pct=1e+18    raw=2.88e+18   int()=2880000000000000000
pct=3.3e+18  raw=9.504e+18  int()=9223372036854775807   <- saturates
pct=1e+30    raw=2.88e+30   int()=9223372036854775807   <- saturates
```

So here the conversion saturates to `math.MaxInt64` and the existing `threshold > totalNodes` clamp already catches it. But you are right that the spec only promises *implementation-dependent*, so an implementation that wrapped to a negative value would slip past that clamp, and `recentCordonedNodes >= negative` is always true, i.e. permanently tripped. Fail-closed rather than fail-open, but still wrong.

Fixed in `c385769e` by doing the comparison in float space, before the conversion:

```go
scaled := math.Ceil(float64(totalNodes) * b.cfg.TripPercentage / 100)
if scaled > float64(totalNodes) {
    threshold, bound = totalNodes, boundFleetSize
} else {
    threshold, bound = int(scaled), boundPercentage
}
```

Now no out-of-range value ever reaches `int()`, so the behaviour is architecture-independent. Regression case added with `percentage: 3.3e18`. Note reachability is limited: `percentage` arrives from Helm as an `int`, so this needs a configured value around 3.3e18, and `validateBounds` already rejects negatives.

fault-quarantine/pkg/breaker/breaker.go
```diff
@@ -45,6 +45,10 @@ var (
 // It prevents cordoning more than a specified percentage of nodes within a time window.
 // The breaker uses a ring buffer with 1-second granularity to track unique cordoned nodes.
 func NewSlidingWindowBreaker(ctx context.Context, cfg Config) (CircuitBreaker, error) {
+	if err := cfg.validateBounds(); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
lfriedman-netllama: Confirmed, and this one is a genuine bug. Measured rather than reasoned about:

```
NaN   passes validateBounds=true   scaled=NaN    clampFires=false  int(scaled)=0
+Inf  passes validateBounds=true   scaled=+Inf   clampFires=true   int(scaled)=9223372036854775807
-Inf  passes validateBounds=false  scaled=-Inf   clampFires=false  int(scaled)=-9223372036854775808
```

So exactly as you describe: NaN passes validation because every comparison against it is false, the fleet-size clamp does not fire because `NaN > 288` is also false, and `int(NaN)` is 0. With `maxNodes: 0` that leaves `recentCordonedNodes >= 0`, i.e. **the breaker is tripped from startup with no cordons at all**, which is the worst possible failure for a safety control.

`+Inf` happens to be saved by the clamp and `-Inf` by the existing negative check, but relying on that is luck rather than design, so both non-finite forms are now rejected outright in `aa1f507e`:

```go
if math.IsNaN(c.TripPercentage) || math.IsInf(c.TripPercentage, 0) {
    return fmt.Errorf("circuit breaker percentage must be a finite number, got %v", c.TripPercentage)
}
```

Test cases added for NaN and +Inf. Reachability is the same caveat as before: `percentage` comes from Helm as an `int` so a config cannot produce NaN, but `Config` is exported and built directly by tests and the e2e suite, which is precisely why @KaivalyaMDabhadkar asked for the validation to live in the constructor. Worth having.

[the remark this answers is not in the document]
lfriedman-netllama: Applied in `aa1f507e`: `return nil, fmt.Errorf("validate circuit breaker bounds: %w", err)`. Matches the wrapping convention used by the other early returns in that constructor.

docs/configuration/fault-quarantine.md
```diff
@@ -110,11 +111,22 @@ fault-quarantine:
...
+
+At least one of `percentage` and `maxNodes` must be positive, and a negative value for either is rejected rather than ignored. The breaker refuses to start otherwise, so it cannot be silently reduced to a no-op while `enabled: true`. For the same reason, a threshold above the fleet size is clamped to the fleet size rather than being left unreachable, and the `bound` label described below reports `fleetSize` when that happens. To turn the breaker off, use `enabled: false` rather than an unreachable threshold.
 
 #### duration
 Time window for tracking cordon events. The circuit breaker counts unique node cordons within this sliding window.
 
+### Observability
+
+`fault_quarantine_breaker_threshold_nodes{bound}` reports the effective threshold in nodes, with `bound` naming what produced it: `percentage`, `maxNodes`, or `fleetSize` when a configured bound exceeded the cluster size and was clamped. `fault_quarantine_breaker_utilization` keeps its existing meaning, the recent cordon count as a fraction of total nodes, so existing dashboards and alerts are unaffected.
```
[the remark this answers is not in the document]
lfriedman-netllama: Fixed in `aa1f507e`, "a fraction of GPU nodes". Same underlying fact as the `percentage` correction above: the denominator is `GetTotalNodes`, which counts nodes carrying `informer.GPUNodeLabel` (`nvidia.com/gpu.present`), not the whole cluster.

lfriedman-netllama:
@lalitadithya this now needs a fresh `/ok to test aa1f507e` when you get a chance. Three rounds of review have landed since the last one, including two real bugs in code paths I added:

1. **A bound above the fleet size silently disabled the breaker** (@XRFXLP). `recentCordonedNodes` can never exceed `totalNodes`, so `maxNodes: 1000` on a 288-node cluster made the threshold unreachable. Now clamped, and reported as `bound="fleetSize"`.
2. **A `NaN` percentage left the breaker permanently tripped from startup** (@coderabbitai). NaN passes every comparison, the clamp does not fire, and `int(NaN)` is 0, so `recentCordonedNodes >= 0` is always true. Non-finite values are now rejected.

Plus: validation moved from the initializer into `NewSlidingWindowBreaker` so direct `Config` builders including the e2e suite cannot bypass it, negatives rejected instead of silently ignored, the float comparison moved ahead of the `int` conversion so behaviour is architecture-independent, the test folded into `breaker_test.go`, the constructor error wrapped, and the docs you asked for across `docs/configuration/fault-quarantine.md`, `docs/circuit-breaker.md` and `docs/METRICS.md` (with two pre-existing inaccuracies fixed: the denominator is GPU nodes, and the breaker trips when the threshold is *reached*, not exceeded).

Verification: full `fault-quarantine` suite green including the ~50s reconciler e2e suite, `golangci-lint` 0 issues across the module. `TestValidateBounds_RejectsConfigsThatCannotLimit` now covers 9 cases and `TestTripThreshold_...` 12.

fault-quarantine/pkg/breaker/breaker.go
```diff
@@ -176,11 +180,54 @@ func (b *slidingWindowBreaker) sumBuckets() int {
...
+}
+
 // IsTripped checks if the circuit breaker should prevent further node cordoning.
 // It returns true if:
 // 1. The breaker is already in TRIPPED state, OR
-// 2. Recent cordon events reach the configured threshold (TripPercentage * GPU nodes)
-// The method automatically trips the breaker if the threshold is exceeded.
+// 2. Recent cordon events reach the configured threshold
+// The threshold is the lower of TripPercentage * GPU nodes and TripMaxNodes, whichever
+// bounds are configured. The method automatically trips the breaker if it is exceeded.
```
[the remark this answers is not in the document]
lfriedman-netllama: Both correct, and both were mine: I rewrote that doc comment in this PR and got it wrong twice. Fixed in `73586d2b`.

- `TripPercentage * GPU nodes` did omit the `/100`.
- "exceeded" contradicted `shouldTrip := recentCordonedNodes >= threshold`, and also contradicted line 228 of the same comment, which already said "reach".

I went slightly further than the suggestion because the comment also predated the fleet-size clamp:

```go
// The threshold is the lower of the configured percentage of GPU nodes and TripMaxNodes,
// clamped to the fleet size. The breaker trips when the recent-cordon count reaches it.
```

@lalitadithya heads up that this landed after your `/ok to test adc29f74`, so the run you kicked off is one commit behind. `73586d2b` is a **comment-only change** with no behaviour difference, so nothing in that run is invalidated in substance; it just needs a fresh `/ok to test 73586d2b` to satisfy the gate. Sorry for the extra round.

Full `fault-quarantine` suite re-run uncached and green, `golangci-lint` 0 issues.

fault-quarantine/pkg/breaker/breaker.go
```diff
@@ -176,11 +180,54 @@ func (b *slidingWindowBreaker) sumBuckets() int {
 	return sum
 }
 
+// Names for the configured bounds, used in logs and the binding-bound metric.
+const (
+	boundPercentage = "percentage"
```
XRFXLP: nit: can we make these named type like `Bound` as this is a small fixed set?

distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
```diff
@@ -65,6 +65,15 @@ circuitBreaker:
...
   percentage: 50
+  # Optional absolute cap on cordoned nodes within the window, independent of fleet size.
+  # 0 disables it. When both are set the lower bound binds, so "percentage: 5, maxNodes: 10"
+  # means 5% but never more than 10 nodes.
+  #
+  # Prefer this when the blast radius you care about is a number of machines rather than a
+  # share of the fleet, since a percentage silently changes meaning as the fleet grows. Note
+  # it is per cluster by nature: an operator running many clusters of differing size will
+  # need to tune it per cluster, or leave it 0 and rely on percentage alone.
+  maxNodes: 0
```
XRFXLP, line 76: nit: can we make the similar change in values-full.yaml?

fault-quarantine/pkg/breaker/breaker.go
lfriedman-netllama: Agreed, done in `ef6f7273`:

```go
// Bound names the configured limit that produced the effective threshold. It appears in
// logs and as the label on fault_quarantine_breaker_threshold_nodes.
type Bound string

const (
	boundNone       Bound = ""
	boundPercentage Bound = "percentage"
	boundMaxNodes   Bound = "maxNodes"
	boundFleetSize  Bound = "fleetSize"
)
```

`tripThreshold` now returns `(int, Bound)`, and the conversion to `string` happens once at the metric boundary. I also added `boundNone` as an explicit zero value rather than a bare `""`, which makes the neither-bound-configured case readable in the table test.

distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
lfriedman-netllama: Good catch, that was an omission. Added in `ef6f7273`.

`values-full.yaml` now documents `maxNodes` alongside `percentage` in the `circuitBreaker` block, with the lower-bound-binds rule, the reason to prefer an absolute bound, and the at-least-one-positive requirement. I also noted on `percentage` that `0` bounds by `maxNodes` alone, since that combination is not obvious from the existing "Range: 1-100" comment.

Verified it reaches the config rather than just sitting in the values file:

```
$ helm template t . --set global.faultQuarantine.enabled=true -f values-full.yaml
      percentage = 50
      maxNodes = 0
```

lfriedman-netllama:
@lalitadithya merged `main` in (`201d0654`) since this had gone out of date, and this now needs an `/ok to test 201d0654`.

Since the last full run, this branch has picked up @XRFXLP's two nits, both applied: `Bound` is now a named string type with an explicit `boundNone` zero value, and `values-full.yaml` documents `maxNodes` alongside `percentage`.

Re-verified on the merged head with envtest assets set: full `fault-quarantine` suite green including the ~50s reconciler e2e suite, `golangci-lint` 0 issues, and `maxNodes = 0` confirmed reaching the rendered config from `values-full.yaml`.

Every review thread on this PR is answered.

XRFXLP (approved):

PR #1720, opened 2026-09-02
perf(kubernetes-object-monitor): derive cache transforms from CEL policy instead of storing everything

### LINKED ISSUE #1718: [Feature]: Derive cache transforms from policy CEL in kubernetes-object-monitor
### Prerequisites

### Code of Conduct

### Feature Summary

kubernetes-object-monitor caches every watched object in full, as `unstructured.Unstructured` with no informer transform. Because the policies that decide what it reads are known configuration, the set of fields each policy touches can be derived from the compiled CEL at startup and used to build a per-GVK cache transform. On a 30,113-node fleet this reduced resident memory from 21.85 GB to 6.15 GB with no change in behaviour.

### Problem/Use Case

`buildCacheOptions` in `health-monitors/kubernetes-object-monitor/pkg/initializer/initializer.go` registers each watched GVK as `unstructured.Unstructured` and sets no `Transform`. Two effects compound:

1. **Every field is retained.** The `NodeNotReady` policy evaluates one expression, `resource.status.conditions.exists(c, c.type == "Ready" && c.status == "False")`, which reads about 8 KB of a 53.6 KB production GPU worker node. The remaining 85 percent -- `managedFields`, roughly 180 labels, a 50-entry image list, annotations -- is cached so that it can be ignored.

2. **`unstructured` costs about 12 bytes of heap per byte of object**, because each field becomes a map entry with a boxed value and its own allocation. Components that cache the same nodes typed pay roughly 2.4x: fault-quarantine retains *more* fields per node (32 KB against 53.6 KB) and uses 6.8 GB where kubernetes-object-monitor uses 58.9 GB.

Measured on an EKS v1.34 benchmark cluster at production node shape:

| nodes | kubernetes-object-monitor RSS | note |
|---|---|---|
| 5,105 | 3.15 GB | |
| 30,113 | 21.85 GB | |
| 90,105 | 58.9 GB | still climbing when the limit was raised |

At 90k nodes it is the largest memory consumer in the system by roughly 9x, and it OOM-killed 65 times against a 32Gi limit before the limit was raised to 64Gi. Each restart is a multi-minute gap in node monitoring while the cache refills.

The cost scales with the size of the objects a policy selects rather than with the work the policy does. A policy scoped to a namespace pays for that namespace; a cluster-wide policy pays for the whole collection. On a cluster with 991k pods, a cluster-wide Pod policy would hold about 15.7 GB of wire bytes in the same untransformed cache, which at the inflation above is on the order of 190 GB resident. Deriving the retained fields from the policy makes the cost proportional to what the expression reads instead, which is what keeps a cluster-wide policy affordable.

### Proposed Solution

Derive the retained field set from the policy expressions and install a per-GVK cache transform.

**1. Extract referenced fields from the compiled CEL AST.** The policies are configuration read at startup, so this constrains nothing an operator may write. Walking the AST for `Select` and index chains rooted at the `resource` variable yields the paths each expression reads, for example `status.conditions` or `spec.nodeName`. cel-go already provides what is needed via `ast.NativeRep()` and `common/ast`; no new dependency.

Three cases need care:

- A recorded path means the whole subtree beneath it is kept. That is what makes comprehensions safe: `status.conditions.exists(c, c.type == x)` records `status.conditions`, and the per-element accesses through the iteration variable need no separate handling.
- Comprehension iteration and accumulator variables must be scope-tracked so a binding that shadows `resource` is not mistaken for the object.
- Where the object is used whole rather than through a field, for example `resource == null` or `size(resource)`, extraction must report failure and that GVK must be cached in full. Pruning on an incomplete field set would silently change evaluation results. A computed index such as `metadata.labels[resource.spec.key]` is safe without special handling, because the whole `metadata.labels` subtree is retained.

**2. Build a transform per GVK** from the union of the paths of that GVK's enabled policies, covering both the predicate and the node-association expressions, plus a mandatory set the informer and reconciler need regardless: `apiVersion`, `kind`, `metadata.name`, `metadata.namespace`, `metadata.uid`, `metadata.resourceVersion`, `metadata.deletionTimestamp`. Outside CEL the reconciler reads only `GetName()` and `GetNamespace()`.

**3. Register every watched GVK in `ByObject`, not only the namespaced ones.** Today a policy without a namespace is routed to `allNamespacesByGVK` and receives no `ByObject` entry at all, which is precisely why the cluster-scoped Node cache is untransformed. Adding an entry with an empty `Namespaces` map preserves cluster-wide caching and provides somewhere to attach the transform.

**Measured result.** A working implementation was deployed to the benchmark cluster and compared against `:main` on the same 30,113-node fleet, read from the kubelet summary API:

| | resident memory |
|---|---|
| current behaviour | 21.85 GB |
| with CEL-derived transform | 6.15 GB |
| reduction | 3.55x, 15.7 GB |

The per-node saving is larger than that ratio suggests. Roughly 3 GB is a fixed floor of Go runtime, workqueues and other informers that the transform cannot affect, so the ratio improves as the fleet grows. Extrapolated to 80k nodes this is about 58 GB against about 16 GB, which is the difference between requiring a 64Gi limit and a 16Gi one.

Behaviour was verified unchanged: CEL evaluates per node against the pruned object, forcing a node to `Ready=False` moved true evaluations from 0 to 4, and the resulting health event was published and sent. A unit test confirms the transform keeps the fields a policy reads while dropping `managedFields`, `labels` and `status.images`, and that informer-critical metadata survives.

The same mechanism applies to every watched kind, not only nodes. A `PodFailed` policy reading `status.phase` and `spec.nodeName` would drop the container spec, which is 78 to 92 percent of a production user pod, and would turn the cluster-wide pod policy cliff above from unbootable into merely expensive.

### Component

Health Monitor

XRFXLP:
## Summary

The Kubernetes object monitor has two problems today.

1. **It stores far more of each object than it needs.** Every watched GVK is registered as `unstructured.Unstructured` with no `Transform`, so a 53.6 KB node is cached whole although the `NodeNotReady` predicate reads only `status.conditions`. Unstructured costs roughly 12 bytes of heap per byte of object, every field being a map entry with a boxed value, so the waste compounds until the monitor is unusable: #1718 reports 3.15 GB at 5k nodes, 21.85 GB at 30k, 58.9 GB at 90k, and 65 OOM-kills against a 32Gi limit.
2. **It calls the API server once per `lookup()`.** A policy that reaches a second object through `lookup()` pays a request on every evaluation of every object, because the GVK the call names was treated as unknowable until the expression ran.

The policies are configuration loaded at startup, so what each expression touches can be known before a single object is cached. This PR replaces blind evaluation with a walk of the compiled CEL AST, and caches only the fields the policies predicate against — those of the watched object, and those of the objects their `lookup()` calls return.

### Implementation

Take a policy that watches pods, reads the node through `lookup()`, and reports against that node:

```yaml
- name: device-plugin-crashlooping
  enabled: true
  resource: { group: "", version: v1, kind: Pod }
  predicate:
    expression: |
      resource.metadata.labels['app'] == 'nvidia-device-plugin' &&
      resource.status.phase != 'Running' &&
      lookup('v1', 'Node', '', resource.spec.nodeName).status.conditions.exists(c, c.type == 'Ready' && c.status == 'True')
  nodeAssociation:
    expression: resource.spec.nodeName
```

#### Step 1. Read the policies before the cache is built.

`InitializeAll` calls `createManager`. `createManager` calls `buildCachePlan`, and `buildCachePlan` returns the `cache.Options` that the manager is then built with. The fields to keep are therefore worked out before the manager exists.

The CEL environment used here comes from `NewCompilerEnvironment`. It can compile an expression and inspect it. It cannot run one, because at this point nothing is connected to the cluster.

#### Step 2. Find the fields each expression reads

`buildCacheEntries` calls `derivePolicyFieldPaths`. That function compiles every expression of every enabled policy. For this policy `policyExpressions` hands it two: the predicate and the nodeAssociation.

Each compiled expression is asked for two lists.

`ResourceFieldPaths` returns the fields the expression reads off the watched pod. The predicate returns `metadata.labels.app`, `spec.nodeName` and `status.phase`. The nodeAssociation returns `spec.nodeName`.

`LookupTargets` returns the kinds the expression reads through `lookup()`, and the fields it reads off each of them. The predicate returns `v1/Node` with `status.conditions`.

`spec.nodeName` is in the first list because the predicate passes it to `lookup()` as an argument. Recording a kind read through `lookup()` does not stop the walk from descending into the arguments of that call.

Both lists come out of one walk of the syntax tree. `resolveChain` takes an expression such as `resource.status.phase` and returns two things: the expression the chain starts at, and the field path read from that starting point. `recordChain` then looks at the starting point. If it is the `resource` variable, the path is a field of the watched pod. If it is a `lookup()` call whose apiVersion and kind are string literals, the path is a field of that kind. Anything else is not recorded.

A recorded path covers every field below it. `status.conditions` is recorded once, and every field of every condition is kept, so `exists(c, ...)` needs no handling of its own. A computed map key behaves the same way: `metadata.labels[resource.spec.nodeName]` records `metadata.labels`, so every label is kept.

Paths are lists of segments rather than dotted strings. A label key such as `nvidia.com/gpu.present` is one segment, dot and all.

#### Step 3. Decide what to cache for each kind

`cacheEntryFor` builds the field list for one kind. It starts with the seven fields that are always kept: `apiVersion`, `kind`, `metadata.name`, `metadata.namespace`, `metadata.uid`, `metadata.resourceVersion` and `metadata.deletionTimestamp`. The informer needs those to key objects and to handle deletes and resyncs. It then adds the fields from step 2, and wraps the finished list in `newFieldPruningTransform`.

`collectWatchScopes` supplies the namespaces. Pod is watched and the policy names no namespace, so Pod is cached cluster-wide. Node is never watched, only looked up, and it is cached cluster-wide as well, because a `lookup()` can name any namespace it likes.

Both kinds get an entry in `cache.Options.ByObject`. Node also goes into `plan.lookupGVKs`, which is the list of kinds `lookup()` is allowed to read from the cache.

Startup logs the outcome:

```
gvk="/v1, Kind=Pod"  servesLookups=false retainedFields="apiVersion kind metadata.deletionTimestamp metadata.labels.app metadata.name metadata.namespace metadata.resourceVersion metadata.uid spec.nodeName status.phase"
gvk="/v1, Kind=Node" servesLookups=true  retainedFields="apiVersion kind metadata.deletionTimestamp metadata.name metadata.namespace metadata.resourceVersion metadata.uid status.conditions"
```

#### Step 4. Wire up the two readers once the manager exists

`buildManagerOptions` sets `client.Options.Cache.Unstructured` to true. controller-runtime leaves it false. While it is false, the reconciler's `Get` of an unstructured object goes to the API server on every reconcile and comes back with an object that nothing pruned.

`InitializeAll` then builds the environment that evaluates the policies, and gives it two ways to read an object. `NewEnvironment(mgr.GetAPIReader())` reads from the API server. `UseCacheForLookups(mgr.GetCache(), plan.lookupGVKs)` reads from the cache, and only for the kinds in that list.

#### Step 5. Run

A pod arrives from the watch. The transform prunes it to the ten fields logged above. Its other labels, its annotations, its containers and the rest of its status are dropped before the cache stores it.

Reconcile reads that pruned pod from the cache and evaluates the predicate against it. The `lookup('v1', 'Node', ...)` inside reaches `getForLookup`. Node is one of the kinds `UseCacheForLookups` was given, so `getForLookup` asks whether the cache can answer for Node yet.

The informer for Node is created by the first `lookup()` that needs it, and it lists every node in the cluster before it can answer anything. Waiting for that list is what has to be avoided, because `Evaluate` runs one evaluation at a time for the whole process. So `getForLookup` reads Node from the API server until the informer has caught up, and from the cache afterwards. A kind whose informer never catches up, for want of `list` and `watch` permission, is read from the API server for the life of the process, and that is logged once.

The nodeAssociation runs against the same pruned pod. `spec.nodeName` is there because step 2 read it off that expression too.

#### When the fields cannot be worked out

Some expressions cannot be reduced to a list of fields. Nothing is pruned and nothing is read from the cache in those cases. Pruning a guess would change how a policy evaluates without saying so.

A kind is cached in full when a policy that watches it uses the whole object, as in `size(resource)`, or when one of its expressions fails to compile.

A `lookup()` reads from the API server in four cases. It computes its apiVersion or kind. Its result is used as a whole. It sits in an expression that uses the watched object as a whole, where the walk stops and cannot see what the calls past that point read. Or it names a kind that is cached for named namespaces only.

One rule holds the two roles together. A kind can be watched by one policy and looked up by another. Both sets of fields go into the same list. Without that, the kind would be pruned to what the watching policy reads, and the `lookup()` would ask for a field that is no longer there. CEL cannot tell a field the cache dropped from a field the object never had, so nothing would report an error and the policy would quietly answer wrongly.

The unit tests establish something narrower than the cluster numbers below. On a synthetic node carrying the field counts of a large cluster's nodes (≈180 labels, ≈38 conditions, ≈50 images, large `managedFields`), pruning takes the object from 118,894 to 8,572 bytes of serialized JSON. That is serialized size, not heap. The typed `*v1.Node` cache that `annotations.Manager` reads through is still stored in full, and is worth pruning separately.

### Validation

Measured on a 50,113-node cluster (EKS v1.34), nodes at production shape — 52,696 B median, ≈180 labels, 40 conditions, 50 images — carrying 551,326 pods.

Policies under test. The second is the shape this change is for: a pod watched cluster-wide, reaching its node through `lookup()` on a field of the watched object.

```toml
[controller]
  maxConcurrentReconciles = 100

[[policies]]
  name = "NodeNotReady"
  enabled = true
  [policies.resource]
    group = ""
    version = "v1"
    kind = "Node"
  [policies.predicate]
    expression = "has(resource.status.conditions) && resource.status.conditions.exists(c, c.type == \"Ready\" && c.status == \"False\")"
  [policies.healthEvent]
    componentClass = "Node"
    isFatal = true
    message = "Node is not ready"
    recommendedAction = "CONTACT_SUPPORT"
    errorCode = ["NODE_NOT_READY"]

[[policies]]
  name = "PodOnUnschedulableNode"
  enabled = true
  [policies.resource]
    group = ""
    version = "v1"
    kind = "Pod"
  [policies.predicate]
    expression = "resource.status.phase == \"Failed\" && lookup('v1', 'Node', '', resource.spec.nodeName).spec.unschedulable == true"
  [policies.nodeAssociation]
    expression = "resource.spec.nodeName"
  [policies.healthEvent]
    componentClass = "Node"
    isFatal = false
    message = "Pod failed on an unschedulable node"
    recommendedAction = "CONTACT_SUPPORT"
    errorCode = ["POD_ON_UNSCHEDULABLE_NODE"]
```

Each arm ran on a fresh pod against the same fleet, with a 80Gi memory limit, for ten minutes.

| | v1.21.0 | this branch |
|---|---|---|
| Resident memory | 70.85 GB | 19.01 GB |
| Restarts | 2, OOMKilled | 0 |
| Node reconciles in the window | 0 | 200,901 |
| Pod reconciles in the window | 0 | 1,654,023 |
| API GETs | 112 | 258 |

The released build does not run this workload. Watching pods cluster-wide caches 551,326 objects whole, which reached 70.85 GB of an 80Gi limit and was OOM-killed twice inside ten minutes. It completed no reconciles in that time, having never finished filling the cache, and its 112 GETs are startup traffic for the same reason. Pruned to `status.phase`, `spec.nodeName` and identity, the same policy set fits in 19.01 GB and processes 1.85 million reconciles over the same window with no restarts.

The memory ratio here understates the difference, because the 70.85 GB is a cache still growing when the process died rather than a settled figure.

An earlier run with a narrower policy set — both policies watching nodes only, the lookup naming the watched object itself — gave 25.75 GB against 10.84 GB, 34,976 reconciles against 155,162, and 70,113 API GETs against 57. That is 2.005 GETs per reconcile on the released build, one for the watched object because unstructured reads bypass the cache without `Cache.Unstructured`, and one for the `lookup()`. At the 58 reconciles per second it sustains, that is about 117 GETs per second against the API server at 50k nodes, scaling with node count to about 187 at 80k.

Both figures are specific to their policy sets. A policy set touching more fields prunes less, and a fixed floor of runtime and other informers is untouched either way.

Refs #1718

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [x] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [x] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Improved Kubernetes resource monitoring with policy-aware cache usage.
  * Added efficient `lookup()` evaluations using synchronized cached data when available, with API-based fallback.
  * Added support for cluster-scoped and all-namespace resources.
  * Reduced cached resource data while retaining fields required for policy evaluation and health reporting.
  * Preserved literal map keys containing dots during field selection.
  * Added safe handling for cache synchronization delays and cache read failures.

* **Documentation**
  * Expanded documentation on cache synchronization, informer initialization, and API-read behavior.

* **Tests**
  * Added coverage for cache pruning, lookups, field selection, transformations, and fallback behavior.


health-monitors/kubernetes-object-monitor/pkg/cel/environment.go
```diff
@@ -23,24 +23,51 @@ import (
...
+	// starts a cluster-wide informer for it on demand and holds it in full.
+	reader client.Reader
+	// objectCache and cachedGVKs back lookup() for the GVKs UseCacheForLookups
+	// named. fellBack records the GVKs already reported as falling back, so a
+	// misconfiguration is logged once rather than per evaluation. All three are
+	// written before the manager starts and read under evalMu, which Evaluate
+	// holds while lookup() runs.
+	objectCache LookupCache
+	cachedGVKs  map[schema.GroupVersionKind]bool
+	fellBack    map[schema.GroupVersionKind]bool
```
lalitadithya: nit: should this be: 

```suggestion
	fallBack    map[schema.GroupVersionKind]bool
```

or is my grammar off?

health-monitors/kubernetes-object-monitor/pkg/cel/fields.go
```diff
@@ -0,0 +1,354 @@
...
+// their own. The same holds for a computed index such as
+// metadata.labels[resource.spec.nodeName], which records metadata.labels and
+// therefore retains whichever key is present at runtime.
+//
+// The second return value reports whether extraction was complete. It is false
+// when the expression uses the object as a whole rather than through a field
+// access, as in size(resource), because no set of paths describes what such an
+// expression reads. Callers must cache the object in full in that case: pruning
+// against an incomplete field set silently changes evaluation results.
+func ResourceFieldPaths(compiled *cel.Ast) ([][]string, bool) {
```
lalitadithya: can we use named return here so that we don't need to have lengthy docs about return values


```suggestion
func ResourceFieldPaths(compiled *cel.Ast) (fieldPaths [][]string, isComplete bool) {
```

XRFXLP: done

health-monitors/kubernetes-object-monitor/pkg/cel/environment.go
XRFXLP: it was past tense so "fell", renamed it make the intent clearer.

lalitadithya (approved):

Issue #1434, opened 2026-06-29
[Feature]: Add optional enrichment pipeline to event-exporter for Pod metadata and business context


ymygogogo:
### Prerequisites

- [x] I searched existing issues

### Code of Conduct

- [x] I agree to follow NVSentinel's Code of Conduct

### Feature Summary

Add an optional enrichment pipeline to event-exporter so exported HealthEvent CloudEvents can include controlled Pod metadata and business correlation fields before being sent to the configured sink.

The feature should be disabled by default. When enabled, event-exporter enriches each HealthEvent with Pod information from the affected node, selected Pod labels/annotations, optional Prometheus-derived context, and enrichment status metadata. The original data.healthEvent payload should remain backward compatible.

### Problem/Use Case

In a shared GPU cloud environment, a node fault event is not sufficient for business-side notification. The downstream service needs to correlate the node fault with user workloads, tasks, tenants, or orders running on the affected node.

Today event-exporter can send NVSentinel HealthEvents to a remote service, but the payload does not include the Pods currently or recently running on the faulty node, nor the business labels/annotations attached to those Pods. The receiver must either query Kubernetes/Prometheus itself or deploy an additional adapter.

For platforms that want event-exporter to produce the final outbound event payload, we need an optional enrichment stage inside event-exporter that can:

Identify the fault node from data.healthEvent.nodeName.
Attach Pod summaries for workloads on that node.
Include only allowlisted business labels/annotations, such as tenant ID, user ID, task ID, job ID, workspace ID, or order ID.
Optionally query Prometheus for historical or metric context.
Preserve event delivery even when enrichment is partial or unavailable.

### Proposed Solution

Add a configurable enrichment pipeline between HealthEvent loading and sink publishing:
```
MongoDB HealthEvents change stream
  -> event-exporter
  -> enrichment pipeline
       -> pod metadata enricher
       -> prometheus enricher optional
       -> field filter / redaction / size limiter
  -> CloudEvents payload
  -> configured sink
```
The pipeline should be optional and default to disabled:

```yaml
event-exporter:
  exporter:
    enrichment:
      enabled: false
```
When disabled, event-exporter behavior and payloads should remain unchanged.

Payload Proposal
Do not change the existing data.healthEvent structure. Add enrichment data under data.enrichment:
```json
{
  "data": {
    "healthEvent": {},
    "metadata": {},
    "enrichment": {
      "schemaVersion": "nvsentinel.event_exporter.enrichment.v1",
[...]

PR #1724, opened 2026-09-02
feat(event-exporter): add a CEL filter for events reaching the sink

### LINKED ISSUE #1702: [Feature]: event-exporter has no event filter, so 99% of exported events are recommendedAction=NONE
### Prerequisites

### Code of Conduct

> **Reworked 2026-09-01** around CEL, per @lalitadithya's review. The original proposal used
> per-property filter fields; that is replaced below by a single CEL expression, which is both
> more expressive and consistent with the rest of the project.

### Feature Summary

`event-exporter` has no event selection. It publishes every HealthEvent from the change stream
to the configured sink. Request: an optional **CEL expression** to select which events reach the
sink, defaulting to today's behaviour of exporting everything.

### Problem/Use Case

`CONTACT_SUPPORT` is the one recommended action that produces no remediation CR: the node is
cordoned and drained and nothing else happens, so no human is notified unless something outside
NVSentinel is watching. `event-exporter` is the natural notification path, and its CloudEvents
payload already carries `recommendedAction`, `customRecommendedAction` and `isFatal`
(`pkg/transformer/cloudevents.go:58-70`).

The obstacle is volume. Measured on a 288 node GB200 fleet over 7 days:

| recommendedAction | isHealthy | events |
| --- | --- | --- |
| `NONE` | false | 797,965 |
| `NONE` | true | 529,081 |
| `CONTACT_SUPPORT` | false | 12,031 |
| `COMPONENT_RESET` | false | 91 |
| `RUN_DCGMEUD` | false | 9 |
| `RESTART_VM` | false | 7 |

About 190,000 events per day, of which **99.1 percent carry `recommendedAction: NONE`**. A
single non-actionable code, XID 45, is about 52,000 per day on its own.

The actionable subset is tiny by comparison: over 3 days, unhealthy `CONTACT_SUPPORT` was 203
events across 13 distinct `(node, errorCode)` pairs, roughly 4 to 5 deduplicated notifications
per day. Delivering those 203 events today means the sink must accept about 570,000.

On the original intent of exporting everything to a data lake and filtering at the destination:
agreed, and that stays the default. The value of a source filter shows up when the sink is a
notification path rather than a lake, which is our case.

### Proposed Solution

A single CEL expression over the HealthEvent:

```yaml
exporter:
  filter:
    # CEL over the HealthEvent. Absent or empty exports everything, which is today's behaviour.
    expression: >-
      event.recommendedAction != 'NONE' && event.errorCode != '45'
```

This is close to free to implement, because the binding already exists.
`platform-connectors/pkg/transformers/overrides/cel.go` evaluates CEL over a `HealthEvent`, and
`buildEventMap` there already exposes exactly the fields a filter needs: `agent`, `checkName`,
`componentClass`, `errorCode`, `isFatal`, `isHealthy`, `recommendedAction`, `nodeName`,
`metadata`, `message`. Reusing it keeps one CEL surface and one field vocabulary across the
project, and avoids the combinatorial growth of per-property options.

### Two implementation points that matter regardless of syntax

1. **A filtered event must still advance the resume token.**
   `pkg/exporter/sequence_tracker.go` advances only for contiguous completions, so a filtered
   event has to be marked completed rather than skipped. If it is merely skipped, one filtered
   event at the head of the stream stalls token progress and a restart redelivers everything
   after it. With a filter that drops 99 percent of events this is the difference between
   working and not working at all.

2. **Compile and validate the expression at startup, and refuse to start on error.**
   A typo must not silently drop every event at runtime. Given the volumes, silent total
   suppression is a worse failure than having no filter.

### Interaction with #1434

#1434 proposes an enrichment pipeline that attaches Pod metadata and optional Prometheus context
per event. Event selection should be able to run **before** it, otherwise a fleet like this one
performs about 190,000 enrichment lookups per day for events that are then discarded:

```text
change stream -> decode HealthEvent -> filter (this issue) -> enrichment (#1434)
  -> CloudEvents transform -> sink
```

### Acceptance criteria

- An absent or empty `filter.expression` exports every event, identical to today.
- An event whose expression evaluates false is not published to the sink.
- The resume token advances past filtered events. No stall, and no redelivery gap after restart.
- An invalid expression fails at startup rather than at runtime.
- Filtered events are counted in a metric, so drops are observable rather than silent.
- Backfill honours the same filter.
- Filtering runs before enrichment when #1434 is present.
- Unit tests cover a passing expression, a failing expression, an empty filter, an invalid
  expression at startup, and token advancement across a run of consecutively filtered events.
- Documentation states that filtering is evaluated before enrichment and before the CloudEvents
  transform.

### Component

Event Exporter

lfriedman-netllama:
Closes #1702, using CEL as you asked rather than the per-property knobs I originally proposed.

## Problem

`event-exporter` publishes every HealthEvent from the change stream. Measured on a 288 node GB200 fleet over 7 days:

| recommendedAction | isHealthy | events |
| --- | --- | --- |
| `NONE` | false | 797,965 |
| `NONE` | true | 529,081 |
| `CONTACT_SUPPORT` | false | 12,031 |
| `COMPONENT_RESET` | false | 91 |
| `RUN_DCGMEUD` | false | 9 |
| `RESTART_VM` | false | 7 |

~190k events/day, **99.1% `NONE`**. Delivering the ~200 actionable events in a 3 day window means the sink must accept about 570,000.

To your point on the issue that the original intent was exporting everything to a data lake and filtering at the destination: agreed, and that stays the default. A source filter earns its place when the sink is a notification path rather than a lake.

## Configuration

```yaml
event-exporter:
  exporter:
    filter:
      expression: "event.recommendedAction != 'NONE' && !('45' in event.errorCode)"
```

Absent or empty exports everything, so nothing changes unless configured.

## On sharing the CEL vocabulary

You suggested reusing `buildEventMap` from `platform-connectors/pkg/transformers/overrides/cel.go`. That turned out not to be directly possible: it is **unexported**, and `platform-connectors` is a binary module that nothing else imports, while `event-exporter` depends only on `commons` and `data-models`.

Rather than copy it and end up with two vocabularies drifting apart, which is what you were trying to avoid, I moved it to **`commons/pkg/celevent`** and pointed both callers at it. The override transformer's behaviour is unchanged, which its existing test suite demonstrates without modification. If you would rather I duplicated it and left `platform-connectors` untouched, say so and I will.

## Three points that mattered

**1. Filtered events still advance the resume token.** `processEvent` returns `nil` for a filtered event, exactly as for a published one, so its sequence completes and the token moves past it. Pleasingly, the `workerPool`'s `processFunc` contract already anticipated this:

> Returning nil means the event was handled (published **or intentionally skipped**)

so no change to `sequenceTracker` was needed. There is a regression test asserting the token advances across three filtered events, because this is the difference between the feature working and not working at all: were a filtered event merely skipped, one at the head of the stream would stall the token, and with a filter dropping 99% of events that means never making progress.

**2. Startup validation, with an honest limit.** The expression is compiled during config validation, so a syntax error or a concrete non-boolean type refuses to start. But `event` is bound as `map[string]dyn`, so **any** expression reading a field is statically `dyn`. That means `event.isFatal` cannot be proven boolean and has to be accepted, which in turn means `event.agent` also compiles and is only rejected when the first event arrives. My own test caught this: I had written the strict check first and it rejected `event.isFatal`, which would have been a worse bug than the gap. Documented rather than hidden. Note the pre-existing `compileRules` in the override transformer has the same characteristic.

**3. It fails open.** An evaluation error exports the event and increments `health_events_exporter_filter_errors_total`. Dropping events because of a filter bug is silent data loss; an extra event is noise the sink already tolerates.

## A correction to the issue

The example in #1702 said `event.errorCode != '45'`. **That is wrong**: `errorCode` is a repeated proto field, so it binds as a list and equality against a string is a type error, not a false match. The correct form is `'45' in event.errorCode`. There is a test documenting this specifically, since it is the kind of thing an operator will get wrong once.

## Interaction with #1434

The filter runs before the CloudEvents transform, so it naturally precedes any future enrichment added by #1434, which is the ordering that issue wants:

```
change stream -> decode -> filter (this PR) -> enrichment (#1434) -> transform -> sink
```

It also applies to the backfill path, so a backfill exports the same subset as the live stream.

## Verification

- **`commons/pkg/celevent`: 20 new test cases** across every bound field, the errorCode list semantics, and both compile-time and evaluation-time rejection paths, plus one asserting `metadata` is cloned so an expression cannot mutate the event.
- **`event-exporter`: 7 filter tests** covering no expression, whitespace-only, the actionable-only filter, errorCode exclusion including a multi-code event, fail-open on a non-boolean result, fail-open on a missing field, and the resume-token advancement test.
- **6 config validation tests**: valid, malformed, non-boolean, and blank-expression cases.
- All three touched modules green uncached: `commons` 18 packages, `event-exporter` 6, `platform-connectors` 12 including the 40s kubernetes envtest suite. `golangci-lint` 0 issues in each.
- `make gomod-lint` passes across all 32 modules, which matters because this adds `cel-go` to `commons`. Every module already on `cel-go` was on v0.31.0, so there is no version skew.
- `helm lint` clean. The configmap omits `[exporter.filter]` entirely by default and renders `expression = "event.recommendedAction != 'NONE'"` with correct TOML quoting when set.

## Docs

`docs/configuration/event-exporter.md` gains an Event Filter section: the field table with types, the errorCode-is-a-list warning, four worked examples, and the operational notes on startup validation, fail-open, and token advancement.

One thing I did **not** do: `event-exporter` has no section in `docs/METRICS.md` at all, so its existing metrics are undocumented there too. I documented the two new counters in the configuration doc instead of inventing a section. Happy to add a full `event-exporter` metrics section as a separate PR if useful.

## Summary by CodeRabbit

- **New Features**
  - Added optional CEL expressions to filter health events before publishing.
  - Supports event fields, metadata, and error-code lists.
  - Filtering applies to live and backfilled events while resume progress continues normally.
  - Added metrics for filtered events and filter evaluation errors.

- **Bug Fixes**
  - Invalid or non-boolean expressions are rejected at startup.
  - Evaluation errors fail open, allowing affected events to be published.

- **Documentation**
  - Added configuration guidance, supported fields, examples, metrics, and filtering behavior.


commons/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.27.0
 
 require (
 	github.com/BurntSushi/toml v1.6.0
+	github.com/google/cel-go v0.31.0
```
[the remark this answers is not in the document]
lfriedman-netllama: Good call, and acted on. The v0.32.0 move to `cel.dev/cel-go` is a real cost to weigh, and putting cel-go into `commons` does widen it, since 23 modules depend on `commons`.

Fixed in `07e4cac6` by making the dependency **encapsulated rather than merely centralised**. `celevent` now exposes an opaque type instead of cel-go's:

```go
type Filter struct { program cel.Program }   // unexported field
func Compile(expression string) (*Filter, error)
func (f *Filter) Matches(event *pb.HealthEvent) (bool, error)
```

Previously `Compile` handed back a `cel.Program`, which forced `event-exporter`'s config and exporter to import cel-go too. Now no caller does:

- `event-exporter`: cel-go is **indirect** in `go.mod`, zero `.go` imports.
- `platform-connectors`: **zero** `.go` imports, where it previously had two. `compiledRule` holds a `*celevent.Filter`, and `buildCELEnvironment` is gone.

So the v0.32.0 migration is now a change to one file, `commons/pkg/celevent/celevent.go`, rather than to every component that filters events. That is a better position than before this PR, when `platform-connectors` imported cel-go directly.

The doc comment on `Filter` records this reason so the encapsulation is not accidentally undone.

event-exporter/pkg/config/config.go
```diff
@@ -162,9 +176,30 @@ func (c *Config) Validate() error {
...
+// Compile returns the compiled filter program, or nil when no expression is configured.
+// A nil program means "export everything".
+func (c *FilterConfig) Compile() (cel.Program, error) {
+	if strings.TrimSpace(c.Expression) == "" {
+		return nil, nil
+	}
+
+	env, err := celevent.NewEnv()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
lfriedman-netllama: Applied in `07e4cac6`. `Compile` now wraps with the expression itself, which is the context that actually helps when several components can carry a filter:

```go
filter, err := celevent.Compile(c.Expression)
if err != nil {
    return nil, fmt.Errorf("filter expression %q: %w", c.Expression, err)
}
```

The separate `NewEnv` error path you were pointing at no longer exists: environment creation moved inside `celevent.Compile`, so there is one error to wrap rather than two, and it already carries its own "failed to create CEL environment" context.

event-exporter/pkg/exporter/filter_test.go
```diff
@@ -0,0 +1,142 @@
...
+// with a filter dropping 99% of events means never making progress.
+func TestWorkerPool_FilteredEvent_StillAdvancesResumeToken(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	source := &mockSource{}
+
+	// Mirrors processEvent's contract: a filtered event returns nil, exactly like a
+	// published one, so its sequence completes.
+	filtered := func(_ context.Context, _ client.Event) error { return nil }
```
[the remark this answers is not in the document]
lfriedman-netllama: You were right, and this one was worth catching. The stub returning `nil` meant the test asserted the worker pool's behaviour, not the filter's, so it would have passed even if `processEvent` had errored or published on the filtered path.

Rewritten in `07e4cac6` to drive the real `HealthEventsExporter.processEvent` through the pool, with a false filter and a small `fakeEvent` implementing `client.Event`. The exporter is built with a **nil sink and nil transformer on purpose**: if a filtered event ever reached `publishWithRetry`, the test panics rather than quietly passing.

I mutation-tested it rather than trusting it. Replacing the filter's early return with `if false`, so filtered events fall through to publish:

```
panic: runtime error: invalid memory address or nil pointer dereference
FAIL github.com/nvidia/nvsentinel/event-exporter/pkg/exporter
```

Restored, and green again. So the test now genuinely fails if the filtered path stops short-circuiting.

lfriedman-netllama:
Three review findings addressed in `07e4cac6`, plus one behaviour decision worth flagging explicitly because it changes what the PR does.

**I reverted my own relaxation of CEL type checking.** My first version accepted dyn-typed expressions so that `event.isFatal` would work. That turned out to **break a pre-existing test**, `TestNewProcessor/non-boolean-expression`, which asserts the override transformer rejects `when: "event.agent"` at startup. Two things followed from that:

- I was silently changing `platform-connectors` behaviour, which was never the intent of this PR.
- The strict rule is simply better for a filter that can legitimately drop 99% of events: catching `event.agent` at startup beats failing open on the first event.

So `celevent.Compile` now requires a statically boolean expression, matching what the override transformer has always done. The cost is that a bare boolean read must be written as a comparison, `event.isFatal == true`, and **the error message says so**, since that is the obvious thing to get wrong. Documented in the config doc as its own callout, and the worked example that used `event.isFatal` is updated.

@lalitadithya one thing to confirm since it touches your module: this PR now leaves `platform-connectors` with **no direct cel-go import at all** (`compiledRule` holds a `*celevent.Filter`, and `buildCELEnvironment` is deleted). Behaviour is unchanged and the existing override test suite passes untouched, which is the evidence. If you would rather I left `platform-connectors` alone and duplicated the vocabulary, that is a one-line revert; I went this way because it means cel-go's v0.32.0 module-path move touches one file instead of every component that filters events.

Also, apologies for some churn on this branch: I force-pushed over your `main` merge (`6ab9a74a`) by fetching immediately before a `--force-with-lease`, which defeats the lease. I merged it back rather than leaving it lost, and `07e4cac6` contains both it and your later `5bf49d7a`. Verified with `git merge-base --is-ancestor` for each. I have stopped fetching before force-pushing.

Re-verified after all of it: `commons` 18 packages, `event-exporter` 6, `platform-connectors` 12, zero failures, `golangci-lint` 0 issues in each, `make gomod-lint` green across all 32 modules.

commons/pkg/celevent/celevent.go
```diff
@@ -0,0 +1,119 @@
...
+// BuildEventMap binds a health event's fields for CEL evaluation.
+//
+// Note errorCode is a repeated field, so it binds as a list: match it with
+// `'45' in event.errorCode`, not `event.errorCode == '45'`.
+func BuildEventMap(event *pb.HealthEvent) map[string]any {
+	return map[string]any{
+		"agent":             event.GetAgent(),
+		"checkName":         event.GetCheckName(),
+		"componentClass":    event.GetComponentClass(),
+		"errorCode":         event.GetErrorCode(),
```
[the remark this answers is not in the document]
lfriedman-netllama: Agreed, and applied in `223d0f56`. `metadata` was cloned and `errorCode` was not, which was an oversight rather than a decision: both are reference types and `BuildEventMap` is exported, so handing back the event's own slice lets a caller mutate the event through the returned map.

```go
"errorCode": slices.Clone(event.GetErrorCode()),
```

Test added mirroring the existing metadata one: mutate `eventMap["errorCode"].([]string)[0]` and assert `event.GetErrorCode()[0]` is untouched. The doc comment now says both are cloned and why.

lfriedman-netllama:
@lalitadithya this needs an `/ok to test 223d0f56`.

Since the last run: the three CodeRabbit findings, plus the `errorCode` clone (`metadata` was cloned and `errorCode` was not, so a caller of the exported `BuildEventMap` could mutate the event through the returned map).

The item worth your eye is still the one in my earlier comment: this PR leaves **`platform-connectors` with no direct cel-go import**, because `celevent` now hands back an opaque `*Filter` rather than a `cel.Program`. Behaviour is unchanged and the existing override tests pass untouched. That was in service of the review point that cel-go moves module path to `cel.dev/cel-go` in v0.32.0: the migration is now one file instead of every component that filters events. Say the word if you would rather I left your module alone and duplicated the vocabulary.

Verified: `commons` 18 packages, `event-exporter` 6, `platform-connectors` 12, zero failures, `golangci-lint` 0 issues each, `make gomod-lint` green across all 32 modules.

commons/go.mod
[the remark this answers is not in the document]
lfriedman-netllama: You are right and my claim was overstated. Corrected in `e1e77931`.

I said confining cel-go to `celevent` "keeps that migration to this file". Verified repo-wide, and it does not. Direct `github.com/google/cel-go` importers on this branch:

```
commons/pkg/celevent/celevent.go
fault-quarantine/pkg/evaluator/rule_evaluator.go
health-monitors/kubernetes-object-monitor/pkg/cel/environment.go
health-monitors/kubernetes-object-monitor/pkg/policy/evaluator.go
labeler/pkg/devicecounts/device_counts.go
preflight/pkg/gang/discoverer/podgroup.go
```

What is actually true is narrower, and that is what the comment now says: this keeps the migration out of the **health-event filter path**. Within that path the PR is a net reduction, since `platform-connectors` previously imported cel-go directly and now does not, and `event-exporter` never gains a direct import.

The other four are **deliberately not routed through `celevent`**, and I would not change that here. They build CEL environments over different inputs: fault-quarantine evaluates over `event` *and* `node` for its ruleSets, kubernetes-object-monitor over arbitrary Kubernetes objects, labeler over device counts, preflight over pod groups. Forcing those through a health-event vocabulary would be worse than the duplicated dependency.

So the v0.32.0 migration remains a repo-wide change across five modules. That is a pre-existing condition this PR neither creates nor fixes, and it is worth its own issue rather than being folded in here. Happy to file it if you would like.

Thanks for pushing on this one; I had asserted something I had only checked for the two modules I touched.

docs/configuration/event-exporter.md
```diff
@@ -235,6 +235,77 @@ The default of `10` handles clusters up to ~3,300 nodes at typical event rates.
...
+### Parameters
+
+#### expression
+CEL over the health event. Must evaluate to a boolean; an event whose expression is false is not published.
+
+### When to use it
+
+The original design exports everything to a data lake and filters at the destination, and that remains the default. A source filter earns its place when the sink is a **notification path** rather than a lake.
+
+For scale, on a 288-node GB200 fleet: about 190,000 events per day, of which **99.1% carry `recommendedAction: NONE`**. Delivering the ~200 actionable events in a 3-day window means the sink must otherwise accept roughly 570,000.
```
lalitadithya: I don't think we need to specify this in docs, we can just mention it in the PR description and it should be fine

docs/configuration/event-exporter.md
```diff
@@ -235,6 +235,77 @@ The default of `10` handles clusters up to ~3,300 nodes at typical event rates.
...
+```
+
+### Parameters
+
+#### expression
+CEL over the health event. Must evaluate to a boolean; an event whose expression is false is not published.
+
+### When to use it
+
+The original design exports everything to a data lake and filters at the destination, and that remains the default. A source filter earns its place when the sink is a **notification path** rather than a lake.
```
lalitadithya: Can we rephrase this to say, something like when you want to drop events before it reaches the sink, common when the sink is used to send out notifications

docs/configuration/event-exporter.md
lfriedman-netllama: Agreed, removed in `05a3623c`. The 190k events/day and 99.1% NONE figures are specific to one fleet and would go stale in the shipped docs. They stay in the PR description, where they are evidence for the change rather than guidance for a reader.

The docs now make the point qualitatively instead: most events recommend no action and are not worth notifying on.

docs/configuration/event-exporter.md
lfriedman-netllama: Reworded in `05a3623c`, close to your phrasing:

> Use it when you want to drop events before they reach the sink. This is common when the sink sends out notifications rather than archiving everything, since most events recommend no action and are not worth notifying on.
>
> Exporting everything to a data lake and filtering at the destination remains the default.

I kept the second line so it stays clear that this is opt-in and the original data-lake intent is unchanged.

lfriedman-netllama:
CI was red with 9 failing checks; **one real cause**, fixed in `c67dccb4`.

`simple-lint (gomod)` failed with:

```
ERROR:   ✗ Module needs 'go mod tidy' to be run
make: *** [Makefile:364: gomod-lint] Error 1
```

`event-exporter/go.mod` still carried `github.com/google/cel-go` as a **direct** require. That went stale in this PR's own last round: once `celevent` started returning an opaque `*Filter` instead of a `cel.Program`, event-exporter stopped importing cel-go in any `.go` file, so tidy wants it marked indirect. My earlier local `make gomod-lint` pass predated that change.

The other 8 red checks were `The operation was canceled`, i.e. fail-fast collateral from that one leg, not independent failures.

Reproduced locally, fixed, and re-verified: `make gomod-lint` now reports all validations passed across all 32 modules, and `commons` (18 packages), `event-exporter` (6) and `platform-connectors` (12) are green with zero failures.

Needs another `/ok to test c67dccb4` when convenient.

event-exporter/pkg/config/filter_test.go
```diff
@@ -0,0 +1,80 @@
...
+	cfg.Exporter.OIDC.ClientID = "client"
+	cfg.Exporter.OIDC.Scope = "scope"
+	cfg.Exporter.ResumeToken.Collection = "resume_tokens"
+	cfg.Exporter.ResumeToken.Database = "nvsentinel"
+	cfg.Exporter.Filter.Expression = expression
+
+	return cfg
+}
+
+func TestValidate_NoFilterExpression_IsValid(t *testing.T) {
```
XRFXLP: Can we make these tests table driven including that of .Compile()? 

Do we even need these tests here? Looks like they are basic syntactial cases that belong to commons/pkg/celevent/celevent_test.go.

docs/configuration/event-exporter.md
```diff
@@ -235,6 +235,77 @@ The default of `10` handles clusters up to ~3,300 nodes at typical event rates.
 
 If your publish latency is lower (e.g., 100ms for a co-located endpoint), each worker handles proportionally more events — divide the latency ratio to estimate your actual throughput.
 
+## Event Filter
```
XRFXLP, line 238: Can we also add the appropriate field in values-full.yaml with empty config?

event-exporter/pkg/exporter/filter_test.go
```diff
@@ -0,0 +1,176 @@
...
+		Agent:             "syslog-health-monitor",
+		CheckName:         "SysLogsXIDError",
+		NodeName:          "node-1",
+		ErrorCode:         codes,
+		RecommendedAction: action,
+		IsFatal:           action != pb.RecommendedAction_NONE,
+	}
+}
+
+func TestShouldExport_NoExpression_ExportsEverything(t *testing.T) {
```
XRFXLP: Should we make these tests table driven with fields containing different CEL expression, health event and expected results?

XRFXLP (changes requested):

event-exporter/pkg/config/filter_test.go
lfriedman-netllama: Both parts taken, in `7f60def2`. You are right that most of these belonged elsewhere.

**Deleted as duplicates:** the malformed and non-boolean cases were re-testing `commons/pkg/celevent`'s rules, which already has `TestCompileBool_MalformedExpression_IsRejected`, `TestCompileBool_ConcreteNonBooleanType_IsRejectedAtCompileTime` and `TestCompile_BareFieldRead_IsRejectedWithGuidance`. Which expressions are valid is celevent's contract, not this package's.

**Kept, because it is this package's own behaviour rather than CEL's:**
- a blank or whitespace-only expression compiles to a **nil program**, meaning export everything. That is `FilterConfig.Compile`'s decision, not celevent's.
- `Config.Validate` **surfaces** a compile failure, so startup stops. That is wiring, and it matters because a filter can legitimately drop almost every event, so a typo found at runtime looks identical to "nothing is happening".
- exactly **one** invalid case, asserting the error contains the offending expression, to prove propagation without re-testing the CEL rule.

Both are now table driven, including `Compile()`. 5 cases plus 4, all confirmed running as subtests.

event-exporter/pkg/exporter/filter_test.go
lfriedman-netllama: Done in `7f60def2`: one `TestShouldExport` table over `(expression, action, codes, want)`, 10 cases covering no expression, whitespace-only, actionable-only across three actions, errorCode exclusion including a multi-code event, and the fail-open path.

These do table-drive cleanly because they vary only by input and expected outcome. Worth noting the contrast with the resume-token test just below, which I left standalone: it drives the real `processEvent` through the worker pool and asserts on `MarkProcessed`, so it has a different shape and would only be obscured by being forced into the same table.

Confirmed 19 subtests running across the two files, module green, `golangci-lint` 0 issues.

docs/configuration/event-exporter.md
lfriedman-netllama: Added in `7f60def2`, with `expression: ""` as the default so behaviour is unchanged. The comment covers the three things that are easy to get wrong: write a comparison rather than a bare field (`event.isFatal == true`), `errorCode` is a list so match with membership, and filtered events still advance the resume token.

Verified it reaches the config rather than just sitting in the values file:

```
$ helm template t . -f values-full.yaml --set global.eventExporter.enabled=true \
    --set "event-exporter.exporter.filter.expression=event.recommendedAction != 'NONE'"
      [exporter.filter]
      expression = "event.recommendedAction != 'NONE'"
```

And with the default empty value the `[exporter.filter]` section is omitted entirely, so nothing changes for existing deployments. `helm lint` clean.

lfriedman-netllama:
@XRFXLP all three addressed in `7f60def2`, and this needs an `/ok to test 7f60def2` when convenient.

Your point about the config tests belonging in `celevent` was the useful one: it made me delete coverage rather than move it, because `commons/pkg/celevent` already tests those exact rules. What is left in `event-exporter/pkg/config` is only what that package itself decides, a blank expression meaning "export everything" and `Validate` stopping startup on a bad one.

Net effect on the suite: fewer test functions, same behaviour covered, and the CEL rules asserted in one place instead of two.

XRFXLP (approved):

PR #1726, opened 2026-09-02
fix(health-events-analyzer): skip XID events with nil GeneratedTimestamp instead of panicking

### LINKED ISSUE #1725: [Bug]: XID burst detector panics on health events with nil GeneratedTimestamp, crash-looping health-events-analyzer
## Summary

`XidBurstDetector.ProcessEvent` in `health-events-analyzer/pkg/analyzer/xid_burst_detector.go` (line 359) unconditionally dereferences `event.GeneratedTimestamp`:

```go
timestamp := time.Unix(event.GeneratedTimestamp.Seconds, 0)
```

If a GPU health event arrives with a nil `GeneratedTimestamp`, this panics. There is no `recover()` anywhere on the event-processing path (`store-client/pkg/client/event_processor.go` `processEvents`/`handleSingleEvent` do not recover), so the panic kills the health-events-analyzer process.

Because `MarkProcessedOnError: false` (the failed event is intentionally **not** checkpointed so it is retried on restart), the same poison event is re-delivered from the resume token on every restart → **permanent crash loop** of health-events-analyzer. No health events are analyzed cluster-wide until the event is manually removed from the store.

## Why nil timestamps are a real occurrence

The codebase already defends against nil `GeneratedTimestamp` elsewhere — `platform-connectors/pkg/connectors/kubernetes/process_node_events.go` implements `safeTimestamp()` specifically to handle *"HealthEvent has nil GeneratedTimestamp, falling back to current time"*. `shouldProcessXidEvent` (reconciler) filters on `Agent`, `ComponentClass`, `IsHealthy`, and `ErrorCode`, but not on the timestamp, so a nil-timestamp event passes straight into `ProcessEvent`.

## Reproduction

Minimal unit test (deterministic, no infra needed):

```go
d := analyzer.NewXidBurstDetector()
ev := &protos.HealthEvent{
    NodeName:       "node-nil-ts",
    Agent:          "gpu-health-monitor",
    ComponentClass: "GPU",
    IsHealthy:      false,
    ErrorCode:      []string{"79"},
    // GeneratedTimestamp intentionally nil
    EntitiesImpacted: []*protos.Entity{{EntityType: "GPU", EntityValue: "0"}},
}
d.ProcessEvent(ev) // panics
```

Result:

```
panic: runtime error: invalid memory address or nil pointer dereference
github.com/nvidia/nvsentinel/health-events-analyzer/pkg/analyzer.(*XidBurstDetector).ProcessEvent(...)
        health-events-analyzer/pkg/analyzer/xid_burst_detector.go:359 +0xfb
```

## Expected behavior

The detector should tolerate a missing timestamp — either skip the event (with a warning) or fall back to the current time, consistent with `safeTimestamp` behavior elsewhere in the codebase.

## Actual behavior

Panic → process exit → restart replays the same event via resume token → infinite crash loop.

## Suggested fix

In `ProcessEvent` (or add the guard to `shouldProcessXidEvent` in the reconciler, matching the existing filter style):

```go
if event.GeneratedTimestamp == nil {
    slog.Warn("XID event missing GeneratedTimestamp; skipping",
        "node", event.NodeName)
    return false, 0
}
```

A regression test (`TestXidBurstDetector_NilGeneratedTimestamp_NoPanic`) is available and currently fails red on this input.

ApexYash11:
﻿## Summary

Fixes #1725 (option (a), as agreed in the issue thread).

`XidBurstDetector.ProcessEvent` dereferenced `event.GeneratedTimestamp` unconditionally (`xid_burst_detector.go`), so a health event without a timestamp crashed the analyzer with a nil pointer panic. Because the event processor deliberately does not checkpoint failed events (`MarkProcessedOnError: false`) and redelivers them via the resume token, and there is no `recover()` anywhere on the processing path, a single such event put the analyzer into a permanent crash loop.

## Change

Skip the event with a warning log instead of panicking:

```go
if event.GeneratedTimestamp == nil {
    slog.Warn("XID event has nil GeneratedTimestamp; skipping burst detection",
        "node", nodeName, "errorCode", xidCode)
    return false, 0
}
```

This mirrors the defensive handling already present in platform-connectors (`safeTimestamp`, which logs `"HealthEvent has nil GeneratedTimestamp, falling back to current time"`), i.e. the codebase already treats a missing timestamp as an expected input on other paths.

## Testing

- Added regression test `TestXidBurstDetector_NilGeneratedTimestamp_NoPanic` (fails with a panic before the fix, passes after).
- `go test ./...` in `health-events-analyzer`: all packages pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented crashes when processing XID health events with missing timestamps.
  - Such events are now safely skipped and logged with relevant node and error-code details.

- **Tests**
  - Updated regression test documentation to reflect the previous crash-loop scenario caused by missing timestamps.


lalitadithya:
hi @ApexYash11 , can you please complete DCO https://github.com/NVIDIA/NVSentinel/pull/1726/checks?check_run_id=100360036365

ApexYash11:
Done — both commits are now signed off (`git rebase --signoff`), so the DCO check should pass: `Signed-off-by: Yash Maheshwari <yashmaheshwari8983@gmail.com>`. Branch force-pushed as af7243ec.

ApexYash11:
@lalitadithya  testing is done!!

lalitadithya (approved):

PR #1734, opened 2026-09-02
feat(lifecycle-manager): implement MaintenanceRequest controller and webhook [2/3]

cbumb:
## Summary

Adds the reconciler and validating webhook for the `MaintenanceRequest` CRD introduced in [Part 1](https://github.com/NVIDIA/NVSentinel/pull/1733). When a MaintenanceRequest is created, the controller claims the node atomically via annotation, emits a health event through the platform-connector pipeline, and on deletion emits a clearing event to retract the fault. The validating webhook enforces spec immutability, startTime-in-future, node existence, and one-MR-per-node constraints.

### What's included
- **Reconciler** (`maintenancerequest_controller.go`) — adds finalizer, claims node atomically via annotation, auto-populates event fields (`id`, `version`, `generatedTimestamp`), stamps traceability metadata (`maintenanceRequestName`, `maintenanceRequestUID`), emits opening health events via platform-connector, handles deletion with clearing event and annotation removal. Includes conflict-retry on status updates to prevent duplicate publishes.
- **Validating webhook** (`maintenancerequest_webhook.go`) — enforces required fields, `isHealthy=false`, `startTime` in the future, node existence, duplicate-MR rejection, and spec immutability on update (table-driven scalar checks + composite field comparison)
- **Webhook registration** (`webhook.go`) — registers `MaintenanceRequestValidator` alongside existing `ValidationRequestValidator`
- **Wiring** (`main.go`) — `--enable-maintenance-controller` and `--platform-connector-socket` flags, `healthpub.Publisher` creation, controller + webhook registration in `setupControllers`

Part 2 of 3 — depends on [Part 1](https://github.com/NVIDIA/NVSentinel/pull/1733) (API types). Merge Part 1 first, then rebase this branch onto main.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring
- [ ] Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: lifecycle-manager

## Testing

- [x] Tests pass locally 
- [x] No breaking changes

Verified on a live multi-node cluster with the CRD, both webhooks and the lifecycle-manager Deployment applied from this branch.

**Backwards compatibility.** With `--platform-connector-token-path` unset, the publisher dials exactly as before: `{"msg":"Dialing platform-connector","socket":"unix:///var/run/nvsentinel.sock","tokenAuthEnabled":false}`

**Webhooks** (`kubectl apply --dry-run=server`). The mutating webhook stamped `id`, `generatedTimestamp` and `metadata.maintenanceRequestName`. The validating webhook accepted a valid request and rejected all four invalid shapes: `isHealthy: true`, empty `nodeName`, a non-existent node, and a past `startTime`. The non-existent-node case queries the real API server, which a fake client cannot exercise.

**Reconcile path.** One persisted MaintenanceRequest against an already-cordoned GPU node. The finalizer was added, the status condition was written as `HealthEventEmitted=False` / `EmitFailed`, and platform-connector rejected the publish as expected:

```text
rpc error: code = PermissionDenied desc = event 0: caller may only report health
events for node "<lifecycle-manager's own node>", got "<target node>"
(cross-node reporting requires an allowlisted service account token)
```

Without a projected token, platform-connector treats the caller as node-local, so a cluster-scoped Deployment can only report for whichever node the scheduler placed its pod on. This is the case for the two follow-ups: mounting the token in the chart, and allowlisting the ServiceAccount for cross-node reporting. Deletion emits a clearing event that is rejected the same way, but the finalizer is still released, so the object does not stick in `Terminating`.

All test resources and the CRD were removed afterwards; the node returned to its pre-existing cordon with no new taints.

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added support for managing `MaintenanceRequest` resources.
  - Added validation for required fields, health-event data, node availability, scheduling, and immutable settings.
  - Added automatic node claiming and release throughout the maintenance lifecycle.
  - Added health-event publication for maintenance start and completion, with status conditions and retry handling.
  - Added configurable maintenance controller, webhook, and health-publisher integrations.
  - Added conversion support for displaying maintenance conditions in Kubernetes-compatible formats.

- **Bug Fixes**
  - Improved handling of missing or blocked nodes, conflicts, retries, and publication failures.

- **Tests**
  - Added comprehensive coverage for controller, webhook, validation, lifecycle, retry, and failure scenarios.


lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+	}
+
+	if mr.Spec == nil || mr.Spec.HealthEvent == nil {
+		log.Error("MaintenanceRequest has nil spec or healthEvent")
+		return ctrl.Result{}, nil
+	}
+
+	nodeName := mr.Spec.HealthEvent.NodeName
+	if nodeName == "" {
+		log.Error("MaintenanceRequest has empty nodeName")
```
lalitadithya: we should return an error here and set a condition so that the failure is visible to the user

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+func (r *MaintenanceRequestReconciler) initializeFinalizer(
+	ctx context.Context, log *slog.Logger, mr *v1alpha1.MaintenanceRequest,
+) (ctrl.Result, error) {
+	controllerutil.AddFinalizer(mr, mrFinalizerName)
+
+	if err := r.Update(ctx, mr); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	r.setCondition(mr, conditionHealthEventEmitted, "Unknown", "Initializing",
```
lalitadithya: I'm not sure what is the purpose of the initializing phase since we really have any initialization work to do, do you think we can drop this?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+
+	if err := r.emitOpeningEvent(ctx, log, mr); err != nil {
+		r.setCondition(mr, conditionHealthEventEmitted, "False", reasonEmitFailed,
+			fmt.Sprintf("Failed to emit health event: %v", err))
+
+		if statusErr := r.Status().Update(ctx, mr); statusErr != nil {
+			log.Error("Failed to update status after emit failure", "error", statusErr)
+		}
+
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
lalitadithya: we should return error here so that we get backoff for retries. As much as possible we should avoid setting requeue after time

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+) error {
+	r.setCondition(mr, conditionHealthEventEmitted, "True", reasonEmitted,
+		"Submitted health event to platform-connector.")
+
+	statusErr := r.Status().Update(ctx, mr)
+	if statusErr == nil || !apierrors.IsConflict(statusErr) {
+		return statusErr
+	}
+
+	if err := r.Get(ctx, client.ObjectKeyFromObject(mr), mr); err != nil {
```
lalitadithya: why are we updating the same condition twice?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+	controllerutil.RemoveFinalizer(mr, mrFinalizerName)
+
+	if err := r.Update(ctx, mr); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// claimNode atomically checks that no other MR owns the node and writes
```
lalitadithya: doesn't janitor already have a node level locking feature, can't we reuse the same here?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+	}
+
+	return true, nil
+}
+
+func (r *MaintenanceRequestReconciler) autoPopulateEventFields(mr *v1alpha1.MaintenanceRequest) {
+	he := mr.Spec.HealthEvent
+
+	if he.Id == "" {
+		he.Id = fmt.Sprintf("he-mr-%s", mr.UID)
```
lalitadithya: why can't it just be the UUID?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+	}
+
+	if he.Version == 0 {
```
lalitadithya: user has to provide this

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+		he.Version = 1
+	}
+
+	if he.GeneratedTimestamp == nil {
+		he.GeneratedTimestamp = timestamppb.Now()
```
lalitadithya: shouldn't this be the creation timestamp of the CR?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+	if !exists || currentMR != mrName {
+		return nil
+	}
+
+	delete(node.Annotations, managed.AnnotationActiveMR)
+
+	return r.Update(ctx, &node)
+}
+
+func (r *MaintenanceRequestReconciler) setCondition(
```
lalitadithya, line 296: I'm fairly certain there is helper for this in controller runtime that we are using in different parts of the code base, can you please check?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,444 @@
...
+		Type:               condType,
+		Status:             status,
+		Reason:             reason,
+		Message:            message,
+		LastTransitionTime: now,
+		ObservedGeneration: mr.Generation,
+	})
+}
+
+func isConditionTrue(mr *v1alpha1.MaintenanceRequest, condType string) bool {
```
lalitadithya, line 314: I'm fairly certain there is helper for this in controller runtime that we are using in different parts of the code base, can you please check?

lifecycle-manager/pkg/webhook/v1alpha1/maintenancerequest_webhook.go
```diff
@@ -0,0 +1,284 @@
...
+			}
+		}
+	}
+
+	return nil, nil
+}
+
+func (v *MaintenanceRequestValidator) ValidateDelete(_ context.Context,
+	obj *v1alpha1.MaintenanceRequest) (admission.Warnings, error) {
+	mrLog.Info("Validating MaintenanceRequest on delete", "name", obj.Name)
```
lalitadithya: if we don't have any validation for this, can we not register the webhook for delete?

lifecycle-manager/pkg/webhook/v1alpha1/maintenancerequest_webhook.go
```diff
@@ -0,0 +1,284 @@
...
+
+// checkNoDuplicateMR is a best-effort early rejection of a second MR for the
+// same node. It uses a List, so two concurrent creates can both pass; the
+// controller's annotation-based claimNode provides the authoritative guard.
+func (v *MaintenanceRequestValidator) checkNoDuplicateMR(ctx context.Context, nodeName, selfName string) error {
+	if v.Client == nil {
+		return nil
+	}
+
+	var list v1alpha1.MaintenanceRequestList
```
lalitadithya: Lists are generally expensive, but since we already have node level locking do we need this?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: Done

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: Addressed

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: Done

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: The condition was being set twice because there was an r.Get() in between that re-fetched a completely fresh mr from the API server. The first set was on the original in-memory copy, the second was on the re-fetched copy (which didn't have our condition yet). It was a one-shot retry for optimistic concurrency conflicts on the status subresource.

That said, we've simplified this now. Instead of doing an inline retry, we just return the error and let controller-runtime's exponential backoff handle the retry on the next reconcile. Same outcome, less code.

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: We can reuse janitor's Lease-based lock so MR and janitor operations are mutually exclusive on the same node. It's a bit involved though since we'd need to move distributedlock to commons, abstract out the janitor metrics dependency, and adapt the unlock flow for MR's finalizer-based lifecycle (MR doesn't have CompletionTime like janitor's CRDs). Going to tackle this as an immediate follow-up PR to keep this one focused on the core controller logic.

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: Done

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: Done

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: Moved the proto-to-metav1 condition converter from janitor/pkg/condition to commons/pkg/condition so both components share it. `setCondition` now delegates to meta.SetStatusCondition and isConditionTrue uses meta.IsStatusConditionTrue from k8s.io/apimachinery/pkg/api/meta, same pattern janitor uses for ExtRR.

lifecycle-manager/pkg/webhook/v1alpha1/maintenancerequest_webhook.go
cbumb: Removed lists

lifecycle-manager/pkg/webhook/v1alpha1/maintenancerequest_webhook.go
cbumb: Addressed

lifecycle-manager/main.go
```diff
@@ -183,10 +193,45 @@ func setupControllers(mgr ctrl.Manager, cfg *config.Config, enableValidationCont
...
+// maintenance controller is disabled.
+func newPublisher(
+	enabled bool, socketTarget string,
+) (*healthpub.Publisher, error) {
+	if !enabled {
+		return nil, nil
+	}
+
+	conn, err := grpc.NewClient(socketTarget,
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
```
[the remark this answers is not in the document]
cbumb: This matches the existing pattern across all health monitors and components in the repo; hardening should be a repo-wide effort.

lifecycle-manager/internal/controller/maintenancerequest_controller.go
lalitadithya: let's do it in this one please

cbumb: Added

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,326 @@
...
+// SetupWithManager registers the reconciler with the manager.
+func (r *MaintenanceRequestReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.MaintenanceRequest{}).
+		Named("maintenancerequest").
+		Complete(r)
+}
+
+// Reconcile handles create, update, and delete of MaintenanceRequest objects.
+func (r *MaintenanceRequestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
lalitadithya, line 66: I believe we have a start time as well, can we handle that as well?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,326 @@
...
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"time"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
lalitadithya, line 27: can we merge the two imports into one?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
```diff
@@ -0,0 +1,326 @@
...
+			fmt.Sprintf("Node %s is locked by another maintenance operation.", nodeName))
+
+		if statusErr := r.Status().Update(ctx, mr); statusErr != nil {
+			log.Error("Failed to update blocked status", "error", statusErr)
+		}
+
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+	}
+
+	r.autoPopulateEventFields(mr)
```
lalitadithya: I'm not a fan of updating the spec in the controller as it can often lead to reconcile storms, can we instead do this in the webhook?

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: In the current design, startTime is informational for this version. The design document states that it is recorded for observability and future scheduling, while node preparation begins when the MR is created. The webhook only validates that a supplied time is not already in the past.

Are you suggesting that we change the behavior in this MR so the controller waits until startTime before acquiring the lock and emitting the event? I can implement that if scheduling should be included in this version.

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: k8s.io/apimachinery/pkg/api/meta provides `SetStatusCondition` and `IsStatusConditionTrue`, while k8s.io/apimachinery/pkg/apis/meta/v1 provides `Condition` and `ConditionStatus`. The helper package does not re-export the v1 types, so both imports are required. Did you have a different consolidation in mind?

lifecycle-manager/main.go
```diff
@@ -183,10 +194,48 @@ func setupControllers(mgr ctrl.Manager, cfg *config.Config, enableValidationCont
...
 	// +kubebuilder:scaffold:builder
 	return nil
 }
 
+// newPublisher creates a healthpub.Publisher backed by a gRPC connection
+// to the platform-connector socket. Returns (nil, nil) when the
+// maintenance controller is disabled.
+func newPublisher(
+	enabled bool, socketTarget string,
+) (*healthpub.Publisher, error) {
```
jtschelling, line 227: we need to provide a token with pushes to platform-connector now https://github.com/NVIDIA/NVSentinel/pull/1636

PTAL at [this doc](https://github.com/NVIDIA/NVSentinel/blob/main/docs/platform-connectors.md#health-event-node-binding) and the linked pull request, you can set it up the same way the other deployment health-monitors work now.

jtschelling (changes requested):

lifecycle-manager/internal/controller/maintenancerequest_controller.go
cbumb: Addressed

lifecycle-manager/main.go
cbumb: I have added the first piece of flag + token on the dial in this MR and the rest will be the part of next MR

lalitadithya (approved):

PR #1735, opened 2026-09-02
feat(lifecycle-manager): add Helm chart support for MaintenanceRequest [3/3]

cbumb:
## Summary

Helm chart changes to deploy the `MaintenanceRequest` feature introduced in [Part 1](https://github.com/NVIDIA/NVSentinel/pull/1733) and [Part 2](https://github.com/NVIDIA/NVSentinel/pull/1734), plus the platform-connector authentication the feature needs in order to work at all. Gated behind `controllers.maintenanceRequest.enabled` (defaults to `false`, opt-in).

Part 3 of 3. Parts 1 and 2 are merged; this branch is rebased onto current `main`.

### Why authentication is in scope here

A `MaintenanceRequest` may name **any** node, but `lifecycle-manager` is a Deployment running on one node. platform-connector pins a publisher to its own node unless that publisher is on the cross-node allowlist **and** presents a ServiceAccount token minted for the configured audience.

Without both halves the failure is silent: the MR is admitted, its `HealthEventEmitted` condition goes `True`, and the node simply never cordons, with the rejection visible only as a `node_mismatch` counter on platform-connector.

## What's included

**Feature deployment**
- **ClusterRole** — RBAC for `maintenancerequests`, `maintenancerequests/status`, `maintenancerequests/finalizers`, and node annotation access
- **Deployment** — `--enable-maintenance-controller` and `--platform-connector-socket` flags, platform-connector socket volume mount
- **Webhook** — validating webhook configuration for `MaintenanceRequest`
- **Values** — `controllers.maintenanceRequest.enabled`, documented inline

**Platform-connector authentication**
- **Chart-local `pcAuth` helpers** — copied from `kubernetes-object-monitor`, named `lifecycle-manager.*` so the subchart renders standalone without overriding the umbrella's `nvsentinel.*` definitions
- **Deployment** — `--platform-connector-token-path` plus the projected ServiceAccount token volume and mount, nested inside the existing feature gate
- **Umbrella** — derives `system:serviceaccount:<namespace>:lifecycle-manager` into `AuthCrossNodeServiceAccounts`

**Quarantine rule** — fault-quarantine ruleset matching the `maintenanceRequestName` metadata key that the mutating webhook stamps on every MR-originated event

**Tests** — 3 helm-unittest suites (8 tests) and an E2E covering create-to-cordon and delete-to-uncordon

**Docs** — `docs/configuration/lifecycle-manager.md`, cross-referencing [authentication.md](docs/configuration/authentication.md)

## Design notes for reviewers

**The allowlist entry is gated on the subchart's own value**, `lifecycle-manager.controllers.maintenanceRequest.enabled`, read from the umbrella via `.Values["lifecycle-manager"]` — not on a new `global` key. One switch drives both the credential and the grant, so they cannot be enabled independently. A second switch is precisely how you end up with a pod holding a valid token that the connector has never heard of.

**The ServiceAccount name is derived, not written out.** It resolves to a fixed `lifecycle-manager` only because the subchart ships `fullNameOverride`; overriding that or `serviceAccount.name` would otherwise leave the allowlist naming an identity that does not exist. Same class of bug as the hardcoded namespace the derived list already exists to prevent.

**The two halves cannot be asserted in one test.** The umbrella suite runs with `--with-subchart=false`, so a subchart template is not selectable from it. Both suites are keyed on the same value instead, with a comment explaining why they are apart.

**Enabling this rolls platform-connector cluster-wide**, because the allowlist lives in its ConfigMap and that ConfigMap's checksum is stamped on the DaemonSet pod template. True of enabling any cross-node publisher, but worth scheduling deliberately. Documented.

**Backwards compatible.** With `global.platformConnectorAuth.enabled=false`, the rendered Deployment is byte-identical to before this change (verified by diff).

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring
- [ ] Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: lifecycle-manager Helm chart

## Testing

- [x] Tests pass locally

### Automated

| Gate | Result |
|---|---|
| `make helm-lint` | Pass |
| `make helm-test` | Pass — 8 new tests across 3 suites |
| `make license-headers-lint` | Pass |
| `make gomod-lint` | Pass — 33 modules |
| `make -C tests lint-test` | Pass — 0 issues |
| `make -C lifecycle-manager lint-test` | Pass |

The allowlist test was mutation-checked: removing the MaintenanceRequest condition from the gate (the exact mistake someone would make while tidying that entry into the table with the other five) fails `must not allowlist lifecycle-manager when the MR controller is disabled` by name.

### Switch matrix

Rendered for all four combinations. The token's three pieces appear only when both switches are on:

| MaintenanceRequest | Auth | Token in pod |
|---|---|---|
| on | on | flag, mount, volume |
| on | off | none |
| off | on | none |
| off | off | none |

### Live cluster evidence

Local kind cluster, all components built from this branch, `lifecycle-manager` scheduled on `nvsentinel-control-plane3` and the MaintenanceRequest targeting `nvsentinel-worker` — a genuinely cross-node publish.

Deployed wiring:

```
--platform-connector-token-path=/var/run/secrets/nvsentinel/platform-connector/token

{"name":"platform-connector-token","projected":{"sources":[{"serviceAccountToken":
  {"audience":"platform-connector.nvsentinel.nvidia.com","expirationSeconds":3600,"path":"token"}}]}}

"enableNodeBindingAuth": "true"
"AuthCrossNodeServiceAccounts": [... ,"system:serviceaccount:nvsentinel:lifecycle-manager", ...]
```

Controller published both events successfully:

```
Successfully emitted opening health event   node=nvsentinel-worker
Successfully emitted clearing health event  node=nvsentinel-worker
```

platform-connector metrics — the decisive evidence, since these record the authentication decision itself rather than the absence of a failure:

```
platform_connector_auth_decisions_total{decision="cross_node"} 2
platform_connector_auth_node_claim_total{result="verified"} 2
```

Two decisions, matching the opening and clearing events. `cross_node` scope is granted only to a verified identity found on the allowlist; `verified` means the token's node claim was validated rather than merely present. No `platform_connector_auth_violations_total` series exists, which is what you would expect when nothing has violated the rule.

### Rendering evidence against a production-shaped installation

Dry-run of this chart against the live values of `nvs-dgxc-k8s-azr-scus-dev1` (chart `v1.16.0`) produced the correct allowlist from that cluster's real configuration — right namespace, right ServiceAccount names, including the one enabled cross-node monitor and omitting the disabled ones:

```
"AuthCrossNodeServiceAccounts": ["system:serviceaccount:nvsentinel:health-events-analyzer",
                                 "system:serviceaccount:nvsentinel:lifecycle-manager"]
```

That same dry-run caught a real misconfiguration. `helm upgrade --reuse-values` does not apply new chart defaults, so `global.platformConnectorAuth.enabled` arrived as `nil`; the strict boolean check in the `pcAuth` helpers refused to render rather than proceeding with authentication in an undefined state.

### Known gaps

- **The cordon was not observed end to end.** On the kind cluster, fault-quarantine's node informer reports zero nodes and its circuit breaker never initialises, so it consumes no events from any publisher. Unrelated to this change; the publish side is proven by the metrics above.
- **The E2E test has not had a full green run.** Its first assertion (controller accepts the MR and emits the opening event) passes; the cordon assertion is blocked by the above. It will run in CI on the mirrored `pull-request/` branch.
- A full staging deployment was attempted and rolled back: `azr-scus-dev1` runs `v1.16.0`, which predates node-binding auth entirely, so testing there would mean upgrading every component on a shared cluster.

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added opt-in MaintenanceRequest support to Lifecycle Manager, including validation, health-event publishing, node targeting, and platform-connector authentication.
  - Added quarantine rules that cordon affected nodes and restore them after maintenance completes.
  - Added cross-node authorization for maintenance events.
- **Bug Fixes**
  - Validation webhooks now process only create and update operations.
- **Documentation**
  - Added Lifecycle Manager configuration guidance, authentication details, examples, and quarantine behavior.
- **Tests**
  - Added coverage for deployment, permissions, authentication, cross-node authorization, and the end-to-end MaintenanceRequest lifecycle.


distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
```diff
@@ -228,6 +228,37 @@ ruleSets:
...
+    # on null, which would fail evaluation for every metadata-free event any
+    # monitor publishes.
+    #
+    # isHealthy is deliberately not matched. Recovery evaluates the clearing event
+    # against these rulesets too, so excluding healthy events would stop the node
+    # un-cordoning when the requester deletes the MR.
+    match:
+      all:
+        - kind: "HealthEvent"
+          expression: "event.metadata != null && 'maintenanceRequestName' in event.metadata"
```
lalitadithya: we typically start the expression with the agent match, can we do the same here as well?

distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
```diff
@@ -228,6 +228,37 @@ ruleSets:
     cordon:
       shouldCordon: true
 
+  - enabled: true
+    version: "1"
+    name: "MaintenanceRequest ruleset"
+    # A MaintenanceRequest (ADR-051) lets an external system tell NVSentinel that
```
lalitadithya: we don't have to mention the ADR everywhere, we can just describe the feature :)

docs/configuration/lifecycle-manager.md
```diff
@@ -0,0 +1,140 @@
+# Lifecycle Manager Configuration
+
+## Overview
+
+The Lifecycle Manager hosts the controllers that manage a node's transitions in and out of
+service. It runs two independent controllers, each with its own feature gate:
+
+- **ValidationRequest controller** — runs validation tests against a node and keeps it gated
+  until they pass. On by default. See [ADR-045](../designs/049-node-validation.md); its Helm
```
lalitadithya: these are using facing docs, we shouldn't link to ADRs or design documents here. The docs should be written in way such that anyone reading it can understand it without having to go through the design details.

distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
cbumb: We can’t reliably match on agent here because it is supplied by the MaintenanceRequest author and lifecycle-manager publishes it unchanged. 
`maintenanceRequestName` is the stable identifier because the webhook adds it and the controller reasserts it before publishing. That’s why this rule matches the metadata key instead of agent.

lalitadithya: oh, this will result in confusion, can we always overwrite the agent to something static before publishing the event?

cbumb: Sure that makes sense. I will investigate and update the changes

docs/configuration/lifecycle-manager.md
cbumb: Addressed

distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
cbumb: Addressed

distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
cbumb: Added the changes

lalitadithya (approved):

PR #1736, opened 2026-09-03
fix(syslog-health-monitor): refresh GPU metadata mappings

lingyue328:
## Problem

`syslog-health-monitor` caches GPU metadata after its first successful read. When `metadata-collector` atomically replaces the metadata file, later syslog events can still resolve a PCI address with the stale GPU UUID mapping.

## Root cause

The metadata reader only reloaded after an initial load failure or, for the driver version, while the cached value was empty. It did not detect successful metadata file replacements.

## Change

- Detect metadata file replacement or version changes before each lookup.
- Build a complete replacement snapshot before swapping metadata and lookup maps under the reader mutex.
- Keep the last successful snapshot when a refresh cannot be read or parsed.
- Cover PCI-to-UUID refresh, malformed replacements, driver-version refresh, and concurrent lookups.

## Validation

- `go test ./pkg/metadata`
- `go test -race ./pkg/metadata`
- `git diff --check`

All checks pass locally with Go 1.27.0.

## Risk and rollback

The change adds one file metadata check per lookup and serializes refresh and lookup through the existing mutex. A continuously missing or malformed replacement may cause repeated refresh warnings while the reader continues serving its last valid snapshot. Revert this commit to roll back.

## Summary by CodeRabbit

* **Bug Fixes**
  * Metadata lookups now refresh automatically when the metadata file changes.
  * GPU UUID and driver-version information remain available when replacement data is malformed.
  * Metadata recovers automatically after valid data becomes available again.
  * Improved reliability during concurrent lookups and file updates.

* **Tests**
  * Added coverage for file replacement, malformed data recovery, refresh behavior, and concurrent access.

### DIFF
--- a/health-monitors/syslog-health-monitor/pkg/metadata/reader.go
+++ b/health-monitors/syslog-health-monitor/pkg/metadata/reader.go
@@ -17,6 +17,7 @@ package metadata
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 	"strings"
@@ -30,6 +31,7 @@ type Reader struct {
 
 	mu     sync.Mutex
 	loaded bool
+	file   os.FileInfo
 
 	metadata *model.GPUMetadata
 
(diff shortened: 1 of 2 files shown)


XRFXLP:
hey @lingyue328, thanks for taking this up!

Can you talk a bit about the specific situation that surfaced the problem of stale metadata reads? I'm guessing this bug will only surface in either of the two conditions: metadata-collector was restarted but syslog health was restarted prior to metadata-collector or never got restarted at all.

lingyue328:
Yes, that matches the practical failure modes we identified.

  `metadata-collector` currently writes the metadata snapshot once during startup. The stale read can occur when:

  1. `syslog-health-monitor` remains running while `metadata-collector` restarts and replaces the file; or
  2. `syslog-health-monitor` restarts first, reads the existing snapshot, and `metadata-collector` replaces it afterward.

  If `syslog-health-monitor` starts after the new snapshot has been written, it reads the current metadata and the issue does not occur.

  This change detects file replacement before each lookup. It also keeps the reader correct if metadata refresh becomes periodic in the
  future.

XRFXLP:
Does it makes to wire `fsnotify` instead of adding state check on hot path? or a debounced `stat()`?

lingyue328:
Good question. I considered both options, but I would prefer to keep the per-lookup `stat()` here for correctness.

  These metadata lookups are not performed for every syslog line. They occur after a relevant XID, SXID, or reset event has been matched,
  so the check is not on the raw log-ingestion path.

  A debounced `stat()` would introduce a stale window: if the metadata file is replaced and the first relevant event arrives within that
  window, the event could still be enriched from the old PCI-to-UUID mapping, which is the failure this change is intended to prevent.

  `fsnotify` would also add lifecycle complexity because the writer uses atomic rename, requiring a directory watch plus handling missed
  events, queue overflow, and watcher recovery. A fallback `stat()` would still be needed for correctness.

  Given that metadata changes are rare and these lookups only occur for relevant health events, the per-lookup check seems like the
  simpler correctness-first tradeoff. Does that match your concern about the hot path?

XRFXLP (approved):

Issue #1709, opened 2026-09-01
[Feature]: expose change stream consumer lag; a consumer 11 days behind presents as completely healthy


lfriedman-netllama:
### Prerequisites

- [x] I searched existing issues

### Code of Conduct

- [x] I agree to follow NVSentinel's Code of Conduct

### Feature Summary

Expose **change stream consumer lag** as a metric in the shared `store-client`, so every
change stream consumer gets it rather than each module reinventing it. Split out of #1704 at
@lalitadithya's suggestion, since it applies to all consumers and not just
health-events-analyzer.

### Problem/Use Case

On a 288 node GB200 fleet, `health-events-analyzer` fell **11 days and 2.4 million events
behind** its change stream. Nothing surfaced it. The pod presented as completely healthy:

* `Running`, **zero restarts**, uptime 6 days
* **2m CPU against a 1000m limit**, 67Mi against 1Gi, so no throttling or OOM signal
* no errors in logs, and its own success counters climbing normally
* `rule_matched_total` incrementing, so it looked like it was working

It was working. Just on events from 11 days earlier. We only found it because the stale
pattern events it emitted looked like a current hardware fault, and chasing that down led to
decoding the resume token by hand.

**The existing metric does not catch this.** `event-exporter` has
`ResumeTokenUpdateTimestamp` (`event-exporter/pkg/metrics/metrics.go:80`, set at
`worker_pool.go:171`), but it records **when the token was last written**, not **where the
stream position is**. Our analyzer was persisting tokens continuously and successfully the
entire time it was 11 days behind, so a freshness gauge on token writes would have read as
perfectly healthy throughout. The two are independent: a consumer can commit progress every
few seconds while falling further behind every minute.

### Proposed Solution

Emit the **stream position** as a gauge from `store-client`, where the token is already
loaded and stored, so it lands once and covers every consumer:

```
changestream_position_timestamp_seconds{client="health-events-analyzer"}
```

Lag is then `time() - changestream_position_timestamp_seconds`, which alerts cleanly:

```promql
time() - changestream_position_timestamp_seconds > 900
```

For MongoDB this needs no extra query. The resume token's `_data` encodes the BSON timestamp
in its leading bytes, so the position is derivable from the token the consumer already holds:
the seconds are `parseInt(_data[2:10], 16)`. We decoded ours exactly that way to confirm the
2026-08-21 position. The PostgreSQL provider tracks a monotonic event ID rather than a
[...]

PR #1665, opened 2026-08-25
feat: datastore proxy design

KaivalyaMDabhadkar:
## Summary

This adds ADR-052, the design for the deployment platform connector, addressing the connection scaling problem in #1595.

The problem: the platform connector is a DaemonSet and each pod opens about 3 MongoDB connections, all of which land on the primary, so connections grow with the fleet. At 100k nodes that is roughly 300k connections on that one member, pushing its memory limit to 91 GiB just to keep them open, while each pod only ever inserts health events.

The design: run the platform connector as a small central Deployment that health monitors publish to directly over gRPC, each request authenticated with the monitor's own projected ServiceAccount token. It runs the same pipeline, node condition updates and datastore writes as today through a small fixed connection pool, so the per-node DaemonSet is no longer needed and the write path drops to about 50 connections. It is the same binary and image behind a mode flag, and the mode is not the default: it is meant for large fleets, and small clusters lose nothing by staying on the DaemonSet setup.

The design covers:

- Delivery and idempotency: OK means accepted and queued, delivery is bounded best-effort with duplicate suppression via per-event keys under a partial unique index, and the ordering consequences for consumers are stated as an explicit switchover gate.
- Admission control with per-node quotas and typed rejections, replica drain behavior, and three small store-client changes.
- Authentication: still exactly one token validation on the event path, every batch pinned to its caller token's node claim, a cross-node allowlist for the four cluster-wide publishers with their placement enforced, fail-closed behavior, and TLS as an invariant.
- Node condition updates running centrally behind a per-fault ordering guard, and the pipeline (including dedup) running centrally.
- What each monitor needs, with the queue and key logic living in the shared Go and Python publishing clients, and the open product decision for custom and token-less publishers.
- Scaling sized at the 100k-node target, with a load test at the modeled rates as the gate.

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally
- [ ] Manual testing completed
- [x] No breaking changes (or documented)

## Checklist
- [x] Self-review completed
- [x] Documentation updated (if needed)
- [x] Ready for review

## Summary by CodeRabbit

[...]

PR #1738, opened 2026-09-03
docs: ADR-054 change stream consumer lag metrics

lfriedman-netllama:
Design for #1709, which @lalitadithya asked for before implementation. **No code, just the ADR**, so the shape can be agreed first.

## The problem, restated

`health-events-analyzer` was **2.4 million events and ~11 days behind** its change stream on a 288 node fleet, and presented as completely healthy: `Running`, zero restarts, 2m CPU against a 1000m limit, no errors. It was found by manually decoding a resume token out of MongoDB.

## Three things the investigation changed about the ask

**1. The capability is half-present, not missing.** `ChangeStreamMetrics.GetUnprocessedEventCount` already exists (`store-client/pkg/client/interfaces.go:104-106`) and is implemented for **both** providers. Exactly one consumer wired it up:

| Consumer | Stream backlog metric |
| --- | --- |
| `fault-quarantine` | **yes, works** |
| `event-exporter` | **declared, never set** — reports `0` forever |
| `health-events-analyzer` | none. This is the one that went 2.4M behind |
| `fault-remediation`, `janitor`, `lifecycle-manager`, `labeler` | none |

So this is mostly "adopt what is already there, from one place", which is cheaper than I first framed it on the issue.

**2. The obvious signal is the wrong one.** "Resume token age", the natural reading of the issue title, means `now - position_time`. That **reports lag on a caught-up but idle consumer**: no events for an hour looks like an hour of lag. It would alert continuously on a quiet cluster and train operators to ignore the metric.

The ADR proposes `head_time - position_time` instead, which compares the consumer against the stream rather than the wall clock, so an idle stream reports zero. Cost is one indexed lookup, not a scan.

**3. Decoding the Mongo token is not safe to ship.** A resume token's `_data` hex encodes the cluster time, and that is genuinely how I measured the 11 days by hand. But it is a driver implementation detail with no public API, so it would break silently on an upgrade. The ADR instead records the position timestamp when the token is saved, in `MarkProcessed`, which already holds the event and its `clusterTime`. PostgreSQL needs nothing new, since `datastore_changelog.changed_at` already exists.

## What is proposed

```
changestream_lag_seconds{client}      gauge   head time minus position time, 0 when caught up
changestream_events_behind{client}    gauge   count after position, -1 when not collected
changestream_position_known{client}   gauge   1 once a position has been saved
```

Emitted from a single poller in the shared `store-client` watcher, labelled by the existing client name, so all eight consumers get it with no per-module wiring. The expensive `COUNT(*)` is on a longer interval and skippable; the cheap time signal is default-on.

It also proposes deleting `event-exporter`'s permanently-zero gauge, and leaving `fault_quarantine_event_backlog_count` for a release so dashboards can migrate.

## Open questions for review

1. **Metric naming**: `changestream_*` unprefixed, or under an existing namespace? These come from `store-client`, which is a library rather than a component, so it has no natural prefix.
2. **Defaults**: I have `lag_seconds` on and `events_behind` off, on the grounds that the cheap correct signal should always be present and the `COUNT(*)` should be opt-in. Happy to flip it if you would rather both be on.
3. **The `positionTime` field** on the Mongo `ResumeTokens` document is a small schema addition. It degrades gracefully (no lag series until the first save under new code, no backfill needed), but it is worth an explicit yes.
4. **Whether to fold in the dead-gauge fix** or keep it as its own small PR, which I am happy to send separately and first.

Every file:line reference in the ADR was verified against `main` at the time of writing.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for shared change-stream lag and document-backlog metrics across MongoDB and PostgreSQL consumers.
  - Clarified provider-specific position capabilities, head derivation, persisted MongoDB position metadata, and known MongoDB limitations.
  - Documented metric cost controls, testing requirements, rollout alerts, and monitoring considerations.
  - Corrected `fault-quarantine` documentation to reflect its existing backlog metric and unsupported-value behavior.


docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,288 @@
...
+changestream_position_known{client}     gauge   1 once the consumer has saved a position, else 0
+```
+
+`client` is the existing `TokenConfig.ClientName` / `fieldClientName`, so the label space is
+the set of consumers and nothing more.
+
+The `-1` convention for `events_behind` follows the precedent `fault-quarantine` already set
+for "interface unavailable", which is better than a missing series or a misleading zero.
+
+### The event-exporter bug
```
lalitadithya: we don't have to document this here, we can track it separately in the PR you opened :)

lfriedman-netllama: Done, removed. Tracking it in #1739, which is already open.

That also resolves the migration-window question raised further down: I verified before opening #1739 that nothing consumes `health_events_exporter_event_backlog_size` (no code reference, absent from `docs/METRICS.md`, and nothing in our own fleet's dashboards or alerts), so there is no window to preserve. Anything that did consume it was graphing a flat zero.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,288 @@
...
+## Implementation
+
+### Where it lives
+
+A single poller in the shared watcher, started from `Start(ctx)`, rather than the current
+pattern of each consumer running its own ticker. `fault-quarantine`'s loop
+(`fault-quarantine/pkg/eventwatcher/event_watcher.go:787-798`) becomes redundant and can be deleted once the
+shared metric is in place, or kept temporarily for dashboard continuity.
+
+```
```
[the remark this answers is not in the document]
lfriedman-netllama: Fixed, both fences now marked `text`. MD040 count on the file is 0.

For what it's worth MD013 is not actionable here: the repo has no markdownlint config and CI does not run it, so the default 80-column rule fires 68 times on ADR-051 and 44 on ADR-012.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,288 @@
...
+// StreamPosition reports where the consumer is and where the stream head is, so lag can be
+// computed without the caller knowing how a position is represented.
+type StreamPosition struct {
+    // PositionTime is the timestamp of the last event the consumer marked processed.
+    PositionTime time.Time
+    // HeadTime is the timestamp of the newest event in the stream.
+    HeadTime time.Time
+    // HasPosition is false before the consumer has ever marked an event processed, in which
+    // case both timestamps are meaningless and no lag is reported.
+    HasPosition bool
```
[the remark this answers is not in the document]
lfriedman-netllama: Valid, and fixed. `StreamPosition` now carries an opaque `PositionID string`, documented as "the provider's opaque identifier for that position, in the form `GetUnprocessedEventCount` expects".

This was a real hole rather than a documentation gap: without it the shared poller has no position identifier to pass, so `changestream_events_behind` genuinely could not be implemented from the contract as written. It is also the reason no consumer other than `fault-quarantine` ever adopted the count metric, since `fault-quarantine` is the only one that keeps a `LastProcessedObjectIDStore`. Carrying the id alongside the timestamps is what makes the capability shared rather than per-module, and keeping it opaque avoids committing the shared layer to a position representation.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,288 @@
...
+}
+
+// ChangeStreamMetrics gains:
+StreamPosition(ctx context.Context) (StreamPosition, error)
+```
```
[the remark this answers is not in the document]
lfriedman-netllama: Correct, and this was the most useful finding on the PR. Fixed by adding a separate `ChangeStreamPosition` interface and leaving `ChangeStreamMetrics` exactly as it is.

The reasoning is now stated in the ADR: these capabilities are satisfied by whole-type assertion, so adding a method to `ChangeStreamMetrics` means an implementation providing only `GetUnprocessedEventCount` stops satisfying it and loses the count metric it already had. My previous "degrades to no lag reported" mitigation was wrong for exactly that reason and has been rewritten. There is now a test case for the direction that would have regressed: a watcher implementing only `ChangeStreamMetrics` must keep reporting the count and emit no lag series.

Verifying this turned up something bigger, which I have filed as #1740. The MongoDB watcher already does not satisfy `ChangeStreamMetrics`, because its method takes a `bson.ObjectID` where the interface wants a `string`. The resume-control wrapper does satisfy it, so `fault-quarantine`'s assertion succeeds, the forwarding assertion at `resume_token.go:79` fails, and the error is swallowed at `Debug` level, leaving the gauge unwritten and reporting 0 on every MongoDB deployment. The `-1` fallback is unreachable. The ADR's adoption table said that metric "works"; it works on PostgreSQL only, and that is corrected.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,288 @@
...
+### Per-provider derivation
+
+**PostgreSQL** is straightforward, because the changelog already carries a timestamp
+(`datastore_changelog.changed_at TIMESTAMPTZ`, `store-client/pkg/datastore/providers/postgresql/datastore.go:382`) and the
+position is a row id:
+
+```sql
+SELECT
+  (SELECT changed_at FROM datastore_changelog WHERE id = $1)              AS position_time,
+  (SELECT MAX(changed_at) FROM datastore_changelog WHERE table_name = $2) AS head_time;
```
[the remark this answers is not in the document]
lfriedman-netllama: Confirmed and fixed. The ADR now specifies `CREATE INDEX IF NOT EXISTS idx_changelog_table_changed_at ON datastore_changelog(table_name, changed_at)` as step 1 of the rollout, ahead of anything that queries it.

Checked all three existing indexes at `datastore.go:428-434`: `idx_changelog_unprocessed` on `(changed_at)` and `idx_changelog_resume` on `(table_name, changed_at, id)` are both partial on `processed = FALSE`, and `idx_changelog_table_record` has no timestamp column. The head lookup deliberately has no `processed` predicate, because the newest event in the stream is usually one the consumer has already processed, so none of them serve it.

This also corrected a cost claim: I had described `lag_seconds` as "two indexed lookups", which was only true given an index that does not exist. Without it the head lookup is a sequential scan of the whole changelog on every poll, making the cheap signal the expensive one. There is now a test asserting the query plan uses the new index.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,288 @@
...
+```
+
+**MongoDB** needs a position timestamp that the token does not readily give up. Two options,
+and the choice matters:
+
+- **Preferred: record the position time when the token is saved.** `MarkProcessed`
+  (`store-client/pkg/datastore/providers/mongodb/watcher/watch_store.go:491`) already holds the change event, whose `clusterTime` is the position
+  time. Adding a `positionTime` field to the `ResumeTokens` document makes the timestamp a
+  first-class stored value rather than something inferred. Head time is
+  `db.<collection>.find().sort({_id: -1}).limit(1)`, which uses the existing `_id` index.
```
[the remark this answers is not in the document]
lfriedman-netllama: Right, and this is not theoretical in this codebase, which I had missed. Fixed.

Health events are updated in place after insertion: `UpdateHealthEventStatus`, `UpdateNodeQuarantineStatus`, `UpdatePodEvictionStatus`, `UpdateRemediationStatus` and `UpdateSpanID` at `store-client/pkg/datastore/providers/mongodb/health_store.go:53-254`. The watcher also opens the stream with `SetFullDocument(options.UpdateLookup)` at `watch_store.go:168`, precisely because update events are expected. So an update to a week-old document advances the stream without advancing `max(_id)`, and a consumer behind on those events would have reported near-zero lag.

The design now takes the **later of two head candidates**: `max(_id)` as a timestamp, and a cursor high-water mark of the greatest `clusterTime` the watcher has decoded. Each is insufficient alone, and the reason matters. The high-water mark alone fails when the watcher's own read loop is blocked, since head then freezes at the position and reports zero lag for a badly-behind consumer, which is the "derive lag from arriving events" trap this ADR rejects elsewhere. `max(_id)` alone fails on update-only traffic. Their maximum is correct whenever either mechanism is live, and both are stale only when the collection is genuinely idle, which is when zero is the right answer. Cluster `operationTime` from `hello` was considered as a third candidate and rejected, since it advances on activity anywhere in the deployment and would report growing lag on an idle collection.

Verifying this also caught an error of my own: I had written that `MarkProcessed` holds the change event. It does not, it receives only `token []byte`. `processNextEvent` at `watch_store.go:434-438` is where the event is decoded into a `bson.M`, so that is where `clusterTime` and `documentKey._id` have to be captured. Tests added for update-oldest-document, delete, non-monotonic `_id`, and the blocked-cursor case.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,288 @@
...
+The `-1` convention for `events_behind` follows the precedent `fault-quarantine` already set
+for "interface unavailable", which is better than a missing series or a misleading zero.
+
+### The event-exporter bug
+
+`health_events_exporter_event_backlog_size` is registered and never written, so it currently
+advertises "no backlog" permanently. That is worse than having no metric, and it is
+independent of the rest of this work. It should either be wired to the shared metric or
+removed. Recommend removing it in favour of `changestream_events_behind{client="event-exporter"}`
+rather than maintaining two names for one thing.
```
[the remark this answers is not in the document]
lfriedman-netllama: This one is resolved by removing the section rather than by adding a migration window, per the maintainer's comment above: the gauge is now tracked in #1739.

On the substance, I did verify consumers before opening that PR, which is what the finding asks for. `health_events_exporter_event_backlog_size` has no code reference outside its declaration, does not appear in `docs/METRICS.md` (event-exporter has no section there at all), and nothing in our own fleet's dashboards or alerts references it. Since it was registered and never written, it reported a flat 0 for its entire existence, so a deprecated alias would only preserve a false signal. The asymmetry with `fault_quarantine_event_backlog_count` is deliberate: that one has real consumers, which is why it keeps its release of overlap.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,288 @@
...
+- Adds a field to the MongoDB `ResumeTokens` document, so a consumer running new code against
+  tokens written by old code has no `positionTime` until it next saves one.
+- Adds a periodic query per consumer, small but not free.
+- `ChangeStreamMetrics` grows a method, so any out-of-tree implementation must add it.
+
+### Mitigations
+
+- **Missing `positionTime` is handled explicitly**: `HasPosition: false`, no lag series, and
+  `changestream_position_known` goes to 1 as soon as the first token is saved under new code.
+  No migration step, no backfill.
```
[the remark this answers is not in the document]
lfriedman-netllama: Fair, and the ADR now says so directly instead of claiming coverage. There is a new section, "The rollout gap for existing tokens, stated plainly", which splits the two cases:

- A consumer that is **behind but still working** marks events processed continuously by definition, so it populates a position within one event and becomes visible almost immediately. That is the #1709 shape, and it is covered.
- A consumer that is **fully stalled** never marks anything processed, so it never populates a position and emits no lag series at all. It stays invisible to `changestream_lag_seconds` indefinitely.

No backfill fixes the second case, because the only place the old position's timestamp exists is inside the opaque resume token, and decoding that is rejected earlier in the ADR. `changestream_events_behind` does not close it either, since `PositionID` is written by the same operation as `positionTime`.

So the mitigation is a metric rather than a migration: **`changestream_position_known == 0` persisting for more than a few poll intervals is itself the alert**, because that condition is exactly "stalled, or brand new". It is now called out as alertable in the rollout step alongside the lag threshold, and the lag alert is qualified by `changestream_position_known == 1" so a zero reading is only ever read as "caught up" when the position is actually known.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,432 @@
...
+`ChangeStreamMetrics` is left exactly as it is.
+
+Lag is then `HeadTime.Sub(PositionTime)`, clamped at zero, computed once in the shared layer.
+
+`PositionID` is what lets the shared poller emit the count signal too. `GetUnprocessedEventCount`
+takes a position identifier, and the poller has no other way to obtain one: only
+`fault-quarantine` keeps a `LastProcessedObjectIDStore`, which is precisely why no other consumer
+could adopt the count metric. Carrying the id alongside the timestamps is what makes the
+capability shared rather than per-module. Its shape is provider-defined and opaque to the
+poller, so this does not commit the shared layer to a position representation.
```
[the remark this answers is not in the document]
lfriedman-netllama: Valid, and fixed in `73817354` by renaming the metric to **`changestream_documents_behind`**.

You are right that it cannot be an event count on MongoDB. The existing implementation is `CountDocuments({_id: {$gt: lastProcessedID}})` at `watch_store.go:563-574`, which counts documents newer than the position and therefore misses every in-place update and delete. Same root cause as the `max(_id)` head problem: a consumer 500 status updates behind on old documents would count zero.

Of the two options you offered I took the second, redefining rather than reaching for the resume token, because a true event backlog on MongoDB means scanning the oplog, which needs privileges the application user should not require.

Worth noting the asymmetry that made naming this awkward: on PostgreSQL `datastore_changelog` holds one row per change, so counting rows after the position *is* the event count and the metric is exact there. Rather than give one metric name two meanings per provider, the name now states the weaker guarantee that holds on both, and the docs say where it is exact. There is a test each way: on MongoDB updating an old document must *not* move the count, on PostgreSQL the same scenario must.

This also reinforces why the count is the secondary, opt-in signal and `changestream_lag_seconds` is the default-on one: lag is the signal that sees update and delete traffic.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,432 @@
...
+
+Non-partial, so the `MAX` is an index-only backward scan. Verify with `EXPLAIN` rather than
+assuming; without it the head lookup is a sequential scan of the whole changelog on every poll,
+which would make the "cheap" signal the expensive one.
+
+**MongoDB** needs a position timestamp that the token does not readily give up. Two options,
+and the choice matters:
+
+- **Preferred: record the position time when the token is saved.** Adding `positionTime` and
+  `positionId` fields to the `ResumeTokens` document makes both a first-class stored value
```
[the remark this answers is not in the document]
lfriedman-netllama: Fair hit, that was a hand-wave in my text: "must be carried alongside the per-event token" without saying how. Specified in `73817354`, both halves.

**Travel.** The handoff is only `EventWithToken{Event, ResumeToken []byte}` (`store-client/pkg/datastore/types.go:106-109`), so there genuinely was no route. It now gains the metadata additively, so existing readers are unaffected; `MarkProcessed` keeps its signature and gains a sibling taking the metadata explicitly, so consumers migrate rather than break. A consumer still calling the old method falls back to an in-flight lookup keyed by token, and **on a miss writes the token with no position**, leaving `HasPosition: false`. A missing lag series is recoverable; a wrong one is not.

**Atomicity.** This turned out to come free, and the ADR now records why rather than asserting it. `MarkProcessed` already persists through a single upsert on the client's own document:

```go
// watch_store.go:537-541
w.resumeTokenCol.UpdateOne(
    ctx,
    bson.M{fieldClientName: w.clientName},
    bson.M{"$set": bson.M{"resumeToken": resumeTokenToStore}},
    options.UpdateOne().SetUpsert(true),
)
```

Adding `positionTime` and `positionId` to that same `$set` makes all three fields one single-document write, which MongoDB guarantees atomically. So there is no window where the stored token and stored position describe different events, and nothing to reconcile after a crash. `TokenDoc` (`watch_store.go:93-95`) gains the two fields to read them back.

Restart acknowledgement test added as you asked: process events, restart mid-stream, assert the resumed position time matches the token actually stored rather than the newest event seen.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,432 @@
...
+   head observation, and distinct from position time, because the cursor reads ahead of the
+   consumer rather than in lockstep with it.
+
+Neither alone is sufficient, and the reason is worth stating because it is the same failure this
+ADR rejects elsewhere. Candidate 2 alone fails when the watcher's own read loop is blocked,
+since `processNextEvent` stops advancing and head freezes at the position, reporting zero lag
+for a consumer that is badly behind. That is the "derive lag from arriving events" trap under a
+different name. Candidate 1 alone fails on update-only traffic. Their maximum is correct
+whenever either mechanism is live, and both are wrong only when the collection is genuinely
+idle, which is when zero is the right answer anyway.
```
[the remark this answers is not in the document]
lfriedman-netllama: Correct, and I had claimed this away too easily. The two failure conditions intersect and I wrote "correct whenever either mechanism is live" without following that through: a blocked read loop **and** update-or-delete-only traffic leaves candidate 2 frozen and candidate 1 unmoved, so head goes stale and lag reads near zero. Worse than a corner case, because a blocked read loop is exactly the state this ADR exists to detect.

Fixed in `73817354` by narrowing the guarantee to what is true, rather than widening the mechanism:

> On MongoDB, `changestream_lag_seconds` detects a consumer that is behind whenever the collection is receiving inserts, or whenever the watcher's read loop is live. It does **not** detect a stalled consumer on a collection receiving only updates and deletes.

Then closing it with the cheaper of the two available options, a **`changestream_reads_total{client}`** counter incremented in `processNextEvent`. A flat counter means the read loop is not advancing, so the blocked case is caught directly rather than inferred from lag. That is also the honest use of position age, which this ADR rejects as a lag metric: a stall detector qualified by liveness.

The exact head, a maintained `updatedAt` making `MAX(updatedAt)` exact for inserts and updates, is named as deliberately out of scope, since it is a change to every writer rather than to the watcher.

Two consequences elsewhere: the rollout now documents **three** alerts rather than one, because no single series covers every way a consumer goes silent, and the update-only blocked-cursor case gets a test that asserts the **known-false** behaviour on purpose. Pinning the gap is what stops someone later widening the guarantee in the docs to match an assumption.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,542 @@
...
+does not say how it travels will either lose it or, worse, pair a position time with the wrong
+token. Both halves need stating:
+
+**Travel.** `EventWithToken` gains the position metadata, additively, so existing readers are
+unaffected. `MarkProcessed(ctx, token)` keeps its signature and gains a sibling that takes the
+metadata explicitly, letting consumers migrate rather than break. For a consumer that still
+calls the old method, the watcher falls back to an in-flight lookup keyed by token, covering
+only events between the cursor and the consumer's position. On a miss it writes the token with
+**no** position and leaves `HasPosition: false`, because a missing lag series is recoverable and
+a wrong one is not.
```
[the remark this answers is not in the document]
lfriedman-netllama: Correct, and it contradicted the atomicity argument sitting two paragraphs below it. Fixed in `7ebd5fab`.

"No position" now has to mean **actively cleared** rather than merely unwritten: the fallback path `$unset`s `positionTime` and `positionId` in the same update that writes the token. One write, three fields, and no combination a reader can mistake for a valid position.

The failure you describe is precisely the one that section exists to prevent, so leaving the miss path as a token-only `$set` would have shipped it through the back door: the document keeps a position from the previous token, and a restart then reports a position belonging to a different event.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,542 @@
...
+> On MongoDB, `changestream_lag_seconds` detects a consumer that is behind whenever the
+> collection is receiving inserts, or whenever the watcher's read loop is live. It does **not**
+> detect a stalled consumer on a collection receiving only updates and deletes.
+
+Two things close it, and both are cheaper than an exact head:
+
+- **A watcher liveness signal.** A `changestream_reads_total{client}` counter incremented in
+  `processNextEvent` makes "the read loop is not advancing" directly observable, so the blocked
+  case is caught by a stalled counter rather than inferred from lag. This is the honest way to
+  use position age: not as lag, but as a stall detector qualified by liveness.
```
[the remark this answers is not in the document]
lfriedman-netllama: You are right, and this is the better catch of the two. Dropped the counter entirely in `7ebd5fab` rather than qualifying it.

The counter is flat on a healthy idle stream and flat on a blocked read loop, so `rate(...) == 0` fires on a quiet cluster. That is the position-age mistake this ADR spends its Context section rejecting, reintroduced by me one section later under a different name.

I also checked whether a different granularity would rescue it, and it does not. The loop blocks inside `w.changeStream.Next(ctx)` (`store-client/pkg/datastore/providers/mongodb/watcher/watch_store.go:411-431`), which does not return until an event arrives or the context is cancelled, so a per-iteration heartbeat sits frozen on an idle stream too and inherits the same ambiguity. A genuine heartbeat needs the loop restructured around `TryNext` with its own tick, which is a change to the read path's shape rather than an observability addition.

What replaces it is an observation that was already implicit and is now explicit: **a blocked read loop is already covered for insert traffic**, because `max(_id)` is queried against the collection independently of the cursor, so head keeps advancing while the loop is stuck and lag rises. Distinguishing blocked from idle is exactly the head-versus-position comparison the ADR is built on.

The residual gap, a blocked loop on **update-only** traffic, stays open and named, with the two ways to close it recorded as future work rather than smuggled in: an exact head via a maintained `updatedAt`, or a real `TryNext` heartbeat.

Consequences: the metric set is back to three, with a note that every candidate fourth had a flat or zero value with two meanings, which is the defect this ADR removes rather than adds; the alert list is back to two; and there is a new test requiring idle and blocked to be **distinguishable on the shipped metrics**, which is the test that would have caught the counter.

lfriedman-netllama:
Correction, and it affects things I said earlier in this review.

I claimed in `38eefb81` that `fault_quarantine_event_backlog_count` is permanently 0 on MongoDB, added a section on it, and changed the adoption table to say the metric is broken. **That was wrong.** Retracted in `1afb1442`, and #1740 is closed as invalid.

The error: I verified that the MongoDB watcher in `providers/mongodb/watcher` does not satisfy `ChangeStreamMetrics`, which is true, then concluded the runtime assertion must fail. That does not follow, because the raw watcher is never the type asserted on. `MongoDBClient.NewChangeStreamWatcher` returns `mongoChangeStreamWatcher` (`store-client/pkg/client/mongodb_client.go:954`), an adapter that takes a `string`, converts with `bson.ObjectIDFromHex`, and forwards to the raw watcher (`mongodb_client.go:1061-1073`). It satisfies the interface, so the forwarding assertion at `resume_token.go:79` succeeds and the gauge is set normally. That adapter is precisely what bridges the signature mismatch I had found.

Confirmed by asserting against the chain's real types instead of reading signatures:

```
*mongoChangeStreamWatcher satisfies ChangeStreamMetrics: true
*resumeControlChangeStreamWatcher satisfies ChangeStreamMetrics: true
forwarding assertion at resume_token.go:79 succeeds: true
```

What this changes in the ADR:

- The adoption table says **yes, works** for `fault-quarantine` again, qualified only by the document-versus-event count point, which does hold and is documented in the count-semantics section.
- The `-1` convention paragraph no longer claims that branch has never fired.
- "Backlog visibility on MongoDB, which today has none" is replaced by the accurate benefit: six consumers with no visibility today get both signals.
- The separate-interface decision is **unchanged**, but its rationale no longer leans on a broken metric. It stands on its own and is arguably stronger this way: `ChangeStreamMetrics` is satisfied by whole-type assertion, so adding a method to it would break `fault-quarantine`'s **working** metric on both providers until every implementation caught up. @coderabbitai's original finding was right for the right reason, and my elaboration on it was wrong.

Everything else from the three review rounds stands: the count rename, the metadata handoff and `$unset` on the fallback miss, the narrowed MongoDB head guarantee, the dropped heartbeat counter, the PostgreSQL index, and the rollout gap for existing tokens.

The lesson I am taking from it, since it is the second time this review has turned on this: a compile-time interface assertion is evidence only about the type you assert on, and in a chain with adapters that is often not the type implementing the method.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,570 @@
...
+
+> On MongoDB, `changestream_lag_seconds` detects a consumer that is behind whenever the
+> collection is receiving inserts, or whenever the watcher's read loop is live. It does **not**
+> detect a stalled consumer on a collection receiving only updates and deletes.
+
+**A blocked read loop is already covered, for insert traffic.** No new metric is needed for it:
+`max(_id)` is queried against the collection and is therefore independent of the cursor, so head
+keeps advancing while the loop is stuck and `changestream_lag_seconds` rises. Distinguishing
+"blocked" from "idle" is exactly the head-versus-position comparison this ADR is built on, and it
+already works whenever inserts are flowing.
```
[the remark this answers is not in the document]
lfriedman-netllama: Valid, and it holds more strongly than the finding states. Named in `9734afc8`.

Health events carry **driver-generated** ObjectIDs, with no explicit `_id` set on insert, and an ObjectID is a 4-byte timestamp followed by a per-process random value and a counter. So across processes **within the same second** the ordering is effectively random rather than chronological, and an insert can land below the current maximum and leave head unmoved. Beyond a one-second window the timestamp dominates and ordering holds, modulo writer clock skew. That is a property of ObjectIDs generally, not only of a hand-supplied `_id`.

The practical magnitude is small: bounded by roughly that one-second window, against a 60s poll and a 900s alert threshold. But the guarantee is now stated as "head advances within about a second of an insert" rather than "head advances on every insert", which is the honest version, and the alert section notes the threshold must stay well clear of that window.

Worth adding that the assumption is **inherited, not introduced**: the existing count metric filters `_id > lastProcessedID` (`store-client/pkg/datastore/providers/mongodb/watcher/watch_store.go:563`), so it has always depended on the same ordering.

Taking your suggestion literally rather than only documenting it: the blocked-cursor test now runs twice, once with ascending `_id`s and once with a non-monotonic `_id` inserted below the current maximum, so the assumption is exercised rather than trusted.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,570 @@
...
+   poller, default on for `lag_seconds` and off for `documents_behind`.
+3. Leave `fault_quarantine_event_backlog_count` and its ticker in place for one release so
+   dashboards can move, then remove the ticker in favour of the shared metric.
+4. Document in `docs/METRICS.md`, as two alerts rather than one, because neither covers the
+   other's blind spot:
+   - `changestream_lag_seconds > 900` for 10m, tuned per fleet, qualified by
+     `changestream_position_known == 1`. The primary "behind" alert, and the one that catches a
+     blocked read loop whenever inserts are flowing.
+   - `changestream_position_known == 0` persisting. Covers a consumer stalled since before
+     rollout, which no lag series can see.
```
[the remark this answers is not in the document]
lfriedman-netllama: Agreed, fixed in `9734afc8`. "Persisting" was doing too much work without a number behind it.

The alert is now `changestream_position_known == 0` for longer than a startup grace period, suggested at **3 poll intervals or 10 minutes, whichever is longer**, and it is explicitly named a **"stalled or uninitialized"** alert rather than a stall alert. As written before it would have paged on every rollout, since a freshly started consumer, and one that has never processed an event on a quiet stream, both sit legitimately at zero.

Naming it for both states rather than trying to separate them is deliberate: there is no cheap way to tell "new" from "stalled" at this layer, and "uninitialized for ten minutes on a busy cluster" is itself worth looking at. Better an honestly-named alert covering two conditions than a precisely-named one that misses the case this ADR exists for.

KaivalyaMDabhadkar:
@lfriedman-netllama Thanks for all the work on this. I have one bigger concern I wanted to raise before any code gets written.

I think the head minus position idea will still give false alarms for most of our consumers. The position only moves forward when a consumer handles an event that its filter lets through. node drainer and fault remediation only care about a few status updates, so on a healthy cluster their position can stay days in the past while new health events keep getting inserted. They would look days behind even though they are fully caught up. That is the same false alarm the ADR is trying to avoid.

There may be a simpler way. Every event already carries the time it happened on the server. And when the stream returns an empty batch, that tells us the consumer is caught up at that moment. So if the mongo read loop used TryNext instead of Next, and the postgres poller took note when a poll came back empty, the watcher could remember two things: when it last saw an empty batch, and the time of the last event it read. The Lag is then the current time minus whichever of those two is more recent, calculated fresh each time.
This seems to cover all the cases. A quiet stream reads zero. A slow consumer shows how far behind it really is. A stuck consumer shows lag that keeps growing once there is something waiting for it. And it works the same whether the consumer watches inserts or updates. It also needs no new fields, no head query, no new index, and nothing to backfill, so the rollout gap goes away. TryNext should not add any cost, since Next already makes the same getMore calls internally.

Would you be open to reshaping the ADR around this?

lfriedman-netllama:
@KaivalyaMDabhadkar you are right, and this is a bigger hole than the one closed two rounds ago. Thank you for catching it before implementation.

I checked the pipelines rather than reasoning about it, and it is worse than "can stay days in the past":

- `node-drainer` watches `BuildNodeQuarantineStatusPipeline` (`store-client/pkg/client/mongodb_pipeline_builder.go:44-71`), which matches **only** `operationType: update` where `nodeQuarantinedStatus` was set to Quarantined, AlreadyQuarantined, UnQuarantined or Cancelled.
- `fault-remediation` watches `BuildQuarantinedAndDrainedNodesPipeline`, similarly narrow.
- `fault-quarantine` watches `BuildAllHealthEventInsertsPipeline`, unfiltered, which is why it is the one consumer the head-minus-position idea happens to suit.

On our own fleet `global.faultQuarantine.enabled` is `false`, so **no node is ever quarantined and those two pipelines match zero events, ever**. Their positions never move off the initial token. `head - position` would report "time since deployment" as lag, permanently, on two completely healthy consumers. Not an edge case, the normal state for a detection-only deployment.

So the primary metric was wrong for most consumers, and wrong in exactly the way this ADR's Context section rejects position age for.

**Your alternative is better on every case I can construct**, and the reason is the part I missed: an empty batch is *affirmative evidence of being caught up*, evaluated against the consumer's **own filtered stream**. That is the property `head - position` cannot have, because head is global while position is filtered. Working through it:

| Scenario | `now - max(last_empty, last_event_read)` |
|---|---|
| Quiet stream, caught up | empty batches keep arriving → ~0 ✓ |
| Genuinely behind | reading old events, no empties → age of last event read ✓ |
| Narrow filter, caught up | empty batches → ~0 ✓ (the case that breaks mine) |
| Read loop blocked | neither advances → grows ✓ |
| Consumer slow, loop blocked on channel send | no reads, no empties → grows ✓ |

It also deletes much of what the draft had accumulated: no head-time query, so no `max(_id)` ordering assumption, no new PostgreSQL index, no update-and-delete blind spot, and `documents_behind` goes away along with its document-versus-event caveat. Strictly simpler and strictly more correct.

Three things the rewrite should be explicit about, none of them objections:

1. **The state is in-memory, so a restart resets it.** Between start and the first empty batch or first event read, lag is genuinely unknown, and it must report unknown rather than 0 or a huge number. This is what `changestream_position_known` was for in the current draft, and it survives into this design for a different reason.
2. **`TryNext` needs its own pacing.** It returns immediately, so the loop needs a tick or backoff or it busy-polls a core on an idle stream. This is the read-path restructuring the draft listed as out of scope; this design moves it in scope, and it is the main implementation cost. Worth measuring idle CPU before and after.
3. **It measures the watcher's progress, not the durable position.** A consumer that processes an event and dies before `MarkProcessed` looks fine by this metric. That is the right trade for a lag signal, but it means this is not also a resume-token correctness check, and the ADR should say so rather than implying broader coverage.

The PostgreSQL half is nearly free, since an empty poll is already a natural signal there; the MongoDB `TryNext` change is the real work.

I will rewrite the ADR around this and drop the head-time machinery, crediting the approach to you. Anything above that you would like framed differently, say so and I will fold it in.

lfriedman-netllama:
Rewritten around @KaivalyaMDabhadkar's design in `c75a8e35`. Net **-216 lines**, because the new approach removes more than it adds.

Gone: the `StreamPosition` interface, `positionTime` and `positionId` on the `ResumeTokens` document, the new non-partial PostgreSQL index, the `max(_id)` head estimate with its ObjectID ordering assumption and update-only blind spot, the cursor high-water mark, and `documents_behind` with its document-versus-event caveat. Two metrics remain: `changestream_lag_seconds` and `changestream_lag_known`.

The Context section now leads with the fact that drove the redesign, that every consumer watches a differently-filtered stream, with the three pipelines tabulated, since that is what makes or breaks any candidate signal. Head minus position is recorded under Alternatives with the specific reason it fails, so the next reader does not retrace it, and "narrow pipeline, caught up" is now a permanent test case described as the one that fails under the old design.

Costs are stated rather than buried: the `TryNext` change touches the read loop's hot path and has an idle-CPU measurement as a merge gate; lag is left **absent** rather than zero until the first read or empty batch; and the ADR says explicitly that this measures watcher progress rather than durable position, so it is not also a resume-token correctness check.

All eight file:line citations re-verified, and all three pipeline builders confirmed to exist. MD040 clean.

@KaivalyaMDabhadkar, the design is credited to you in the Decision section. Worth a look at whether the pacing discussion and the unknown-state handling match what you had in mind, since those are the two places I made choices you did not specify.

docs/designs/054-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+4. **time since the last evidence of being caught up** — the chosen signal, below.
+
+What all three rejected signals lack is any **affirmative evidence** that a consumer is caught
+up. They infer it from position, which is filtered, or from head, which is not the consumer's
+business. The stream offers direct evidence instead: **an empty batch means "there is nothing
+further for you right now"**, evaluated server-side against that consumer's own pipeline.
+
+## Decision
+
+Emit lag from the **shared `store-client` watcher layer**, labelled by the existing client name,
```
KaivalyaMDabhadkar, line 82: fault-remediation only exposes controller-runtime's metrics registry, while store-client registers its metrics on the default registry. As written, this gauge would never appear on fault-remediation's metrics endpoint. Could you add a short section explaining how the gauge reaches fault-remediation, for example by letting store-client register into a caller-supplied registry?

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+} else if err := w.changeStream.Err(); err != nil {
+    w.handleChangeStreamError(err)
+    return
+} else {
+    w.recordCaughtUp()                        // lastEmptyBatchAt
+    // pace the loop; see below
+}
+```
+
+**`TryNext` needs pacing.** It returns immediately, so a bare loop busy-polls a core whenever the
```
KaivalyaMDabhadkar: I believe TryNext does not need pacing. It issues a getMore, and the server holds that getMore for its await window, about one second unless MaxAwaitTime is set. Next makes the same calls in an internal loop, so idle cost should stay the same. Could you replace this paragraph with a note that the loop sets MaxAwaitTime explicitly, and remove the sleep, the idle CPU test and the merge gate?

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+### Every consumer watches a different, filtered stream
+
+This is the fact that determines the whole design, so it comes before the candidate signals.
+
+Each consumer opens its change stream with **its own aggregation pipeline**, and most of them
+are narrow:
+
+| Consumer | Pipeline | Admits |
+| --- | --- | --- |
+| `fault-quarantine` | `BuildAllHealthEventInsertsPipeline` | every health event insert |
```
KaivalyaMDabhadkar: fault-quarantine uses the "processable inserts" pipeline, which filters on processing strategy. The unfiltered pipeline belongs to event-exporter. Could you correct this row, add rows for health-events-analyzer and event-exporter, and update the sentence in Alternatives that says only fault-quarantine is unfiltered?

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+
+### Current adoption
+
+| Consumer | Stream backlog metric | Notes |
+| --- | --- | --- |
+| `fault-quarantine` | **yes, works** | ticker at `fault-quarantine/pkg/eventwatcher/event_watcher.go:787-798` sets `fault_quarantine_event_backlog_count` |
+| `event-exporter` | **declared, never set** | `health_events_exporter_event_backlog_size` is created at `event-exporter/pkg/metrics/metrics.go:73` and assigned nowhere in the module, so it reports `0` forever. Removal in #1739 |
+| `health-events-analyzer` | none | no reference to `ChangeStreamMetrics` anywhere in the module. This is the consumer that went 2.4M events behind |
+| `fault-remediation` | none | |
+| `janitor` | none | |
```
KaivalyaMDabhadkar: janitor, lifecycle-manager and labeler do not read a change stream today, and node-drainer does but is not listed here. Could you remove those three rows, add node-drainer, and update the "seven consumers" and "eighth" wording in the Rationale to match five consumers?

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+- **Restart**: assert `changestream_lag_seconds` is **absent** and `changestream_lag_known` is
+  `0` before the first read or empty batch, and that both become live immediately afterwards.
+  Assert lag is never reported as `0` while unknown.
+- **Idle CPU**: measure the watcher's CPU on an idle stream before and after the `TryNext`
+  change, and assert the pacing interval bounds it. The one performance-sensitive part of this
+  design deserves a number rather than a claim.
+- **Both providers**: every case above against MongoDB and PostgreSQL, since the empty-batch
+  signal is derived differently in each.
+- **Integration**: with `envtest` plus a real MongoDB, stop a consumer, insert events, restart
+  it, and assert lag rises and then returns to near zero as it drains. This reproduces the #1704
```
KaivalyaMDabhadkar: When a stored token is too old for the oplog, the watcher deletes it and reopens from now, so lag reads near zero exactly when the most data was skipped. Could you add a sentence saying that the existing recoveries counter covers this case and should be read together with the lag metric?

docs/designs/054-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+so every consumer gets it without per-module wiring.
+
+The watcher tracks two timestamps in memory:
+
+- **`lastEmptyBatchAt`** — when the stream last returned no events, meaning the consumer was
```
KaivalyaMDabhadkar, line 87: The stream uses SecondaryPreferred without a max staleness setting, so an empty batch only means caught up with that secondary. If the secondary falls behind the primary, lag still reads zero. Could you add a sentence stating that replication lag is a separate signal not covered by this metric?

docs/designs/054-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+
+```go
+if w.changeStream.TryNext(ctx) {
+    w.recordEventRead(clusterTimeOf(event))   // lastEventReadAt
+    w.processNextEvent(ctx)
+} else if err := w.changeStream.Err(); err != nil {
+    w.handleChangeStreamError(err)
+    return
+} else {
+    w.recordCaughtUp()                        // lastEmptyBatchAt
```
KaivalyaMDabhadkar, line 178: If the server closes the stream, TryNext keeps returning false with no error, so this branch would record "caught up" on every tick. Could you add an ID check here that treats a closed stream as an error and reopens it? The current Next loop has the same spin today, so this would fix both.

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+unfiltered head. Both are covered under Alternatives.
+
+This design was raised by @KaivalyaMDabhadkar in review on #1738.
+
+## Implementation
+
+### Where it lives
+
+```text
+store-client/pkg/client/
```
KaivalyaMDabhadkar: pkg/client already imports the Mongo watcher package, so the watcher cannot import a reporter in pkg/client without an import cycle. Could you change the layout so the watcher exposes the two timestamps and the pkg/client wrapper owns the gauge?

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+milliseconds on a live stream, but it must be explicit rather than incidental.
+
+A consumer whose watcher never starts, or is wedged before its first read, therefore shows
+`changestream_lag_known == 0` persistently, which is itself alertable and is the only signal that
+covers that state.
+
+### Metrics
+
+```text
+changestream_lag_seconds{client}  gauge  now - max(lastEmptyBatchAt, lastEventReadAt); absent until known
```
KaivalyaMDabhadkar: store-client already emits change_stream_resume_token_recoveries_total. Could you rename these two metrics to use the same change_stream_ prefix?

docs/designs/052-changestream-lag-metrics.md
```diff
@@ -0,0 +1,347 @@
...
+  send, assert lag grows. This distinguishes "reading fine, processing slowly" from healthy.
+- **Restart**: assert `changestream_lag_seconds` is **absent** and `changestream_lag_known` is
+  `0` before the first read or empty batch, and that both become live immediately afterwards.
+  Assert lag is never reported as `0` while unknown.
+- **Idle CPU**: measure the watcher's CPU on an idle stream before and after the `TryNext`
+  change, and assert the pacing interval bounds it. The one performance-sensitive part of this
+  design deserves a number rather than a claim.
+- **Both providers**: every case above against MongoDB and PostgreSQL, since the empty-batch
+  signal is derived differently in each.
+- **Integration**: with `envtest` plus a real MongoDB, stop a consumer, insert events, restart
```
KaivalyaMDabhadkar: envtest provides a kubernetes api server, not a mongodb. Could you point this integration test at the kind plus Tilt suite under tests/ instead?

KaivalyaMDabhadkar:
Small conflict with https://github.com/NVIDIA/NVSentinel/pull/1665 which also uses ADR-052. Could this ADR move to 053?

docs/designs/052-changestream-lag-metrics.md
lfriedman-netllama: Correct on both counts, fixed in `d7a4d0ea`. Verified by grepping each module for change-stream construction: `janitor`, `labeler` and `lifecycle-manager` have zero references, `node-drainer` does open one. The table now lists the five real consumers and the Rationale says four of five lack visibility rather than "seven consumers... the eighth".

docs/designs/052-changestream-lag-metrics.md
lfriedman-netllama: Right, and my source was a stale comment rather than the code. Fixed.

`fault-quarantine/pkg/initializer/init.go:71` uses `BuildProcessableHealthEventInsertsPipeline`, and `event-exporter/pkg/initializer/initializer.go:141` is the one using `BuildAllHealthEventInsertsPipeline`. `health-events-analyzer/main.go:99` uses `BuildProcessableNonFatalUnhealthyInsertsPipeline`. All three now have rows with their pipelines and what each admits, and the Alternatives sentence says **event-exporter** is the only unfiltered consumer.

Worth flagging separately: the comment above `BuildAllHealthEventInsertsPipeline` still says "This is used by fault-quarantine to detect all new health events without filtering", which is what I trusted. That comment is wrong and will mislead the next reader too.

docs/designs/052-changestream-lag-metrics.md
lfriedman-netllama: Accepted, and the paragraph is gone rather than softened. You are right that `TryNext` issues a `getMore` the server holds for its await window and that `Next` makes the same calls in an internal loop, so the idle cost argument I built was false.

Removed the sleep, the idle-CPU test and the merge gate that depended on it. Replaced with setting `MaxAwaitTime` **explicitly**, since the stream is currently opened with only `SetFullDocument(options.UpdateLookup)` (`watch_store.go:168`) and sets no await time, and that interval is what bounds the resolution of `lastEmptyBatchAt`.

This also removes the largest stated risk from the Consequences section, so the design got cheaper as well as simpler.

docs/designs/054-changestream-lag-metrics.md
lfriedman-netllama: Good catch, and adopted. The empty branch now checks `w.changeStream.ID() == 0` and treats a closed cursor as an error to reopen, rather than recording "caught up" against a dead stream on every tick.

Your last sentence is the important part and I have written it into the ADR: the existing `Next` loop spins the same way today, so this is a live bug fix rather than a guard for the new metric. It is now called out in the Rationale and has its own regression test in the test plan, so it does not get lost as an implementation detail of an observability change.

docs/designs/052-changestream-lag-metrics.md
lfriedman-netllama: Correct, and I hit this exact cycle earlier today putting a test in `pkg/client`: `imports .../providers/postgresql ... imports .../pkg/client ... import cycle not allowed in test`.

Layout changed as you suggest: the providers own the two timestamps and expose accessors with no metrics, and the `pkg/client` wrapper owns the gauges and reads them. The ADR now leads that section with the cycle as the reason, so the split reads as a constraint rather than a preference.

docs/designs/054-changestream-lag-metrics.md
lfriedman-netllama: Verified and fixed. `fault-remediation` registers everything with `promauto.With(crmetrics.Registry)` (`fault-remediation/pkg/metrics/metrics.go:37-66`) and passes `metrics.WithRegisterer(crmetrics.Registry)` at `main.go:109`, while `store-client` uses plain `promauto` on the default registry. So the gauge would have been missing from exactly one consumer's endpoint, silently.

The ADR now has a section on it: `store-client` takes a caller-supplied registerer defaulting to the default registry, `fault-remediation` passes `crmetrics.Registry`, and this is step 1 of the rollout so nothing depends on a gauge that cannot be scraped. There is also a test asserting both gauges appear on **fault-remediation's** endpoint specifically, since a test against the default registry would pass while the real endpoint stayed empty.

docs/designs/054-changestream-lag-metrics.md
lfriedman-netllama: Added. The stream opens `SecondaryPreferred` with no max staleness (`watch_store.go:263` and `:291`), so an empty batch means caught up with **that secondary**, and a lagging secondary still reads zero.

It is now one of three named blind spots in the ADR rather than a footnote, since each is a way to be behind while the gauge reads zero. Replication lag is stated as a separate signal this metric does not cover.

docs/designs/052-changestream-lag-metrics.md
lfriedman-netllama: Added, and it is the sharpest of the three blind spots because the failure is inverted: lag reads near zero precisely when the most data was skipped.

The ADR now states that when a stored token is too old for the oplog the watcher deletes it and reopens from now (`watch_store.go:485`), and that `change_stream_resume_token_recoveries_total` is what covers that case. The rollout says the two must be read together: a lag of zero is only reassuring if the recoveries counter has not moved.

docs/designs/052-changestream-lag-metrics.md
lfriedman-netllama: Done. Renamed to `change_stream_lag_seconds` and `change_stream_lag_known`, matching `change_stream_resume_token_recoveries_total`. Updated throughout, including both alert expressions in the rollout.

docs/designs/052-changestream-lag-metrics.md
lfriedman-netllama: Correct, that was confused on my part: `envtest` provides a Kubernetes API server and etcd, not a datastore, so "envtest plus a real MongoDB" made no sense. The integration test now points at the `kind` plus Tilt suite under `tests/`, which has a real MongoDB.

lfriedman-netllama:
Renumbered to **ADR-053** in `d7a4d0ea`. Confirmed the conflict: #1665 merged and owns `052-deployment-platform-connector.md`, and 053 was free.

@KaivalyaMDabhadkar all ten inline points are addressed and replied to individually. I verified each against the code, and all ten hold up. Two of them changed the design rather than the wording:

- Dropping the `TryNext` pacing removed the largest stated risk in the ADR, along with the idle-CPU test and merge gate built on top of it.
- The closed-cursor check turns out to fix a **live bug** in the existing `Next` loop, which spins against a dead stream today. That is now in the Rationale with its own regression test, rather than sitting inside an observability change as an implementation detail.

Two others would have shipped broken and been awkward to diagnose: the registry mismatch would have left the gauge absent from fault-remediation's endpoint alone, and the import cycle would have blocked the package layout as drafted.

Your pipeline comment also uncovered an error of mine. The consumer table listed `janitor`, `labeler` and `lifecycle-manager`, none of which read a change stream, and omitted `node-drainer`, which does. Five consumers, not seven.

One thing worth fixing outside this PR: the comment above `BuildAllHealthEventInsertsPipeline` still reads "This is used by fault-quarantine to detect all new health events without filtering". That is what led me to the wrong attribution, and it will mislead the next reader. Happy to send a one-line correction separately.

The empty-batch design is credited to you in the Decision section.

lfriedman-netllama:
Renumbered again, to **ADR-054** in `16691192`, and the PR title follows.

053 collided with the derived-condition-recovery ADR that #1706 added at 12:34 UTC, eight minutes after this branch took the number at 12:26. Effectively simultaneous and neither is merged, so I moved rather than contested it: #1706 is a 51-file feature PR under active review and this is a single file, so renumbering here is the cheaper side. 054 is free on `main` and unused in #1706.

@KaivalyaMDabhadkar this is the second renumber off your original note about #1665, so if there is a convention for reserving ADR numbers ahead of writing, that would save the next person the same round trip.

KaivalyaMDabhadkar (approved):

PR #380, opened 2025-11-20
feat: add event exporter for db

XRFXLP:
## Summary

For design details: https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/012-health-events-exporter.md

### Timelines tested
- [x] Regular event via changestream: 

          1. Wait for exporter to running state
          2. Inject health event
          3. Validate event is exported to sink

- [x] Resume token functionality

          1. Scale down exporter to 0 replicas after it created resume token
          2. Inject health event
          3. Scale back exporter to 1 replica
          4. Validate event is exported to sink

- [x] Backfill functionality

          1. Scale down exporter to 0 replica
          2. Delete resume token from mongodb
          3. Inject a health event
          5. Wait for 2 minutes
          6. Scale up exporter
          7. Validate that event is exported to sink with the correct event generated timestamp

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: ____________

## Testing
- [x] Tests pass locally
- [x] Manual testing completed
- [x] No breaking changes (or documented)

## Checklist
- [x] Self-review completed
- [ ] Documentation updated (if needed)
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added event-exporter build target and end-to-end exporter: TOML-driven config with validation/defaults, OIDC token provider, HTTPS CloudEvents sink with token auth, Prometheus metrics endpoint, initializer-driven startup with graceful shutdown, backfill workflow with resume-token tracking, event transformer and sink/transformer interfaces, and exporter run orchestration.

* **Observability**
  * Introduced Prometheus metrics for publish/transform/backfill/token activities.

* **Tests**
  * Added unit tests for CloudEvent transformation, OIDC token handling, and HTTP sink publishing.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1739, opened 2026-09-03
fix(event-exporter): remove the permanently-zero event backlog gauge

lfriedman-netllama:
Small, self-contained bug fix, split out of #1709 as offered there. **One file**, one deletion.

## The bug

`health_events_exporter_event_backlog_size` is registered at `event-exporter/pkg/metrics/metrics.go:73` and **assigned nowhere in the module**. A registered gauge that is never written reports `0`, so event-exporter has been advertising "no backlog" for every deployment since the metric was added.

That is worse than having no metric: a dashboard or alert built on it reads as healthy by construction.

## It was never wired, at any point

Checked across all history and all branches, since the natural assumption is that this regressed from a refactor:

- In the introducing commit (`ee6c06bb`, #380) the symbol count in `event-exporter/` goes from **0 to exactly 1**, and that occurrence is the declaration itself.
- Only three other commits touch the symbol in this module: `d6c5c46d` (repo restructure, drops the tree to 0), `fe4ab63b` (restores it at 1, same declaration, same line), and this one. The count is never above 1.
- `git log --all -S 'EventBacklogSize.Set' -- event-exporter/` is empty, so no commit ever contained a setter here.
- `git log --all -S 'GetUnprocessedEventCount' -- event-exporter/` is empty, so the count method was never called from this module either.

So this is an unimplemented item, not a regression.

## Why remove rather than wire it up

The reason it was never wired is visible in the interface:

```go
GetUnprocessedEventCount(ctx context.Context, lastProcessedID string) (int64, error)
```

It needs a **document id**, and event-exporter tracks only resume tokens. `fault-quarantine` can call the same method because it maintains a `LastProcessedObjectIDStore`.

Wiring it here means threading a document id through the concurrent worker pool so that "last processed" respects the contiguous-completion ordering `sequenceTracker` already owns. That is a feature rather than a bug fix, and one that #1709 / ADR-052 supersedes by computing this in the shared layer that already has the position.

So this PR deletes the misleading metric now, and real backlog visibility for every consumer arrives with #1709.

## Blast radius

Checked before removing:

- **No code references it** anywhere besides the declaration. `fault-quarantine` has a same-named Go symbol in its own package and is untouched. (That one has its own problem, filed separately as #1740.)
- **Not in `docs/METRICS.md`** — event-exporter has no section there at all.
- **No dashboard or alert in our own fleet repo references it.** Anything that did would have been graphing a flat zero.

An earlier revision of this PR also annotated ADR-012's metric list. That is reverted at the maintainer's request, since design docs here are point-in-time records; the file is byte-identical to `main`.

## Verification

Full `event-exporter` module green, `golangci-lint` 0 issues, `make gomod-lint` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the `EventBacklogSize` Prometheus metric from exported monitoring metrics.


docs/designs/012-health-events-exporter.md
lalitadithya, line 1: can we not change this file, please? this is typically a point in time decision that was taken. 

for this specific change I think we had supported this when the event exporter was single threaded, but this metric was likely removed when it made multi threaded, we can possibly check from the commit history

lfriedman-netllama: Done, reverted in `8ab25cb2`. The PR is now a single-file change to `event-exporter/pkg/metrics/metrics.go`, and ADR-012 is byte-identical to `main`. Point taken on ADRs being point-in-time records rather than living documents; I will leave them alone.

On the history question, I checked as you suggested, and the hypothesis does not hold: the gauge was **never wired**, not before or after any threading change.

In the commit that introduced it (`ee6c06bb`, #380) the symbol count in `event-exporter/` goes from **0 to exactly 1**, and that one occurrence is the declaration itself:

```
ee6c06bb:event-exporter/pkg/metrics/metrics.go:73:	EventBacklogSize = promauto.NewGauge(
```

Across all history and all branches, only three other commits touch the symbol in this module: `d6c5c46d` (repo restructure, drops the tree to 0), `fe4ab63b` (restores it at 1, same declaration, same line), and mine. The count is never above 1, and `git log --all -S 'EventBacklogSize.Set' -- event-exporter/` is empty, so no commit ever contained a setter here. `git log --all -S 'GetUnprocessedEventCount' -- event-exporter/` is empty too, so the count method was never called from this module either.

You are right that the difference is the threading, though as a cause rather than a chronology: `GetUnprocessedEventCount` needs a `lastProcessedID`, and wiring one here means threading a document id through the worker pool so "last processed" respects the contiguous-completion ordering `sequenceTracker` owns. That is why it is worth doing once in the shared layer, per ADR-052, rather than per module.

Happy to drop this PR entirely if you would rather the removal ride along with #1709's work. It is only useful as a standalone if the misleading zero is worth clearing sooner.

docs/designs/012-health-events-exporter.md
```diff
@@ -556,7 +556,12 @@ health_events_exporter_publish_errors_total{cluster="...",error_type="..."}
 health_events_exporter_token_refresh_errors_total{cluster="..."}
 
 // Queue/Backlog
-health_events_exporter_event_backlog_size{cluster="..."}
+// NOT IMPLEMENTED. The gauge was registered but never set, so it reported 0 for every
+// deployment, and it was removed rather than left reporting a false "no backlog".
+// event-exporter tracks only resume tokens and has no per-document id to pass to
+// ChangeStreamMetrics.GetUnprocessedEventCount, which is why it was never wired.
+// Backlog and lag for every change stream consumer are addressed by ADR-052 and #1709.
+// health_events_exporter_event_backlog_size{cluster="..."}
```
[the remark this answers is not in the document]
lfriedman-netllama: Resolved differently: the whole ADR-012 change is reverted in `8ab25cb2` at the maintainer's request, since design docs here are point-in-time records rather than living documents. That file is now byte-identical to `main`, so the metrics list is untouched and this finding no longer applies.

Worth noting the finding had the right instinct but inverted the situation. The line was not advertising a metric that had been retired; that entry has been a planned-but-never-implemented item since the ADR was written, which the annotation was trying to record. Verified: no commit on any branch ever contained a setter for it in `event-exporter/`.

lalitadithya:
@lfriedman-netllama , can you please resolve conflicts?

lfriedman-netllama:
Done in `83e383e2`, now `MERGEABLE`.

The conflict came from #1724 adding `EventsFiltered` and `FilterErrors` to the same `var` block this PR removes `EventBacklogSize` from. Resolution keeps both new filter metrics and drops only the dead gauge, so the net diff against `main` is 7 deleted lines in `event-exporter/pkg/metrics/metrics.go` and nothing else.

Merged `main` in rather than rebasing, to preserve your earlier merge commit on this branch. Re-verified after resolving: module builds, all 6 test packages pass, `golangci-lint` 0 issues, `make gomod-lint` clean, and `EventBacklogSize` has 0 references left in `event-exporter/`.

Needs a fresh `/ok to test` at `83e383e2` when you get a chance.

lalitadithya (approved):

PR #1741, opened 2026-09-03
chore: upgrade percona database

### LINKED ISSUE #1737: [Bug]: bundled psmdb subcharts are mismatched: operator 1.23.0 with crVersion 1.21.1, and the jump skips a Percona minor
### Prerequisites

### Code of Conduct

### Bug Description

The bundled Percona subcharts ship a mismatched pair: **`psmdb-operator` is at 1.23.0 while `psmdb-db` is still at 1.21.1**, so the chart renders operator image `1.23.0` driving a CR with `crVersion: 1.21.1`.

Percona documents both of the things this breaks.

### What the chart currently renders

Neither value is overridden anywhere in the parent chart (`values.yaml`, `values-full.yaml` and `charts/mongodb-store/values.yaml` all leave them alone), so these subchart defaults are effective:

| File | Value |
| --- | --- |
| `charts/mongodb-store/charts/psmdb-operator/values.yaml:10` | `tag: 1.23.0` |
| `charts/mongodb-store/charts/psmdb-db/values.yaml:24` | `crVersion: 1.21.1` |
| `charts/mongodb-store/charts/psmdb-db/values.yaml:54` | mongod `tag: 8.0.12-4` |

Chart metadata agrees: `psmdb-operator/Chart.yaml` is `version`/`appVersion` **1.23.0**, `psmdb-db/Chart.yaml` is **1.21.1**.

### Why this is a problem

**1. It skips a minor version on upgrade, which Percona does not support.** From [Upgrade considerations](https://docs.percona.com/percona-operator-for-mongodb/update-operator.html):

> You can only upgrade the Operator to the nearest `major.minor` version. For example, from `1.18.0` to `1.19.0`.

> To upgrade the Operator from `1.17.0` to `1.19.1`, you need to first upgrade it to `1.18.0`, and then to `1.19.1`.

Any cluster already running a 1.21.x operator that adopts this chart jumps straight to 1.23.0, skipping 1.22.

**2. `crVersion` is meant to track the operator version.** Percona's own upgrade procedure patches `crVersion` to the operator version being upgraded to. Shipping operator 1.23.0 against `crVersion: 1.21.1` is a two-minor gap in the pair.

It is currently *tolerated* rather than correct, because [CRD compatibility](https://docs.percona.com/percona-operator-for-mongodb/update-operator.html) is:

> CRD supports the last 3 minor versions of the Operator. This means it is compatible with the newest Operator version and the two previous minor versions.

1.23, 1.22, 1.21 is exactly that window, so it works today. The next operator bump to 1.24 would push `crVersion: 1.21.1` out of support.

**3. mongod is behind the certified set.** Per [Versions compatibility](https://docs.percona.com/percona-operator-for-mongodb/versions.html), operator 1.23.0 certifies Percona Server for MongoDB `8.0.26-11`; the chart pins `8.0.12-4`.

### How it happened

The operator subchart was bumped twice and the db subchart was never moved with it:

- #1656 `fix(mongodb-store): bump psmdb-operator to 1.22.0`
- #1662 `chore: upgrade percona and cert-manager to latest` (to 1.23.0)
- `psmdb-db/Chart.yaml` last changed in #1530, which was unrelated (`customRecommendedAction` policy template)

So this looks like an oversight in "upgrade to latest" rather than a deliberate pairing.

### Impact observed

On a 288 node GB200 fleet adopting v1.21.0 we pinned the operator back to `1.21.0` specifically to avoid this, because the unpinned chart would have:

- pulled `percona-server-mongodb-operator:1.23.0`, which our mirror did not carry, so the operator would have gone `ImagePullBackOff`, and
- jumped two Percona minors on a live datastore in a single sync.

Live state with the pin is a consistent pair: operator `1.21.0`, `crVersion: 1.21.1`, `state: ready`. In other words the pinned-back configuration is closer to Percona's guidance than the chart default.

### Suggested fix

Either of these makes the chart self-consistent; the first is presumably the intent:

1. **Bump `psmdb-db` in step with the operator**: `crVersion: 1.23.0` and a mongod tag from 1.23.0's certified set (`8.0.26-11`). Note this still leaves existing 1.21.x installs skipping 1.22, so the release notes should say an intermediate hop is required, or the chart should land 1.22.0 first.
2. **Pin `psmdb-operator` back to 1.21.x** until the pair can move together.

Worth stating explicitly in either case: because operator upgrades cannot skip a minor, moving this chart's Percona version by two minors at once is a breaking change for existing deployments and should be called out in release notes rather than arriving inside a `chore: upgrade to latest`.

### Component

Other

lalitadithya:
## Summary

upgrade the actual database post the operator upgrade in the previous release. 

Note for release notes team: We need to call this out in the release notes and that customers should check the version they are currently running and ensure their upgrade process is line with the Percona documentation

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [X] 🔨 Build/CI

## Component(s) Affected
- [X] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added MongoDB ClusterSync configuration and migration support.
  * Added configurable Vector Search, storage autoscaling, Vault integration, probes, hook scripts, and environment settings.
  * Added expanded backup storage options for GCS, OSS, MinIO, Azure Blob, and OCI.
  * Added optional TLS, arbiter, logging, and monitoring configuration.

* **Enhancements**
  * Updated MongoDB Operator and chart versions to 1.23.0.
  * Improved namespace and client connection handling.

* **Documentation**
  * Added upgrade guidance, Kubernetes 1.32+ guidance, configuration examples, and updated image defaults.


KaivalyaMDabhadkar (dismissed):

distros/kubernetes/nvsentinel/charts/mongodb-store/charts/psmdb-db/templates/cluster.yaml
```diff
@@ -6,24 +6,61 @@ metadata:
 {{ .Values.annotations | toYaml | indent 4 }}
   {{- end }}
   name: {{ include "psmdb-database.fullname" . }}
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ include "psmdb-database.namespace" . }}
   labels:
 {{ include "psmdb-database.labels" . | indent 4 }}
-  {{- if .Values.finalizers }}
```
KaivalyaMDabhadkar, line 12: This drops the `if .Values.finalizers` guard we carry on this vendored file (#1230). Our parent chart sets `finalizers: []`, so the CR renders an empty list, the apiserver strips it, and argo keeps the PerconaServerMongoDB resource OutOfSync forever. Nothing breaks at runtime, but every argocd managed install regresses.

I checked the values-only alternative (`finalizers: null`) and it is not really safe: on helm 3.16, which ArgoCD 2.14 bundles, it silently brings back percona's default finalizer. So the guard is still the only fix that works on every helm version.

```suggestion
  {{- if .Values.finalizers }}
  finalizers:
{{ .Values.finalizers | toYaml | indent 4 }}
  {{- end }}
```

I will raise a PR against percona-helm-charts to add this guard upstream, so once that lands in a released chart we can vendor it untouched again. Until then we need to keep carrying it here.

lalitadithya: done, please take a look

KaivalyaMDabhadkar: I have raised this PR, we shouldn't need to patch this guard going forward once this is merged: https://github.com/percona/percona-helm-charts/pull/948

KaivalyaMDabhadkar: Can you check the lint failures?

lalitadithya: yeah, they are all apt repo outages, I'm rerunning them to get them to pass. they are failing on other PRs and main too

KaivalyaMDabhadkar (approved):

Issue #1743, opened 2026-09-03
[Bug]: the Kubernetes connector drops health events on write failure while the store and grpcsink connectors retry, so a lost recovery event latches a node condition forever


lfriedman-netllama:
### What happened

The Kubernetes connector is the only one of the three platform connectors that **drops** a health event when processing fails. `grpcsink` and `store` both requeue with backoff. Because the Kubernetes connector is the one whose writes carry *recovery* semantics, a single terminal write failure can leave a node condition `True` permanently, with no in-band path to correct it.

### The asymmetry

```go
// platform-connectors/pkg/connectors/kubernetes/k8s_connector.go:136
r.ringBuffer.HealthMetricEleProcessingFailed(queuedHealthEvents)
```

```go
// platform-connectors/pkg/ringbuffer/ring_buffer.go:126-129
func (rb *RingBuffer) HealthMetricEleProcessingFailed(data *QueuedHealthEvents) {
	rb.healthMetricQueue.Forget(data)   // reset backoff
	rb.healthMetricQueue.Done(data)     // and discard
}
```

`Forget` + `Done` with no re-add means the event is gone. Compare the other two connectors, which use the requeue method that already exists beside it (`AddRateLimited`, `ring_buffer.go:131-134`):

- `platform-connectors/pkg/connectors/store/store_connector.go:135` — requeues with exponential backoff up to `maxRetries`, then logs "Max retries exceeded, dropping health events permanently" (`store_connector.go:138`).
- `platform-connectors/pkg/connectors/grpcsink/grpc_sink_connector.go:116` — same pattern.

So the bounded-retry-then-drop pattern is already established in this codebase; the Kubernetes connector simply skips the retry half.

### Why the in-process retry does not cover it

`processNodeEvents` does wrap the read-modify-write in `retry.OnError` (`platform-connectors/pkg/connectors/kubernetes/process_node_events.go:72-84`), classifying conflicts and temporary errors as retriable:

```go
err := retry.OnError(retry.DefaultRetry, func(err error) bool {
	isRetriable := apierrors.IsConflict(err) || isTemporaryError(err)
	...
```

That handles the common `Conflict` case well. But once `retry.DefaultRetry` is exhausted, or the error is classified non-retriable, `processHealthEvents` returns an error and the event is discarded at `k8s_connector.go:136`. An API server that is unavailable for longer than `DefaultRetry`'s short window (five attempts, tens of milliseconds of backoff) exhausts it easily.

[...]

PR #1747, opened 2026-09-04
fix(platform-connectors): retry Kubernetes writes in order

Saibernard:
## Summary

Reduces Kubernetes health-event loss during transient API failures by retrying failed writes before consuming the next batch. Successful writes are not replayed, and permanent errors do not prevent independent transient failures from retrying.

For the node-local Kubernetes queue, the retry count defaults to 25 per write. The whole-batch deadline defaults to one minute and is configurable up to five minutes. API calls and backoff share that deadline; cancellation and shutdown interrupt both. Uncertain Event writes are reconciled by stable fault name and persisted timestamp before retrying, avoiding false drops and duplicate counts. The integration with main preserves on-change node updates, Event suppression, ordered recovery clearing, and the synchronous connector API. Added drop counters by reason, explicit queue discard semantics, configuration coverage, and real Kubernetes API regressions.

Retries remain bounded. Outages beyond the configured window and pod restarts can still lose recovery events, and pending batches accumulate in memory. Related to #1743; this does not guarantee recovery delivery through arbitrary outages.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Service
- [x] Deployment/Config
- [x] Documentation/CI

## Testing
- [x] Tests pass locally (platform-connectors and Helm; repository-wide limitation below)

- Full platform-connectors race suite on macOS ARM64: 567 tests/subtests passed across 14 packages, including real Kubernetes envtest regressions.
- Build, vet, dependency tidy, and golangci-lint passed.
- 109 Helm tests across 21 suites and Helm lint passed.
- Regression coverage includes independent write retries, deadlines, shutdown, uncertain creates, lost AlreadyExists/refresh responses, restart replay, and fault/recovery/recurrence ordering.
- Repository protobuf generation, license checks, and all 33 Go module validations passed.
- `make lint-test-all` stopped at three unchanged syslog tests requiring Linux’s `/proc/sys/kernel/random/boot_id`, which is absent on macOS. The full repository gate has not passed locally.
- Fresh cluster E2E and CI are pending for this update.

## Checklist
- [x] Self-review completed
- [x] Ready for review


platform-connectors/pkg/connectors/kubernetes/k8s_connector.go
```diff
@@ -121,24 +150,131 @@ func (r *K8sConnector) FetchAndProcessHealthMetric(ctx context.Context) {
...
+		}
+
+		retryCount++
+		slog.WarnContext(ctx, "Kubernetes health event processing failed; retrying in order",
+			"error", err,
+			"retryCount", retryCount,
+			"maxRetries", r.config.MaxRetries,
+			"retryDelay", retryDelay)
+
+		if err := waitForKubernetesRetry(ctx, retryDelay); err != nil {
```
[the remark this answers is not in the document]
Saibernard: Addressed in 94ec3807. The retry wait now observes `stopCh` as well as `ctx.Done()`, returns promptly on shutdown, and still flows through the existing queued-batch cleanup. I added a long-backoff regression test that asserts shutdown returns `context.Canceled` before another write; it passed 50 consecutive runs and the full race suite on both ARM64 and AMD64.

platform-connectors/pkg/connectors/kubernetes/k8s_connector_retry_test.go
```diff
@@ -0,0 +1,325 @@
...
+	return &K8sConnector{
+		config:         K8sConnectorConfig{MaxRetries: maxRetries},
+		processEvents:  process,
+		retryBaseDelay: time.Nanosecond,
+		retryMaxDelay:  time.Nanosecond,
+	}
+}
+
+// TestNewK8sConnectorDefaultsMaxRetries verifies the default outer retry limit.
+func TestNewK8sConnectorDefaultsMaxRetries(t *testing.T) {
```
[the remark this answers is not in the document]
Saibernard: Addressed in 4fcc668a. All six retry tests now use the requested `TestFunctionName_Scenario_ExpectedBehavior` naming pattern, with matching GoDoc comments. The renamed test set passes under `-race` on Linux ARM64 and AMD64, and golangci-lint v2.13.1 reports zero issues.

platform-connectors/main.go
```diff
@@ -143,9 +143,19 @@ func initializeK8sConnector(
 		return nil, fmt.Errorf("failed to convert K8sConnectorBurst to int: %v", config["K8sConnectorBurst"])
 	}
 
+	maxRetries := int64(k8sconnector.DefaultMaxRetries)
+	if configuredMaxRetries, configured := config["K8sConnectorMaxRetries"]; configured {
+		maxRetries, ok = configuredMaxRetries.(int64)
+		if !ok || maxRetries <= 0 {
```
deesharma24, line 149: this is contradictory. here you are saying maxRetries <=0 is not okay whereas at the top  https://github.com/NVIDIA/NVSentinel/pull/1747/changes#diff-c4e88a6358531c39c9929197b5a18b70dae5ebbeee8bbba2eb5e033051bc3343R79 you have configured it like if maxRetries ==0 , go with default which is 3.Can we have the consistency pls?

lfriedman-netllama:
Reviewed this as the author of #1743. The structure is right and it addresses the concerns from that issue, including one I had raised as an open question rather than answered.

**Ordering: handled better than what I suggested.** #1743 proposed `AddRateLimited`, and flagged as a design question that requeueing puts the batch behind newer ones, which can reorder a fault and its recovery for the same condition. Holding the batch and retrying in place avoids that entirely, and the comment on `processHealthEventsWithRetry` says so explicitly. That is the stronger choice; `aggregateEventMessages` is last-write-wins per condition, so a reordered stale fault re-asserting a cleared condition was a real risk.

**Permanent vs transient: present and comprehensive.** `isKubernetesConnectorRetryableError` returns immediately for non-retryable errors instead of burning the budget, and `isTemporaryError` covers what matters for this failure mode: `IsTimeout`, `IsServerTimeout`, `IsServiceUnavailable`, `IsTooManyRequests`, `IsInternalError`, plus network timeouts, syscall errors and EOF. I went looking for 429 and 503 specifically and they are covered.

**Cancellation ordering is correct.** `ctx.Err()` is checked before the retryable test, so a cancelled context exits rather than retrying, even though `isContextError` treats `context.Canceled` as temporary. `waitForKubernetesRetry` selecting on `ctx.Done()`, `stopCh` and the timer means shutdown is not delayed by a pending backoff. The three-way terminal log split makes the shutdown case distinguishable from a genuine drop, which matters when reading logs after the fact.

**One substantive point: the default horizon is short relative to the failure this targets.**

With `DefaultMaxRetries = 3` and the ring buffer defaults (`DefaultBaseDelay` 500ms, `DefaultMaxDelay` 3s), the outer backoff totals roughly **3.5s** (500ms + 1s + 2s) on top of client-go's inner window.

The case #1743 described was an API server unavailable for longer than the inner window, and a control-plane restart or rolling upgrade is usually tens of seconds. 3.5s is a large improvement over 0s, but it does not span that. Since the delay is capped at 3s, raising the default is cheap: `MaxRetries` of 20 gives about 57s, still bounded and still ordered.

Worth considering because the asymmetry from #1743 still applies. Dropping an unhealthy event is largely self-correcting, since the fault is usually still present and re-reported. Dropping a **healthy** event is not: nothing re-sends a recovery, so the node condition stays `True` until someone patches it by hand. That is the case this PR exists to prevent, and it argues for erring toward more retries.

Either raising `DefaultMaxRetries` or documenting the horizon the default buys, so operators can size it against their own control-plane behaviour, would resolve it. Not blocking either way.

Nothing else from me. Thanks for splitting it out and for picking up the issue.

platform-connectors/main.go
Saibernard: @deesharma24 fixed this. Both paths now use three retries when the setting is missing or zero. Positive values override the default, and invalid values are rejected. Added tests for these cases too.

Saibernard:
@lfriedman-netllama added the retry timings and a longer-retry configuration example to the docs, including what happens if a recovery event exhausts its retries. Kept the default at three.

Saibernard:
@lalitadithya @deesharma24 the review fixes are pushed and all review threads are resolved. Could one of you please run `/ok to test` for the latest changes?

platform-connectors/pkg/connectors/kubernetes/k8s_connector.go
```diff
@@ -121,24 +157,137 @@ func (r *K8sConnector) FetchAndProcessHealthMetric(ctx context.Context) {
...
+	}
+
+	maxRetryDelay := r.retryMaxDelay
+
+	if maxRetryDelay <= 0 {
+		maxRetryDelay = ringbuffer.DefaultMaxDelay
+	}
+
+	for {
+		err := processEvents(ctx, healthEvents)
```
deesharma24, line 247: This retries the whole batch, including events already written successfully. For example, if event A succeeds and event B fails temporarily, the retry writes A again. updateCachedNodeEvent then increases A's Count from 1 to 2, even though A happened only once.

Please retry only failed writes, or ensure that retrying a successful write does not count it again.

lfriedman-netllama:
I filed #1743, so I had a look at this. The retry machinery here is good, and better than what I proposed in the issue. But I do not think it closes #1743 yet, for one specific reason.

## What is clearly right

- **Retrying the current batch before consuming the next** is the correct shape, and it is what preserves fault-then-recovery ordering. That was the part of the issue I cared most about.
- **Permanent errors are classified and not retried**, so a `400` does not burn the budget.
- **Cancellation interrupts the backoff.** Worth calling out, because `wait.ExponentialBackoff` consults no context at all; I got caught by exactly that on #1778 and had to add a derived deadline.
- Rejecting a negative `MaxRetries` rather than reinterpreting it, and defaulting `0`, is the right treatment of the knob.
- 379 lines of tests covering success, exhaustion, ordering, permanent errors and cancellation.

## Why I think #1743 is still open

On exhaustion the batch goes to `HealthMetricEleProcessingFailed`, and that function is:

```go
func (rb *RingBuffer) HealthMetricEleProcessingFailed(data *QueuedHealthEvents) {
	rb.healthMetricQueue.Forget(data)
	rb.healthMetricQueue.Done(data)
}
```

which is **byte-identical to `HealthMetricEleProcessingCompleted`**. `Forget` clears the rate-limiter history and `Done` closes the item out; nothing re-enqueues. So the batch is still dropped permanently, and the log line says so: "Max retries exceeded, dropping Kubernetes health events permanently".

The issue's title is that a lost recovery event latches a node condition forever. After this change that still happens, just less often.

**And I think 3.5 seconds is short for the failure modes that motivated the issue.** Some numbers from our fleet, 288 GB200 nodes:

- A kubeconfig rotation on 2026-09-07 produced failures spread across **roughly an hour**: 88 of 288 metadata-collector pods exited between 15:26 and 16:27 UTC, in clusters of a few at a time.
- That is the same class of event that #1767 was about, and the reason the fix there widened a ~40ms budget: the old one could not span an authentication window, so more retries at that scale changed nothing.

An API server rollout is tens of seconds. 3.5 seconds of outer delay covers a blip, not a window, and a window is when recovery events get lost.

## The honest tension

`AddRateLimited` already exists on `RingBuffer`, right below the two functions above, and is what the store and grpcsink connectors use. That is the requeue path the issue proposed, and it is not used here.

But I do not think that is an oversight, and I do not want to just tell you to switch: **requeueing puts the batch behind newer batches, which breaks the strict ordering this PR works to preserve.** That is a genuine tradeoff, so it is your call which side to land on. The options I can see:

1. **Requeue with a bounded `NumRequeues`.** Ordering is only disturbed on the failure path, which is already abnormal, and nothing is lost. Closest to what store and grpcsink do.
2. **Keep in-line retry but size the budget to a real outage window** rather than 3.5s, for example a whole-batch deadline of a minute or two with capped exponential delays.
3. **Apply backpressure**: stop consuming while the API is unreachable rather than draining the queue into the floor. Preserves ordering perfectly and loses nothing, at the cost of the buffer filling.

My preference is 1 or 3, and either would close the issue. 2 narrows it without closing it.

## One thing I would ask for regardless

**A counter on the drop.** The PR records `retry_count` and `max_retries` as span attributes, which is useful for tracing a known incident but is not alertable, and there is no metric for a dropped batch. Given the entire subject of #1743 is *silent* event loss, the drop needs to be countable so an operator can see it happened, ideally labelled by reason so exhaustion is distinguishable from a permanent error.

## Small aside, possibly the root cause

`HealthMetricEleProcessingFailed` doing exactly what `HealthMetricEleProcessingCompleted` does is, I suspect, how this bug came to exist: a caller reading that name would reasonably assume it requeues. Whatever is decided above, renaming it or making it live up to its name would stop the next connector making the same assumption.

Happy to be told I am wrong about the budget, and happy to test any of this against our cluster if that helps.

platform-connectors/pkg/connectors/kubernetes/k8s_connector.go
```diff
@@ -121,24 +157,137 @@ func (r *K8sConnector) FetchAndProcessHealthMetric(ctx context.Context) {
...
+		if err == nil {
+			return retryCount, nil
 		}
+
+		if ctx.Err() != nil {
+			return retryCount, ctx.Err()
+		}
+
+		if !isKubernetesConnectorRetryableError(err) || retryCount >= r.config.MaxRetries {
```
deesharma24, line 256: processHealthEvents tries all writes in the batch but returns only the first error. This error then decides whether the whole batch is retried.

For example, if write A fails permanently and write B fails temporarily, A's error stops the retry, so B never gets another chance. If B fails first, the whole batch is retried, including A, which cannot succeed. 

Can we distinguish between which writes to retry vs which can be skipped so that so we don't write same healthevent multiple times.

deesharma24:
@Saibernard Regarding @lfriedman-netllama  comment, we should use a combination of 2 and 3 with configurable cap for write retries and can go upto 5 minutes.

Saibernard:
@deesharma24 @lfriedman-netllama I’ve addressed the latest review comments and pushed the changes: retries now apply only to failed writes, the batch deadline defaults to one minute and allows up to five, and drops have counters by reason. The connector race suite and Helm tests pass locally.

For outages that outlast the cap, I’d propose persisting pending node-condition updates before acknowledging them, then replaying them after API recovery or a restart. We’d preserve fault/recovery order per node and make replay safe when a write succeeds just before a crash. A full backlog would pause intake, and invalid updates would be reported separately.

That would give accepted recovery updates a path to complete after longer outages, instead of relying on a larger timeout. Would you be comfortable keeping this PR focused on the bounded-retry fix and me working through that durable-recovery design as a follow-up?

platform-connectors/pkg/connectors/kubernetes/k8s_connector.go
Saibernard: Fixed. Retries now stay with the failed write, so successful writes aren’t repeated. The real Kubernetes API test checks that A stays at Count=1 while B recovers.

platform-connectors/pkg/connectors/kubernetes/k8s_connector.go
Saibernard: Fixed. Permanent failures are recorded and skipped, while transient failures get their own retries. Tests cover both error orders.

distros/kubernetes/nvsentinel/values-full.yaml
```diff
@@ -321,6 +321,14 @@ platformConnector:
   # Kubernetes connector configuration
   # Handles updating Kubernetes node status and conditions
   k8sConnector:
+    # Retry each failed Kubernetes write in place; completed writes are not repeated.
+    # Zero or omission selects 100 retries per write. Negative values are invalid.
+    # The batch deadline below can stop retries before the count is reached.
+    maxRetries: 100
```
deesharma24: Basic question. From where does this 100 number came from? isn't it too high? I feel 25 is enough which roughly translates to ~70 sec.

lalitadithya: i believe this is what gives us 5 minutes, no?

deesharma24: Yes, 100 retries add up to approximately 5 minutes of backoff. is this ~5 min upper bound intentional?I felt like ~1 min is good enough,hence the comment

lalitadithya: yeah, we could go either way since it is configuration folks that needs more retries can add

deesharma24: not a major thing. but yeah, we can start simple(~1 min/25 retries) as that is hardcoded in code also and if someone feels, they need more retries, they can always change.

deesharma24:
@Saibernard Can you please take care of above maxretries and coderabbit comment ?

distros/kubernetes/nvsentinel/values-full.yaml
Saibernard: Changed the default to 25 in the code and both Helm values, and updated the docs and tests. The one-minute batch deadline still applies, so it stops retries before the full 69.5 seconds of backoff. Operators can raise both settings when they need a longer window.

platform-connectors/pkg/connectors/kubernetes/k8s_connector.go
```diff
@@ -121,24 +172,199 @@ func (r *K8sConnector) FetchAndProcessHealthMetric(ctx context.Context) {
...
+		if err == nil {
+			return retries, nil
+		}
+
+		if ctx.Err() != nil {
+			return retries, ctx.Err()
+		}
+
+		if !isKubernetesConnectorRetryableError(err) || retries >= r.config.MaxRetries {
+			return retries, err
```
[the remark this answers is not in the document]
Saibernard: Fixed. After an uncertain Create, the retry looks up that exact Event name and checks that it matches the pending occurrence. A matching Event is accepted and cached without increasing Count; a different object is rejected. Another Create is attempted only if the lookup confirms the name is absent.

Added real API tests for lost responses, a failure before the write reaches Kubernetes, failed lookups across retry rounds, name collisions, and deletion races. They also check that the next real occurrence still increments Count normally.

Saibernard:
@deesharma24 @lalitadithya both review fixes are pushed, and I have resolved the conflicts with main while preserving the new Event deduplication behavior. The connector race suite and Helm tests pass locally. Could one of you run `/ok to test` again and take another look?

deesharma24 (approved):

lalitadithya (approved):

PR #1749, opened 2026-09-04
feat: add support for post-remediation validation

### LINKED ISSUE #1427: [Feature]:  Add post-remediation validation gate before clearing node/GPU fault state
### Prerequisites

### Code of Conduct

### Feature Summary

NVSentinel currently handles fault detection, quarantine, drain, and remediation, but there is no explicit post-remediation validation gate before
  a node or GPU is considered recovered.

  For GPU-related faults, remediation success should not only mean that the remediation action has completed. The platform should also verify that
  the affected GPU stack is actually usable again before removing quarantine state or allowing workloads to be scheduled back.

### Problem/Use Case

 In real GPU clusters, a remediation action may complete successfully while the node or GPU is still unhealthy. Examples include:

  - GPU disappeared and later reappeared, but CUDA workloads still fail.
  - Driver/NVML is available, but DCGM diagnostics fail.
  - Device plugin reports capacity again, but NCCL or basic CUDA runtime tests fail.
  - A node reboot or GPU reset completes, but the GPU is still unstable.

  Without a post-remediation validation step, NVSentinel may clear fault state too early and allow new workloads onto a node or GPU that has not
  been proven healthy.

### Proposed Solution

  Add an optional post-remediation validation stage before clearing quarantine / fault state.

  The validation stage could run one or more configurable checks, for example:

  - `dcgmi diag` basic diagnostic
  - `nvidia-smi` / NVML availability check
  - CUDA sample or basic CUDA runtime test
  - NCCL loopback / allreduce smoke test
  - device plugin allocatable GPU consistency check
  - custom validation command or Kubernetes Job

  Only when the configured validation checks pass should NVSentinel mark the fault as recovered and allow quarantine removal.

## Expected Flow

  ```text
  HealthEvent detected
    -> fault-quarantine cordon/taint
    -> node-drainer drains or waits
    -> fault-remediation executes recovery action
    -> post-remediation validation runs

  If validation fails:
    -> keep node/GPU quarantined
    -> emit validation-failed HealthEvent
    -> optionally escalate to CONTACT_SUPPORT or manual remediation

  ## Configuration Idea

  postRemediationValidation:
    enabled: true
    timeoutSeconds: 600
    retryCount: 3
    validators:
      - name: dcgmi-short
        type: command
        command: ["dcgmi", "diag", "-r", "1"]

      - name: cuda-smoke-test
        type: kubernetesJob
        image: nvcr.io/nvidia/k8s/cuda-sample:latest
        command: ["./vectorAdd"]

      - name: nccl-smoke-test
        type: kubernetesJob
        image: <nccl-test-image>
        command: ["./all_reduce_perf", "-b", "8", "-e", "128M", "-f", "2", "-g", "1"]

### Component

Fault Management

natherz97:
## Summary
This PR adds support for post-remediation from the fault-quarantine module. It automates the creation of ValidationRequests when a given node has an unquarantine event. This ValidationRequest will include tests from any event in the node's quarantine session which required validation. Tests are derived from Validation.RuleSets which map a HealthEvent to a list of tests which will be included in the ValidationRequest. 

```
  ruleSets:
    - enabled: true
      version: "1"
      name: "Post-reboot validation"
      match:
        all:
          - kind: "HealthEvent"
            expression: >-
              event.componentClass == 'GPU' && event.isFatal == true &&
              event.recommendedAction in [15, 24]
      tests:
        - dcgm-diag-test
```

#### How is a ValidationRequest is generated during a quarantine session?

We use the quarantineValidationHealthEvent annotation to track all events which occurred during a quarantine session that require >=1 test. Specifically, if a new HealthEvent is added to the quarantineHealthEvent annotation from either a Quarantined or AlreadyQuarantined event, we will also evaluate if that unhealthy HealthEvent requires a validation test. If one of these events requires one or more validation tests, we will add the event to the quarantineValidationHealthEvent annotation.

We inherit the same de-duping behavior which is provided by the quarantineHealthEvent annotation to only track HealthEvents with a unique HealthEventKey (which is a subset of HealthEvent properties). In other words, even if we process all new unhealthy HealthEvents, we will only consider it for validation tests if it resulted in us updating our annotation on either a Quarantined or AlreadyQuarantined event. As a result, we will not capture any tests from duplicated events which are derived from properties outside of what's included in HealthEventKey.

Recall that we do not remove events from quarantineValidationHealthEvent throughout the quarantine session compared to the quarantineHealthEvent where we remove events as soon as a matching healthy event clears it. This could result in us tracking duplicated events for the same HealthEventKey if an event recovers and goes back to unhealthy as part of the same session. To avoid this, we would need quarantineValidationHealthEvent to maintain its own de-duplicating on the HealthEventKey in the future. Since we only track unhealthy HealthEvents which result in Quarantined or AlreadyQuarantined events, we will not allow events to contribute tests from the validation rule-set if they do not satisfy the primary fault-quarantine rule-set.

We have multiple levels of test de-duplication to track:
- HealthEvent de-duplication: as discussed above, we de-duplicate HealthEvents with the same HealthEventKey by
leveraging the quarantineHealthEvent.
- Test de-duplication within 1 HealthEvent: if a validation rule-set evaluation results in the same test being
required multiple times for the same fault, we will ensure that only unique tests are requested per event.
- Test de-duplication across HealthEvents: this isn't relevant to the node annotation update here and is covered
in FetchValidationTestsFromQuarantineSession when the node is unquarantined. When processing all HealthEvents
which required a validation test, we will build up a unique list of tests requested across all events to include
as part of a single validation test.

#### Overview of changes

- Implemented the automatic creation of ValidationRequests from fault-quarantine based on configurable validation rule-sets.
- The existing fault-remediation CRD template rendering and node-drainer tracking of previously completed drains were moved to a common package so that fault-quarantine could re-use the template rendering logic during ValidationRequest creation and re-use the previous drain completion check to determine if a full drain for the current node completed prior to creating the ValidationRequest.
- Improved TestProvider object deletion to always use foreground deletion which ensures that object deletion is blocked on children objects being garbage collected or on any custom finalizers being removed
- Added logic to dcgm-diag-test to retry dcgmi diag if it encounters a DCGM_ST_IN_USE error
- Added validation-controller metrics for validation_requests_total, validation_requests_completed_total, and validation_requests_duration_seconds.

## Testing

1. New unit tests for fault-quarantine:
- TestE2E_ValidationRequestSkippedWhenValidationDisabled
- TestE2E_ValidationRequestSkippedWhenNoRuleSetMatches
- TestE2E_ValidationRequestSkippedWhenDrainIsPartial
- TestE2E_ValidationRequestSkippedWhenNoEventDrained
- TestE2E_ValidationRequestCreatedWhenEventDrained
- TestE2E_ValidationRequestSkippedWhenManuallyUncordoned

2. e2e test coverage for post-remediation validation:
- Extended the existing TestFatalHealthEvent in smoke_test.go to verify that a ValidationRequest is created and succeeds prior to the node being uncordoned.

3. UAT test coverage:
- Extended the existing test_gpu_monitoring_dcgm and test_xid_monitoring_syslog cases (which both trigger a RESTART_VM unhealthy event) to ensure that a successful ValidationRequest exists after the node was uncordoned.

4. Manual testing executed against a GCP GB200 cluster:
- TestFailed is correctly reported when DCGM diag detects an unhealthy GPU
- TestFailed occurs when DCGM diag job object is deleted while test is running
- TestTimeout occurs if DCGM diag job exceeds its test timeout
- 1 ValidationRequests targeting multiple nodes will run DCGM diag jobs concurrently
- 2 ValidationRequests targeting different nodes will will run DCGM diag jobs concurrently
- 2 ValidationRequests targeting the same node will will run DCGM diag jobs sequentially

Logs from a manual UAT test execution:
```
% ./tests.sh
[2026-09-03 10:12:42] Starting NVSentinel UAT tests...
[2026-09-03 10:12:42] Checking if circuit breaker is TRIPPED...
Error from server (NotFound): configmaps "circuit-breaker" not found
[2026-09-03 10:12:43] =========================================
[2026-09-03 10:12:43] Test 1: GPU monitoring via DCGM
[2026-09-03 10:12:43] =========================================
[2026-09-03 10:12:45] Selected GPU node: gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-827s
...
[2026-09-03 10:23:58] Node gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-827s is uncordoned and ready ✓
[2026-09-03 10:23:58] Checking for a succeeded ValidationRequest for node gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-827s...
[2026-09-03 10:23:58] ValidationRequest gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-827s-validation-b02b9f33514b for node gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-827s succeeded
[2026-09-03 10:23:58] Test 1 PASSED ✓
[2026-09-03 10:23:58] Waiting for syslog-health-monitor to initialize (60s)...
[2026-09-03 10:24:58] ======================================================
[2026-09-03 10:24:58] Test 2: XID monitoring via syslog triggers RESTART_VM
[2026-09-03 10:24:58] ======================================================
[2026-09-03 10:25:00] Selected GPU node: gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-5tw3 (has healthy syslog-health-monitor)
...
[2026-09-03 10:37:43] Node gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-5tw3 is uncordoned and ready ✓
[2026-09-03 10:37:43] Checking for a succeeded ValidationRequest for node gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-5tw3...
[2026-09-03 10:37:43] ValidationRequest gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-5tw3-validation-5d6ded699532 for node gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-5tw3 succeeded
[2026-09-03 10:37:44] Node debug pod deleted
[2026-09-03 10:37:44] Test 2 PASSED ✓
...
[2026-09-03 10:37:44] All tests PASSED =========================================
```

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [X] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added optional post-remediation validation using configurable templates and health-event rules.
  - Validation requests can preserve node scheduling gates until validation completes.
  - Added metrics for validation requests initiated, completed, and timed.
  - Added the `vr` short name for ValidationRequests.
  - Improved recognition and reuse of completed full and partial node drains.
  - Validation requests now support deduplicated tests and health-event tracking.
  - Added validation-related cordon and uncordon metadata.

- **Bug Fixes**
  - Diagnostic commands now retry transient runtime errors before failing.
  - Improved provider-resource cleanup and retry handling.
  - Improved conflict-safe node cordon and uncordon operations.


distros/kubernetes/nvsentinel/values-tilt-arm64.yaml
```diff
@@ -19,3 +19,13 @@ mongodb-store:
       repository: "dlavrenuek/bitnami-mongodb-arm"
       tag: "8.0.4"
 
+lifecycle-manager:
+  config:
+    tests:
+      dcgm-diag-test:
+        image: alpine:3.21
```
lalitadithya: can we consider using https://gallery.ecr.aws/docker/library/alpine here to avoid rate limits?

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1839,6 +1898,9 @@ func (r *Reconciler) performUncordon(
 		return false, nil
 	}
 
+	annotationsToBeRemoved, taintsToBeRemoved, isUnCordon = r.triggerValidationOnUncordon(ctx, span, event, annotations,
```
lalitadithya: nit: 

```suggestion
	annotationsToBeRemoved, taintsToBeRemoved, isUnCordon = r.triggerValidationOnUnquarantine(ctx, span, event, annotations,
```

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1884,6 +1946,78 @@ func (r *Reconciler) performUncordon(
...
+		taintsToBeRemoved = nil
+		isUnCordon = false
+
+		span.SetAttributes(attribute.Bool("fault_quarantine.validation_request.created", true))
+	}
+
+	return annotationsToBeRemoved, taintsToBeRemoved, isUnCordon
+}
+
+func (r *Reconciler) checkIfValidationRequested(ctx context.Context, event *protos.HealthEvent,
```
lalitadithya: nit: we need to come up with a better name for this function since it not only checks if validation was requested, but it also creates the validation as well

distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
```diff
@@ -205,3 +205,48 @@ ruleSets:
...
+  version: "v1alpha1"
+  kind: "ValidationRequest"
+  templateFileName: "validation-request-template.yaml"
+  templates:
+    "validation-request-template.yaml": |
+      apiVersion: nvsentinel.nvidia.com/v1alpha1
+      kind: ValidationRequest
+      metadata:
+        name: {{ .NodeName }}-validation-{{ .SessionID }}
+        labels:
```
lalitadithya, line 249: can we a label to indicate that this was created by the fault quarantine module? This will make it easy for an operator select all the validation requests created post remediation. Would selection by the node name also be possible? For the latter, I think we can add it as future scope as well

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1884,6 +1946,78 @@ func (r *Reconciler) performUncordon(
...
+	annotations map[string]string, annotationsToBeRemoved []string, taintsToBeRemoved []config.Taint,
+	isUnCordon bool) ([]string, []config.Taint, bool) {
+	if _, exists := annotations[common.QuarantineValidationHealthEventAnnotationKey]; !exists {
+		return annotationsToBeRemoved, taintsToBeRemoved, isUnCordon
+	}
+
+	annotationsToBeRemoved = append(annotationsToBeRemoved, common.QuarantineValidationHealthEventAnnotationKey)
+
+	validationRequestCreated, err := r.checkIfValidationRequested(ctx, event, annotations)
+	if err != nil {
```
XRFXLP, line 1989: If this fails, do we still want to remove the annotation `QuarantineValidationHealthEventAnnotationKey`?

fault-quarantine/pkg/config/config.go
```diff
@@ -66,10 +66,27 @@ type RuleSet struct {
 	Taint    Taint  `toml:"taint"`
 	Label    Label  `toml:"label"`
 	Cordon   Cordon `toml:"cordon"`
+	// Tests lists post-remediation validation test names. This is only populated for Validation.RuleSets
```
XRFXLP: RuleSet is shared between rule-sets (quarantine) and validation.ruleSets with no load-time validation, so Taint/Label/Cordon under validation.ruleSets and Tests under rule-sets are silently ignored without any error/log

The one field isn't safe to ignore though: `Match` is evaluated in both sections, just against different outcomes (taint/cordon vs. test selection). So a copy-pasted or stale CEL expression under validation.ruleSets won't no-op and it'll misfire test selection instead. Given how easy that copy-paste is between two identical-looking blocks, could we split into distinct QuarantineRuleSet/ValidationRuleSet structs, or add a load-time check rejecting fields each section doesn't use?

fault-quarantine/pkg/validation/validation.go
```diff
@@ -0,0 +1,325 @@
...
+a. HealthEvent de-duplication: as discussed above, we de-duplicate HealthEvents with the same HealthEventKey by
+leveraging the quarantineHealthEvent.
+b. Test de-duplication within 1 HealthEvent: if a validation rule-set evaluation results in the same test being
+required multiple times for the same fault, we will ensure that only unique tests are requested per event.
+c. Test de-duplication across HealthEvents: this isn't relevant to the node annotation update here and is covered
+in FetchValidationTestsFromQuarantineSession when the node is unquarantined. When processing all HealthEvents
+which required a validation test, we will build up a unique list of tests requested across all events to include
+as part of a single validation test.
+*/
+func (c *ValidationClient) UpdateQuarantineValidationAnnotation(ctx context.Context,
```
XRFXLP, line 142: can you use key-ing mechanism like we've for quarantineHealthEvent so it doesn't grow without bounds on flappy errors?

lifecycle-manager/internal/controller/validationrequest_provider.go
```diff
@@ -248,37 +257,46 @@ func (r *ValidationRequestReconciler) renderTestGroupObject(vr *v1alpha1.Validat
...
-	if len(suffix) >= validation.DNS1123LabelMaxLength {
-		return strings.TrimRight(suffix[:validation.DNS1123LabelMaxLength], "-")
-	}
+	return hashTruncate(baseObjectName, maxBaseObjectName) + attemptSuffix
+}
 
-	remaining := validation.DNS1123LabelMaxLength - len(suffix) - 1
-	if remaining <= 0 {
-		return suffix
+func hashTruncate(base string, maxBase int) string {
```
XRFXLP, line 286: we've kinda a similar mechanism in https://github.com/NVIDIA/NVSentinel/blob/main/janitor/pkg/controller/gpureset_controller.go#L889 and I believe also in preflight. Should we consolidate in commons?

commons/pkg/drain/drain.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		entity, err := getPartialDrainEntity(healthEvent)
+		if err != nil {
+			return false, fmt.Errorf("evaluating partial drain entity for health event %s on node %s: %w",
+				event.Id, nodeName, err)
+		}
+
+		drainStatus := status.HealthEventStatus.UserPodsEvictionStatus.Status
+		drainCompletedForEvent := drainStatus == datastore.StatusSucceeded || drainStatus == datastore.AlreadyDrained
```
KaivalyaMDabhadkar: This now treats AlreadyDrained the same as Succeeded. ND also writes AlreadyDrained for events with drainOverrides.skip, where nothing was actually evicted. A later fatal event on that node would then skip its real drain and we could reboot with workloads still running. Could we ignore skip events here or only trust AlreadyDrained when it came from a real drain?

distros/kubernetes/nvsentinel/charts/fault-quarantine/templates/configmap.yaml
```diff
@@ -71,3 +71,50 @@ data:
...
+      templateMountPath = "/etc/config"
+    {{- end }}
+
+    {{- range .Values.validation.ruleSets }}
+    [[validation.ruleSets]]
+      enabled = {{ if hasKey . "enabled" }}{{ .enabled }}{{ else }}true{{ end }}
+      version = {{ .version | quote }}
+      name = {{ .name | quote }}
+      priority = {{ .priority | default 0 }}
+      tests = {{ .tests | toJson }}
```
KaivalyaMDabhadkar: If a rule set has no tests, this renders tests = null which is invalid TOML, and FQ crashloops on startup. This block also renders when validation is disabled. Could we default to `[]` and move the range inside the enabled check?

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1711,12 +1745,37 @@ func (r *Reconciler) addEventToAnnotation(
...
+// updateValidationAnnotation adds or updates a HealthEvent to the node's quarantineValidationHealthEvent if
+// validation is enabled, one or more validation.ruleSets exist, and the given HealthEvent requires one or more tests
+// when evaluated against validation.ruleSets.
+func (r *Reconciler) updateValidationAnnotation(ctx context.Context, annotationsMap map[string]string,
+	event *protos.HealthEvent) error {
+	if r.validationClient == nil {
+		return nil
+	}
+
+	existing := annotationsMap[common.QuarantineValidationHealthEventAnnotationKey]
```
KaivalyaMDabhadkar, line 1780: `existing` is always empty here because annotationsMap is a fresh map. If two events for the same node race here the second overwrites the validation annotation and the first event's tests are lost. Could this key be merged like quarantineHealthEvent is or run inside the node update like Case 2?

fault-quarantine/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: When a VR is created the node stays cordoned but this still cleans up as if the node had been released. It removes the cordon-by, cordon-reason, cordon-timestamp and state labels, sets current_quarantined_nodes to 0 and counts the node as unquarantined. Later when lifecycle manager uncordons the node, nothing writes the uncordon-by label. Could we keep the labels and metrics as they are when a VR was created?

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1884,6 +1946,78 @@ func (r *Reconciler) performUncordon(
...
+func (r *Reconciler) triggerValidationOnUncordon(ctx context.Context, span trace.Span, event *protos.HealthEvent,
+	annotations map[string]string, annotationsToBeRemoved []string, taintsToBeRemoved []config.Taint,
+	isUnCordon bool) ([]string, []config.Taint, bool) {
+	if _, exists := annotations[common.QuarantineValidationHealthEventAnnotationKey]; !exists {
+		return annotationsToBeRemoved, taintsToBeRemoved, isUnCordon
+	}
+
+	annotationsToBeRemoved = append(annotationsToBeRemoved, common.QuarantineValidationHealthEventAnnotationKey)
+
+	validationRequestCreated, err := r.checkIfValidationRequested(ctx, event, annotations)
```
KaivalyaMDabhadkar: If a user had already cordoned the node we still create a VR and LM will uncordon it when the test passes, undoing the user cordon. And if a new fatal event quarantines the node again while the VR is running, LM still uncordons on success and FQ then thinks an user did it and drops the new session. Could we skip the VR when the cordon was pre-existing, and have LM re-check quarantine before uncordoning?

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -1884,6 +1946,78 @@ func (r *Reconciler) performUncordon(
...
+		metrics.ProcessingErrors.WithLabelValues("create_validation_request_error").Inc()
+		tracing.RecordError(span, err)
+		span.SetAttributes(
+			attribute.String("fault_quarantine.error.type", "create_validation_request_error"),
+			attribute.String("fault_quarantine.error.message", err.Error()),
+		)
+	}
+
+	if validationRequestCreated {
+		taintsToBeRemoved = nil
```
KaivalyaMDabhadkar: When we create a VR we leave the FQ taints on the node but delete the annotation that records them. LM only removes the cordon by default, so once validation passes the node is uncordoned but keeps its NoSchedule taint forever, and FQ has no way to find it later. Could we remove the taints here and hand over only the cordon, or keep the taint annotation until LM is done? We can also add a e2e test that the taint is gone after the VR succeeds

fault-quarantine/pkg/reconciler/reconciler.go
```diff
@@ -2084,33 +2219,23 @@ func (r *Reconciler) handleManualUncordon(nodeName string) error {
 	slog.DebugContext(ctx, "Retrieved node annotations for manual uncordon",
 		"node", nodeName, "annotationCount", len(annotations))
 
-	annotationsToRemove := []string{}
+	// Remove the applied taints annotation (but keep the taints themselves on the node).
+	annotationsToRemove := appendIfPresent(annotations, nil,
+		common.QuarantineHealthEventAppliedTaintsAnnotationKey,
+		common.QuarantineHealthEventAnnotationKey,
+		common.QuarantineHealthEventIsCordonedAnnotationKey,
```
KaivalyaMDabhadkar: handleManualUncordon clears the validation annotation here but handleManualUntaint below still leaves it behind. A later unrelated quarantine could then create a VR from those old events. Could handleManualUntaint use the same list?

lifecycle-manager/internal/controller/validationrequest_controller.go
```diff
@@ -669,6 +687,38 @@ func (r *ValidationRequestReconciler) startPendingTestGroup(ctx context.Context,
...
+	}
+
+	previousAttempt := currentPendingTestGroup.Attempts[len(currentPendingTestGroup.Attempts)-1]
+
+	deleted, err := r.checkTestGroupObjectDeleted(ctx, currentPendingTestGroup, previousAttempt.ObjectName)
+	if err != nil {
+		return false, fmt.Errorf("checking previous attempt %q is deleted: %w", previousAttempt.ObjectName, err)
+	}
+
+	return !deleted, nil
```
KaivalyaMDabhadkar, line 732: A retry now waits until the old job is fully deleted. If its pod cannot terminate for example because the kubelet on that node is down, the job never goes away and the VR stays running forever with nothing to wake it up. Could we put a time limit on this wait?

tests/uat/tests.sh
```diff
@@ -362,11 +368,52 @@ wait_for_gpu_reset() {
     wait_for_node_unquarantine "$node"
 }
 
+verify_validation_request() {
+    local node=$1
+    local current_ts=$2
+
+    log "Checking for a succeeded ValidationRequest for node $node..."
```
lalitadithya, line 383: Can we check if the feature is enabled before testing for it?

distros/kubernetes/nvsentinel/values-tilt-arm64.yaml
natherz97: Done.

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: Good callout, updated to createValidationRequestIfRequested.

distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml
natherz97: Updated ValidationRequests created by fault-quarantine to have the following labels:
```
      labels:
        app.kubernetes.io/managed-by: nvsentinel
        nvsentinel.nvidia.com/created-by: fault-quarantine
```
For listing ValidationRequests by node, we could add a label like nvsentinel.nvidia.com/node-name that is only added by fault-quarantine since that client currently only targets a single node per request. I think we can punt on that since the validation-session annotation per node can already be used to extract all blocking ValidationRequests per node. If we want to differentiate this by fault-quarantine later, we can add the additional label.

fault-quarantine/pkg/validation/validation.go
natherz97: We get the same de-duping behavior from the HealthEventKey used to de-dupe quarantineHealthEvent by only calling UpdateQuarantineValidationAnnotation when a new event is added to the existing quarantineHealthEvent annotation. Specifically, we only call UpdateQuarantineValidationAnnotation if a new unhealthy event (either Quarantined or AlreadyQuarantined) actually required updating the quarantineHealthEvent annotation. I opted for this approach since it's simpler and doesn't require us to add re-implement HealthEventKey de-duping logic to our new annotation. 

The one downside of this approach is that if a health event continually fires and resolves AND this does not result in an unquarantine event (meaning some other unhealthy event existing as part of quarantineHealthEvent annotation), it could be possible that we get multiple events tracked for the same HealthEventKey. If there was a single flappy health event, it wouldn't matter since we'd create the ValidationRequest immediately and remove quarantineValidationHealthEvent at that point.

Let me know your thoughts on this since we'd be adding annotation management complexity to satisfy this specific case for 1 persistent error occurring with 1 flapping error.

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: That is a good callout. If we go with that behavior and keep QuarantineValidationHealthEventAnnotationKey, it seems like we should skip removing all annotations and maintain any existing cordon/taint, I don't think we'd get a benefit from only keeping the QuarantineValidationHealthEventAnnotationKey present. If we skipped removing annotations and kept the node cordoned, we'd need an oncall to manually uncordon or require the breakfix pipeline to be re-triggered for it to retry the ValidationRequest creation. What do you think?

fault-quarantine/pkg/config/config.go
natherz97: I went with the first recommendation and added 2 separate types for QuarantineRuleSet and ValidationRuleSet where shared fields are included in an embedded struct named RuleSetMeta:
```
type QuarantineRuleSet struct {
  RuleSetMeta
  Taint  Taint  `toml:"taint"`
  Label  Label  `toml:"label"`
  Cordon Cordon `toml:"cordon"`
}

type ValidationRuleSet struct {
  RuleSetMeta
  Tests []string `toml:"tests"`
}
```
We could also make our TOML decoding stricter and reject any config with fields that could not be unmarshaled to our Go struct but will hold off on that for now since we'd be adding a backwards incompatibility risk to fault-quarantine where an existing config with unrecognized fields would start failing.

lifecycle-manager/internal/controller/validationrequest_provider.go
natherz97: Yes, we should. I'm planning a smaller follow-up PR to add improved logging and documentation. If that works for you, let me do this consolidation in that follow-up PR to avoid making this PR more complex and also touching preflight/GPU reset.

commons/pkg/drain/drain.go
natherz97: Great catch! Updated the check for AlreadyDrained to also require that the event wasn't skipped.
```
wasSkipped := healthEvent.DrainOverrides != nil && healthEvent.DrainOverrides.Skip
drainCompletedForEvent := drainStatus == datastore.StatusSucceeded || (drainStatus == datastore.AlreadyDrained && !wasSkipped)
```

distros/kubernetes/nvsentinel/charts/fault-quarantine/templates/configmap.yaml
natherz97: Thanks for catching this, I should've enabled validation in e2e tests and then in a follow-up for UAT to ensure that Helm chart is still valid if validation is not enabled. I made both fixes in the revision.

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: Can you clarify what you mean by, "Could we keep the labels and metrics as they are when a VR was created?"

I think we should continue with the existing behavior except skip writing the uncordon-by label  and I can add logic to the validation-controller to start writing that when the uncordon actually occurs.

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: > If a user had already cordoned the node we still create a VR and LM will uncordon it when the test passes, undoing the user cordon.

Compared to fault-quarantine, we need to support LM uncordoning nodes from pre-existing cordons. Specifically, we will have NodeWright apply a cordon to new nodes which we want LM to remove after validation succeeds. LM also supports a mode where it can automatically cordon nodes if not already cordoned so I think we could only preserve pre-existing cordons if we're running in that mode. Additionally, a user could configure LM to never uncordon on success so if they need to keep nodes cordoned after validation they could use that setting (but it would apply to all validation requests).

> And if a new fatal event quarantines the node again while the VR is running, LM still uncordons on success and FQ then thinks an user did it and drops the new session.

LM will never uncordon a node that has the quarantineHealthEvent annotation so we should never run into this case where fault-quarantine detects a manual uncordon from LM. This is configured in the as part of the NodeReadinessCritiera where we fail an existing test attempt and block a new attempt from starting if this criteria is violated. So if a new unhealthy event fires during validation, LM will fail the current attempt and then be stuck pending until NVSentinel completes remediation. At this point, it will retry the old request as long as there are remaining attempts. NVSentinel does preserve pre-existing cordons so regardless of whether this new fault required validation, it will keep the node cordoned after the unquarantine event and LM will still be responsible for the uncordon.

Let me know if that answers your question.

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: My current plan was if someone applied taints as part of fault-quarantine, they could configure the validation-controller to remove both the cordon and those taints after success like this:
```
config:
  schedulingGate:
    cordon:
      remove: true
    taints:
      - key: "nvidia.com/gpu-xid-error"
        value: "true"
        effect: "NoSchedule"
        remove: true
```
So it would require configuring the same set of taints in both fault-quarantine + validation-controller but it would avoid the need for validation-controller to consume the annotation specifying the taints from fault-quarantine. To keep this simple, maybe we can start with only keeping the node cordon in place for post-remediation validation and as a follow-up we can add this behavior to preserve taints? So we'd only do something like this and not touch taintsToBeRemoved:

```
	if validationRequestCreated {
		isUnCordon = false
	}
```

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: Fixed! I added a shared variable called manualCleanupQuarantineAnnotationKeys with a list of all annotations so we don't have to keep maintaining the annotations in the 2 separate functions.

lifecycle-manager/internal/controller/validationrequest_controller.go
natherz97: I'm trying to think what we would do after that timeout if the resource is stuck deleting indefinitely. It's possible that a follow-up TestGroup targets the same node which has an old resource. In that case, it would be preferable to fail the ValidationRequest rather than proceed with the request (since it may be possible that the object stuck deleting actually succeeded its test). 

There's other cases that can also block a ValidationRequest:
- A node referenced in the request has a NodeReadinessViolation
- A node referenced in the request already has a different running ValidationRequest.
- A previous TestGroup attempt has a resource stuck deleting. This could be caused by foregroundDeletion waiting on

We enforce a timeout on the underlying test resource completing but we don't have an overall timeout (specifically if we're blocked for a long period between starting tests). For example, it's hard to estimate how long an NVSentinel remediation which overlaps with validation will take so right now we support waiting indefinitely until the node exits NVSentinel remediation and we can retry any test. 

What if we kept the current behavior and then we can add an overall ValidationRequest timeout later if we find it necessary (which would capture the stuck deletion case you're calling out along with the node not ready issue)? Since nothing is waiting on the ValidationRequest status, there's no user behavior difference (other than potentially losing retries) if we keep the request pending since the node will be stuck cordoned in either case.

tests/uat/tests.sh
natherz97: Yes, good call. Here's a UAT run that successfully skips the ValidationRequest check based on the fault-quarantine configmap not enabling the feature:
```
[2026-09-08 12:59:15] =========================================
[2026-09-08 12:59:15] Test 1: GPU monitoring via DCGM
[2026-09-08 12:59:15] =========================================
[2026-09-08 12:59:17] Selected GPU node: gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-827s
...
[2026-09-08 13:04:21] Node gke-xmnvzter-dgxc-k8s-gc-customer-gpu-eb86dce7-827s is uncordoned and ready ✓
[2026-09-08 13:04:21] Post-remediation validation is not enabled, skipping ValidationRequest check
[2026-09-08 13:04:21] Test 1 PASSED ✓
```

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: Good catch, I updated QuarantineNodeAndSetAnnotations to merge QuarantineValidationHealthEventAnnotationKey in a manner that's consistent with QuarantineHealthEventAnnotationKey.

fault-quarantine/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: Sorry I could have been clearer. What I meant was that after this step the node is still cordoned, but we remove cordon by, cordon-reason, cordon timestamp and the state label and set current_quarantined_nodes to 0. So during validation the FQ labels and metrics say that the node was released, even though it is still cordoned. LM's validation-session annotation does explain it, so it is not invisible but just inconsistent.
Your proposal sounds good to me though. If LM writes uncordon by and uncordon timestamp when it actually lifts the cordon (using the same label prefix FQ uses) then that covers what I was after

fault-quarantine/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: Yes that makes sense, thanks for explaining

fault-quarantine/pkg/reconciler/reconciler.go
KaivalyaMDabhadkar: Agreed, keeping only the cordon for now sounds good

lifecycle-manager/internal/controller/validationrequest_controller.go
KaivalyaMDabhadkar: > What if we kept the current behavior and then we can add an overall ValidationRequest timeout later

cool cool, sounds good

natherz97: I added this comment in the controller file so that we keep this context:

There is no overall timeout applied to a ValidationRequest. Our 2 initial clients (node-validation-controller and fault-quarantine) do not poll for the ValidationRequest status to enter a terminal state. A node will remain cordoned whether the ValidationRequest entered a terminal status or is blocked (on either a NodeReadinessViolation or stuck TestProvider resource deletion). In the future, we could add an optional timeout to requests if there are clients watching for ValidationRequests entering a terminal status.

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: Sounds good. To summarize this is the label/annotation management we'll follow:

#### 1. With validation enabled:

1a. Without a pre-existing cordon:
```
# On quarantine by fault-quarantine:
labels:
  cordon-by: NVSentinel
  cordon-reason: Basic-Match-Rule
  cordon-timestamp: 2026-06-30T...
  nvsentinel-state: quarantined
annotations:
  quarantineHealthEvent: '{"events":[...]}'
  quarantineHealthEventIsCordoned: "True"
  quarantineHealthEventAppliedTaints: '[{"key":"AggregatedNodeHealth","value":"False","effect":"NoSchedule","preExisting":true}]'
  quarantineValidationHealthEvent: '[{"...","tests":["dcgm-diag-test"]}]'
spec:
  unschedulable: true

# On unquarantine by fault-quarantine:
labels:
  cordon-by: NVSentinel
  cordon-reason: Basic-Match-Rule
  cordon-timestamp: 2026-06-30T...
spec:
  unschedulable: true

# On uncordon by validation-controller:
labels:
  uncordon-by: NVSentinel
  uncordon-timestamp: 2026-06-30T...
  uncordon-reason: Basic-Match-Rule-removed
```

1b. With a pre-existing cordon:
```
# On quarantine by fault-quarantine:
labels:
  nvsentinel-state: quarantined
annotations:
  quarantineHealthEvent: '{"events":[...]}'
  quarantineValidationHealthEvent: '[{"...","tests":["dcgm-diag-test"]}]'
  quarantineHealthEventIsCordoned: "True"
  quarantineHealthEventCordonPreExisting: "True"
  quarantineHealthEventAppliedTaints: '[{"key":"AggregatedNodeHealth","value":"False","effect":"NoSchedule","preExisting":true}]'
spec:
  unschedulable: true   

# On unquarantine by fault-quarantine:
labels: {}
annotations: {}
spec:
  unschedulable: true 

# On uncordon by validation-controller:
labels:
  uncordon-by: NVSentinel
  uncordon-timestamp: 2026-06-30T...
  uncordon-reason: Basic-Match-Rule-removed
```

1c. Manual uncordon before validation
```
# On quarantine by fault-quarantine
labels:
  cordon-by: NVSentinel
  cordon-reason: Basic-Match-Rule
  cordon-timestamp: 2026-06-30T...
  nvsentinel-state: quarantined
annotations:
  quarantineHealthEvent: '{"events":[...]}'
  quarantineValidationHealthEvent: '[{"...","tests":["dcgm-diag-test"]}]'
  quarantineHealthEventIsCordoned: "True"
  quarantineHealthEventAppliedTaints: '[{"key":"AggregatedNodeHealth","value":"False","effect":"NoSchedule","preExisting":true}]'
spec:
  unschedulable: true

# On manual uncordon:
labels:
  cordon-by: NVSentinel
  cordon-reason: Basic-Match-Rule
  cordon-timestamp: 2026-06-30T...
annotations:
  quarantinedNodeUncordonedManually: "True"
spec:
  unschedulable: false
```
#### 2. With validation disabled:
2a. Without a pre-existing cordon:
```
# On quarantine by fault-quarantine:
labels:
  cordon-by: NVSentinel
  cordon-reason: Basic-Match-Rule
  cordon-timestamp: 2026-06-30T...
  nvsentinel-state: quarantined
annotations:
  quarantineHealthEvent: '{"events":[...]}'
  quarantineHealthEventIsCordoned: "True"
  quarantineHealthEventAppliedTaints: '[{"key":"AggregatedNodeHealth","value":"False","effect":"NoSchedule","preExisting":true}]'
spec:
  unschedulable: true

# On unquarantine by fault-quarantine:
labels:
  uncordon-by: NVSentinel
  uncordon-timestamp: 2026-06-30T...
  uncordon-reason: Basic-Match-Rule-removed
spec:
  unschedulable: false
```

2b. With a pre-existing cordon:
```
# On quarantine by fault-quarantine:
labels:
  nvsentinel-state: quarantined
annotations:
  quarantineHealthEvent: '{"events":[...]}'
  quarantineHealthEventIsCordoned: "True"
  quarantineHealthEventCordonPreExisting: "True"
  quarantineHealthEventAppliedTaints: '[{"key":"AggregatedNodeHealth","value":"False","effect":"NoSchedule","preExisting":true}]'
spec:
  unschedulable: true   

# On unquarantine by fault-quarantine:
labels: {}
annotations: {}
spec:
  unschedulable: true 
```

2c. Manual uncordon:
```
# On quarantine by fault-quarantine:
labels:
  cordon-by: NVSentinel
  cordon-reason: Basic-Match-Rule
  cordon-timestamp: 2026-06-30T...
  nvsentinel-state: quarantined
annotations:
  quarantineHealthEvent: '{"events":[...]}'
  quarantineHealthEventIsCordoned: "True"
  quarantineHealthEventAppliedTaints: '[{"key":"AggregatedNodeHealth","value":"False","effect":"NoSchedule","preExisting":true}]'
spec:
  unschedulable: true

# On manual uncordon:
labels:
  cordon-by: NVSentinel
  cordon-reason: Basic-Match-Rule
  cordon-timestamp: 2026-06-30T...
annotations:
  quarantinedNodeUncordonedManually: "True"
spec:
  unschedulable: false
```

KaivalyaMDabhadkar: Yes, that sounds good. Thanks for the summary!

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: Updated behavior to always remove taints and only preserve a cordon. This behavior is tested by TestE2E_ValidationRequestCreatedWhenEventDrained.

fault-quarantine/pkg/validation/validation.go
natherz97: Discussed offline, we will go with the current approach while limiting the total number of events tracked in quarantineValidationHealthEvent to 10. Long-term we can consider if we want to implement event deduplication directly for this annotation.

fault-quarantine/pkg/reconciler/reconciler.go
natherz97: As discussed offline, we will keep the node quarantined if we fail to create a ValidationRequest.

tests/uat/tests.sh
```diff
@@ -362,11 +368,60 @@ wait_for_gpu_reset() {
...
+        local created=$(echo "$vr" | jq -r '.metadata.creationTimestamp')
+        local vr_node=$(echo "$vr" | jq -r '.spec.nodes[0].name')
+
+        if [ -z "$created" ] || [ "$created" == "null" ]; then
+            continue
+        fi
+        local created_ts=$(get_epoch_time "$created")
+
+        if [ "$created_ts" -gt "$current_ts" ] && [ "$vr_node" == "$node" ]; then
+            matching_vr="$vr"
```
XRFXLP, line 400: Do we need to pick up the latest VR here? what if there are multiple VRs that are of the same node all that are created after current time?

XRFXLP (approved):

KaivalyaMDabhadkar (approved):

tests/uat/tests.sh
natherz97: We're checking that a ValidationRequest for the given was created between the cordon and uncordon time for the given node. If multiple ValidationRequests existed, the node wouldn't be uncordoned until both are in successful status so it shouldn't matter which ValidationRequest we are checking for success. The main purpose of this check was to ensure that the validation-controller is the entity doing the uncordon and not fault-quarantine so a single ValidationRequest in successful status should accomplish that. The test for GPUReset is doing the same time-based check without ensuring that a single CRD exists within that time. I'll plan to keep it consistent for now between both of these time-based checks for CRD creation.

PR #1751, opened 2026-09-05
feat(node-drainer): select drain modes with pod labels

### LINKED ISSUE #1689: [Feature]: Node Drainer configuration should use k8s selectors
### Prerequisites

### Code of Conduct

### Feature Summary

Configurability of https://docs.nvidia.com/nvsentinel/configuration/node-drainer/#user-namespaces is a little bit simple at the moment, we can configure actions per namespace. I would suggest replacing the namespace-based matching with more complex rules (and deprecating the simple per-namespace matching mechanism).

### Problem/Use Case

One thing that I would have liked, when looking at this configuration, would have been to match Pods belong to StatefulSets, or matching pods with particular labels. As an example, very often a StatefulSet can be immediately terminated and will restart on a different node so I would be tempted to first have a "if stateful, then immediate" rule.

Adjacently, I'm not sure that all pods in a namespace can have the same behavior so matching per behavior is a bit coarse, however i'm convinced that I could convince users who care, to label their pods with labels to opt in or out of a behavior.

### Proposed Solution

I wonder if the configuration language can / should be expanded to use selectors-based resource matching, similarly to how other abstractions do this. As an example Kyverno is matching objects with something looking like https://htmlpreview.github.io/?https://github.com/kyverno/kyverno/blob/main/docs/user/crd/index.html#kyverno.io/v1.MatchResources and this allows a very flexible.

Maybe something like:

```yaml
- mode: Immediate
  match:
    any:
      - resources:
          selector:
            matchLabels:
              nodedrainer: Immediate
      - resources:
          selector:
            kinds:
              - StatefulSet
- mode: DeleteAfterTimeout
  match:
    any:
      - resources:
          selector:
            matchLabels:
              nodedrainer: DeleteAfterTimeout
```

Where the first rule which applies matches?

If you're not familiar with the syntax https://kyverno.io/docs/policy-types/cluster-policy/match-exclude/ is how those things look like.

I'm not attached to this particular solution shape but in general I think you could maybe match objects through this kind of mechanism and make it a lot easier to customize behaviors.

Thanks!

### Component

Deployment/Config

Saibernard:
## Summary

Workloads in the same namespace can have different shutdown needs. This adds `podDrainPolicies` so, for example, a worker can be evicted immediately while a training job in the same namespace is allowed to finish.

Policies match pod labels and are checked in order. The first match sets the drain mode. Operators choose either `userNamespaces` or `podDrainPolicies`; configurations containing both are rejected. Helm users enabling policies set `userNamespaces: []` to clear the default namespace rule. Unmatched pods remain outside the drain scope. These rules also apply when a drain is retried or node-drainer restarts.

The design document describes the memory impact and includes repeatable cache-allocation measurements. This adds the pod-label selection discussed in #1689.

## Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected

- [ ] Core Services
- [x] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing

The full node-drainer suite passed with race detection and integration enabled: 131 passed, with one existing demonstration test skipped. This includes the revised policy cases and legacy namespace-only regressions.

All 59 Helm tests passed, along with both Go module lint targets, node-drainer vet/build, dependency checks and Helm rendering. The documented Helm example, default configuration and updated E2E fixture all load successfully through the TOML validator. The new conflicting-configuration regression fails when its rejection guard is removed. Cache-allocation results were consistent across three benchmark runs.

Before this review follow-up, local end-to-end testing passed for 13 selected PostgreSQL scenarios and five MongoDB scenarios using simulated GPU devices. Those cluster runs have not been repeated for the either/or change; both updated E2E binaries compile. The API regression suite covers mixed modes, restart recovery, timeouts, unmatched-pod preservation, partial GPU drains, dry runs and relabelling.

The earlier local stack also hit an existing PostgreSQL aggregation error in health-events-analyzer. That component is unchanged by this PR.

The PR also corrects the CRD fixture path so the existing custom-drain integration test can run.

- [x] Tests pass locally

## Checklist

- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added configurable pod-drain policies based on pod labels and optional namespace patterns.
  * Supports ordered matching with Immediate, AllowCompletion, and DeleteAfterTimeout modes.
  * Policies respond to label changes and support force-drain, partial-drain, dry-run, and restart scenarios.
  * Unmatched pods are excluded from draining when policy mode is enabled.
  * Added safeguards against actions on stale or replaced pods.
  * Pod policies cannot be combined with user namespace rules or custom draining.

* **Documentation**
  * Added configuration guidance and design documentation for pod-drain policies.


docs/designs/055-pod-drain-policies.md
lalitadithya, line 1: We have also done a lot of optimization around memory usage in the drainer recently, can we add a section here that talks about what the impact of this change (if any) would be on the overall memory footprint of the component?

docs/designs/055-pod-drain-policies.md
```diff
@@ -0,0 +1,59 @@
+# ADR-055: Node Drainer — Pod label policies
+
+## Context
+
+[Issue #1689](https://github.com/NVIDIA/NVSentinel/issues/1689) describes workloads with different interruption requirements sharing a namespace. Namespace-level drain configuration cannot express those differences. The existing drainer also passes namespace groups through immediate eviction, deadline deletion and completion checks, so filtering only during initial evaluation would still allow an action to affect unrelated pods.
+
+## Decision
+
+Add an optional ordered `podDrainPolicies` list. Each policy matches a standard Kubernetes pod label selector and an optional namespace-name glob, then selects an existing drain mode. The first matching policy wins; unmatched pods fall back to `userNamespaces`.
```
lalitadithya: Can we make this either/or to make the implmentation simpler? As in, the user can either specify `userNamespaces` or `podDrainPolicies` but not both?

Saibernard:
Thanks @lalitadithya, addressed both points in 63bd3596. `userNamespaces` and `podDrainPolicies` are now mutually exclusive, and the ADR includes the memory impact and allocation measurements. Local tests passed.

Could you run /ok to test when you get a chance?

docs/configuration/node-drainer.md
```diff
@@ -148,9 +148,39 @@ annotations:
 - **When disabled (`false`)**: All eligible pods in configured namespaces are evicted (default behavior)
 - Pods without GPU requests are preserved, maintaining critical infrastructure services
 
+## Pod Drain Policies
```
XRFXLP, line 151: need update here as well: https://github.com/NVIDIA/NVSentinel/blob/main/docs/node-drainer.md and in values-full.yaml (we can keep it empty).

node-drainer/pkg/evaluator/pod_policies.go
```diff
@@ -0,0 +1,88 @@
...
+	force := healthEvent.HealthEvent.GetDrainOverrides().GetForce()
+
+	filters := make(map[config.EvictMode]informers.PodFilter)
+	for _, mode := range []config.EvictMode{
+		config.ModeImmediateEvict, config.ModeDeleteAfterTimeout, config.ModeAllowCompletion,
+	} {
+		filters[mode] = e.podModeFilter(mode, force)
+	}
+
+	action := e.getAction(ctx, namespaces{
```
XRFXLP: nit: `allNamespaces` gets passed into all three mode buckets here, so `getAction` scans each namespace 3x (once per mode) plus a 4th time in the recheck below but only the filter changes per mode, not the namespace list. All local cache reads so no big deal, but could just list pods once per namespace and bucket them by mode instead of re-scanning 3-4x.

Saibernard:
Thanks @XRFXLP, both comments are addressed and the fixes are pushed. Could you run /ok to test when you get a chance?

XRFXLP (approved):

Issue #1704, opened 2026-08-31
[Bug]: health-events-analyzer re-ingests its own output, so the XID 31 rules self-trigger indefinitely and emit isFatal RUN_DCGMEUD for a fault that ended 10 days earlier


lfriedman-netllama:
### Prerequisites

- [x] I searched existing issues

### Code of Conduct

- [x] I agree to follow NVSentinel's Code of Conduct

### Describe the bug

`health-events-analyzer` writes its derived events into the same `HealthEvents` collection
it consumes, and those derived events carry the `errorcode` of the pattern they describe.
No rule stage filters on `healthevent.agent`, so a derived event is indistinguishable from a
detector event when rules match. Once one derived `errorcode: 31` event exists, it satisfies
the trigger condition of the XID 31 rules and keeps their rolling 24 hour window populated,
so those rules re-fire indefinitely on their own output.

On a 288 node GB200 fleet this produced a self-sustaining stream of pattern events for a
single node whose real XID 31 activity had ended 10 days earlier, including 10 events
carrying `RUN_DCGMEUD` with `isFatal: true`.

### Steps to reproduce

Observed in production rather than constructed, so this is the timeline as recorded in the
datastore for one node. Times are UTC.

1. The node produced 94 genuine XID 31 events, `agent: syslog-health-monitor`,
   `checkName: SysLogsXIDError`, from **2026-08-04 14:09:44 to 2026-08-21 10:10:06**, and
   then stopped. The node was rebooted around 2026-08-25, and its `dmesg` contains no XID 31
   at all.
2. **2026-08-30 08:12:37**: an unrelated XID 45 burst arrives on that node.
3. **2026-08-30 08:30:52**: `XIDErrorSoloNoBurst` matches and emits a derived event with
   `errorcode: 31`, `recommendedAction: NONE`, `isFatal: false`.
4. **2026-08-30 11:04:36**: `RepeatedXID31OnDifferentGPU` matches and emits, `errorcode: 31`.
5. **2026-08-30 11:41:39**: `RepeatedXID31OnSameGPU` matches and emits, `errorcode: 31`,
   `recommendedAction: RUN_DCGMEUD`, `isFatal: true`.
6. Still emitting as of **2026-08-31 10:08:11**, more than 24 hours later, with all three
   rules now sharing the same timestamps.

Totals for the analyzer across its entire lifetime, every one of them for this single node:

| checkName | recommendedAction | isFatal | errorcode | events |
[...]

PR #1762, opened 2026-09-07
feat(store-client): report change stream consumer lag

### LINKED ISSUE #1709: [Feature]: expose change stream consumer lag; a consumer 11 days behind presents as completely healthy

lfriedman-netllama:
Implements [ADR-054](docs/designs/054-changestream-lag-metrics.md) (merged in #1738). Closes #1709.

Four of the five change stream consumers had no way to tell "idle" from "stuck": `event-exporter`, `health-events-analyzer`, `node-drainer` and `fault-remediation` exposed nothing at all, and #1704 was diagnosed only after the fact.

Kept as one PR per @lalitadithya's note on #1709: it is one feature that needs end-to-end testing to confirm there are no regressions.

## What it measures

```
lag = now - max(lastEmptyBatchAt, lastEventReadAt)
```

An empty batch is the stream telling *this* consumer it is caught up, evaluated against its own filtered pipeline. That is what position age and head-minus-position both get wrong for the four filtered consumers, which are also the ones most likely to sit idle for weeks.

```text
change_stream_lag_seconds{client}  gauge  absent until known, then computed on each scrape
change_stream_lag_known{client}    gauge  0 until the first observation, then 1
```

The lag series is **absent rather than zero** while unknown, so a watcher that never started or is wedged before its first read shows up as `change_stream_lag_known == 0` instead of a healthy zero.

## Changes

- **`pkg/lagstate`** — the two timestamps, as a `Tracker` both providers embed, so they report lag identically. `pkg/client` already imports the mongo watcher, so a reporter there cannot be imported by a provider; providers own the facts, `pkg/client/lagreporter.go` owns the metrics.
- **MongoDB** — `Next` becomes `TryNext` with an explicit `MaxAwaitTime`, which is what makes the empty batch observable at all. Idle cost is unchanged: both issue a `getMore` the server holds for the await window.
- **Closed-cursor check** — `TryNext` reports a server-closed cursor as "no event and no error". The existing `Next` loop spins on that today, so this fixes a live bug as well as preventing the loop from recording "caught up" against a dead stream.
- **PostgreSQL** — no loop change. A poll that returns no rows already *is* the empty batch.
- **Caller-supplied registerer** — `store-client` defaults to the default registry; `fault-remediation` passes `crmetrics.Registry`. Without that, the one signal this adds would be missing entirely from the endpoint of a consumer that needs it.
- **Wrapper pass-throughs** — both watcher chains wrap two or three deep. Every wrapper passes `LagState` through, with compile-time `var _ lagstate.Provider` assertions so adding a wrapper without it fails the build rather than silently reporting an unknown lag.
- **`docs/METRICS.md`** — the metrics, two suggested alerts, and the three cases where lag reads zero while a consumer is still behind (replication lag, data skipped after a resume-token recovery, durable position).

## No new interface method

`ChangeStreamMetrics` is satisfied by whole-type assertion at `store-client/pkg/client/resume_token.go:79` and `fault-quarantine/pkg/eventwatcher/event_watcher.go:787`, so adding a method there would drop any implementation providing only `GetUnprocessedEventCount` and cost it the count metric it has. Lag state is a separate optional interface.

## Testing

`store-client` and `fault-remediation` suites pass, `golangci-lint` clean on both. The other four consumers build unchanged.

Unit tests cover: the tracker's monotonicity and concurrency, lag absent until observed, `lag_known` flipping, the more-recent-observation rule, clock skew clamped to zero, lag growing between scrapes, registration landing on a supplied registry and not the default one, per-client idempotence, the PostgreSQL empty poll / rows / out-of-order / drain-then-empty / failed-poll cases, `clusterTimeOf`'s decode type, the closed-cursor classification, and the adapter pass-through.

`fault-remediation/pkg/initializer` asserts both metrics reach the real `crmetrics.Registry`, since a test against the default registry would pass while that endpoint stayed empty.

Not covered by unit tests, and flagged in the ADR for the `kind`/Tilt suite: the live MongoDB cases (idle stream, narrow pipeline under load, a seeded backlog draining, a wedged loop). `envtest` cannot cover these, as it provides an API server and no datastore.

## Summary by CodeRabbit

* **New Features**
  * Added change-stream lag metrics for MongoDB and PostgreSQL watchers.
  * Metrics report consumer lag status, current lag, and recovery activity.
  * Added support for custom Prometheus registries and simultaneous reporting for multiple clients.
  * Metrics can be registered through the configured monitoring registry for easier integration with existing dashboards and alerts.

* **Documentation**
  * Added guidance for interpreting change-stream metrics, configuring Prometheus alerts, and understanding replication lag, skipped data, and durable positions.


lfriedman-netllama:
@lalitadithya this is the ADR-054 implementation for #1709, kept as one PR per your note there. Could you kick off CI with `/ok to test 105f61cc` when you get a chance?

lfriedman-netllama:
Both addressed in 7833740a.

**The registration collision is real, and worse than a duplicate-series risk.** I checked before changing anything: registering `client-a` then `client-b` against one registry produced a *single* series, labelled `client-a`. Because `client` is a variable label, per-client collectors carry identical descriptors, and Prometheus identifies a collector by its descriptors, so the second `Register` returned `AlreadyRegisteredError` and my caller swallowed it as success. That consumer's lag would simply never appear, which is the same false-healthy failure ADR-054 exists to remove, one layer up. Thank you for catching it.

Went with your first suggestion, one collector per registry: it now holds a client-to-provider map and emits a series per client at scrape time. I preferred that over making `client` a const label because it keeps one descriptor pair per registry and makes the per-client bookkeeping explicit, rather than relying on const-label values to keep descriptor IDs distinct. `TestRegisterChangeStreamLag_TwoClients_ExportsBoth` registers two clients and asserts both label values are exported for both metrics.

Test names: renamed, all 19 new ones across the six files, not just `lagstate_test.go`.

docs/METRICS.md
```diff
@@ -344,6 +345,54 @@ The CSP health monitor tracks cloud provider maintenance events and node health
 
 ---
 
+## Change Stream Metrics
+
+Emitted by `store-client`, so they appear on every module that reads the change stream:
```
KaivalyaMDabhadkar, line 350: Right now this is only true for the four postgres consumers. On mongo nothing is registered yet, and event exporter is not registered on either provider. Once the registration fixes land this becomes true, so please update it together with them

fault-remediation/pkg/initializer/lag_metrics_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (stubLagProvider) LagState() (lastEmptyBatch, lastEventRead time.Time) {
+	return time.Now(), time.Time{}
+}
+
+// This service serves only controller-runtime's registry, so store-client's change stream
+// metrics have to be registered there. A test against the default registry would pass while
+// /metrics stayed empty, which is the failure this covers.
+func TestRegisterChangeStreamLag_ControllerRuntimeRegistry_ExportsBothMetrics(t *testing.T) {
+	client.RegisterChangeStreamLag(crmetrics.Registry, t.Name(), stubLagProvider{})
```
KaivalyaMDabhadkar: This test uses a stub, so it passes even though the real mongo chain registers nothing. Could you wrap the watcher the same way production does before registering? There is a similar test in store-client/pkg/client/resume_token_test.go that could be a starting point

store-client/pkg/client/lagreporter.go
```diff
@@ -0,0 +1,203 @@
...
+	collectors   = map[prometheus.Registerer]*lagCollector{}
+)
+
+// RegisterChangeStreamLag exports the lag metrics for watcher, if it reports lag state. Watchers
+// that do not are left alone, so this is safe to call on any watcher.
+//
+// reg may be nil, in which case the default registry is used. Callers that serve a different
+// registry must pass it: a consumer serving only controller-runtime's registry would otherwise
+// never see these metrics.
+func RegisterChangeStreamLag(reg prometheus.Registerer, clientName string, watcher any) {
```
KaivalyaMDabhadkar, line 150: event exporter never reaches this function. Its watcher comes from the helper and the client factory, and neither calls RegisterChangeStreamLag. Could you add the call inside the factory's CreateChangeStreamWatcher? The default registry is fine for event exporter

store-client/pkg/datastore/providers/mongodb/adapter.go
```diff
@@ -163,6 +165,8 @@ func (a *AdaptedMongoStore) CreateChangeStreamWatcher(ctx context.Context, clien
 		return nil, fmt.Errorf("failed to create change stream watcher: %w", err)
 	}
 
+	client.RegisterChangeStreamLag(a.config.MetricsRegisterer, clientName, watcher)
```
KaivalyaMDabhadkar, line 168: The watcher passed in here is the resume-control wrapper, because that is what the factory returns. That wrapper does not have a LagState method, so the check inside RegisterChangeStreamLag fails and nothing gets registered. In practice no mongo consumer gets these metrics. Could you add LagState to the resume-control wrapper, and cover this path in a test?

lfriedman-netllama:
All four addressed in 22cc141f. Thank you — the fourth one means the feature did not work on MongoDB at all, which is the provider we actually run, and my own tests were the reason I could not see it.

**The resume-control wrapper.** You were exactly right. I proved it before changing anything: wrapping a watcher that *does* satisfy `lagstate.Provider` in `NewChangeStreamWatcherWithResumeControl` yields a value that does not, so the assertion inside `RegisterChangeStreamLag` returned early and nothing registered for any MongoDB consumer. `resumeControlChangeStreamWatcher` now passes `LagState` through, with a `var _ lagstate.Provider` assertion so a future wrapper cannot silently drop it again.

That is the third wrapper in this chain to catch me out, so the honest lesson is that adding pass-throughs to the wrappers I knew about was never sufficient — I should have traced what the factory actually returns rather than what the adapters return.

**event-exporter.** Also right, and it got nothing on either provider, not just MongoDB. I added the registration inside the factory's `CreateChangeStreamWatcher`, right after the wrap, on the default registry as you suggested. That is the one path event-exporter takes. Consumers that also register through their provider adapter with their own registry are unaffected, since registration is idempotent per registry and client.

**The fault-remediation test.** Fair hit: the stub made it pass while the real chain registered nothing, so it was testing a path production never takes — the exact failure it existed to catch. It now builds the chain the factory builds, wrapping the watcher in the resume-control wrapper before registering, and asserts `lagstate.Provider` is satisfied first so the reason is obvious if it regresses. Thanks for the `resume_token_test.go` pointer.

**docs/METRICS.md.** Corrected, and I have kept it in this commit as you asked rather than separately. It now states both providers explicitly, and notes that `fault-remediation` passes its own registry while the others use the default.

Two regression tests cover the wrapper path in `store-client/pkg/client`: one asserts both metrics are exported *through* the wrapper, and one asserts a wrapper over a watcher that does not report lag yields unknown rather than a spurious zero.

`golangci-lint` clean, full `store-client` and `fault-remediation` suites pass.

lfriedman-netllama:
@lalitadithya the push for Kaivalya's review findings moved the head to `22cc141f`, so the run at `a1ca6147` is superseded. Could you re-trigger with `/ok to test 22cc141f` when convenient?

KaivalyaMDabhadkar (approved):

PR #1763, opened 2026-09-08
feat: Add janitor label provider

### LINKED ISSUE #1730: [Feature]: Add a label-based remediation provider to Janitor
### Prerequisites

### Code of Conduct

### Feature Summary

The existing set of `janitor-provider`(s) currently supports either the cloud
provider APIs (`aws`, `gcp`, `azure`, `oci`, `nebius`), or performs the reboots on the host directly via a privileged pod. There's no way for NVSentinel to delegate the remediation action to the platform it's running on. 

Concretely, if you were running NVSentinel in an on-prem bare metal environment where the actions aren't directly exposed via API, we've got limited options for remediation. If we add a `label` provider that implements `model.CSPClient` by writing node labels and observing completion, then we could have any infrastructure provider take over the remediation actions natively. 

### Problem/Use Case

The problem I'm solving is to decouple the infrastructure provider from Janitor's remediation action. So if you wanted Janitor to function against OpenStack provisioned nodes, NICo, EC2 etc, you could do so by applying a label and letting an external system watch the K8s API (the one true common data surface) and act on it. This way Janitor never needs to be fed platform credentials either since K8s creds are sufficient. 

### Proposed behaviour

`SendRebootSignal` — patch the configured reboot label onto the Node with the
configured reason value. 

`IsNodeReady` — report completion when the node has actually rebooted.
We could compare `node.Status.NodeInfo.BootID` against the pre-reboot value as that's what `generic` does I think too. 

`SendTerminateSignal` — patch the configured terminate label. Completion is the node object being removed.

Other things - 
* We should probably handle cases where we time out if the external system doesn't act in a configured amount of time -  maybe emitting a K8s event or so in this case?
* The actual labels and annotation keys being put on the nodes should be configurable as every external system may want to look at something different. 

### Component

Plugins

rupalis-nv:
## Summary

Adds a generic **label** CSP provider so Janitor can request node reboot and terminate by labeling the Node. An external controller (for example NKE) watches those labels and performs the action. The node is treated as ready only after the reboot label is removed.

Helm: `janitor-provider.csp.provider: label`, with configurable `csp.label.rebootKey`, `terminateKey`, and `value` (NKE defaults). When the provider is `label`, the chart grants node `patch`/`update`.

```yaml
janitor-provider:
  csp:
    provider: "label"
    label:
      rebootKey: nke.nvidia.com/reboot
      terminateKey: nke.nvidia.com/terminate
      value: requested-by-nvsentinel
```

## Testing

Injected a real GPU fault and let the stack create the RebootNode (did not create the CR by hand).

1. From the gpu-health-monitor pod, inject XID 95
2. GPU health monitor reported a fatal Gpu error (XID 95).
3. Fault quarantine cordoned the node, Fault remediation created a RebootNode, Janitor called the label provider and set `nke.nvidia.com/reboot=requested-by-nvsentinel.`
4. NKE rebooted the node 
5. NKE set `reboot-status=Completed` and removed the reboot label.
6. `RebootNode: SignalSent=True, NodeReady=True` (Node reached ready state post-reboot).

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [x] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added support for managing node reboot and termination requests through configurable Kubernetes labels.
  - Added `key=value` label specifications with sensible defaults for reboot and termination requests.
  - Reboot readiness now requires removal of the reboot label and a changed, non-empty node boot ID.
  - Added label-provider configuration and required Kubernetes node permissions.

- **Documentation**
  - Documented label-based remediation, supported providers, configuration settings, and external controller requirements.


distros/kubernetes/README.md
```diff
@@ -316,6 +316,20 @@ janitor:
       principalId: "ocid1.principal.oc1..aaa..."
 ```
 
+##### Label provider
+
+Requests reboot and terminate by labeling the Node. An external controller such as NKE watches those labels and performs the action.
```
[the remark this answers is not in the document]
rupalis-nv: updated

janitor-provider/pkg/csp/labelprovider/labelprovider.go
```diff
@@ -0,0 +1,177 @@
...
+		"node", node.Name, "label", c.config.TerminateKey, "value", c.config.Value)
+
+	return model.TerminateNodeRequestRef(c.config.TerminateKey), nil
+}
+
+func (c *Client) applyLabel(ctx context.Context, nodeName, key, value string) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		current, err := c.k8sClient.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get node %s: %w", nodeName, err)
```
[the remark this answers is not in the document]
rupalis-nv: fixed

distros/kubernetes/nvsentinel/charts/janitor-provider/values.yaml
```diff
@@ -131,6 +132,18 @@ csp:
     # Write a reboot attribution message to the node's syslog before rebooting
     writeSyslog: false
 
+  # Label provider configuration (only used when provider=label)
+  # Requests reboot/terminate by labeling the Node. An external controller
+  # such as NKE watches the labels and performs the action. The node is
+  # considered ready once the reboot label is removed.
+  label:
+    # Node label used to request a reboot
+    rebootKey: "nke.nvidia.com/reboot"
```
lalitadithya: can we make this key=value such as:


```suggestion
    rebootKey: "nke.nvidia.com/reboot=requested-by-nvsentinel"
```

because the value may be different for terminate v/s reboot

distros/kubernetes/README.md
```diff
@@ -316,6 +316,20 @@ janitor:
       principalId: "ocid1.principal.oc1..aaa..."
 ```
 
+##### Label provider
+
+Requests reboot and terminate by labeling the Node. An external controller such as NKE watches those labels and performs the action. After reboot completion, that controller must remove the reboot label.
```
lalitadithya:
```suggestion
Requests reboot and terminate by labeling the Node. An external controller running on the Kubernetes control plane can watch those labels and perform the required action. After reboot completion, that controller **MUST** remove the reboot label.
```

janitor-provider/pkg/csp/labelprovider/labelprovider.go
```diff
@@ -0,0 +1,177 @@
...
+	ctx context.Context, node corev1.Node, _ string,
+) (model.ResetSignalRequestRef, error) {
+	if err := c.applyLabel(ctx, node.Name, c.config.RebootKey, c.config.Value); err != nil {
+		return "", fmt.Errorf("failed to set reboot label on node %s: %w", node.Name, err)
+	}
+
+	slog.InfoContext(ctx, "Set reboot label on node",
+		"node", node.Name, "label", c.config.RebootKey, "value", c.config.Value)
+
+	return model.ResetSignalRequestRef(c.config.RebootKey), nil
```
lalitadithya: Can we use the node's boot id as the return value here?

janitor-provider/pkg/csp/labelprovider/labelprovider.go
```diff
@@ -0,0 +1,177 @@
...
+
+// IsNodeReady reports completion when the reboot label is no longer on the node.
+// The live Node is read so this follows the external controller removing the label.
+func (c *Client) IsNodeReady(ctx context.Context, node corev1.Node, _ string) (bool, error) {
+	current, err := c.k8sClient.CoreV1().Nodes().Get(ctx, node.Name, metav1.GetOptions{})
+	if err != nil {
+		return false, fmt.Errorf("failed to get node %s: %w", node.Name, err)
+	}
+
+	if _, present := current.Labels[c.config.RebootKey]; present {
```
lalitadithya: To make this check stronger, can we also compare that the boot ID has changed?

janitor-provider/pkg/csp/labelprovider/labelprovider.go
```diff
@@ -0,0 +1,177 @@
...
+		k8sClient: k8sClient,
+		config:    withDefaults(config),
+	}
+}
+
+// SendRebootSignal patches the configured reboot label onto the node.
+func (c *Client) SendRebootSignal(
+	ctx context.Context, node corev1.Node, _ string,
+) (model.ResetSignalRequestRef, error) {
+	if err := c.applyLabel(ctx, node.Name, c.config.RebootKey, c.config.Value); err != nil {
```
lalitadithya: What if the label is already present, should we fail here? Also what about the case when the label is present, but the value is different, should we overwrite?

janitor-provider/pkg/csp/labelprovider/labelprovider.go
rupalis-nv: yes updated

janitor-provider/pkg/csp/labelprovider/labelprovider.go
rupalis-nv: yes updated

distros/kubernetes/README.md
rupalis-nv: addressed

distros/kubernetes/nvsentinel/charts/janitor-provider/values.yaml
rupalis-nv: addressed

janitor-provider/pkg/csp/labelprovider/labelprovider.go
rupalis-nv: If the label is already present with the same value, we treat it as idempotent and succeed (no update). That covers a retry of the same reboot/terminate request.

If it is present with a different value, we fail and do not overwrite. Another controller may already own that request, and replacing its value would hide that.

distros/kubernetes/README.md
```diff
@@ -316,6 +316,19 @@ janitor:
       principalId: "ocid1.principal.oc1..aaa..."
 ```
 
+##### Label provider
+
+Requests reboot and terminate by labeling the Node. An external controller running on the Kubernetes control plane can watch those labels and perform the required action. After reboot completion, that controller **MUST** remove the reboot label.
```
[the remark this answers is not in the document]
rupalis-nv: fixed

lalitadithya (approved):

PR #1776, opened 2026-09-08
test(uat): support dry-run deployments and clusters without an nvidia RuntimeClass

suket22:
## Summary

The UAT suite assumed NVSentinel always cordons and reboots on the faults it detects. That does not hold on clusters running NVSentinel in **dry-run mode**: `fault-quarantine` applies no cordon and `fault-remediation` creates no `RebootNode`/`GPUReset`/`ExternalRemediationRequest`, so `wait_for_node_quarantine` and `wait_for_boot_id_change` wait for something that never happens.

This detects dry-run from the `fault-remediation` deployment args and asserts the decisions NVSentinel still records:

| module | action suppressed in dry-run | observable this asserts instead |
| --- | --- | --- |
| `fault-quarantine` | cordon/taint | non-empty `quarantineHealthEvent` annotation (the cordon annotation is not written in dry-run) |
| `fault-remediation` | remediation custom resource | `dgxc.nvidia.com/nvsentinel-state=remediation-succeeded` |

A cluster left in dry-run by accident still fails if those decisions are never reached.

Dry-run cannot reboot, so Test 1 restarts the hostengine that owns the injected XID (embedded GPU health monitor on loopback, otherwise the node-local `nvidia-dcgm` pod) and strips leftover FQ/FR node metadata. Syslog tests (2–3) are skipped in dry-run because recovery needs a real reboot. The debug pod always uses `runtimeClassName: nvidia`.

Recovery failures are warnings so the EXIT trap can still reset node state.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [x] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: UAT test suite

## Testing
- [ ] Tests pass locally

Ran `tests/uat/tests.sh` against Lambda QA (`lambda-nvidia-qa-cluster-ctloypud5ftvhhau`) with NVSentinel in dry-run:

- Test 1 passed: DCGM inject 392 (events) → XID 95 (`GpuAllWatch`) → `quarantineHealthEvent` → `nvsentinel-state=remediation-succeeded` → hostengine restart → node metadata cleared.
- Tests 2–3 skipped as designed in dry-run.
- Node left with no leftover quarantine annotations / state label; GPU conditions healthy.

Default live (non-dry-run) waits are unchanged for clusters that still cordon and reboot.

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dry-run cleanup to reliably restore affected node state after testing.
  - Ensured injected hardware error conditions and related metadata are cleared during recovery.
  - Made debug resource cleanup safe to repeat without causing failures.
  - Prevented dry-run test runs from executing incompatible syslog scenarios.
  - Improved monitoring initialization and cleanup when interrupted.
  - Added warning handling for recovery failures so remaining cleanup can continue.
  - Refined quarantine detection and error-condition checks for more accurate results.


tests/uat/tests.sh
```diff
@@ -271,13 +348,48 @@ wait_for_node_unquarantine() {
     error "Timeout waiting for node $node to be uncordoned"
 }
 
+# In dry-run, fault-remediation returns before rendering the template, so no
+# RebootNode, GPUReset or ExternalRemediationRequest is created and nothing
+# downstream runs. It still logs the decision per node, which is the last
+# observable of the remediation half.
+wait_for_dry_run_remediation_decision() {
```
lalitadithya, line 434: Can we instead check for the nvsentinel-state label? `dgxc.nvidia.com/nvsentinel-state` which should be applied irrespective of dry run mode

tests/uat/tests.sh
lalitadithya, line 925: I believe you will need to have the failing health check recover here else the annotation and label state will not be cleared. This is easy in the case of DCGM, because we can just restart the pod and it will result in a healthy signal, but it is more tricky in the case of syslog since we need a boot id change

tests/uat/tests.sh
suket22: Good idea. Turns out the remediation module sets labels under dry-run but quarantine module returns earlier and only sets annotations. So checking for both.

tests/uat/tests.sh
suket22: Yeah so just skipping the syslog tests for now as we discussed offline.

lalitadithya (approved):

PR #1774, opened 2026-09-08
fix(postgres): avoid race condition for datastore trigger recreation on startup

dentiny:
## Summary

Closes https://github.com/NVIDIA/NVSentinel/issues/1773

Disclaimer: I don't think DDL should be executed in application code, it should be managed externally via, say, terraform; but this PR is a strict improvement to achieve DDL statements idempotency and avoid startup failure.
For discussion and detailed proposals, please refer to https://github.com/NVIDIA/NVSentinel/issues/1772

## Type of Change
- [X] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [x] Other: Deployment/config

## Testing
- [X] Tests pass locally
- [ ] Manual testing completed
- [X] No breaking changes (or documented)

## Checklist
- [X] Self-review completed
- [X] Documentation updated (if needed)
- [X] Ready for review

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved PostgreSQL trigger setup to avoid unnecessary trigger replacement.
  - Prevented setup failures caused by concurrent attempts to create the same triggers.
  - Improved reliability when initializing event change tracking in parallel environments.

- **Tests**
  - Added coverage verifying reliable creation of maintenance and health event change-tracking triggers, including concurrent setup scenarios.

PR #1777, opened 2026-09-08
refactor(postgresql): centralize schema in versioned migrations

### LINKED ISSUE #1772: [Feature]: Manage postgres schema in a versioned manner
### Prerequisites

### Code of Conduct

### Feature Summary

Hi team, in the current implementation, we're managing postgres schema in [a single file](https://github.com/NVIDIA/NVSentinel/blob/main/docs/postgresql-schema.sql), it's not ideal for several reasons
- (major) No schema version support
  + With one single file, users cannot track version history, making version rollback and forward, incident diagnosis manual
- (major) Application startup performs DDL
  + The datastore executes schema creation and modification during application startup, which means (1) application credentials require DDL privileges; (2) multiple replicas can attempt schema changes concurrently; (3) DDL failures prevent application startup; (4) Index or trigger changes may introduce locks during a rollout; (5) Schema deployment is coupled to application deployment without an explicit migration gate.
- (medium) The pg schema has been largely duplicate for multiple copies in the codebase, and they're not identical
  + Example-1: datastore golang code adds [two timestamp columns](https://github.com/NVIDIA/NVSentinel/blob/984940e344bb18bb6c348f2ec103fbc1f01dfba4/store-client/pkg/datastore/providers/postgresql/datastore.go#L401-L404) but they're not in SQL schema
  + Example-2: datastore golang code adds [an index](https://github.com/NVIDIA/NVSentinel/blob/984940e344bb18bb6c348f2ec103fbc1f01dfba4/store-client/pkg/datastore/providers/postgresql/datastore.go#L437-L441) but they're not in SQL schema
- (minor) The init schema [values-tilt-postgresql.yaml](https://github.com/NVIDIA/NVSentinel/blob/984940e344bb18bb6c348f2ec103fbc1f01dfba4/distros/kubernetes/nvsentinel/values-tilt-postgresql.yaml#L100-L352) is not very meaningful
  + It doesn't have any impact to an existing database
  + It's almost a duplicate with [datastore](https://github.com/NVIDIA/NVSentinel/blob/984940e344bb18bb6c348f2ec103fbc1f01dfba4/store-client/pkg/datastore/providers/postgresql/datastore.go#L342) code, which adds unnecessary maintenance burden

### Problem/Use Case

I use NV sentinel in production and use postgres as datastore.

### Proposed Solution

There're a few general DB management principles we should follow
- Manage the database schema in one single place, instead of replicating here and there
- Version the schema and store it in a persistent storage (i.e., a system postgres table)
- DDL and DML generally involves different procedure and permission, only perform DML in the application; DDL should be managed separately

### Component

Deployment/Config

dentiny:
## Summary

Related to https://github.com/NVIDIA/NVSentinel/issues/1772

In the above-mentioned issue, I described a few pain points for postgres datastore management, including missing schema management, permission management for DDL in application code, etc.

**Disclaimer: this PR is made to express my thoughts for datastore schema management improvement, and I'd love to hear feedbacks from NV team, and discuss what's the best way to move forward.**

Core principle for the code change
- There's one canonical schema source: ordered migrations. No duplicated DDL in Go, Helm values, or docs.
- Use expand/contract migrations for rolling upgrades, instead of application code.
- Every progressive PR must leave deployment workflows functional.

This PR include a few changes
- Centralizes PostgreSQL DDL in ordered, versioned migration files.
- Removes duplicated schema creation from Go startup code and Helm/Tilt values.
- Adds a read-only schema compatibility check before PostgreSQL-backed applications start.
- Reconciles runtime-only columns, indexes, triggers, and pg_notify behavior in migration 00002.

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [X] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Documentation/CI
- [ ] Fault Management
- [ ] Health Monitors
- [ ] Janitor
- [X] Other: deployment/config

## Testing
- [ ] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added versioned PostgreSQL migrations with schema version tracking.
  * Added startup validation to ensure the database meets the required schema version.
  * Added migration guidance for manual, ordered application and external database workflows.

* **Changes**
  * PostgreSQL schema initialization is no longer performed automatically by NVSentinel, Helm, or Tilt.
  * Updated documentation and troubleshooting guidance to reflect migration-based schema management.
  * Removed the legacy embedded schema, synchronization script, and canonical schema file.


KaivalyaMDabhadkar:
@dentiny could you please address the code rabbit reviews?

dentiny:
> @dentiny could you please address the code rabbit reviews?

Hey @KaivalyaMDabhadkar thanks for the quick reply! This PR is mainly to align with the team to see whether you're happy with the progressive centralization approach; if you think it's on the right path let me formalize this PR into production ready!

KaivalyaMDabhadkar:
> > @dentiny could you please address the code rabbit reviews?
> 
> Hey @KaivalyaMDabhadkar thanks for the quick reply! This PR is mainly to align with the team to see whether you're happy with the progressive centralization approach; if you think it's on the right path let me formalize this PR into production ready!

Thanks @dentiny, I like where this is going, one set of versioned SQL files as the single source of the schema, applied outside the app, with a version check when the app starts. I compared the files against the current schema and they line up, with a few extra indexes on top.
There is one thing that needs to change before this is ready. A fresh install has to work without manual steps and right now the database comes up empty and every postgres backed service exits, which is why the postgres e2e jobs are failing. A job in the chart that applies the migrations, along the lines of the existing mongo init job, would solve that. It could be on by default, with a switch to turn it off for databases managed outside the cluster. Could you also mark this as a breaking change, add a short upgrade note, and list the exact grants the app role needs? It also needs USAGE on the changelog sequence. Please also make sure to remove the new trigger test from #1774 as well, since the function it calls goes away here

dentiny:
> I like where this is going, one set of versioned SQL files as the single source of the schema, applied outside the app, with a version check when the app starts. I compared the files against the current schema and they line up, with a few extra indexes on top.

Thanks for the quick feedback! I will address the CI issues and review comments then.

Issue #1767, opened 2026-09-08
[Bug]: metadata-collector exits on a single failed poll tick, so a credential rotation restarts it fleet-wide


lfriedman-netllama:
## Problem

`metadata-collector` exits the process when a single iteration of its 30-second polling loop fails. A routine credential rotation therefore restarts it on every node in the cluster.

`runMapper` is a ticker loop, but a failed tick returns rather than continuing:

```go
// metadata-collector/main.go
ticker := time.NewTicker(defaultPodDeviceMonitorPeriod)
for {
    select {
    case <-ctx.Done():
        return nil
    case <-ticker.C:
        numUpdates, err := podDeviceMapper.UpdatePodDevicesAnnotations()
        if err != nil {
            return fmt.Errorf("could not update mapper pod devices annotations: %w", err)   // <-- aborts the loop
        }
        ...
    }
}
```

and `main` turns that into an exit:

```go
if err := runMapper(ctx); err != nil {
    slog.Error("Pod device mapper failed", "error", err)
    cancel()
    os.Exit(1)      // main.go:65
}
```

So any transient failure of the kubelet `/pods` call is fatal, even though the very next tick 30 seconds later would almost certainly succeed.

## Observed

288-node GB200 cluster, NVSentinel v1.21.0. The cluster's kubeconfig was rotated to address a security issue. During the rotation window, **88 of 288 metadata-collector pods exited and restarted**, in clusters of a few pods at a time across roughly an hour (24 distinct exit timestamps, e.g. 6 pods at `15:26:46Z`, 3 at `15:26:48Z`, 4 at `16:27:02-03Z`).

Every one exited `1` with:

```
{"level":"ERROR","msg":"Pod device mapper failed","module":"metadata-collector","version":"v1.21.0",
 "error":"could not update mapper pod devices annotations: error listing pods from Kubelet pods endpoint:
  got a non-200 response code from /pods endpoint: 401, body: Unauthorized"}
```

The preceding log lines are an unbroken 30-second cadence of successful `Device mapper pod annotation updates` for six days, then one 401, then exit.

**The other three DaemonSets were unaffected by the same rotation:** `platform-connectors` 0 terminations, `syslog-health-monitor` 0, `gpu-health-monitor` 1. They reach the API server through client-go, which handles credential reload and backoff. `metadata-collector` is the only component that talks to the kubelet endpoint directly over raw HTTP, and the only one that exited.

[...]

PR #1778, opened 2026-09-09
fix(metadata-collector): tolerate transient poll failures instead of exiting

lfriedman-netllama:
Closes the safety half of #1767. Per the agreed split, the metrics endpoint follows in a second PR.

## What went wrong

`runMapper` is a 30s ticker loop, but a failed tick **returned**, and `main.go` turns that into `os.Exit(1)`. So one transient fault killed the process. Since metadata-collector is a DaemonSet, a *fleet-wide* transient fault killed it everywhere: 88 of 288 pods on the cluster where this was found, in bursts over about an hour, after a cluster credential was rotated.

## What this changes

**1. The exit is now threshold-based, not first-failure.** Failing loudly is still the intent, so the threshold keeps it: a collector that is genuinely broken still dies with a non-zero status, while a transient fault no longer takes the fleet with it. Consecutive failures are counted and logged with the streak and the threshold, the count resets on any success, and the process exits only when the threshold is reached.

Default 10, which at the 30s cadence is 5 minutes of a collector that cannot do its job. Settable via `metadata-collector.podMapper.maxConsecutiveFailures`; 1 restores the old exit-on-first-failure behaviour. A value below 1 is refused rather than reinterpreted.

**2. The kubelet `/pods` backoff is widened.** It used `retry.DefaultRetry`, whose own doc comment describes it as the retry for resource-version conflicts. It is `{Steps: 5, Duration: 10ms, Factor: 1.0}`, so **about 40ms of total waiting** — it cannot span any credential or restart window. It now waits roughly 7.5s across the same five steps, still comfortably inside the 30s poll period so ticks cannot pile up.

**3. The token and the request are built per attempt, not once per call.** This one is easy to miss and the widened backoff does not work without it. Previously the token was read once and the `*http.Request` built once, then reused for all five attempts, so every attempt presented the same credential. Widening the wait would have changed nothing for the failure the wait exists to absorb. `TestListPods_TokenRotatedMidRetry_SecondAttemptSendsTheNewToken` pins this: it fails if the token is read once and passes when it is re-read.

Building a fresh request per attempt also drops the reuse of one `*http.Request` across `RoundTrip` calls, which was never guaranteed safe.

## Tests

`metadata-collector/main_test.go` is new. `pollPodDevices` is extracted from `runMapper` so the loop is drivable by hand, without waiting on a 30s ticker or standing up a kubelet:

- failures below the threshold keep polling
- consecutive failures reaching the threshold return, wrapping the underlying error
- **a success between failures resets the streak** — without this, a collector failing once every few hours eventually exits for no reason
- a cancelled context returns nil without polling
- a threshold below 1 is refused without polling
- the default threshold rides out at least a minute

The existing kubelet-client tests now inject a fast backoff. The production backoff is asserted separately, as a duration range rather than exact values, so it is checked to outlast a rotation *and* to finish inside the poll period.

Full module green with `-race`, `golangci-lint` 0 issues, `gomod-lint` clean, `helm lint` clean.

## Not in this PR

- **The metrics** (`metadata_collector_pod_mapper_failures_total`, `metadata_collector_pod_mapper_consecutive_failures`). metadata-collector has no metrics surface at all today: no `client_golang` in its `go.mod`, no registry, no handler, no `--metrics-port`. That means a dependency, an endpoint and a container port on a 288-pod DaemonSet, which deserves its own review.
- **`retryAllErrors` returns `true` unconditionally**, so a permanent misconfiguration such as a wrong `bearerTokenPath` is retried and then counted as a transient failure. Worth distinguishing retryable from permanent, but it is not what caused this and it interacts with the threshold, so it belongs in its own change.
- **`patchPodWithRetry` also uses `retry.DefaultRetry`**, though with a proper `isRetryableError` predicate, and it goes through client-go, which reloads credentials on its own. Left alone deliberately; say the word if you want it widened too.

## Summary by CodeRabbit

* **New Features**
  * Added configurable handling for consecutive pod device-mapper polling failures, defaulting to 10 failures.
  * Added retry support with exponential backoff for pod-list requests.
  * Retries reread authentication credentials, allowing rotated tokens without restarting the collector.
  * Added a 20-second timeout for pod-list requests, including retries and response processing.

* **Bug Fixes**
  * Transient polling and request failures no longer immediately terminate collection.
  * Successful polling resets the consecutive-failure count.


lfriedman-netllama:
Ready for review. Fork PR, so CI needs a maintainer trigger: `/ok to test ed88464f897dfe9c812de8868611dda602e67b3a` when you get a chance.

@lalitadithya this is the safety half we agreed to split. One thing worth a look beyond what the issue described: the token and the request were built once per `ListPods` call and reused across all five retry attempts, so widening the backoff on its own would have changed nothing for a rotated credential. Every attempt now builds its own. `TestListPods_TokenRotatedMidRetry_SecondAttemptSendsTheNewToken` fails without that and passes with it, so the behaviour is pinned rather than asserted.

lfriedman-netllama:
All three findings applied in `2ede91da`. Each premise checked against the client-go/apimachinery v0.37.0 source first rather than taken on trust, and all three held.

**The duration bound is a real gap and I had it wrong in a comment.** `wait.ExponentialBackoff` sleeps with `time.Sleep` and consults no context, the transport's timeouts cover only the header exchange, and `io.ReadAll` has no deadline. So widening the backoff bounded the *waiting* while nothing bounded the *work*, and my comment claiming the call finishes inside the poll period was only ever true of the sleeps.

Fixed by deriving one deadline for the whole call and using it for every attempt, so it covers the requests, their body reads and the sleeps between them. The retriable predicate returns false once it expires, which is what stops `OnError` sleeping past it. I kept `retry.OnError` rather than moving to `ExponentialBackoffWithContext` because `OnError` returns the underlying error while `ExponentialBackoff` returns `ErrWaitTimeout`, and the poll log needs the real cause. 20s, inside the 30s poll period. A call slower than that is already failing to keep up, and the new failure threshold absorbs it rather than exiting.

Pinned by `TestListPods_HungKubelet_ReturnsWhenTheCallDeadlinePasses`, which is decisive: without the deadline it hangs until the test binary's own timeout kills it, with the deadline it finishes in 0.8s.

**The backoff arithmetic was wrong in my test.** `ExponentialBackoff` breaks before sleeping on the last step, so five steps sleep four times: 7.5s, which is what the code comment said, not the 15.5s the test loop computed. The assertion now checks the invariant chain that actually matters: sleeps outlast a rotation, sleeps leave room inside the call deadline, deadline sits inside the poll period.

**Token error wrapped** with the path that failed.

One knock-on worth flagging since it touches an existing test: `TestListPodsWithNewRequestError` set `client.ctx = nil`, which no longer reaches `http.NewRequestWithContext` because the deadline is derived from that context first, and deriving from nil panics. It uses an unparseable URI instead, which hits the same error path and is a state the client could actually be in.

`retryAllErrors` is kept, since the gRPC client still uses it, with its caveat written down there.

Module green with `-race`, `golangci-lint` 0 issues, `gomod-lint` clean. @lalitadithya this needs a fresh `/ok to test 2ede91da873216aad8e05fa51a97835750cd82c6`, sorry for the churn.

lalitadithya (approved):

PR #1779, opened 2026-09-09
chore: doc changes to improve OSS scorecard

lalitadithya:
## Summary

fix findings reported by the OSS scorecard

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [X] 📚 Documentation
- [ ] 🔧 Refactoring
- [X] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Health Monitors
- [ ] Fault Management
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Janitor
- [X] Documentation/CI
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

- **Documentation**
  - Added contributor, governance, maintainer, security, release, changelog, and architecture decision record guidance.
  - Documented official distribution channels, supported versions, image verification, and contribution requirements.
  - Consolidated coding-agent guidance into a canonical reference.

- **Security & Reliability**
  - Pinned container base images to immutable digests across project components.
  - Expanded automated dependency monitoring and increased Docker update frequency.

- **Maintenance**
  - Updated image listings and release verification examples for current registry and release assets.
  - Corrected the generated image list for the Slinky drainer.
  - Improved automated review configuration and local instruction-file handling.


ArangoGutierrez (commented):
Thanks for this, digest-pinning the base images and giving the agent instructions one canonical home in AGENTS.md are both real improvements.

Two strands in one change: the agent instructions consolidated into a root AGENTS.md with .github/copilot-instructions.md reduced to a pointer, and a supply-chain pass that digest-pins the Dockerfile bases and moves the Dependabot docker group to weekly so those digests actually get refreshed. The slinky-drainer rename in scripts/build-image-list.sh is the one functional change here and is easy to lose among the docs, so it is worth naming in the PR description.

- Since the docker group's cadence is already being revisited here, a cooldown block would also satisfy the minimum-release-age check that scorecard scanners look for, and nothing in this file sets one today. (.github/dependabot.yml:128)
- The maintainer-list check wants named individuals in MAINTAINERS.md, OWNERS, CODEOWNERS, GOVERNANCE.md or CONTRIBUTING.md, and the only names in the repo are lalitadithya and dims under sig-maintainers in OWNERS_ALIASES, which is not one of those files. Listing the current maintainers here alongside the team link would cover it, since the team page is not readable from outside the org. (MAINTAINERS.md:3)
- Contribution guidance is also scored on explaining how to extend the project, and CONTRIBUTING.md does not mention adding a health monitor, a store backend or a platform connector. A short section pointing at an existing monitor as the template would cover it, and that dimension carries more weight than the agent-readiness one this PR spends most of its lines on. (CONTRIBUTING.md:53)
- A separate check looks for an embargo and pre-disclosure process, which SECURITY.md does not describe. Naming the disclosure window and who gets pre-notified, next to this new section, would round it out. (SECURITY.md:26)

README.md
```diff
@@ -259,6 +259,20 @@ Validated on NVIDIA Volta, Ampere, Hopper, Ada Lovelace and Blackwell architectu
...
+
+| Channel | Location |
+|---|---|
+| Helm chart (OCI) | `oci://ghcr.io/nvidia/nvsentinel` |
+| Container images | `ghcr.io/nvidia/nvsentinel/<component>` — e.g. `ghcr.io/nvidia/nvsentinel/labeler` |
+| Releases | [GitHub Releases](https://github.com/NVIDIA/NVSentinel/releases), each with a `versions.txt` pinning every component image for that version |
+| Source | [github.com/NVIDIA/NVSentinel](https://github.com/NVIDIA/NVSentinel) |
+| Documentation | [docs.nvidia.com/nvsentinel](https://docs.nvidia.com/nvsentinel/) |
+
+Container images are signed with Sigstore cosign and carry SPDX SBOM and SLSA provenance attestations. See [SECURITY.md](SECURITY.md) for how to verify them.
```
ArangoGutierrez: The publish path attaches a CycloneDX SBOM (anchore/sbom-action with format cyclonedx-json, then cosign attest --type cyclonedx) and buildko.sh passes --sbom=none to ko, so no SPDX document is produced for a user to fetch. There is also no cosign sign anywhere in the pipeline, only signed attestations, so say CycloneDX here and describe the signing as attestations.

RELEASE.md
```diff
@@ -63,19 +63,10 @@ graph LR
...
-- `fault-quarantine`
-- `labeler`
-- `node-drainer`
-- `fault-remediation`
-- `log-collector`
-- `file-server-cleanup`
+**Container Images** are published under `ghcr.io/nvidia/nvsentinel/`. This document deliberately does not list them — a static inventory drifts. There are two sources of truth:
+
+- **For a released version**: the `versions.txt` asset on that [GitHub release](https://github.com/NVIDIA/NVSentinel/releases), which pins every image and tag actually published.
+- **For what a release will contain**: [`scripts/build-image-list.sh`](scripts/build-image-list.sh), which generates `versions.txt`. Adding a component means adding it there.
```
ArangoGutierrez, line 69: build-image-list.sh is now the stated source of truth for what a release contains, but its dynamic_images array omits lifecycle-manager, maintenance-notifier, nic-health-monitor and preflight, all four of which buildko.sh builds and pushes, and values.yaml carries no repository entry for any of them to pick up in the static merge. Adding the four names keeps versions.txt honest now that this document points at it.

AGENTS.md
```diff
@@ -0,0 +1,373 @@
...
+| Node labels, annotations, conditions | External automation keys off these — e.g. `nvsentinel.dgxc.nvidia.com/kata.enabled`. Treat them as public API |
+| Metrics names and labels | Renaming one silently breaks dashboards and alerts |
+| CLI flags and environment variables | Keep the old spelling working |
+
+**CI will not catch this for you.** `make protos-lint` only verifies that generated files are up to date — there is no `buf breaking` check, no API-diff gate, and no schema compatibility test. The `v1alpha1` on the CRDs describes their maturity, not permission to break them.
+
+Since nothing gates it, check the surfaces yourself before opening a PR. Derive the paths rather than hardcoding them, so a CRD added later cannot fall outside the check:
+
+```bash
+SURFACES="data-models distros/kubernetes/nvsentinel/values.yaml $(ls -d */api/v1alpha1)"
```
ArangoGutierrez: ls -d */api/v1alpha1 only matches one directory level, so it returns janitor, lifecycle-manager and preflight but drops plugins/slinky-drainer/api/v1alpha1, which declares two +kubebuilder:object:root=true types. That is exactly the CRD surface the paragraph above promises a derived path will catch, so use find . -type d -path '*/api/v1alpha1' instead.

.github/dependabot.yml
```diff
@@ -100,8 +100,16 @@ updates:
         patterns: ["*"]
 
   # Docker
+  # Base images with a static tag are pinned to digests, so these updates are
+  # what refresh the digests when upstream rebuilds a tag — keep the interval at
+  # weekly so base image CVE fixes are picked up promptly.
+  # Exception: the DCGM and PyTorch base images in gpu-health-monitor, dcgm-diag,
```
ArangoGutierrez: The exception list is short by one real case and overstates another: tilt/simple-health-client's runtime stage is FROM ${IMG}:${TAG} with ARG TAG=v3.1.13, a single static base that no Makefile overrides, and nccl-loopback's Makefile never passes PYTORCH_VERSION either (only nccl-allreduce does). Both can be digest-pinned, which would leave the DCGM images and nccl-allreduce as the only genuine exceptions.

Also at preflight-checks/nccl-loopback/Dockerfile:66.

RELEASE.md
lalitadithya: Good catch, and it mattered more because this PR made `versions.txt` the stated source of truth.

Confirmed all four are built and pushed by `buildko.sh` and absent from the `dynamic_images` array, with no `values.yaml` entry to pick them up in the static merge. Added `lifecycle-manager`, `maintenance-notifier`, `nic-health-monitor` and `preflight`.

Verified by regenerating the list with `SAFE_REF_NAME=v1.22.0` and resolving every entry against the registry — all 23 exist, so these four have been missing from published `versions.txt` for a while.

README.md
lalitadithya: You're right on both counts — I wrote that from `SECURITY.md` rather than from the pipeline.

`.github/actions/sbom-and-attest/action.yml` uses `anchore/sbom-action` with `format: cyclonedx-json` and attests with `cosign attest --type cyclonedx`, and `buildko.sh` passes `--sbom=none` to ko, so no SPDX document exists. There is no `cosign sign` anywhere — only signed attestations plus `actions/attest-build-provenance`.

Reworded to: "Container images carry a Sigstore-signed CycloneDX SBOM attestation and a SLSA build provenance attestation."

Worth flagging separately: `SECURITY.md` has a whole "SPDX SBOM" section claiming SPDX v2.3 with a sample SPDX document. That has the same problem, and the README now points readers there to verify. I left it alone since it predates this PR and fixing it means rewriting the sample and the verification commands — happy to do it here or in a follow-up, whichever you prefer.

AGENTS.md
lalitadithya: Correct, and the failure was exactly the one the paragraph warns about — the command silently dropped the surface it promised to catch. `ls -d */api/v1alpha1` returns 3 directories; `find . -type d -path '*/api/v1alpha1'` returns 4, adding `plugins/slinky-drainer/api/v1alpha1`.

Switched to `find` and left a short comment above it saying why, so it doesn't regress back to the glob.

.github/dependabot.yml
lalitadithya: Verified both — `nccl-loopback`'s Makefile never references `PYTORCH_VERSION` (only `nccl-allreduce` does, 3 times), and nothing overrides `IMG`/`TAG` for `simple-health-client`.

So rather than just fixing the comment, I pinned both:
- `nccl-loopback` → `nvcr.io/nvidia/pytorch:26.04-py3@sha256:192d749b…`
- `simple-health-client` → `nvcr.io/nvidia/distroless/go:v3.1.13@sha256:694780ff…`

Dropped the now-dead `ARG`s, confirmed both digests are multi-arch (amd64+arm64) so the multi-platform builds still work, and both pass `docker buildx build --call=check`.

Genuine exceptions are now only the three DCGM images and `nccl-allreduce`; comment updated to say exactly that.

README.md
lalitadithya: Fixed `SECURITY.md` here too — but digging into it turned up something worth a separate look.

Doc fixes: SPDX → CycloneDX, dropped the invented sample SPDX document, and removed the `$IMAGE_SBOM` / `.sbom` retrieval flow. That flow was ko's SPDX output and cannot work now that `buildko.sh` passes `--sbom=none` — I confirmed the `.sbom` tag no longer exists. Added a `gh attestation verify` example I actually ran; it returns `https://slsa.dev/provenance/v1` from `publish.yml@refs/tags/v1.22.0`.

**The part worth checking:** I could not find the CycloneDX SBOM attestation for a released image. For `fault-quarantine:v1.22.0` there is no `.att` tag, no OCI referrers, and GitHub's attestation store holds only the SLSA provenance. The release's sole asset is `versions.txt`, so `upload-release-assets: auto` didn't attach one either. `sbom-and-attest` predates v1.22.0, so it should have run.

SLSA provenance verifies fine. The SBOM attestation may not be reaching the registry — probably worth its own issue rather than this PR.

AGENTS.md
```diff
@@ -0,0 +1,375 @@
...
+defer resp.Body.Close()
+```
+
+**Prefer informers over polling the API server.** Direct `Get`/`List` in a loop hammers the apiserver at cluster scale; NVSentinel runs on large GPU fleets where that is a real cost. Register event handlers on a shared informer factory and extract handler setup into its own method.
+
+**Other conventions the linters enforce.** `.golangci.yml` enables `wsl_v5` (strict whitespace/statement cuddling), `lll` (line length), `cyclop` and `gocognit` (complexity), `gosec`, `nilerr`, `exhaustive` (switch over enums), `dupl`. When a complexity linter fires, split the function — do not add a `//nolint`. Every source file needs the Apache 2.0 header; `make license-headers-lint` checks it.
+
+**Python** (`gpu-health-monitor`, `dcgm-diag`, `nccl-allreduce`, `log-collector`): Poetry for dependencies, PEP 8, Black for formatting, type hints on all functions.
+
+**Protobuf**: edit `.proto` files under `data-models/protobufs/` and run `make protos-generate`. Never hand-edit generated code — CI diffs the regenerated output and fails on drift.
```
ArangoGutierrez: The protobuf rule names only data-models/protobufs/, but api/proto/ is a second generated tree (device/v1alpha1/gpu.proto and csp/v1alpha1/provider.proto) that the root Makefile drives via make -C api protos-generate at line 306. Add it here and as a Repository Map row, which has no api/ entry at all today even though MAINTAINERS.md lists api/ under Core Infrastructure.

AGENTS.md
```diff
@@ -0,0 +1,375 @@
...
+
+# Individual module (every Go module has the same interface via make/go.mk)
+make -C labeler lint-test    # vet + lint + test for one module
+make -C labeler vet          # go vet ./...
+make -C labeler lint         # golangci-lint with the repo .golangci.yml
+make -C labeler test         # gotestsum, race detector on
+make -C labeler coverage     # coverage report
+
+# Single test
+cd labeler && go test -race -run TestLabeler_KataEnabled ./...
```
ArangoGutierrez: TestLabeler_KataEnabled does not exist in the labeler module; the Kata tests are TestKataLabelDetection, TestKataLabelOverride and TestKataLabelOverrideIsolation. go test -run on a pattern matching nothing prints "no tests to run" and exits 0, so anyone copying this line gets a green that ran nothing.

SECURITY.md
```diff
@@ -38,61 +42,49 @@ Export variables for the image you want to verify, for example:
...
-      ]
-    }
-  ]
-}
+**Retrieve the SBOM attestation**:
+
+```shell
+cosign verify-attestation \
+  --type cyclonedx \
+  --certificate-identity-regexp '^https://github.com/NVIDIA/NVSentinel/' \
```
ArangoGutierrez: This identity regexp accepts any workflow in the repository, while must-have-sbom.yaml and must-have-slsa.yaml both pin the publish.yml workflow identity. The tighter pattern still verifies against fault-quarantine at v1.22.0, so the documented check can match the policy the cluster actually enforces.

Also at SECURITY.md:83.

CodeRabbit flagged this too.

ArangoGutierrez (approved):
Thanks for chasing each of these back to the pipeline rather than to the docs, the digest pinning and the versions.txt sweep both landed further than what was asked.

The five fixes check out against the pipeline and not just the prose: the SURFACES snippet now picks up plugins/slinky-drainer, both newly pinned digests are multi-arch, and the SBOM wording matches what publish.yml actually attaches. What is left is smaller and none of it blocks.

- The sweep picked up the four ko-built images but not the three Dockerfile-built ones publish.yml also pushes: preflight-dcgm-diag, preflight-nccl-allreduce and preflight-nccl-loopback. All three resolve at ghcr.io/nvidia/nvsentinel/<name>:v1.22.0 and are chart images in values-full.yaml, and the static merge cannot recover them because it reads values.yaml. (scripts/build-image-list.sh:67)
- cosign attest signs and pushes an attestation rather than reading one, and it needs a local sbom-<component>.cdx.json plus push access to the registry, so a reader following on from the Setup block cannot run it. Marking it as what the release pipeline does would keep the consumer path pointed at verify-attestation. (SECURITY.md:59)
- Only gpu-health-monitor builds variants: its docker-publish fans out to docker-publish-dcgm3 and docker-publish-dcgm4 with two DCGM_VERSION values and two tags. dcgm-diag and nccl-allreduce each have one IMAGE_NAME, one tag and one ARG value, so both could take a tag@sha256 pin like the bases you just fixed. (preflight-checks/dcgm-diag/Dockerfile:41)
- 030-grpc-tls-authentication.md is headed ADR-029 and 049-node-validation.md is headed ADR-045, so ADR-029 currently appears twice, which is the number reuse the next bullet forbids. Twelve further records carry no # ADR- heading at all, so the convention as stated does not hold for 14 of the 54. (docs/designs/README.md:9)

docs/designs/README.md
```diff
@@ -0,0 +1,70 @@
+# Architecture Decision Records
+
+Each record states one decision: the context that forced it, what was decided, what it costs, and which alternatives were rejected and why. That reasoning is rarely recoverable from the code alone.
+
+## Conventions
+
+- The file name carries the number: `NNN-short-description.md`, and the record's `# ADR-NNN:` heading matches it.
+- New records follow [template.md](template.md): Context, Decision, Implementation, Rationale, Consequences (Positive / Negative / Mitigations), Alternatives Considered, Notes, References.
+- Take the next free number. **Never reuse a number**, even if a record was withdrawn — a reused number silently breaks every reference pointing at the original.
+- Superseding a decision is normal; deleting the record is not. Leave the old record in place, mark it superseded at the top, and have the new record name what it supersedes in its References section. The decision trail is the point.
```
XRFXLP, line 10: can we explicitly call out that we generally avoid changing ADR for new feature and rather do the changes in general docs and configuration?

AGENTS.md
```diff
@@ -0,0 +1,394 @@
...
+
+<!-- GOOD - a reworded sentence is a one-line diff -->
+The node drainer evicts workloads from a quarantined node. It respects pod disruption budgets and will not evict pods that have no controller managing them.
+```
+
+This applies to new files and to sections you add. When editing a file that is already hard-wrapped throughout — `CODE_OF_CONDUCT.md`, for instance, which carries upstream Contributor Covenant text — match the surrounding style rather than leaving one reflowed paragraph in an otherwise wrapped document.
+
+Also: fenced code blocks need a language (` ```bash `, ` ```go `, ` ```text ` for diagrams), and blank lines go before and after headings, lists and fences.
+
+## Testing
```
XRFXLP, line 317: nit: also avoid raw `time.Sleep` in tests, use conditionals like `require.Never` and `require.Eventually`

AGENTS.md
```diff
@@ -0,0 +1,394 @@
...
+| Hardcoding a tool or image version in a script | Read it from `.versions.yaml` |
+| Swallowing an error with `_ =` or a bare `continue` | Handle it, or `slog.Warn` with the error and say why continuing is safe |
+| Broad refactor bundled into a bug fix | Separate PRs — this repo squash-merges, so the fix and the refactor become one unreviewable commit |
+| Assuming MongoDB | The store is pluggable; PostgreSQL is a supported backend and is covered in the E2E matrix |
+| Implementing an approach an ADR already rejected | Read `docs/designs/` first; if the rejection no longer holds, say so and propose superseding the ADR |
+| Treating `health-events-analyzer` as a distinct pipeline stage | It is a health monitor like any other — same gRPC publish path |
+| Renaming a proto field, Helm value, metric, or node label | Add the new one alongside; keep the old working. Nothing in CI catches this |
+| Adding retry/fallback around an internal call "just in case" | Only add retries where a real transient failure exists; invented resilience hides bugs |
+
+## Pull Request Requirements
```
XRFXLP, line 377: Should we add to make sure to open the issue with all the necessary details before opening PR? I'm not fully sure about this because it is slightly ambiguous on when this is going to be consumed by agents.

XRFXLP (commented):
All nits, nothing blocking.

docs/designs/README.md
lalitadithya: good callout, updated

AGENTS.md
lalitadithya: added, ptal

AGENTS.md
lalitadithya: yeah, I added it nonetheless since I think we are starting to see agents open issues as well make PRs so good to have it in here

XRFXLP (approved):

PR #1780, opened 2026-09-09
feat(metadata-collector): expose pod mapper failure metrics

### LINKED ISSUE #1767: [Bug]: metadata-collector exits on a single failed poll tick, so a credential rotation restarts it fleet-wide

lfriedman-netllama:
Closes #1767. PR 2 of 2, the metrics half of the split you agreed to; PR 1 was #1778 (threshold + widened backoff), merged.

Two metrics, the names proposed on the issue:

| Metric | Type | For |
| -- | -- | -- |
| `metadata_collector_pod_mapper_failures_total` | counter | alerting on a rising failure rate |
| `metadata_collector_pod_mapper_consecutive_failures` | gauge | the streak the exit threshold acts on, so an alert can fire *before* the container dies |

Both are the "increasing failures" signal you asked for, and the gauge is the more directly alertable of the two since it is the exact quantity compared against `--pod-mapper-max-consecutive-failures`.

## Standing the endpoint up

metadata-collector had **no metrics surface at all** — no `client_golang`, no registry, no handler, no port. `commons/pkg/server` already does this, so the endpoint is:

```go
server.NewServer(server.WithPort(*metricsPort), server.WithPrometheusMetrics()).Serve(groupCtx)
```

on `--metrics-port`, default **2112**, matching every other component. The mapper loops until the context ends, so the server runs beside it in an `errgroup`; either goroutine returning cancels the other, which keeps the threshold's non-zero exit intact.

## The chart needed less than expected

**The PodMonitor already selected these pods.** It matches on the release's `app.kubernetes.io/instance` label and scrapes a container port *named* `metrics` — metadata-collector had no such port, so it was selected and silently never scraped. The DaemonSet just needed the port. No PodMonitor change.

**One thing worth your judgement:** metadata-collector is the only component with `hostNetwork: true`. The other five bind `metrics: 2112` inside their own pod netns; this one binds it on the node. I checked the fleet where this was found — nothing claims host `2112` there (node-exporter is 9100, cilium 9963/9964, vast-csi 9090/9091, etcd 2381) — so 2112 is free and consistent. But it is a different exposure from the rest, and if you would rather it not be reachable on the node IP, say so and I will give this component its own port value instead of `global.metricsPort`.

## Tests

- failures counted independently of the streak: `fail, fail, succeed, fail` gives **3** on the counter and **1** on the gauge, not 3 and 3
- a success resets the gauge to 0
- reaching the threshold records the final streak **before** returning, so the scrape preceding the exit still shows why
- both documented names are exposed, so a rename breaks a test rather than someone's alert
- **the registry wiring is pinned**: `main` registers on `prometheus.DefaultRegisterer` and the endpoint serves `prometheus.DefaultGatherer`, so a test asserts the metrics are gatherable from the default gatherer specifically. This is the mismatch that made the change-stream lag metric invisible in #1762 while every unit test passed against its own registry, so it seemed worth pinning rather than assuming. Verified decisive: swapping in a private registry fails it.

A note on test synchronisation, since the first version was wrong: assertions happen only after `pollPodDevices` has returned. A received tick is always processed before the loop reaches `select` again, so once the function returns exactly the sent ticks have been fully handled. Reading a metric between sends races with the tick still in flight.

Module green with `-race`, `golangci-lint` 0 issues, `gomod-lint` clean, `helm lint` clean.

## For the release team

let's call out that the users have to watch for this change and tune the port if required or disabled based on their preference. 

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for device-mapping poll failures and consecutive failure streaks.
  * Added a configurable metrics endpoint, enabled by default on port 2112.
  * Successful updates reset the consecutive-failure indicator, while failures are counted independently.
  * Kubernetes deployments publish the configured metrics port for monitoring discovery.
  * The metrics endpoint can be disabled by setting its port to `0`.

* **Bug Fixes**
  * Metrics server errors no longer terminate the metadata collector or cause a failed exit.
  * Invalid metrics configuration types are rejected during chart rendering.


lfriedman-netllama:
Fork PR, so CI needs a maintainer trigger: `/ok to test 067980741e8b398b83c19e14a19c2a802e719f01` when convenient.

@lalitadithya this closes #1767. Two things I would value your call on rather than deciding myself:

1. **The hostNetwork port.** metadata-collector is the only component with `hostNetwork: true`, so `global.metricsPort` binds 2112 on the node rather than in a pod netns like the other five. Nothing claims host 2112 on the cluster where this was found, so it works and stays consistent, but it is a genuinely different exposure. Happy to give this component its own port value if you would rather it not be reachable on the node IP.
2. **Whether the gauge is the right alerting primitive.** `consecutive_failures` is the exact quantity compared against the exit threshold, so an alert on it fires before the container dies, which I think is what you wanted from "alert on increasing failures". The counter is there for rate-based alerting across the fleet. If you would rather have only one of the two, the gauge is the one I would keep.

distros/kubernetes/nvsentinel/charts/metadata-collector/templates/daemonset.yaml
```diff
@@ -55,6 +55,13 @@ spec:
           args:
             - --output-path={{ .Values.global.metadataPath }}
             - --pod-mapper-max-consecutive-failures={{ .Values.podMapper.maxConsecutiveFailures }}
+            - --metrics-port={{ .Values.global.metricsPort }}
+          ports:
+            # The chart's PodMonitor selects on the release instance label and scrapes a port
+            # named "metrics", so this name is what makes the pods scraped. Note this container
+            # is hostNetwork, unlike the other components, so the port binds on the node.
+            - name: metrics
+              containerPort: {{ .Values.global.metricsPort }}
```
lalitadithya: Can we have a chart specific override for this one? As you rightly pointed out this is host networked which mean that it might need a different port than others. I'm also thinking we may want to consider an option to disable this since it looks like 2112 is also used by kube-vip

lalitadithya:
hey @lfriedman-netllama , can you please check the pipeline failure?

lfriedman-netllama:
Pipeline fixed, and your chart point implemented, in `465497ac` (plus a `main` merge).

**The build failure was mine and simple.** The Dockerfile ran `go build ... main.go` — a single *file*, not the package — so `metrics.go` was never compiled and `main.go`'s references to it were undefined. Reproduced it exactly:

```
$ go build -o /tmp/mc main.go
./main.go:96:30: undefined: newPodMapperMetrics
./main.go:108:46: undefined: podMapperMetrics
./main.go:127:63: undefined: podMapperMetrics
```

Every other gate passed because `go build ./...`, the tests and the linters all compile the package, so nothing local caught it. It now builds `.`, which also stops the next file added to package main breaking the image. Worth noting `health-monitors/syslog-health-monitor/Dockerfile` has the same latent form — fine today with one file, would fail the same way.

**On the port, agreed on both counts, and thank you for the kube-vip pointer** — I had checked that nothing on our cluster claimed host 2112, but a conflict on someone else's is exactly the case I could not see. Now chart-local:

```yaml
metrics:
  enabled: true
  port: 2112
```

`metrics.enabled: false` renders `--metrics-port=0` and omits the `ports:` block entirely. Passing 0 rather than dropping the flag is deliberate: omitting it would leave the binary on its own default, so `enabled: false` would not actually disable anything. Verified all three renders (default, disabled, custom port).

**One change beyond what you asked, which I would rather you overrule than not raise.** A failed metrics listener no longer takes the process down: the error is logged and the mapper carries on. On the host network a port collision would otherwise crash-loop this DaemonSet fleet-wide, which is precisely the failure #1767 exists to prevent — losing metrics is worth strictly less than losing the collector. It does cut against failing loudly for the endpoint itself, so say the word if you would rather it stay fatal and I will revert that part.

Needs a fresh `/ok to test 8ee04ce31898fc3f8b39909c3d161ee175e92bdb`.

lfriedman-netllama:
Addressed in `b20f2a59`. The finding was valid and I reproduced it before changing anything:

```
--set        metrics.enabled=false  ->  --metrics-port=0     (correctly disabled)
--set-string metrics.enabled=false  ->  --metrics-port=2112  (silently ENABLED)
```

**I went further than the suggested comment, because this chart already has a stronger precedent.** `nvsentinel.pcAuth.enabled` in the parent chart's `_helpers.tpl` *fails the render* on anything that is not a real boolean, with reasoning that transfers word for word: quoted strings, null and numbers would otherwise silently enable or disable the thing. Documenting a footgun leaves it loaded, and it matters a little more here than for an ordinary flag because this endpoint binds a port on the **node**, so a mistyped `enabled` is how you end up in the kube-vip collision you flagged.

So `metadata-collector.metrics.enabled` now goes through a `metadata-collector.metricsEnabled` helper that refuses non-booleans. All four bad inputs verified:

| input | result |
| -- | -- |
| `"false"` (string) | refused: `got string` |
| `null` | refused: `got invalid` |
| `0` (int) | refused: `got int` |
| `metrics` block absent | refused: `got invalid` |

Real booleans behave as before, and `metrics.port: 2113` still moves both the flag and the container port. `helm lint` clean.

Values now carries the two-line note the review asked for — the boolean requirement and a custom-port example — just kept short since the helper is what actually enforces it.

Needs a fresh `/ok to test b20f2a5992e59eb02c56f1ee7776dc32eb9b2abc`. The two open questions from earlier are still yours whenever you get to them: whether the non-fatal metrics listener should stay non-fatal, and whether you want both metrics or just the gauge.

lalitadithya (approved):

Issue #1782, opened 2026-09-09
[Feature]: GPU reset cannot be extended without forking janitor


sugarraysam:
### Prerequisites

- [x] I searched existing issues

### Code of Conduct

- [x] I agree to follow NVSentinel's Code of Conduct

### Feature Summary

Janitor executes three remediation actions. Two of them are pluggable and one is not.

`RebootNode` and `TerminateNode` delegate the work to a gRPC provider through `CSPProviderService`. `GPUReset` runs a fixed state machine that creates a Kubernetes Job from a template built in Go. An operator cannot replace the reset procedure, and cannot adjust the Job that performs it.

This asks for the GPU reset equivalent of the reboot and terminate seam, so that GPU reset can be extended without forking janitor.

### Problem/Use Case

[ADR-017](https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/017-remediation-plugins.md) chose gRPC delegation for remediation backends, and stated the goal directly: end users must be able to supply their own remediation backend without forking janitor. It was implemented for reboot and terminate. GPU reset was designed on a separate track in [ADR-019](https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/019-janitor-gpu-reset.md) and [ADR-020](https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/020-nvsentinel-gpu-reset.md) and never received the same treatment.

**The reset Job is built in code.** The signature of `getDefaultGPUResetJobTemplate` in `janitor/pkg/config/default.go` is the entire configuration surface:

```go
func getDefaultGPUResetJobTemplate(namespace string, image string, secrets []ImagePullSecret,
	resources ResourceRequirements, runtimeClassName string, writeSyslogEvent bool,
	uploadURL string) (*batchv1.JobTemplateSpec, error)
```

Everything else in the pod spec is a compile-time constant, including the host paths:

```go
HostDevPath    = "/dev"
DriverRootPath = "/run/nvidia/driver"
HostSysPath    = "/sys"
```

**There are no hooks.** `reconcileHelper` advances through seven condition-gated steps. There is no hook before the reset, none after it, and `GPUResetConditionType` and `GPUResetReason` in `janitor/api/v1alpha1/gpureset_types.go` are closed enums, so a downstream cannot add a condition to report its own state.

[...]

PR #1783, opened 2026-09-09
docs(janitor): add ADR-056 for GPU reset provider delegation

sugarraysam:
## Summary

Draft ADR proposing a provider seam for GPU reset. Documentation only — no code, no proto, no chart changes in this PR.

Refs #1782.

Janitor executes three remediation actions. `RebootNode` and `TerminateNode` delegate to a gRPC provider through `CSPProviderService`, per [ADR-017](https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/017-remediation-plugins.md). `GPUReset` does not. It runs a fixed state machine that creates a Job from a template built in Go, with host paths as compile-time constants and no pre-reset or post-reset hook. The issue describes the impact: we maintain a patched janitor build to run GPU reset on Blackwell bare metal, and the patches divide along four missing extension points.

ADR-056 proposes two additive, opt-in changes. Default behaviour does not change.

1. Add `SendGPUResetSignal` and `IsGPUResetComplete` to the existing `CSPProviderService`. When `gpuResetController.cspProviderHost` is set, the GPUReset controller delegates the reset to the provider instead of creating a Job. This replaces steps 4 and 5 of the seven-step state machine in `reconcileHelper`; the CR lifecycle, node lock, maintenance contention, finalizer, metrics and tracing all stay in janitor.
2. Add `gpuResetController.resetJob.podTemplateOverride`, a strategic merge patch over the built-in Job pod template, for deployments that only need a different host layout.

Opening as a draft to get direction on approach before writing any implementation. Per AGENTS.md, an agent may draft an ADR but a maintainer owns the decision, so I am not building against this yet.

**Two things I would like a maintainer view on:**

- `gpuResetController.cspProviderHost` renders into the ConfigMap today and no code reads it. Reading the existing key is tidier and the field was clearly added for this purpose, but a cluster that already sets it would start delegating after an upgrade. Introducing a separate key is the safer option. I do not have a strong view and the record says so.
- Whether the seam should be delegation at all. If maintainers prefer pre/post hooks around the existing Job, or want only the `podTemplateOverride` half, say so and I will rework the record. The ask is the seam, not this particular design. The Alternatives Considered section covers five rejected options, including routing through `ExternalRemediationRequest` and a full `jobTemplate` replacement.

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Health Monitors
- [ ] Fault Management
- [ ] Deployment/Config
- [x] API/Interface
- [ ] Preflight
- [ ] Plugins
- [x] Janitor
- [x] Documentation/CI
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

Documentation-only change: two Markdown files, no code paths touched. I verified the record follows `docs/designs/template.md` section by section, that all six relative links resolve, that 056 is the next free number, and that the prose stays under the 3000-word guidance in AGENTS.md.

I did not get a clean local `make lint-test-all` run, and I would rather say so than imply a pass I did not see. The target begins with `protos-lint`, which invokes `protos-generate`; that needs `grpcio-tools` and `addlicense`, and neither is present without `make dev-env-setup` on this machine. The change touches no Go, Python, protobuf, Helm or shell file, and CI's path filters skipped the lint-test matrix for the same reason. DCO passes.

The record's own compatibility section covers the surfaces the eventual implementation would touch: the proto change is additive with `codes.Unimplemented` as a defined fallback, there is no CRD change, no condition rename, and one new Helm key.

## Checklist
- [x] Self-review completed
- [ ] Ready for review

### DIFF
--- a/docs/designs/056-gpu-reset-provider-delegation.md
+++ b/docs/designs/056-gpu-reset-provider-delegation.md
@@ -1,0 +1,254 @@
+# ADR-056: Janitor — GPU Reset Provider Delegation
+
+## Context
+
+Janitor executes three remediation actions, each driven by its own CR: `RebootNode`, `TerminateNode` and `GPUReset`. Two of the three are pluggable. The third is not.
+
+`RebootNode` and `TerminateNode` delegate the work to a gRPC service. The controller dials `CSPProviderService` and calls `SendRebootSignal`, `IsNodeReady` or `SendTerminateSignal`. Janitor keeps the CR lifecycle, the node lock, the metrics and the timeout. The provider decides how to reboot or terminate the machine. [ADR-017](017-remediation-plugins.md) made this decision, and stated the goal plainly: end users must be able to supply their own remediation backend without forking janitor.
+
+`GPUReset` never received the same treatment. It was designed separately in [ADR-019](019-janitor-gpu-reset.md) and [ADR-020](020-nvsentinel-gpu-reset.md) as a Kubernetes Job, and the Job is built in Go:
+
+```go
+// janitor/pkg/config/default.go
+func getDefaultGPUResetJobTemplate(namespace string, image string, secrets []ImagePullSecret,
+	resources ResourceRequirements, runtimeClassName string, writeSyslogEvent bool,
+	uploadURL string) (*batchv1.JobTemplateSpec, error)
+```
+
+The signature is the whole configuration surface. Everything else in the pod spec is a compile-time constant, including the host paths:
+
+```go
+HostDevPath    = "/dev"
+DriverRootPath = "/run/nvidia/driver"
+HostSysPath    = "/sys"
+```
+
+This produces two gaps. An operator cannot replace the reset procedure, and an operator cannot adjust the Job that performs it. There is no hook before the reset, no hook after it, and no way to add a condition to the CR.
+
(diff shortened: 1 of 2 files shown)


docs/designs/056-gpu-reset-provider-delegation.md
```diff
@@ -0,0 +1,254 @@
...
+| Add a client barrier, a timeout latch and durable host-service restoration to the reconciler | There is no post-reset hook, and the condition and reason enums are closed |
+| Extend the Helm chart and RBAC to carry the above | Follows from the three above |
+
+The specific policy in that fork is deployment-specific and does not belong upstream. The absence of any seam to hold it is the defect this record addresses.
+
+## Decision
+
+Make GPU reset pluggable through two additive changes.
+
+1. **Add `SendGPUResetSignal` and `IsGPUResetComplete` to the existing `CSPProviderService`.** When `gpuResetController.cspProviderHost` is set, the GPUReset controller delegates the reset to the provider instead of creating a Job. This completes ADR-017 for the third action.
```
lalitadithya, line 47: just to make sure I understand, what you are suggesting here is that we will keep the code where it lives today and only provide an option to delegate to the gRPC service? The other option would be to move the code to the janitor provider which would become a breaking change since users will have to add new flags to retain the behaviour they already have

PR #1789, opened 2026-09-10
docs(scale-tests): add end-to-end scale test report

### LINKED ISSUE #1705: Run scale test end-to-end
## Feature Summary

Given that almost all of the pull requests addressing the action items that came out from microbenchmarking are merged, we should be in position to run the full scale end to end test.

As part of this issue: Run a single graded end-to-end scale test of NVSentinel as a whole system from 1k to 100k nodes, under continuous moderate load with periodic bursts sustained over 1+ hour per scale point, and publish a report.

## Problem / Use Case

We can currently describe how each component behaves alone. We cannot answer the questions a customer actually asks:

- How much CPU and memory does NVSentinel need for a fleet of N nodes with P pods per node?
- How much load does it put on the Kubernetes API server, and does that compete with the customer's own workloads?
- If a GPU fails, how long until the node is cordoned? Drained? Remediated? Labelled?
- Do those numbers hold at 100k nodes, or only at 1k?

Without agreed numbers we cannot publish SLAs, and field teams are left extrapolating from component benchmarks that do not compose linearly.

### Part A — Report format

Notation used in the tables below, so each number's basis is visible where it appears:

| Marker | Meaning |
|---|---|
| `[M]` | **Measured** on the test cluster. A real observation. |
| `[S]` | **Simulated** — a harness constant, not a property of the system. |
| `[I]` | **Input** — a parameter the reader supplies for their own environment (object shape, hardware reboot time). |
| `[V]` | **To verify** during the run — an assumption for now. |

Four mandatory sections. Proposed table shapes below so we are agreeing on something concrete rather than on headings.

#### A1. Component sizing

Per component, at each scale point, expressed **per node and per pod** so it can be extrapolated:

| Component | Nodes | Pods | CPU (med/peak) `[M]` | RSS (med/peak) `[M]` | Recommended request | Recommended limit | Retained bytes/node `[I]` | Retained bytes/pod `[I]` | Fixed overhead per object `[M]` | Inflation factor `[M]` |
|---|---|---|---|---|---|---|---|---|---|---|

Memory scales with which fields each component's informer retains, not with total object size. Transforms keep some fields and drop others, so two objects of the same size can cost very different amounts — the same padded pod is ~9.3 KB in KOM (no transform) and ~200 B in janitor (transform drops annotations). A single "bytes per node" figure would describe the test fixture's padding rather than the component, so the table carries the underlying terms instead.

Sizing formula:

```
M ≈ base + N × (fixed_overhead_per_object + retained_bytes × inflation_factor)
```

- `fixed_overhead_per_object` — cache entry, pointers, headers. Content-independent.
- `inflation_factor` — Go in-memory size ÷ wire bytes for retained fields.
- `retained_bytes` — computed per component by applying its transform to the reference object profile (published in Part B). Customers substitute their own object shapes.

Measured by sweeping padding placement at a fixed node count and regressing RSS against retained bytes. Sweeping stripped-field padding in parallel should yield slope ≈ 0 for components that strip it, which doubles as a transform-correctness check.

Plus a QPS table, because client-side rate limits turned out to dominate throughput:

| Component | `--kube-api-qps` used | Burst | API calls per operation | Throughput achieved | Recommended setting |
|---|---|---|---|---|---|

For the **labeler**, "throughput achieved" means nodes fully labelled per second — all of `driver.installed`, `dcgm-bootstrap-completed`, `dcgm.version`, `nvidia.com/gpu.count`/`present`/`product`, `gpu.count.current`/`expected`, `nic.count.current`/`expected`, `kata.enabled` — measured on initial fleet ramp and on re-label after a GPU/NIC count change.

#### A2. Load on external components

| Metric | Idle | Continuous load | During burst | Notes |
|---|---|---|---|---|
| API server request rate (by verb) | | | | |
| API server request rate (by component) | | | | |
| APF seats consumed / rejections (429s) | | | | |
| etcd write rate, DB size growth | | | | |
| MongoDB: ops/s, connections, oplog churn, storage | | | | |
| Node update (heartbeat) event rate `[M]` `[V]` | | | | drives reconcile load on every node-watching component; KWOK-vs-real ratio to be confirmed |

The node heartbeat rate is the input load for every component that watches Nodes. What each one pays per update differs, and the report should break it out rather than quote a single figure:

| Component | Node watch | Cost per node update |
|---|---|---|
| kubernetes-object-monitor | `NewControllerManagedBy(...).For(gvk)`, no event filter | full reconcile enqueued per event (when a Node policy is configured) — the heaviest |
| fault-quarantine | informer + `UpdateFunc`, no predicate | handler runs on every update |
| labeler | informer + `UpdateFunc` guarded by `nodeRequiresReconciliation` | wakes, early-returns unless a relevant field changed |
| node-drainer | informer + transform, no event handler | cache update only (decode + transform + store write) |
| janitor | Node cache with transform, no Node controller | cache update only |

Components without handlers still pay the informer floor: every event is decoded, transformed and written to the store. At 100k nodes that cost is incurred continuously, with no faults occurring.

Fidelity caveat to verify during the run: KWOK patches the node object every 30 s, whereas real kubelets with NodeLease update the *lease* every ~10 s and the node object only every ~5 min (leases do not wake node informers). If that holds, KWOK produces roughly 10× more node-object churn than a real cluster and our watch-load figures are pessimistic; we will state the ratio.

Object-creation load from the harness (ramping nodes/pods) is genuinely artificial and is excluded from customer-facing figures.

#### A3. Customer-facing SLAs

The headline section. Percentiles, not averages — customers care about the tail.

| SLA | P50 | P90 | P99 | Max | Conditions / caveats |
|---|---|---|---|---|---|
| Time to cordon (fault detected → node cordoned) `[M]` | | | | | |
| Time to label (`dgxc.nvidia.com/nvsentinel-state`) `[M]` | | | | | |
| Time to drain (cordoned → drain-succeeded) `[M]` | | | | | |
| Time to remediate (drained → remediation CR created) `[M]` | | | | | excludes the simulated reboot wait |
| **NVSentinel-attributable MTTR** `[M]` | | | | | sum of the measured phases |
| Simulated reboot wait `[S]` | | | | | harness constant, excluded from totals |
| Customer end-to-end MTTR `[I]` | | | | | = NVSentinel-attributable MTTR + reader's own reboot time |
| Sustained fault rate supported (nodes/hour) `[M]` `[S]` | | | | | inflated by the short simulated reboot |
| Burst absorption (N nodes at once → time to clear) `[M]` `[S]` | | | | | same caveat |

#### MTTR decomposition

Reboots are simulated end to end: the `kind` provider sleeps 3-6 s in `SendRebootSignal` and then reports the node ready. No node is taken out of service, so the reboot phase is a chosen constant rather than a measurement. The measurable part is the work our components do either side of it:

| Phase | Measured? | Notes |
|---|---|---|
| Detect (event ingested) → cordon | yes | |
| Cordon → drained | yes | scales with pods/node |
| Drained → remediation CR created | yes | |
| CR created → provider returns | **no — simulated constant** | `kind` provider sleeps 3-6 s; configurable, carries no physical meaning |
| Provider returns → node marked back in service | yes | node state is simulated throughout |

So the headline figure is **NVSentinel-attributable MTTR**: the sum of the measured phases, with the simulated wait stated separately and excluded. Readers add their own reboot time to get a total for their hardware.

Each SLA must state the conditions it holds under — at minimum fleet size, pods per node, eviction mode, and for MTTR the simulated remediation duration.

### Part B — Methodology (report section A4)

Both how we run the test and what gets written up as the report's fourth mandatory section.

**To document:**

- Cluster topology, node types, versions (NVSentinel, K8s, MongoDB)
- **Reference node and pod field composition, with byte counts** — every sizing number in A1 is relative to this, so it must be published for the table to be reproducible or transferable
- KWOK simulation approach and its known divergences from real nodes (including the node-object update frequency ratio noted in A2)
- Load generation: tools, rates, burst shape, event content
- Measurement: which metric or timestamp defines each SLA boundary, and its resolution
- Simulated remediation timings (see MTTR decomposition)
- What was **not** tested, stated plainly
- Reproduction steps

**Scale points (graded):** 1k → 5k → 10k → 25k → 50k → 100k nodes. Pods per node held at a fixed realistic value (proposal: **10**), since pods-per-node — not node count — drives drain cost. One additional pods-per-node variation at a single scale point to validate the extrapolation.

**Load profile per scale point**, sustained 1+ hour:

- **Continuous:** moderate fault rate, well below saturation. Proposal: a fixed **0.5 nodes/s** fleet-wide, which is far above any realistic fleet failure rate and still leaves headroom.
- **Bursts:** every 15 minutes, inject a correlated failure (proposal: **100, 250, 500, 1000 nodes** simultaneously, one size per burst) to simulate a rack or switch failure.
- **Steady state between bursts** so recovery is observable.

We are explicitly **not** testing to destruction. The question is "does the system meet SLA under realistic load", not "where does it break".

XRFXLP:
## Summary

Preview at: [NVsentinel end-to-end scale test report](https://github.com/XRFXLP/NVSentinel/blob/scale-test-report/tests/scale-tests/scale-test-report.md)

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Health Monitors
- [ ] Fault Management
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Janitor
- [ ] Documentation/CI
- [ ] Other: ____________

## Testing
- [ ] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Updated the NVSentinel 100,000-node scale-test report with revised CPU and memory measurements.
  * Refreshed resource-sizing recommendations, component resource tables, benchmark configuration, and deployed settings.
  * Added Kubernetes object-monitor scale sweeps and findings on working-set behavior and pod-transformation costs.
  * Documented health-event analysis concurrency, index-dependent throughput, and fault-quarantine circuit-breaker results.
  * Updated SLA, external-system, infrastructure, methodology, operational-trap, and measurement findings, including queue-memory and latency calculations.


tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,714 @@
...
+- [A1. Component sizing](#a1-component-sizing)
+    - [kubernetes-object-monitor](#kubernetes-object-monitor)
+    - [fault-quarantine](#fault-quarantine)
+    - [labeler](#labeler)
+    - [node-drainer](#node-drainer)
+    - [preflight](#preflight)
+    - [fault-remediation](#fault-remediation)
+    - [health-events-analyzer](#health-events-analyzer)
+    - [janitor](#janitor)
+  - [QPS](#qps)
```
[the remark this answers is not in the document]
XRFXLP: should be resolved

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,714 @@
...
+
+<img width="1330" height="770" alt="Image" src="https://github.com/user-attachments/assets/fdcfa365-68f1-4e96-b6b9-02c4cd121448" />
+
+One note should be called here: memories in the graph above were considered to that of peak and not steady state, this sometimes differs by large margin because of the pruning that we do inside the modules like FQ (strip `node.status.*`) which makes it temporarily store the full raw object. As an example:
+
+| Component        | peak    | settled | peak per node | vs raw wire |
+| ---------------- | ------- | ------- | ------------- | ----------- |
+| labeler          | 6.54 GB | 2.0 GB  | 247,550 B     | 4.5x        |
+| fault-quarantine | 4.09 GB | 1.68 GB | 151,570 B     | 2.7x        |
+
```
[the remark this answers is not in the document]
XRFXLP: It is already added down below

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,720 @@
...
+| Burst absorption                                       | see below                                        |            |            |            |                                                                                                                                                                                                             |
+
+
+Percentiles are computed from per-document timestamps, so they are exact rather than snapped to Prometheus histogram buckets.
+
+Drain is 99.1% of the measured MTTR, and it is a wait rather than work: node-drainer evicts the pod, requeues at its 10 s base backoff, confirms the pod is gone and marks the node drained. Everything NVSentinel does outside that wait totals 91 ms, two orders of magnitude below the backoff constant, so MTTR at this fleet size is set by that constant rather than by anything that grows with node count.
+
```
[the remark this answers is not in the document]
XRFXLP: this is resolved now

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,739 @@
...
+
+### QPS
+
+The last column is the one to compare across rows: a client limit only means something once it is divided by what that component spends per node. A component with a high limit and an expensive path can be the bottleneck while one with a low limit and a single call is not.
+
+| Component | `--kube-api-qps` / burst | API calls per node | Nodes/s at the limit | Throughput measured |
+|---|---|---|---|---|
+| fault-quarantine | 100 / 200 | 1 PATCH per cordon | **100** | 41.2 cordons/s (100-node burst) |
+| node-drainer | 400 / 800 | 3.7 mean per node over a 100-node burst (audit logs, mostly nodes with nothing to evict); the eviction path itself costs 3.0 per evictable pod, so 15 at 5 pods/node | **108** at 3.7, **27** at 5 pods/node | 19.2 evictions/s = 3.85 nodes/s (1,000-node burst) |
```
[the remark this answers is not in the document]
XRFXLP: isn't this resolved now?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,734 @@
...
+
+Two components are most of the component total. At 100,000 nodes with a pod on every node, kubernetes-object-monitor is 33.9 GB and fault-quarantine 17.9 GB; labeler is 6.6 GB, janitor 4.9 GB, node-drainer 1.0 GB, and nothing else exceeds 0.6 GB.
+
+CPU was never a constraint. The busiest component peaked at 2.00 cores, four cores each is sufficient with headroom, and nothing recorded a single throttled CFS period.
+
+One component does not keep up at fleet scale, for reasons unrelated to its resources. health-events-analyzer consumes its event stream serially and takes about 17.6 ms per event, so it saturates near 55 events/s, while 100,000 nodes at 0.1 events per node per second would offer 10,000. Neither more CPU nor more replicas moves that as the component is built today.
```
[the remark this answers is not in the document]
XRFXLP: fixed

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,799 @@
...
+| 75,005  | ~100,000 | 0.13 / 0.29        | 3.52 G                           | 4 Gi         | 8 Gi       |
+| 100,005 | 731      | 0.05 / 0.09        | 4.17 G                           | 6 Gi         | 12 Gi      |
+| 100,005 | ~100,000 | 0.10 / 0.36        | 4.90 G                           | 6 Gi         | 12 Gi      |
+
+
+### Queue memory
+
+Both components hold one entry per pending event, so a backlog is memory the sizing tables above do not include. Two points each:
+
+| Component | Backlog source | Queued events | Working set | Per event `[M]` |
```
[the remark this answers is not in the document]
XRFXLP: resolved?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,799 @@
...
+| --- | --- | --- | --- | --- |
+| node-drainer | replay | ~200,000 | 182 → 315 MB | 0.67 KB |
+| node-drainer | replay | 1,048,415 | 255.7 → 979.7 MB | 0.69 KB |
+| fault-remediation | cold start | 200,668 | 41.5 → 133.7 MB | 0.47 KB |
+| fault-remediation | live stream | 1,389,136 | 791 → 9,463 MB | 7.79 KB |
```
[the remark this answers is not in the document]
XRFXLP: resolved?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+| drained → remediation dispatched | 0.08 s                         | 10.52 s                  |
+
+
+P90 sits within a second of P50 in every continuous row and within eight seconds in every burst row; all 500 burst nodes and all 400 continuous nodes completed.
+
+Under continuous load a drain costs one node-drainer recheck cycle. It evicts the pod, requeues at its 10 s base backoff, sees the pod gone, and marks the node drained, which is why the whole distribution sits in a 0.8-second band around 10 s rather than spreading out. Under a burst the same cycle repeats while the node waits its turn, giving 52 s for five hundred nodes arriving together.
+
+### Burst absorption
+
+N nodes fail at once, on brand-new KWOK nodes with no quarantine history and no pods, so the two stages measured here are fault-quarantine's cordon and fault-remediation's CR creation without drain in between. Each size was injected as a single insert and run in isolation: fresh node names, and nodes, events and CRs deleted before the next size started.
```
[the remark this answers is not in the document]
XRFXLP: The section already states each size ran once in isolation, and the run-count is now implicit in the recomputed basis. Adding the disclaimer is the defensive framing that has been stripped from this document repeatedly.

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+### NetworkPolicy enforcement broke and stayed broken `[M]`
+
+The AWS VPC CNI expands each `NetworkPolicy` into `PolicyEndpoint` objects, sharded by how many pods the selector matches. The sharding is strictly linear: driving a namespace-wide selector from 1,000 to 51,000 pods produced 1, 6, 21 and 51 shards at those points, exactly **1,000 pod endpoints per shard**.
+
+![PolicyEndpoint shards against selected pods](results/policyendpoint-sharding.png)
+
+NVSentinel's `metrics-access` policy selects every pod in its namespace, and the benchmark put roughly 158,000 pods there. Eighty-three shards existed and no more appeared, and the CNI went on enforcing what it had last programmed: MongoDB was refused on 27017 by a source-IP list naming pods that no longer existed, and policies deleted by `helm uninstall` sat `Terminating` for three days while still isolating their pods. Deleting the 83 stale shards restored connectivity in about two minutes.
```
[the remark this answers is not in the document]
XRFXLP: That paragraph records what happened during the benchmark; it is not a runbook, and the section was deliberately condensed to defer to #1792. Adding a deny-by-default recovery procedure puts remediation guidance in a findings narrative.

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+---
+
+## Summary
+
+NVSentinel has been tested upto 100k nodes, resource consumption grows predictably with fleet size and stays within ordinary limits with reasonable throughputs.
+
+The whole control plane costs about 235 GB of memory at 100,000 nodes. NVSentinel's own components are 65 GB of that and MongoDB is the remaining 170 GB combined across all replicas, which it spends on per-node database connections rather than on data. [ADR-052](../../docs/designs/052-deployment-platform-connector.md) removes those connections, and once it lands the same fleet is projected at about 75 GB.
+
+Two components are most of the component total. At 100,000 nodes with a pod on every node, kubernetes-object-monitor is 33.9 GB and fault-quarantine 17.9 GB; labeler is 6.6 GB, janitor 4.9 GB, node-drainer 1.0 GB, and nothing else exceeds 0.6 GB.
```
lalitadithya, line 52: kubernetes-object-monitor can we have a measurement which shows what would be the uses if there are no policies that watches pods?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+CPU was never a constraint. The busiest component peaked at 2.00 cores, four cores each is sufficient with headroom, and nothing recorded a single throttled CFS period.
+
+Cordon completes in 26 ms P50 and 165 ms P99 under continuous load, and a node carrying one evictable pod is drained about ten seconds after that, which is one of node-drainer's recheck cycles. Detection to drained is 10.11 s P50 and 10.29 s P99; five hundred nodes failing at once stretches that to 55.5 s with every node completing.
```
lalitadithya, line 56: > which is one of node-drainer's recheck cycles

is this a constant or configurable? If configurable can the customer increase it to make this go faster? What would be the impact?

lalitadithya, line 56: > Detection to drained is 10.11 s P50 and 10.29 s P99; five hundred nodes failing at once stretches that to 55.5 s with every node completing.

can we break this down to just have time to cordon and time to remediate?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+What broke during testing was the infrastructure around NVSentinel, not NVSentinel: the AWS VPC CNI stopped rebuilding `PolicyEndpoint` objects and went on enforcing stale rules, the EBS CSI provisioner ran out of memory and stopped attaching volumes cluster-wide, and etcd's 16 GB threshold put the cluster into read-only. Each took MongoDB down with it. The first two are described in the appendix; etcd's budget is in A2.
```
lalitadithya, line 58: > Each took MongoDB down with it.

why would the etcd outage take down MongoDB?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+
+| Nodes   | Control plane | MongoDB       | NVSentinel components |
+| ------- | ------------- | ------------- | --------------------- |
+| 10,000  | ~22 GB        | ~17 GB `[M]`  | 4.6 GB `[M]`          |
+| 25,000  | ~54 GB        | ~43 GB `[M]`  | 11.4 GB `[M]`         |
+| 50,000  | ~113 GB       | ~85 GB `[M]`  | 27.5 GB `[M]`         |
+| 100,000 | **~235 GB**   | ~170 GB `[M]` | **64.9 GB** `[M]`     |
+
+<img width="1330" height="770" alt="Image" src="https://github.com/user-attachments/assets/7e472dc9-8b7b-40f8-b141-3a59091fe00b" />
```
lalitadithya, line 76: can we only have images that are added to the PR used in the doc?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+| preflight                 | ~4,200        | ~5,200     | ~110                                 |
+
+
+Lets walk over component by component:
+
+<img width="1260" height="770" alt="Image" src="https://github.com/user-attachments/assets/ceadafc2-b5c8-49ea-8d3b-a1a2fea86ba1" />
+
+### kubernetes-object-monitor
+
+Node + Pod watches, one per enabled policy; CEL-derived transform (#1720). Per-node cost 174,175 B/node. Per pod cost: ~3,700 B/pod settled, 4,400 B at peak
```
lalitadithya, line 114: it would be good to break this down into just node watches and node + pod watches

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+| 100,005 | 0.29 / 0.46        | 16.17 G                | 12 Gi        | 18 Gi      |
+| 100,005 | **0.18 / 0.32**    | **17.88 G**            | 20 Gi        | 40 Gi      |
+
+
+### labeler
+
+Eager informers: one for Nodes with a fixed field projection, and **four pod informers**, each label-scoped -- `app in (dcgm, driver)`, the driver-component label excluding that app, `k8s-app=<gke-installer>`, and ResourceSlice objects. 
+
+
+| Nodes   | Labelled pods           | Settled    | Sync peak  | CPU med/peak | Rec. request | Rec. limit |
```
lalitadithya, line 150: Can we be a bit more conservative with the limits if we know the peak is much lesser than the actual limit?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+| ------- | -------- | ------------------ | ---------------------- | ------------ | ---------- |
+| 10,005  | 0        | 0.01 / 0.01        | 0.14 G                 | 512 Mi       | 1 Gi       |
+| 50,005  | 0        | 0.08 / 0.16        | 0.715 G                | 512 Mi       | 1 Gi       |
+| 75,005  | 0        | 0.09 / 0.21        | 0.898 G                | 512 Mi       | 1 Gi       |
+| 75,005  | ~100,000 | 0.19 / 0.48        | 1.00 G                 | 1 Gi         | 2 Gi       |
+| 100,005 | 0        | 0.13 / 0.20        | 1.01 G                 | 1 Gi         | 2 Gi       |
+| 100,005 | ~100,000 | 0.10 / 0.38        | 1.04 G                 | 2 Gi         | 4 Gi       |
+
+
+These figures are with an empty queue; see queue memory below.
```
lalitadithya, line 183: below where? can we link?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+### janitor
+
+Node informer created **lazily** on the first `TerminateNode` reconcile, so an idle janitor is **flat at 0.022 GB regardless of fleet size**. Once triggered it costs **19,616 B/node.**
+
+Retained bytes/node `~10,120` computed `[I]`; the measured 19,616 B is 1.94x that, the gap being Go map and pointer overhead. **Size for the triggered case**: 1 GB at 50k nodes, not the 22 MB an idle pod shows. Recommended **2 Gi / 4 Gi** at 50k.
+
+Per scale point, for the triggered case. An untriggered janitor reads 0.022-0.055 G at every fleet size, so the idle row is not a sizing figure.
+
+
+| Nodes   | Pods     | CPU med/peak `[M]` | Working set peak `[M]`           | Rec. request | Rec. limit |
```
lalitadithya, line 244: Same comment as earlier in the labeler about the limits, can we be more conservative here?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+| ------- | -------- | ------------------ | -------------------------------- | ------------ | ---------- |
+| any     | any      | 0.01 / 0.01        | 0.03 G untriggered               | —            | —          |
+| 25,005  | 731      | 0.02 / 0.04        | 0.93 G sync peak, 0.50 G settled | 1 Gi         | 2 Gi       |
+| 50,005  | 731      | 0.03 / 0.07        | 2.41 G                           | 2 Gi         | 4 Gi       |
+| 75,005  | ~100,000 | 0.13 / 0.29        | 3.52 G                           | 4 Gi         | 8 Gi       |
+| 100,005 | 731      | 0.05 / 0.09        | 4.17 G                           | 6 Gi         | 12 Gi      |
+| 100,005 | ~100,000 | 0.10 / 0.36        | 4.90 G                           | 6 Gi         | 12 Gi      |
+
+
+### Queue memory
```
lalitadithya, line 254: this section feels a bit out of place, can we merge these into the component sections itself?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+| fault-quarantine          | 100 / 200                | 1 PATCH per cordon                                                                                                                           | **100**                                              | 41.2 cordons/s (100-node burst)                    |
+| node-drainer              | 400 / 800                | 3.7 mean per node over a 100-node burst (audit logs, mostly nodes with nothing to evict), plus 3.0 per evictable pod, so 18.7 at 5 pods/node | **108** with nothing to evict, **21** at 5 pods/node | 19.2 evictions/s = 3.85 nodes/s (1,000-node burst) |
+| labeler                   | 500 / 1000               | 1 PATCH per relevant event                                                                                                                   | **500**                                              | 53,513 nodes labelled in 18.3 min                  |
```
lalitadithya, line 280: why can't the labeler also be unlimited?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+## A2. Load on external components
+
+*Measured at 50,021 nodes carrying 642,243 pods*
+
+The three columns are conditions on that fleet, not scale points -- idle, continuous load at 0.47 nodes/s, and a 100-node burst. 
+
+### Kubernetes API
+
+Rates per second. Idle and continuous differ mainly in KWOK lease traffic, not in NVSentinel's own load.
```
lalitadithya, line 298: We should remove KWOK and every other component in this measurement else it feels like we have results for components other than NVSentinel

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+| growth                  | none over 139 s | +288 objects / 638 s (0.45 obj/s, ~449 B/s) | +97 objects per 97 remediated nodes, ~94 KiB |
+
+
+All growth under continuous load is RebootNode CRs at 993 B median; pods and nodes are unchanged. `apiserver_storage_size_bytes` varies by up to 983 MB between consecutive scrapes, because each API server instance reports its own etcd backend's file size and those files are allocated independently. It also never shrinks, since freed space is reused inside the file rather than returned. The etcd threshold is 16 GB across tiers, but in-use size is published only through CloudWatch, and CloudWatch itself degrades under the load that matters. Its peak reading of 14.59 GB, about 91% of the 4XL tier's 16 GB, is therefore the only signal available rather than an authoritative measurement, and neither that figure nor the 91% derived from it should be used as a margin. The gap is genuine -- no in-cluster metric reports in-use size, since `apiserver_storage_size_bytes` is file-allocated:
+
+<img width="3136" height="736" alt="Image" src="https://github.com/user-attachments/assets/d861781c-403f-4f6e-9691-b13e6a795e56" />
+
+
+
+### MongoDB
```
lalitadithya, line 331: which version of MongoDB was used here, can we add a note to indicate that? Also are both versions of MongoDB equally performant?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+A remediated node costs **6.1 oplog entries, about 4 KB, and 1.96 stored documents** -- the event itself and its status record. Allocated storage does not move at this size, because those documents fit inside an extent the collection already holds. `[M]`
+
+### What etcd actually holds `[M]`
+
+etcd holds the live objects plus every revision written in the last five minutes, so its size is live data plus five minutes of churn. At 25,000 nodes that churn is about 2 GB:
+
+
+| Source                         | Rate      | Bytes per 5 min | Share   |
+| ------------------------------ | --------- | --------------- | ------- |
+| `leases/update`                | 1,114.6/s | 290 MB          | 14%     |
```
lalitadithya, line 357: are these leases NVSentinel creates?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+What a remediation costs, per node, during a 100-node burst:
+
+![API calls per remediated node, by verb](results/api-calls-per-node.png)
+
+
+| Component          | Requests per node | Breakdown                                                                                                   |
+| ------------------ | ----------------- | ----------------------------------------------------------------------------------------------------------- |
+| fault-remediation  | **9**             | 5 get, 3 update, 1 create                                                                                   |
+| janitor            | **6**             | 2 get, 2 update, 1 create, 1 delete                                                                         |
+| node-drainer       | 3.7               | get                                                                                                         |
```
lalitadithya, line 381: doesn't the drainer also call the evict api?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+An update matching no policy costs less again: patching 5,000 nodes with an irrelevant label drove 17,803 items through kubernetes-object-monitor's workqueue at **0.045 ms of handler time each** and moved no other component's count, since the update reaches their caches without becoming an event. That is 3.56 queue items per node write, so the 250 node writes per second this fleet does when idle come to about 890 items per second and 40 ms of handler time per second. This is serial handler wall time rather than a CPU counter, so it bounds rather than measures the core cost.
+
+---
+
+## A3. Customer-facing SLAs
+
+### Continuous load `[M]`
+
+Measured in burst-free windows, so the tails are steady-state rather than burst contention. The cordon row is a 314-node window on the 50,000-node fleet, where 0.47 nodes/s is the rate at which nodes completed rather than a rate events were offered at; the input rate was not controlled. The drain, remediation and MTTR rows come from a later run that injected at a set **0.5 nodes/s** across 400 nodes, each carrying a drain-eligible pod, and all 400 completed -- so there input and output rates are the same.
```
lalitadithya, line 434: why did we pick 50k nodes here instead of 100k?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+| Time to drain                        | 10.09 s                                          | 10.14 s     | 10.27 s     | 10.92 s     | cordon → drained, 400 nodes at 0.5 nodes/s, one drain-eligible pod each `[M]`. The 10 s is node-drainer's recheck backoff, not eviction time                 |
+| Time to remediate                    | 0.08 s                                           | 0.09 s      | 0.17 s      | 0.24 s      | drained → remediation dispatched across 400 nodes at 0.5 nodes/s `[M]`. Under a 200-node burst the same stage is 3.20 s, essentially all of it change-stream queue wait                                                                                                                  |
+| **NVSentinel MTTR**                  | **10.11 s**                                      | **10.16 s** | **10.29 s** | **10.94 s** | detect → drained, same 400-node run `[M]`. Almost all of it is the drain recheck cycle; detection to cordon is 17 ms at the median                                                                                                                                                       |
```
lalitadithya, line 443: Can we always define MTTR as excluding drain time and just have that as the only measuirement everywhere?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+**Restart the component before you measure it, or the number is wrong.** Reading the same six components at 50,000 nodes without restarting them — after the fleet had been reduced from 100,000 — gave 31.2 GB against 24.7 GB restarted, because none of them returns memory when the fleet shrinks.
+
+**Driving faults.** Health events are inserted directly into MongoDB, bypassing platform-connector. The document shape is unforgiving and fails silently when wrong: the event nests under `healthevent`, field names are their Go names lowercased, `generatedtimestamp` must be a `{seconds, nanos}` subdocument, `recommendedaction` is the enum integer, `errorcode` is an **array**, and `healtheventstatus.userpodsevictionstatus` must be an empty document rather than null. `agent` must match the deployed fault-quarantine ruleset.
```
lalitadithya, line 635: why did we bypass the platform connector, can we add that info to the doc?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+    kind = "HealthEvent"
+    expression = "event.agent == 'gpu-health-monitor' && event.componentClass == 'GPU' && event.isFatal == true"
+  [[rule-sets.match.all]]
+    kind = "Node"
+    expression = "!('k8saas.nvidia.com/ManagedByNVSentinel' in node.metadata.labels && node.metadata.labels['k8saas.nvidia.com/ManagedByNVSentinel'] == 'false')"
+  [rule-sets.cordon]
+    shouldCordon = true
+```
+
+The circuit breaker is configured at 50% over 5 minutes but disabled on the command line (`--circuit-breaker-enabled=false`), so none of the burst results were shaped by it. An event whose agent is anything other than `gpu-health-monitor` matches nothing and produces no cordon, which from outside is indistinguishable from a stalled pipeline.
```
lalitadithya, line 703: Can we also test that the circuit breaker trips at the correct node count at scale and reset of the breaker works as expected (in both modes)?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+Documents predating the schema fix store `generatedTimestamp` as a BSON date and abort aggregations doing arithmetic on it; queries need a numeric type guard.
+
+A component that cannot reach MongoDB is not necessarily a MongoDB problem. Read the `PolicyEndpoint` objects (`networking.k8s.aws/v1alpha1`) before investigating the database, the certificates or DNS: they carry the source-IP allow-lists that are actually enforced, and they go stale silently. A useful one-line discriminator is that a port with no listener returns a RST ("Connection refused") whereas a port blocked by policy times out; if an unlisted port times out, enforcement is active and the allow-list is the thing to read.
+
+`fault-quarantine`, `fault-remediation` and `node-drainer` expose readiness on an HTTP metrics server that comes up before the datastore connection is attempted, so `ready=True` does not mean the component reached MongoDB. Confirm from the log line `Successfully pinged database to confirm connectivity`. Trusting readiness produced two false recoveries during this run.
```
lalitadithya, line 766: I feel like this is a bug, no?

tests/scale-tests/scale-test-report.md
XRFXLP: An approximation of this should already be there in the rows with minimal pods: 
image

tests/scale-tests/scale-test-report.md
XRFXLP: This is a constant with exponential backoff ceiling at 2m:
```
		queue: workqueue.NewTypedRateLimitingQueueWithConfig(
			workqueue.NewTypedItemExponentialFailureRateLimiter[NodeEvent](10*time.Second, 2*time.Minute),
			workqueue.TypedRateLimitingQueueConfig[NodeEvent]{
				DelayingQueue: delayingQueue,
			},
		),
```

> If configurable can the customer increase it to make this go faster?

Yes, but that just means the events turnaround time would be faster, not very useful if we've AllowCompletion or DeleteAfterTimeout mode.

tests/scale-tests/scale-test-report.md
XRFXLP: Bad phrasing. Only CSI provisioner and network policy did, not all three. etcd blowup did trigger did blocked writes to the cluster.

tests/scale-tests/scale-test-report.md
lalitadithya: I saw that, but it is worth calling that out as "kubernetes-object-monitor is 33.9 GB (21.2 GB with minimal pods)"

but I think it would be good to have the numbers without the pod watch since that is not something we have in the default policies

tests/scale-tests/scale-test-report.md
XRFXLP: We should already have an approximation, but yeah, we can add this cleaner.

tests/scale-tests/scale-test-report.md
XRFXLP: https://github.com/XRFXLP/NVSentinel/blob/scale-test-report/tests/scale-tests/scale-test-report.md#queue-memory, will add anchor.

tests/scale-tests/scale-test-report.md
XRFXLP: Nope. These are kwok stuffs.

tests/scale-tests/scale-test-report.md
lalitadithya: yes, all of this only applies to immediate mode, now if we were to change 10s to 1s, what would it harm, I guess asked differently what the floor is for requeue time after which we will start to lose out?

tests/scale-tests/scale-test-report.md
XRFXLP: That is unrelated to the continuous load measurement. It is just that 100k nodes with pods is close to etcd 16G limit. Reducing the node size would not help much since majority of traffic is from harness so we had to use smaller fleet.

tests/scale-tests/scale-test-report.md
lalitadithya: can we drop those?

tests/scale-tests/scale-test-report.md
XRFXLP: yeah, I found ND and FR dead without any signal until restarted.

tests/scale-tests/scale-test-report.md
XRFXLP: Okay, will remove MTTR including drain.

tests/scale-tests/scale-test-report.md
lalitadithya: I see, can we just call that out in the document? Would it be possible to extrapolate these numbers to 100k at all? 

Also say more about majority of the traffic being from the harness

tests/scale-tests/scale-test-report.md
XRFXLP: Platform connector is already something that will be re-run since it is development at this point and it is just faster to inject the events if we know exact shape.

tests/scale-tests/scale-test-report.md
lalitadithya: okay, let's open up a bug for this

tests/scale-tests/scale-test-report.md
lalitadithya: right, let's just add a blurb about this to the doc since this will come up when others review as well

tests/scale-tests/scale-test-report.md
XRFXLP: Yes, lot of these numbers are linear as stated in burst absorption section. Difference between continuous and burst fatality is just the event rate and we already know that things like time to cordon and remediate are linear in respect to the bust size.

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,769 @@
...
+
+| Event rate    | CPU peak `[M]` | Working set `[M]` | Rec. request | Rec. limit |
+| ------------- | -------------- | ----------------- | ------------ | ---------- |
+| idle          | 0.002          | 25.5 MB           | 128 Mi       | 256 Mi     |
+| 36.9 events/s | **0.214**      | 19.6-20.1 MB      | 128 Mi       | 256 Mi     |
+
+
+Wall time per event held between 17.5 and 19.4 ms across measured rates from 1.8 to 55 events/s, so the ceiling is flat rather than degrading with load.
+
+Memory is a fixed cost of about 25 MB at any rate. Throughput is set by wall time per event, not by CPU: the component consumes its stream in a single serial loop (`store-client/pkg/client/event_processor.go`) and spends most of each event blocked on a MongoDB aggregation, so the **17.6 ms** its own `health_event_analyzer_event_handling_duration_seconds` histogram reports per event gives a ceiling near **55 events/s**. CPU is about 5.8 ms of that 17.6 ms, which is why the container draws 0.2 cores at saturation and why more cores do not raise the ceiling. A 100,000-node fleet at 0.1 events per node per second offers 10,000 events/s.
```
kwombach12: are we saying max throughput on the health events analyzer is 55events per second? Do we view that as sufficient at a 100k node scale?

XRFXLP: Yes, that was the case because health events analyzer was a single routine module. This has been fixed as part of https://github.com/NVIDIA/NVSentinel/pull/1817 and docs/configuration has been updated with matrix of how many workers do we need for different cluster sizes.

tests/scale-tests/scale-test-report.md
XRFXLP: done

tests/scale-tests/scale-test-report.md
XRFXLP: fixed.

tests/scale-tests/scale-test-report.md
XRFXLP: removed.

tests/scale-tests/scale-test-report.md
XRFXLP: removed.

tests/scale-tests/scale-test-report.md
XRFXLP: merged to their existing sections.

tests/scale-tests/scale-test-report.md
XRFXLP: handled, ptal

tests/scale-tests/scale-test-report.md
XRFXLP: fixed by moving queue memory into their own section.

tests/scale-tests/scale-test-report.md
XRFXLP: handled, ptal

tests/scale-tests/scale-test-report.md
XRFXLP: done.

tests/scale-tests/scale-test-report.md
XRFXLP: fixed.

tests/scale-tests/scale-test-report.md
XRFXLP: fixed.

tests/scale-tests/scale-test-report.md
XRFXLP: fixed.

XRFXLP: 2s is better sweet spot as 2s buys a 4.8x latency improvement for 1.4x the eviction calls, where 1 s buys 9x for 2.3x.

tests/scale-tests/scale-test-report.md
XRFXLP: should be fixed now.

tests/scale-tests/scale-test-report.md
XRFXLP: added section around circuit breaker.

tests/scale-tests/scale-test-report.md
XRFXLP: At the moment, the restricting part is the serialized bootstrapping in the labeler and not the QPS. We can make it faster just like we did for HEA by parallelizing it.

tests/scale-tests/scale-test-report.md
XRFXLP: added.

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,846 @@
...
+| 25,005  | 0.14 / 0.19        | 4.11 G                 | 4 Gi         | 6 Gi       |
+| 50,005  | 0.20 / 0.24        | 8.55 G                 | 8 Gi         | 12 Gi      |
+| 75,005  | 0.14 / 0.19    | 13.35 G            | 13 Gi        | 20 Gi      |
+| 100,005 | 0.29 / 0.46        | 16.17 G                | 16 Gi        | 24 Gi      |
+| 100,005 | 0.18 / 0.32    | 17.88 G            | 17 Gi        | 26 Gi      |
+
+
+### labeler
+
+Eager informers: one for Nodes with a fixed field projection, and **four pod informers**, each label-scoped -- `app in (dcgm, driver)`, the driver-component label excluding that app, `k8s-app=<gke-installer>`, and ResourceSlice objects.
```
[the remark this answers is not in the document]
XRFXLP: done

tests/scale-tests/scale-test-report.md
XRFXLP: opened: https://github.com/NVIDIA/NVSentinel/issues/1839

tests/scale-tests/scale-test-report.md
lalitadithya: okay, cool, can we add this to the document? I suspect we need some additional dev work to get this exposed right?

tests/scale-tests/scale-test-report.md
lalitadithya: I see, this is good info, can we add this to the doc?

tests/scale-tests/scale-test-report.md
XRFXLP: yes, we need some additional dev work.

tests/scale-tests/scale-test-report.md
```diff
@@ -46,31 +47,33 @@ Markers: `[M]` measured, `[S]` simulated harness constant, `[I]` reader-supplied
...
 
-Two components are most of the component total. At 100,000 nodes with a pod on every node, kubernetes-object-monitor is 33.9 GB and fault-quarantine 17.9 GB; labeler is 6.6 GB, janitor 4.9 GB, node-drainer 1.0 GB, and nothing else exceeds 0.6 GB.
+Two components are most of the component total. At 100,000 nodes with a pod on every node, kubernetes-object-monitor is 19.8 GB (15 GB with minimal pods) and fault-quarantine 17.9 GB; labeler is 6.6 GB, janitor 4.9 GB, node-drainer 1.0 GB, and nothing else exceeds 0.6 GB.
 
-CPU was never a constraint. The busiest component peaked at 2.00 cores, four cores each is sufficient with headroom, and nothing recorded a single throttled CFS period.
+CPU is a constraint for exactly one component. kubernetes-object-monitor needs four cores at 75,005 nodes, where it draws 0.99 to 3.24 depending on how many pods it watches, and eight at 100,005, where it draws 4.09 to 5.48. Sized that way it is throttled on 0.9-6.2% of CFS periods at 75,005 and 2.0-5.2% at 100,005, each throttled period costing 0.3-0.5 ms, which is under a millisecond of stall per second. Every other component peaked well inside its limit with no throttling.
```
lalitadithya: nit: "where it uses", power is drawn, CPUs are used :)

tests/scale-tests/scale-test-report.md
XRFXLP: Added that in the doc.

tests/scale-tests/scale-test-report.md
XRFXLP: Added

tests/scale-tests/scale-test-report.md
XRFXLP: Should be added.

tests/scale-tests/scale-test-report.md
XRFXLP: fixed and simplified.

tests/scale-tests/scale-test-report.md
```diff
@@ -187,6 +187,24 @@ Eager informers, five in total: one for Nodes with a fixed field projection, **t
 | 10,005  | 20,000 (DCGM + driver)  | 0.59 G | 0.59 G | 0.01 / 0.19  | 768 Mi       | 2 Gi       |
 
 
+The bootstrap sweep is serial, one node and one PATCH at a time, and labeler runs raw client-go informers rather than a controller-runtime manager, so there is no concurrency setting to raise.
```
lalitadithya, line 215: can we use a golang bounded parallel workqueue here or would that cause contention between adds and updates in the informer?

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,863 @@
...
+    - [Deployed configuration](#deployed-configuration)
+    - [Traps](#traps)
+
+---
+
+## Summary
+
+NVSentinel has been tested upto 100k nodes, resource consumption grows predictably with fleet size and stays within ordinary limits with reasonable throughputs.
+
+The whole control plane costs about 221 GB of memory at 100,000 nodes. NVSentinel's own components are 51 GB of that and MongoDB is the remaining 170 GB combined across all replicas, which it spends on per-node database connections rather than on data. [ADR-052](../../docs/designs/052-deployment-platform-connector.md) removes those connections, and once it lands the same fleet is projected at about 63 GB.
```
lalitadithya, line 54: Can we add a blurb about what is the cost per GPU node, that is, how much CPU/Memory do the daemonsets consume under input pressure, i.e., very spammy journal, lots of pods on a node to see how the metadata collector reacts and so on

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,863 @@
...
+
+Two components are most of the component total. At 100,000 nodes with a pod on every node, kubernetes-object-monitor is 19.8 GB (15 GB with minimal pods) and fault-quarantine 17.9 GB; labeler is 6.6 GB, janitor 4.9 GB, node-drainer 1.0 GB, and nothing else exceeds 0.6 GB.
+
+CPU constrains one component. kubernetes-object-monitor uses 0.99 to 3.24 cores at 75,005 nodes and 4.09 to 5.48 at 100,005, so it needs four cores and eight respectively. Every other component stayed well inside its limit with no throttling.
```
lalitadithya, line 58: Are we ever going to be network bandwidth or port constrained? I suspect not, but good to confirm and document.

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,863 @@
...
+
+Under continuous load, time to cordon is 26 ms P50 and 165 ms P99, and time to remediate -- drained to the remediation CR being created -- is 0.08 s P50 and 0.17 s P99. Everything NVSentinel does outside the drain totals 91 ms P50. The drain itself is the one term that does not belong to NVSentinel's speed: a node carrying one evictable pod takes about ten seconds, which is one of node-drainer's recheck cycles rather than eviction time, and five hundred nodes failing at once stretches it to 52 s with every node completing.
+
+What broke during testing was the infrastructure around NVSentinel, not NVSentinel. Two failures took MongoDB down with them. The AWS VPC CNI stopped rebuilding `PolicyEndpoint` objects and went on enforcing stale rules, which blocked MongoDB's traffic; the EBS CSI provisioner ran out of memory and stopped attaching volumes cluster-wide, which stranded MongoDB's volumes. A third failure did not touch MongoDB: etcd crossing its 16 GB threshold put the cluster into read-only, so the Kubernetes API refused writes while the running database carried on serving. The first two are described in the appendix; etcd's budget is in A2.
+
+---
```
lalitadithya, line 64: one more thing I think that is worth a callout is correctness, for example under a lot of load would the system ever end up cordoning an incorrect node or would it ever end up rebooting a node that was never drained. I think for a system like this every incorrect operation is an impact to the customer. This shouldn't include false positives from the health monitors themselves. 

Given the design and architecture I know this is near impossible for the system to be incorrect, but I think it is worth a callout and some quick test

tests/scale-tests/scale-test-report.md
```diff
@@ -0,0 +1,863 @@
...
+| cold start | 1,037,329 | 0.55 KB |
+| live stream | 1,113,804 | 1.40 KB |
+
+Each per-event figure is the working-set increase over that run's own baseline divided by the events queued, and the same axes put node-drainer's two paths beside them:
+
+![Memory held by a queued backlog](results/queue-memory.png)
+
+Recommended **2 Gi / 4 Gi**. The working set is a rounding error, but a queued backlog is not: 4 Gi covers roughly three million events on the live path.
+
+### preflight
```
lalitadithya, line 300: For preflight don't we also need throughput? as in how many pods can it process per second under load?

tests/scale-tests/scale-test-report.md
XRFXLP: We can do that, tracking issue for this: https://github.com/NVIDIA/NVSentinel/issues/1846

tests/scale-tests/scale-test-report.md
XRFXLP: created: https://github.com/NVIDIA/NVSentinel/issues/1847

tests/scale-tests/scale-test-report.md
XRFXLP: added for additional webhook latency for non-gang pods and gang admission throughput, PTAL.

tests/scale-tests/scale-test-report.md
XRFXLP: Added, PTAL.

tests/scale-tests/scale-test-report.md
XRFXLP: Checked on the harness cluster, it is nowhere near limit despite all the kwok lease traffic.

tests/scale-tests/scale-test-report.md
XRFXLP: Added, PTAL.

tests/scale-tests/scale-test-report.md
```diff
@@ -536,16 +619,28 @@ The drain figures come from a workload simulator (`--mode=workload` in `k8s-obje
...
+
+Outside that envelope there are three scenarios in which a module acts against its own stated intent, and none of them is closed by configuration.
+
+1. **A consumer that falls off the oplog skips events permanently.** `ChangeStreamHistoryLost` is classified with corrupt-token errors, so the resume token is deleted and the stream reopens at the current cluster time. The gap is never replayed and nothing re-emits it, because monitors commit their "already reported" state on platform-connector's gRPC ack rather than on the MongoDB write. A skipped fatal event is a node never quarantined; a skipped healthy event is a node cordoned until someone runs `kubectl uncordon`. node-drainer and fault-remediation recover part of a gap through their cold-start queries; fault-quarantine, health-events-analyzer and event-exporter have none. There is no knob: the chart sets no `oplogSizeMB`, so retention is MongoDB's implicit default against the 8 Gi PVC, and `change_stream_resume_token_recoveries_total` carries only `(client, phase)`, so data loss cannot be distinguished from a corrupt token.
```
lalitadithya: can we reword this a bit please? I'm a bit confused on where the error is. Is this because the platform connector didn't write to MongoDB or because some module didn't cold start correctly? If it is the former, won't that be resolved by the new mode of platform connector?

lalitadithya:
not to add too much scope, but one more thing that came up was what would be the impact of an upgrade? Say we are in the middle of a burst and NVSentinel rolls out would that result in a spike? Would it be possible to test that after we are done with all our open items?

tests/scale-tests/scale-test-report.md
XRFXLP: clarified. This was about a slower module and has happened with event exporter in the past where due to export metric latency it was already behind the event rate produced in the cluster.

XRFXLP:
> not to add too much scope, but one more thing that came up was what would be the impact of an upgrade? Say we are in the middle of a burst and NVSentinel rolls out would that result in a spike? Would it be possible to test that after we are done with all our open items?

Added a section about rollout upgrade in appendix for 2000 node burst. As expected, given that we've checkpoints over changestream, none of the module missed their intent. Circuit breaker failed to trip as expected given that we track it completely in-memory. 

One minor caveat  is we cannot rely on the prometheus metrics during these situations as it polls every 60 second and FQ can cordon a lot of nodes in those 60 seconds.

image

lalitadithya:
one thing that came up during review was how do we know the tests we have done with KWOK compare to an actual production cluster and what data points gives us the confidence that what we have tested will hold in production, I think it would be good to add a blurb/para for this somewhere close to the top

Issue #1471, opened 2026-07-09
[Bug]: DCGM_FR_IMEX_UNHEALTHY reported as GpuNvlinkWatch/CONTACT_SUPPORT on single-node NVSwitch systems where IMEX is not applicable


njhensley:
### Prerequisites
- [x] I searched existing issues
- [x] I can reproduce this issue

### Bug Description

On single-node NVSwitch GPU systems (AWS `p5.48xlarge` — 8× H100 SXM + on-board NVSwitch), NVSentinel persistently reports the node condition `GpuNvlinkWatch=True` with:

```
ErrorCode:DCGM_FR_IMEX_UNHEALTHY GPU:0 IMEX daemon status is -3 (not READY)
Check IMEX installation, configuration, domain and daemon status, and network
connectivity. Recommended Action=CONTACT_SUPPORT;
```

This is a **false positive**. IMEX (Internode Memory Exchange) brokers NVLink memory across nodes in a **multi-node** NVLink domain (GB200 NVL72 / NVLink-network). On a standalone single-node NVSwitch box there is no inter-node NVLink domain, so the IMEX daemon is intentionally **not configured** and does not run. DCGM's IMEX field diagnostic returns "not ready" (`-3`), and the mapping added in v1.9.0 (PR #1369) surfaces it as an unhealthy, `CONTACT_SUPPORT` condition on healthy hardware.

Two mislabelings compound it:
1. **Wrong subsystem in the name** — surfaced under `GpuNvlinkWatch` (implying NVLink *link* health), but the check is IMEX daemon readiness, a different layer. This also collides with the genuine NVLink signals now mapped under the same condition (`DCGM_FR_NVLINK_EFFECTIVE_BER_THRESHOLD`, `DCGM_FR_NVLINK_SYMBOL_BER_THRESHOLD`).
2. **Wrong severity** — "IMEX not configured" is *expected state* on this topology, not a fault, yet it is `CONTACT_SUPPORT`.

This is the same class of severity/topology mismap as #465 (GpuPowerWatch → CONTACT_SUPPORT should be NONE) and #1363 (DCGM resource-in-use misclassified as fatal), and the same cloud-shape-unaware false-positive pattern as the NIC monitor in #1361 / #1379.

**Distinct from (please don't mis-triage):**
[...]

PR #1793, opened 2026-09-10
fix: incorrect supression of DCGM errors

### LINKED ISSUE #1791: [Bug]: suppressedErrorCodes drops every incident for a GPU+watch, so the #1471 IMEX workaround masks real NVLink faults on GPU 0
### Prerequisites

### Bug Description

`suppressedErrorCodes` does not suppress an *incident*. It deletes the whole **(watch, GPU)** entry, so suppressing one error code also discards any **other, unsuppressed** incident DCGM reported for that same GPU and watch.

That matters because `suppressedErrorCodes: [DCGM_FR_IMEX_UNHEALTHY]` is the workaround this project **recommends** for #1471. Anyone who took that advice has silently lost NVLink fault detection on whichever GPU reports IMEX — GPU 0 on every node, in our case. We hit it during a real incident.

Two mechanisms combine:

**1. The accumulator keeps only the first error code per (watch, GPU).**

`health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py:643`

```python
if accumulator_key not in gpu_failures_accumulator:
    gpu_failures_accumulator[accumulator_key] = {"code": error_code, "messages": []}

gpu_failures_accumulator[accumulator_key]["messages"].append(error_msg)   # :647
```

Later incidents on the same GPU and watch contribute their **message** but their **code is dropped**. Consolidation at `:653` then stores exactly one `ErrorDetails`, carrying that first code.

**2. Suppression then deletes by GPU, not by incident.**

`.../dcgm_watcher/dcgm.py:347-368`, called at `:1105` — i.e. *after* consolidation:

```python
suppressed_gpu_ids = [
    gpu_id for gpu_id, failure in details.entity_failures.items()
    if failure.code in self._suppressed_error_codes
]
for gpu_id in suppressed_gpu_ids:
    ...
    del details.entity_failures[gpu_id]        # :364  <- the whole GPU, not the incident
```

So if GPU 0's stored code is the suppressed one, GPU 0 vanishes from that watch entirely — including a genuine fault whose message was merged in at `:647`.

Worth noting the comment at `:621` states the opposite contract:

> Per-incident suppression: a suppressed incident must neither degrade the watch status nor land in the accumulator, **while other incidents on the same GPU and watch are kept.**

That holds for `_is_nvlink_down_false_positive`, which `continue`s per incident inside the loop. It does **not** hold for the config-driven suppression, which runs later on the consolidated dict.

### Reproduction, from a live incident

A GB200 NVL72 rack (18 nodes, 4 GPUs each) lost power. On return, NVLink was genuinely degraded — confirmed independently of NVSentinel:

- `nvidia-smi nvlink -s`: **links 2 and 3 `<inactive>` on all 4 GPUs**, the other 16 at full 53.125 GB/s
- `DCGM_FI_DEV_NVLINK_BANDWIDTH_TOTAL`: **0 on all 4 GPUs**, against **~342,600** on an unaffected rack
- `DCGM_EXP_GPU_HEALTH_STATUS{health_watch="NVLINK"}`: severity **20 on all 4 GPUs**, code `DCGM_FR_FABRIC_PROBE_STATE`. On the unaffected rack only GPU 0 is non-zero.

NVSentinel reported `GpuNvlinkWatch` on **3 of the 4 GPUs**. GPU 0 was absent on all 18 nodes, for 9+ hours, while DCGM was flagging it the whole time.

Cause: GPU 0's stored code is `DCGM_FR_IMEX_UNHEALTHY` (IMEX is deliberately not run on this cluster, exactly the #1471 situation), so mechanism 2 deleted GPU 0 wholesale and took the fabric fault with it.

Measured confirmation that it is precisely one GPU per poll:

```
PollIntervalSeconds = 15
dcgm_health_check_suppressed_incidents_total{error_code="DCGM_FR_IMEX_UNHEALTHY"}
  t0 = 2155
  t1 = 2159   (+4 over 64s -> one suppression every 16.0s)
```

One suppression per poll cycle, i.e. exactly one GPU, on every node in the rack.

### Why this is not #1471, #656 or #1701

- **#1471** (closed) is that `DCGM_FR_IMEX_UNHEALTHY` is a false positive on topologies without IMEX, and recommends `suppressedErrorCodes` to silence it. This issue is that **the recommended workaround has a blast radius nobody stated**: it blinds the whole watch for that GPU. #1471 does note IMEX "collides with the genuine NVLink signals now mapped under the same condition", but only as a naming concern — not that suppression discards them.
- **#656** is a stale condition from `;` message parsing. Here nothing is stale: the gauge is recomputed every poll and DCGM is still reporting.
- **#1701** (closed, fixed by #1715) is transient NVLink-down after SXM boot. This is not transient — 9+ hours, zero NVLink bandwidth, and `dcgm_health_check_debounced_incidents_total` is **0**, so the debounce is not involved. Separately: #1715's default threshold is 1, which `dcgm.py:251-253` filters out (`if polls > 1`), so the debounce is inert unless an operator raises it. That may be worth its own look.

### Suggested fix

Make config-driven suppression per-incident, matching the documented contract and the existing `_is_nvlink_down_false_positive` behaviour: test `error_code in self._suppressed_error_codes` inside the incident loop before `:643` and `continue`, so a suppressed incident never enters the accumulator and other incidents for that GPU survive.

That also needs the accumulator to stop collapsing codes, or a suppressed-first / real-second ordering still loses the real one. Options: key the accumulator on `(watch, gpu, code)`, or keep a list of codes per `(watch, gpu)`. The second is a wider change to `ErrorDetails`, so the per-incident `continue` plus `(watch, gpu, code)` keying is probably the smaller correct step.

A regression test worth having: one GPU, one watch, two incidents where the **suppressed** code is reported first — assert the unsuppressed incident still surfaces.

### Impact

Any deployment using `suppressedErrorCodes` has an undetectable blind spot on the affected GPU and watch, and the project recommends that setting in #1471. In our case a real, rack-wide NVLink fault went unreported on 18 of 72 GPUs.

lalitadithya:
## Summary

The `suppressedErrorCodes` option removed too much data. It deleted the full entry for a GPU and a watch. So, it also removed the other incidents for that GPU. Two parts of the code caused this fault:
1. The accumulator keeps only the first error code for each (watch, GPU) pair. A later incident adds its message. The code of that incident is lost.
2. Suppression ran after this step. It examined only the first error code. If that code was in the list, the code deleted the full entry.

As a result, an operator who suppresses `DCGM_FR_IMEX_UNHEALTHY` loses all NVLink fault detection on each GPU that reports IMEX. This is the workaround that #1471 recommends.

This PR moves suppression into the incident loop. Each incident is now examined on its own. A suppressed incident does not hide the other incidents for the same GPU and the same watch.

Fixes #1791

## Type of Change
- [X] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [X] Health Monitors
- [ ] Fault Management
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Janitor
- [ ] Documentation/CI
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## For release team

dcgm_health_check_suppressed_incidents now counts per incident, not per (watch, GPU) — release-note material, since anyone calibrated on the old cadence sees a step.

## Summary by CodeRabbit

* **New Features**
  * Configured GPU health incidents can now be suppressed individually, preventing them from affecting health status or failure tracking.
  * Suppressed incidents are isolated per GPU and reported through suppression metrics.

* **Bug Fixes**
  * Suppressed incidents no longer advance health-check debounce streaks or obscure unsuppressed incidents.


deesharma24 (approved):

lfriedman-netllama:
Verified — resolves #1791. Detail on the issue.

With these test files on `main`'s unmodified `dcgm.py`, `test_suppressed_code_keeps_other_incidents_on_the_same_gpu` and `test_suppressed_code_does_not_build_a_debounce_streak` both fail; on this branch 19 pass and the full suite is 179 passed. The reproduction ordering is the right one (IMEX first, `DCGM_FR_FABRIC_PROBE_STATE` second, survivor keeps its own code).

One thing worth a comment in the code: `_suppress_configured_error_codes` must stay, because the thermal-margin (`:751`) and power-brake (`:861`) monitors write `entity_failures` outside the incident loop. Otherwise someone deletes it as dead code later.

KaivalyaMDabhadkar (approved):

PR #1804, opened 2026-09-10
fix(gpu-health-monitor): preserve all DCGM incident codes

### LINKED ISSUE #1795: [Bug]: the incident accumulator keeps only the first error code per (watch, GPU), so a second code on the same GPU is silently dropped
### Prerequisites

### Bug Description

Split out of #1791 at @lalitadithya's request. #1793 fixes the suppression path; this is the underlying data loss that remains once suppression is out of the picture.

When DCGM reports **two or more incidents for the same (watch, GPU)**, only the **first** error code survives. Later incidents contribute their message but their code is discarded, so the reported code depends on DCGM's iteration order.

`health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py:643`

```python
if accumulator_key not in gpu_failures_accumulator:
    gpu_failures_accumulator[accumulator_key] = {"code": error_code, "messages": []}

gpu_failures_accumulator[accumulator_key]["messages"].append(error_msg)   # :647
```

Consolidation at `:653` then writes a single `ErrorDetails`, and `types.ErrorDetails` holds exactly one `code` plus one `message` — so the structure cannot represent two codes even if the loop wanted to.

### Why it matters

**The code is what drives remediation.** The messages are merged, so an operator reading the event text may see both problems, but the `recommendedAction` mapping keys off the single surviving `code`. A GPU reporting, say, `DCGM_FR_FABRIC_PROBE_STATE` and `DCGM_FR_NVLINK_DOWN` under the NVLINK watch is actioned as whichever DCGM happened to return first.

It also makes the reported code non-deterministic across polls if DCGM's ordering is not stable, which would show up as an event flapping between codes on a GPU whose hardware state has not changed.

We have not been bitten by this in production — our GPU 0 case was the *suppression* interaction, which #1793 fixes. This is the latent half.

### Suggested fix

Two options, in increasing order of correctness and cost:

1. **Key the accumulator on `(watch, gpu, code)`.** Each distinct code becomes its own incident. Cheap, but `entity_failures` is `dict[gpu_id, ErrorDetails]`, so it cannot hold two entries for one GPU — this needs the container to change too.
2. **Make `entity_failures` hold a list per GPU**, i.e. `dict[int, list[ErrorDetails]]`. Correct, and it is the shape the data actually has, but it touches every consumer of `HealthDetails`.

A cheaper interim that removes the non-determinism without restructuring: pick the surviving code deterministically rather than by arrival — for example the highest-severity code, or the first in a documented precedence order — and say so in a comment. That does not fix the loss, but it makes behaviour predictable and reviewable.

### Test worth having

One GPU, one watch, two unsuppressed incidents with different codes, asserted in **both** arrival orders — the reported code should be the same either way. That test fails today for one of the two orders whichever precedence you pick, which is the point.

sylvesterkaczmarek:
## Summary

Fixes #1795.

Preserve every distinct error code for a GPU within a health watch. Group messages by watch, GPU and code; publish each code with its corresponding remediation; retain all pending codes in the event cache. No protobuf or configuration changes.

## Validation

- GPU health monitor make lint-test: 197 tests and 5 subtests passed.
- Seven regression cases fail on the original code. Coverage includes both arrival orders, per-code actions, repeated polls, suppression and failed-publish retries.
- Validation: https://github.com/sylvesterkaczmarek/nccl-tests/actions/runs/34522297146
- Full-repository make lint-test-all stopped because protoc was unavailable. Live GPU execution was not run.

Signed-off-by: Sylvester Kaczmarek <16242628+sylvesterkaczmarek@users.noreply.github.com>

## Summary by CodeRabbit

* **New Features**
  * GPU health incidents now report and track multiple error codes independently for each GPU.
  * Repeated messages for the same error are combined, while distinct errors retain separate details and remediation actions.
  * Per-error suppression and debounce behavior allows unaffected incidents to continue reporting.
  * Incident reporting behavior is documented.

* **Bug Fixes**
  * Improved incident delivery, retry, recovery, and cache handling across multiple simultaneous GPU failures.

### DIFF
--- a/health-monitors/gpu-health-monitor/README.md
+++ b/health-monitors/gpu-health-monitor/README.md
@@ -1,3 +1,10 @@
 # GPU Health Monitor
 
 Health monitor for monitoring the health of GPUs
+
+
+## Incident reporting
+
+Each GPU and watch reports every distinct error code. Repeated incidents for the same code share one event with their combined messages. Each event retains its code-specific remediation action.
+
+Suppression and debounce apply separately to each code. A healthy event clears the watch only when that GPU has no remaining reported incidents. Cache updates occur after successful delivery.
--- a/health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
+++ b/health-monitors/gpu-health-monitor/gpu_health_monitor/dcgm_watcher/dcgm.py
@@ -360,22 +360,20 @@ class DCGMWatcher:
             return
 
         for watch_name, details in health_status.items():
-            suppressed_gpu_ids = [
-                gpu_id
-                for gpu_id, failure in details.entity_failures.items()
-                if failure.code in self._suppressed_error_codes
-            ]
-            for gpu_id in suppressed_gpu_ids:
-                error_code = details.entity_failures[gpu_id].code
-                log.debug(
-                    f"Suppressing incident for watch={watch_name} entity={gpu_id} "
-                    f"error_code={error_code}: high-frequency non-actionable event"
-                )
-                metrics.dcgm_health_check_suppressed_incidents.labels(error_code).inc()
-                del details.entity_failures[gpu_id]
+            had_failures = bool(details.entity_failures)
+            for gpu_id, failures in list(details.entity_failures.items()):
+                remaining = [
+                    failure
+                    for failure in failures
+                    if not self._is_suppressed_error_code(watch_name, gpu_id, failure.code)
+                ]
+                if remaining:
+                    details.entity_failures[gpu_id] = remaining
+                else:
+                    del details.entity_failures[gpu_id]
 
             # A watch with no remaining failures is healthy again.
-            if suppressed_gpu_ids and not details.entity_failures:
+            if had_failures and not details.entity_failures:
                 details.status = types.HealthStatus.PASS
 
     def _is_nvlink_down_false_positive(self, watch_name: str, gpu_id: int, error_code: str) -> bool:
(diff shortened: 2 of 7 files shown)


deesharma24:
@sylvesterkaczmarek Can you please take a look at coderabbit comments and CI pipeline failure?

sylvesterkaczmarek:
@deesharma24 Addressed both CodeRabbit items in `308d4f3f`: the synthetic DCGM incident records now use dataclasses, and all `processor` fixture parameters are typed. Black, `py_compile`, and `git diff --check` pass. The repository-wide `make lint-test-all` cannot reach tests on this macOS host because protobuf generator tooling is missing during setup.\n\nThe E2E failure is unrelated to this change: a `main` E2E run from the same window (`34569895388`) shows the same NVIDIA repository certificate verification failure, missing `preflight-webhook-tls`, and readiness timeouts. Please `/ok to test` the new head when convenient.

deesharma24:
LGTM. @lalitadithya Can you please trigger the CI again.I don' t have permissions.Already tried my luck this morning above.

deesharma24 (approved):

sylvesterkaczmarek:
Thanks for the review and approval.

sylvesterkaczmarek:
The current E2E failures on `deda3f2` are outside the GPU-health-monitor change. Three failing jobs stop in `Setup build environment`; the AMD64 Percona job reaches cluster setup but fails on infrastructure/dependency setup (including a KWOK chart 504 and an oversized Percona CRD annotation) before exercising this regression path. The GPU-health-monitor lint/container checks are green, so I do not see a code change needed for these E2E failures.

lalitadithya (approved):

sylvesterkaczmarek:
Thanks for merging this.

PR #1806, opened 2026-09-10
fix(janitor-provider/oci): add retry to transient failures

### LINKED ISSUE #1805: [Bug]: OCI janitor provider doesn't retry
### Prerequisites

### Code of Conduct

### Bug Description

Hi team, when using OCI as janitor provider, more often than not, I found RebootNode CR fails due to transient failures, including request timeout and conflict.

- Checking through the [code](https://github.com/NVIDIA/NVSentinel/blob/main/janitor-provider/pkg/csp/oci/oci.go), I found we don't have retry implemented
- When CSP client fails to request reboot, janitor provides returns [INTERNAL error](https://github.com/NVIDIA/NVSentinel/blob/20a4ed1698b401c2e8cf1263c576529f328933fa/janitor-provider/main.go#L89-L98), which is not retriable from [janitor's perspective](https://github.com/NVIDIA/NVSentinel/blob/20a4ed1698b401c2e8cf1263c576529f328933fa/janitor/pkg/controller/rebootnode_controller.go#L332-L351)

### Component

Other

### Steps to Reproduce

N/A

### Environment

- NVSentinel version: v1.22.0
- Kubernetes version: v1.34.2 (Oracle Kubernetes Engine, Phoenix production)
- Deployment method: Argo CD ApplicationSet + NVSentinel OCI Helm chart (ghcr.io/nvidia/nvsentinel), managed through GitOps in ml-infra

### Logs/Output

_No response_

dentiny:
## Summary

Closes https://github.com/NVIDIA/NVSentinel/issues/1805

This PR adds retry mechanism to OCI request, so it's tolerant of transient failures. I've met several CR failures due to timeout or conflict operations, so it's more like a bug fix rather than a feature request from my perspective.

The reason why I added retry inside of oci, which is a detailed janitor provider, instead of somewhere more high-level, is because each provider has its own (slightly) different error code and message.

## Type of Change
- [X] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Core Services
- [ ] Health Monitors
- [ ] Fault Management
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [X] Janitor
- [ ] Documentation/CI
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## Summary by CodeRabbit

* **Bug Fixes**
  * OCI reboot requests now use stable idempotency tokens for safer request handling.
  * Transient OCI failures are reported as temporarily unavailable, while other failures retain instance-specific context.
  * Reboot, readiness, and termination errors now preserve recognized service status codes instead of being reported uniformly as internal errors.

* **Tests**
  * Added coverage for stable request handling, transient and permanent failures, and service status translation.

### DIFF
--- a/janitor-provider/pkg/csp/oci/oci.go
+++ b/janitor-provider/pkg/csp/oci/oci.go
@@ -16,6 +16,7 @@ package oci
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -27,6 +28,8 @@ import (
 	"github.com/oracle/oci-go-sdk/v65/common"
 	"github.com/oracle/oci-go-sdk/v65/common/auth"
 	"github.com/oracle/oci-go-sdk/v65/core"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/nvidia/nvsentinel/commons/pkg/auditlogger"
@@ -36,8 +39,6 @@ import (
 var (
 	_ model.CSPClient = (*Client)(nil)
 )
-
-const rebootRetryAttempts = 6
 
 // Compute provides a wrapper around a subset of the OCI Compute client interface,
 // to enable mocking/stubbing for testing.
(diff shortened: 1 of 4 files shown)


janitor-provider/pkg/csp/oci/oci.go
```diff
@@ -149,6 +165,31 @@ func (c *Client) IsNodeReady(ctx context.Context, node corev1.Node, requestID st
...
+	if common.IsErrorRetryableByDefault(err) {
+		return true
+	}
+
+	var serviceErr common.ServiceError
+	if !errors.As(err, &serviceErr) {
+		return false
+	}
+
+	return serviceErr.GetHTTPStatusCode() == http.StatusConflict &&
```
dentiny, line 193: Conflict status error is not included in the OCI retriable errors by default, but I met it for quite a few times in production.

lalitadithya:
hey @dentiny , what are you thoughts on hooking into the existing retry mechanism we have in https://github.com/NVIDIA/NVSentinel/blob/20a4ed1698b401c2e8cf1263c576529f328933fa/janitor/pkg/controller/rebootnode_controller.go#L535? So, the provider can return a error that is marked as retriable and we can let the controller runtime drive the backoff and retries using it's queue. The advantage of doing it this way means that all providers can benefit from this change and also at the same time we don't increase the duration of the grpc calls. What do you think?

dentiny:
Thanks @lalitadithya for the quick reply and suggestion! I was actually thinking about the same thing -- reworking the PR now. :)

janitor-provider/pkg/csp/oci/oci.go
```diff
@@ -149,6 +159,42 @@ func (c *Client) IsNodeReady(ctx context.Context, node corev1.Node, requestID st
...
+
+	token := fmt.Sprintf("%x", sha256.Sum256([]byte(providerID+"\x00"+crName)))
+
+	return &token
+}
+
+func translateRebootError(providerID string, err error) error {
+	contextualErr := fmt.Errorf("send soft reset action for OCI instance %q: %w", providerID, err)
+	if isRetryableRebootError(err) {
+		return status.Error(codes.Unavailable, contextualErr.Error())
```
dentiny, line 177: This is a coarse-granular error translation, it's a tradeoff compared with fine-granular error catagorization.

lalitadithya (approved):

PR #1812, opened 2026-09-11
feat: add opt-in rule_matched_entity_total

### LINKED ISSUE #1746: [Feature]: rule_matched_total omits the entity the rule selected on, so RUN_DCGMEUD findings cannot be acted on without querying MongoDB
`rule_matched_total` is labelled `{rule_name, node_name}` (`health-events-analyzer/pkg/reconciler/metrics.go:52-58`), but several rules select on an **entity within the node**, and that entity is not exported. The result is a metric that tells you a rule fired and gives you no way to act on it.

`RepeatedXID13OnSameGPCAndTPC` is the clearest case. Its whole purpose is locality:

```toml
name = "RepeatedXID13OnSameGPCAndTPC"
description = "Detect if XID 13 occurred 2 or more times on the same GPC and TPC within 24 hours"
message = "if DCGM EUD tests pass, run field diagnostics"
recommended_action = "RUN_DCGMEUD"
```

The rule matches on `this.healthevent.entitiesimpacted`, comparing `entitytype`/`entityvalue` pairs for GPC and TPC. So the analyzer knows exactly which GPU, GPC and TPC it selected on, then discards all of it when it increments the counter.

An operator sees this and nothing more:

```
rule_matched_total{rule_name="RepeatedXID13OnSameGPCAndTPC", node_name="...r05-13..."}  302
rule_matched_total{rule_name="RepeatedXID13OnSameGPCAndTPC", node_name="...r02-05..."}  117
```

The recommended action is "run field diagnostics", and the metric cannot say against which unit. Answering that requires querying MongoDB directly, which defeats the purpose of having the metric.

The same applies to any rule whose aggregation keys on an entity: `RepeatedXIDErrorOnSameGPU` and `RepeatedXID31OnSameGPU` identify a GPU and export only the node.

### Precedent, and interaction with rule thresholds

This is the same defect class as #1685, closed by #1714: partial drain was visible as "a drain happened" with no way to see the scope or which GPU it targeted, so the metric could not drive a decision. Same shape here.

It also interacts badly with rule thresholds. `RepeatedXID13OnSameGPCAndTPC` fires at **2 occurrences in 24h**. Under a large XID 13 burst, many GPC/TPC pairs will cross that threshold by volume alone, and without entity labels there is no way to tell "one unit repeatedly" from "every unit twice" — which is exactly the distinction between a localized hardware defect and an application fault scattered across the die. Both currently look identical in Prometheus.

### Proposed change

Add the matched entity to the signal, treating cardinality as a first-class concern.

**Preferred:** a separate metric, so the existing one keeps its current cardinality, gated by a flag defaulting to off:

```
rule_matched_entity_total{rule_name, node_name, entity_type, entity_value}
```

- `entity_type` / `entity_value` come straight from the entity the rule keyed on, so no new collection is needed.
- Emitted only for rules whose aggregation selects on an entity; others are unaffected.
- Default off, following the precedent set in #1685/#1714, where the maintainer asked for the GPU-UUID dimension to sit behind a flag defaulting to off because of cardinality.

**Cardinality, stated plainly.** GPU index, GPC and TPC are small bounded integers, so per node the fan-out is GPUs × GPCs × TPCs. Bounded, but not small on a large fleet, which is why this is opt-in and a separate series rather than extra labels on `rule_matched_total`.

**What not to add:** the GPU UUID, for the same reason `errorCode` was excluded from `health_events_total` in #1687: it is effectively unbounded from Prometheus's point of view, and a replaced GPU changes it. The stable slot identity (GPU index or PCI address) is the useful key, which is the same reasoning `recoveryEntities` already applies in the Kubernetes connector when it prefers a stable slot id over the UUID.

### Alternative considered

Exposing a gauge of currently-implicated entities, rather than a counter of matches, would be lower cardinality and arguably more useful for "what should I run diagnostics on right now". It needs a definition of when an entity stops being implicated, which is the same recovery-semantics problem as #1553, so it is probably better as a follow-up than a prerequisite.

### Context

Found while auditing a fleet dashboard where `RepeatedXID13OnSameGPCAndTPC` had fired several hundred times across two nodes, with no way to tell from metrics whether a specific SM was implicated. Everything needed is already on the health event; only the export is missing.

rupalis-nv:
## Summary

Fixes #1746.

`rule_matched_total` is labeled `{rule_name, node_name}` only, so entity-keyed rules (especially `RepeatedXID13OnSameGPCAndTPC`) fire without saying which GPU, GPC, TPC, or NIC matched.

This adds an opt-in counter. `rule_matched_total` is unchanged. The new series is registered only when the flag is on (lazy `CounterVec`, nil until enabled). GPU UUID is never exported; PCI / GPU index / GPC / TPC / NVLINK / NIC / NICPort are. SM and register entities are dropped. Node-scoped rules such as `MultipleRemediations` do not emit the series even if the triggering event carries GPU entities.

```yaml
health-events-analyzer:
  ruleMatchedEntityMetricEnabled: false
```

When enabled, a match on an entity-keyed rule increments:

`rule_matched_entity_total{rule_name, node_name, entity_type, entity_value}`

Labels come from the triggering event's EntitiesImpacted. No extra Mongo query.

## Testing

Ran this on a dev cluster. Source test events were STORE_AND_ANALYSE so they reached the analyzer without driving quarantine/drain. Analyzer output used STORE_ONLY so derived events were observability-only.

1. Flag off: a matching entity-keyed rule incremented `rule_matched_total. rule_matched_entity_total` was not registered and no entity series appeared.
2. Flag on: analyzer logged that `rule_matched_entity_total` was registered.
3. `RepeatedXID13OnSameGPCAndTPC` match with PCI, GPU UUID, GPC, TPC, and SM on the event: rule_matched_total incremented; PCI, GPC, and TPC were exported; GPU UUID and SM were not.
4. GPU-keyed rule match with PCI, GPU UUID, and GPU index: PCI and GPU index exported; UUID was not.
5. NIC-keyed rule match: NIC and NICPort exported; UUID was not.
6. Node-scoped `MultipleRemediations` match on an event that still carried PCI/GPC/TPC/UUID: `rule_matched_total` incremented; no entity series for that rule.
7. Entity-keyed match whose only entities were UUID / SM / REGFILE: `rule_matched_total` incremented; no entity series.
8. Healthy events, `STORE_ONLY` source events, and `health-events-analyzer` agent events were ignored by the watch.

## Type of Change
- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Health Monitors
- [ ] Fault Management
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Janitor
- [ ] Documentation/CI
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added an optional rule-matched entity metric with rule, node, entity type, and entity value labels. It is disabled by default and applies to entity-keyed rules, including `RepeatedXidError`.
  - Entity metrics support recognized entity types, PCI canonicalization, bounded network and switch values, and GPU index values while excluding empty, malformed, duplicate, and sensitive GPU UUID values.

- **Documentation**
  - Documented supported entity types, label behavior, producer spelling, GPU UUID redaction, node-scoped rule exclusions, and `RepeatedXidError` handling.


health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,117 @@
...
+
+		out = append(out, entity)
+	}
+
+	return out
+}
+
+func recordMatchedEntityMetric(ruleName, nodeName string, event *protos.HealthEvent) {
+	for _, entity := range metricSafeEntities(event) {
+		recordRuleMatchedEntity(ruleName, nodeName, entity.GetEntityType(), entity.GetEntityValue())
```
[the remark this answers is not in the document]
rupalis-nv: addressed

health-events-analyzer/pkg/config/rules_test.go
```diff
@@ -0,0 +1,83 @@
...
+// limitations under the License.
+
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
tanishagoyal2: Shall we remove this rules_test.go file? Its only checking the behaviour when flag is enabled

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	return false
+}
+
+func canonicalMetricEntityType(entityType string) (string, bool) {
+	canonical, ok := metricSafeEntityTypes[strings.ToLower(entityType)]
+
+	return canonical, ok
+}
```
tanishagoyal2: why do we need this mapping? Can we directly use entity_type value?

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,159 @@
...
+		recordRuleMatchedEntity(ruleName, nodeName, entity.GetEntityType(), entity.GetEntityValue())
+	}
+}
+
+// eventHasGPUUUID reports whether the burst detector keyed on a GPU UUID.
+// Events without one fall back to the unknown-GPU bucket and are not entity-keyed.
+func eventHasGPUUUID(event *protos.HealthEvent) bool {
+	if event == nil {
+		return false
+	}
```
tanishagoyal2: event can't be nil, right?

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+	}
+
+	return false
+}
+
+func recordMatchedEntityMetricForXidBurst(nodeName string, event *protos.HealthEvent) {
+	if !eventHasGPUUUID(event) {
+		return
+	}
```
tanishagoyal2: why do we need to check for GPUUUID presence. If I understand correctly we want to capture all entitiesImpacted of an event for the matched rule, right? If GPU_UUID is absent but some other entities are present then we can capture that. If no entities present then recordRuleMatchedEntity won't be emitted.

health-events-analyzer/pkg/reconciler/metrics.go
```diff
@@ -75,3 +89,30 @@ var (
...
 	)
 )
+
+// EnableRuleMatchedEntityMetric registers rule_matched_entity_total. Call once at
+// startup, before the reconciler runs, when the operator has opted in. Repeat
+// calls are ignored so registering twice cannot panic.
+func EnableRuleMatchedEntityMetric() {
+	if ruleMatchedEntityTotal != nil {
+		return
+	}
```
tanishagoyal2: when can this be empty? Do we need this nil check?

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,159 @@
...
+	return canonical, ok
+}
+
+// metricSafeEntities returns the triggering event's impacted entities that are
+// safe Prometheus labels: stable slot identity, not GPU UUID, SM, or register
+// values. Types are rewritten to the documented canonical spelling so mixed
+// case cannot split a series. Duplicates are dropped. The returned entities
+// are copies and do not mutate the triggering event.
+func metricSafeEntities(event *protos.HealthEvent) []*protos.Entity {
+	if event == nil {
```
XRFXLP: Can a nil event ever reach this function call?

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil
+	}
+
+	seen := make(map[string]struct{}, len(event.GetEntitiesImpacted()))
+	out := make([]*protos.Entity, 0, len(event.GetEntitiesImpacted()))
+
+	for _, entity := range event.GetEntitiesImpacted() {
+		if entity == nil {
```
XRFXLP: is this needed?

health-events-analyzer/pkg/reconciler/metrics.go
```diff
@@ -19,6 +19,14 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
+const (
+	labelRuleName               = "rule_name"
+	labelNodeName               = "node_name"
+	labelEntityType             = "entity_type"
+	labelEntityValue            = "entity_value"
+	metricNameRuleMatchedEntity = "rule_matched_entity_total"
```
tanishagoyal2: Since these constants are used in single metric ruleMatchedEntityTotal. Shall we remove constant from here and directly use string?

health-events-analyzer/pkg/reconciler/metrics.go
```diff
@@ -75,3 +89,30 @@ var (
...
+		[]string{labelRuleName, labelNodeName, labelEntityType, labelEntityValue},
+	)
+}
+
+// recordRuleMatchedEntity records a match against the entity the rule selected
+// on. It is a no-op unless EnableRuleMatchedEntityMetric has been called.
+func recordRuleMatchedEntity(ruleName, nodeName, entityType, entityValue string) {
+	if ruleMatchedEntityTotal == nil {
+		return
+	}
```
tanishagoyal2: when can this be nil?

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,159 @@
...
+
+// eventHasGPUUUID reports whether the burst detector keyed on a GPU UUID.
+// Events without one fall back to the unknown-GPU bucket and are not entity-keyed.
+func eventHasGPUUUID(event *protos.HealthEvent) bool {
+	if event == nil {
+		return false
+	}
+
+	for _, entity := range event.GetEntitiesImpacted() {
+		if entity == nil {
```
XRFXLP: same comment as earlier.

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,159 @@
...
+func recordMatchedEntityMetricForXidBurst(nodeName string, event *protos.HealthEvent) {
+	if !eventHasGPUUUID(event) {
+		return
+	}
+
+	recordMatchedEntityMetric("RepeatedXidError", nodeName, event)
+}
+
+func recordMatchedEntityMetricForRule(rule config.HealthEventsAnalyzerRule, event *protos.HealthEvent) {
+	if event == nil || !ruleSelectsOnEntity(rule) {
```
XRFXLP: Don't we already filter nil event prior to calling this function?

health-events-analyzer/pkg/reconciler/entity_metric.go
tanishagoyal2: Why do we need this mapping? Can we directly use entity type value?

tanishagoyal2 (changes requested):

health-events-analyzer/pkg/config/rules_test.go
rupalis-nv: Keeping this file. It does not test metric emission with the flag on,  it is the only coverage that ruleMatchedEntityMetricEnabled parses from TOML (omitted → false, true, explicit false). The reconciler tests do not load config.

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: yeah right... fixed

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: Removed with that helper. The burst path already has a non-nil event from shouldProcessXidEvent.

health-events-analyzer/pkg/reconciler/metrics.go
rupalis-nv: Kept. promauto.NewCounterVec panics if the same metric is registered twice. Tests call EnableRuleMatchedEntityMetric more than once, so this is the idempotent guard.

health-events-analyzer/pkg/reconciler/metrics.go
rupalis-nv: Inlined the labels that only this metric used (entity_type, entity_value). rule_name and node_name stay as constants because they appear on three counters; goconst fails CI if they are inlined.

health-events-analyzer/pkg/reconciler/metrics.go
rupalis-nv: Kept. This is the default-off path: the vec is nil until ruleMatchedEntityMetricEnabled is set. Recording without this check panics when the flag is false.

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: Removed. Proto getters are nil-safe.

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: Removed. GetEntityType / GetEntityValue already treat a nil entity as empty, which we skip.

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: That helper is gone. Proto getters cover nil entities.

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: removed

health-events-analyzer/pkg/reconciler/metrics.go
```diff
@@ -75,3 +81,30 @@ var (
...
+	if ruleMatchedEntityTotal != nil {
+		return
+	}
+
+	ruleMatchedEntityTotal = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "rule_matched_entity_total",
+			Help: "Total number of times a rule matched, labeled by the entity it selected on.",
+		},
+		[]string{"rule_name", "node_name", "entity_type", "entity_value"},
```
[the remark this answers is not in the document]
rupalis-nv: addressed

health-events-analyzer/pkg/config/rules_test.go
tanishagoyal2: okay, but I dont think so it is adding any value. It will be better to remove it

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: Kept. It is the allowlist (drop SM / REG* / anything else) and canonical casing so PCI and pci do not split a series. Using the raw type would export mixed-case labels and unbounded types.

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,222 @@
...
+	"pci":     entityTypePCI,
+	"gpc":     entityTypeGPC,
+	"tpc":     entityTypeTPC,
+	"nvlink":  entityTypeNVLINK,
+	"nic":     entityTypeNIC,
+	"nicport": entityTypeNICPort,
+}
+
+var nicMetricValuePattern = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9._-]{0,63}$`)
+
```
tanishagoyal2: why do we need this pattern check?

health-events-analyzer/pkg/reconciler/metrics.go
tanishagoyal2: In what cases is it possible to register metric twice?

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,224 @@
...
```
tanishagoyal2: why do we need this pattern?

distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
```diff
@@ -44,6 +44,12 @@ workers: 1
 # Maximum number of uncheckpointed in-flight events before applying backpressure (default: 1000)
 maxInFlight: 1000
 
+# Registers rule_matched_entity_total{rule_name, node_name, entity_type, entity_value}.
+# Off by default: entity labels add a series per GPU/GPC/TPC on each node.
+# rule_matched_total already reports that a rule fired without this.
+# GPU UUID is never used as entity_value; PCI or GPU index is exported instead.
+ruleMatchedEntityMetricEnabled: false
```
XRFXLP, line 51: nit: please update values-full.yaml as well.

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,224 @@
...
+// to carry GPU or NIC entities.
+func ruleSelectsOnEntity(rule config.HealthEventsAnalyzerRule) bool {
+	for _, stage := range rule.Stage {
+		if strings.Contains(strings.ToLower(stage), entitiesImpactedFieldName) {
+			return true
+		}
+	}
+
+	return false
+}
```
tanishagoyal2: Can we remove this helper function as only MultipleRemediation is the only rule without entitiesImpacted grouping?

health-events-analyzer/pkg/reconciler/metrics.go
tanishagoyal2: Can we instead check ruleMatchedEntityMetricEnabled?

tanishagoyal2 (changes requested):

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,224 @@
...
+// Prometheus label spelling. GPU UUID is excluded: it is unbounded from
+// Prometheus's point of view, and a replaced GPU changes it.
+var metricSafeEntityTypes = map[string]string{
+	"gpu":     entityTypeGPU,
+	"pci":     entityTypePCI,
+	"gpc":     entityTypeGPC,
+	"tpc":     entityTypeTPC,
+	"nvlink":  entityTypeNVLINK,
+	"nic":     entityTypeNIC,
+	"nicport": entityTypeNICPort,
```
XRFXLP: nvswitch is missing, right? Moreover this hardcoding seems very fragile given that we've ability to plug arbitrary entity type and value. Do we need this static hardcoded map in the code?

health-events-analyzer/pkg/reconciler/metrics.go
```diff
@@ -75,3 +86,30 @@ var (
 		},
 	)
 )
+
+// EnableRuleMatchedEntityMetric registers rule_matched_entity_total. Call once at
+// startup, before the reconciler runs, when the operator has opted in. Repeat
+// calls are ignored so registering twice cannot panic.
+func EnableRuleMatchedEntityMetric() {
+	if ruleMatchedEntityTotal != nil {
```
tanishagoyal2: Actually we can remove this whole function. Even if we register the metric when flag is disabled then also its fine, right? its just that no series will be emitted because before calling  recordMatchedEntityMetric("RepeatedXidError", event.NodeName, event) we can keep a check to check if flag is enabled or not.

distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
rupalis-nv: updated

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,153 @@
...
+	case "gpu", "gpc", "tpc", "nvlink", "nicport":
+		return canonicalBoundedIndex(entityValue)
+	case "pci":
+		return canonicalPCIValue(entityValue)
+	default:
+		if len(entityValue) == 0 || len(entityValue) > 64 {
+			return "", false
+		}
+
+		return entityValue, true
```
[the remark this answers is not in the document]
rupalis-nv: addressed

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,168 @@
...
+
+func ruleSelectsOnEntity(rule config.HealthEventsAnalyzerRule) bool {
+	for _, stage := range rule.Stage {
+		if strings.Contains(strings.ToLower(stage), "entitiesimpacted") {
+			return true
+		}
+	}
+
+	return false
+}
```
tanishagoyal2: Do we need this check? MultipleRemediation is the only rule where no grouping is done on entities. Can we instead exclude ruleMatchedEntityTotal metric emission for multipleRemediation?

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,153 @@
...
+		out = append(out, &protos.Entity{
+			EntityType:  entityType,
+			EntityValue: canonicalValue,
+		})
+	}
+
+	return out
+}
+
+func (r *Reconciler) recordMatchedEntityMetric(ruleName, nodeName string, event *protos.HealthEvent) {
```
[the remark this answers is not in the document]
rupalis-nv: addressed

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,168 @@
...
+		return "", false
+	}
+
+	return fmt.Sprintf("%04x:%02x:%02x", domain, bus, device), true
+}
+
+func canonicalMetricEntityValue(entityType, entityValue string) (string, bool) {
+	switch strings.ToLower(entityType) {
+	case "gpu", "gpc", "tpc", "nvlink", "nicport":
+		return canonicalBoundedIndex(entityValue)
```
tanishagoyal2: Why did we exclude SM? that is also present in entitiesImpacted, right?

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: addressed

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		return 0, false
+	}
+
+	return parsed, true
+}
+
+// canonicalBoundedIndex accepts the decimal GPU / GPC / TPC / NVLINK / NICPort
+// identifiers the XID handlers copy from syslog metadata.
+func canonicalBoundedIndex(value string) (string, bool) {
```
tanishagoyal2: why do we need this check? Why can't we pass on the exact value received through event?

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,168 @@
...
+		return entityValue, true
+	default:
+		return "", false
+	}
+}
+
+func ruleSelectsOnEntity(rule config.HealthEventsAnalyzerRule) bool {
+	for _, stage := range rule.Stage {
+		if strings.Contains(strings.ToLower(stage), "entitiesimpacted") {
+			return true
```
tanishagoyal2, line 93: Do we need this check? MultipleRemediation is the only rule where no grouping is done on entities. Can we instead exclude ruleMatchedEntityTotal metric emission for multipleRemediation?

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// metricSafeEntities returns copies of the triggering event's impacted
+// entities that are safe Prometheus labels. Only PCI, GPU, GPC, TPC, NVLINK,
+// NIC, NICPort, and NVSwitch are kept, using the producer spelling. GPU UUID
+// is omitted. Values are rewritten to a bounded form; malformed values and
+// duplicates are dropped.
+func metricSafeEntities(event *protos.HealthEvent) []*protos.Entity {
+	seen := make(map[string]struct{}, len(event.GetEntitiesImpacted()))
+	out := make([]*protos.Entity, 0, len(event.GetEntitiesImpacted()))
```
tanishagoyal2, line 107: Why do we want to create seen struct? We only want to exclude GPU_UUID, right? Can we directly check if current entity_type is GPU_UUID or not and exclude it?

tanishagoyal2 (changes requested):

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: Yes, SM is on XID 13 entitiesImpacted. The XID 13 rule matches GPC and TPC, not SM, so we only label GPC and TPC.

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: right... removed

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: MultipleRemediations is not the only one. RepeatedNICDriverError also does not group on entities, and the event can still have a NIC. This check skips both so we don’t emit entity labels for node-scoped rules.

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: UUID is already skipped with entityType == GPU_UUID below. seen is only so we don’t increment twice if the same entity is listed twice. It is not what excludes UUID. Happy to remove it if you want.

tanishagoyal2: But isn't entitiesImpacted is already a map? When can we have same entities twice in an event?

health-events-analyzer/pkg/reconciler/entity_metric.go
tanishagoyal2: in the issue, user mentioned the use case to find the rule triggered count by SM. 

`Found while auditing a fleet dashboard where RepeatedXID13OnSameGPCAndTPC had fired several hundred times across two nodes, with no way to tell from metrics whether a specific SM was implicated. Everything needed is already on the health event; only the export is missing.
`
@lalitadithya WDYT? Should we add SM in the labels? or will that increase the cardinality?

health-events-analyzer/pkg/reconciler/entity_metric.go
rupalis-nv: entitiesImpacted is a list, not a map. After skipping UUID, current events don’t repeat PCI/GPC/TPC, so I dropped seen.

health-events-analyzer/pkg/reconciler/entity_metric.go
lalitadithya: isn't SM a bounded number per GPU? I believe the GPC and TPC are bounded right? We can add it as long as there will only ever be a fixed set of values for the label

rupalis-nv: SM is a small bounded decimal from syslog, same as GPC/TPC. I’ll add it to the allow-list.

tanishagoyal2 (approved):

health-events-analyzer/pkg/reconciler/entity_metric.go
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
XRFXLP, line 1: We've lot of parsing and string manipulation in this file, can we please add some tests?

rupalis-nv: added

XRFXLP (approved):

PR #1814, opened 2026-09-11
feat(store-client): idempotency keys, unique partial index and idempotent inserts for deployment platform connector

KaivalyaMDabhadkar:
## Summary

Part of ADR 052, the deployment platform connector. This PR adds the datastore side of the design and nothing else: a per-event idempotency key, the unique partial index that enforces it, inserts that treat a resend as already done, and the index management the connector's index Job and readiness check will call. No component uses the new methods yet; the platform connector adopts them in a later PR. It is independent of #1796 and can merge before or after it.

### What changes

- `datastore` gains the contract both providers share: the metadata field `idempotencyKey`, the index name `healthevent_idempotency_key_unique`, the errors `ErrIndexMissing`, `ErrIndexMismatch` and `ErrIndexBuilding`, and the failure type `BulkWriteFailure`, which names the one document an insert stopped at, with its position, the violated index when it is a duplicate, and the server's message, so the reason reaches the logs.
- `DatabaseClient` gains three methods: `InsertManyIdempotent`, `EnsureHealthEventIdempotencyIndex` and `VerifyHealthEventIdempotencyIndex`. `InsertManyResult` gains `DuplicateCount`, the number of documents skipped because they were already stored under the key. A resend is therefore a success result, not an error; only a real failure is an error.
- MongoDB: `InsertManyIdempotent` is an ordered insert that resumes after a duplicate on the idempotency index, so a resent batch stores only its missing events and a monitor's events land in the order it sent them. A duplicate on any other unique index, or any other per-document failure, stops the insert and is returned as `BulkWriteFailure`. The index is a unique partial index on the key field, covering only documents that carry it, so existing records need no backfill. An index of the same name with another definition is dropped and recreated. Verification checks the name, key path, uniqueness, partial predicate and a completed build.
- PostgreSQL: `InsertManyIdempotent` inserts one row at a time, in order, each outside a transaction, skipping duplicates on the idempotency index and stopping at any other constraint answer. SQLSTATE classes 22 and 23 count as answers about a document; anything else fails the whole batch. The index is created `CONCURRENTLY` outside a transaction so the build never blocks writers; a leftover invalid build or a mismatched definition is dropped concurrently and rebuilt. Verification checks the normalised definition and `pg_index.indisvalid`.
- Connection pool bound: one resolver, `datastore.MaxConnections`, reads the `maxConnections` option first and the existing `DATASTORE_MAX_CONNECTIONS` variable second. PostgreSQL applies it on both the store and the client factory path (default 25); MongoDB caps the driver pool with it (default 100).
- The two test fakes of `DatabaseClient` outside this module (health-events-analyzer, platform-connectors) gain the three methods so their tests keep compiling. A small test pins that `APP_NAME` reaches the MongoDB driver's appName, which the connector relies on to identify itself.

### What is not in this PR

The deployment platform connector, the Helm chart, the publishers and the documentation come in the other PRs of the series.

### Tests

- `mongodb_idempotency_test.go`: bulk write classification (duplicate with the index name, from the message or the key pattern; a non-duplicate keeps the server's message; write concern and non-bulk errors are not classified); the resuming insert (a resend stores its missing events in order with one round trip per duplicate, a real failure stops the insert, a duplicate on another index after a resume stops it, a duplicate at the end costs no extra round trip); index specification checks; the pool cap reaching the driver options.
- `postgresql_idempotency_test.go`: the one-at-a-time insert (a resend continues, another constraint stops it, a transport failure fails the batch, a duplicate on another index stops it), SQLSTATE classification including an administrator shutdown, and literal-preserving SQL normalisation.
- `database_client_idempotency_test.go`: a duplicate met through the health-event insert route, the pool resolver and pool configuration.
- `datastore/idempotency_test.go` and `config_test.go`: the constants that form the schema contract, the failure message, `DuplicateOn`, `AsBulkWriteFailure`, `MaxConnections` and option loading.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] Core Services
- [ ] Health Monitors
- [ ] Fault Management
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Janitor
- [ ] Documentation/CI
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added idempotent health-event batch insertion for MongoDB and PostgreSQL.
  - Duplicate events are skipped and counted, while other write failures include document-level details.
  - Added automatic creation, verification, and repair of health-event idempotency indexes.
  - Added configurable database connection-pool limits through configuration options or environment variables.
  - MongoDB and PostgreSQL clients apply configured pool limits while retaining safe defaults.
  - Event processing continues after duplicate entries while preserving insertion order.

### DIFF
--- a/store-client/pkg/client/mongodb_idempotency.go
+++ b/store-client/pkg/client/mongodb_idempotency.go
@@ -1,0 +1,418 @@
+// Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"regexp"
+	"strings"
+
+	"go.mongodb.org/mongo-driver/v2/bson"
+	"go.mongodb.org/mongo-driver/v2/mongo"
+	"go.mongodb.org/mongo-driver/v2/mongo/options"
+
+	"github.com/nvidia/nvsentinel/store-client/pkg/datastore"
+)
+
+// healthEventIdempotencyKeyDocumentPath is the dotted document path of the
+// idempotency key inside a stored health event document.
+const healthEventIdempotencyKeyDocumentPath = "healthevent.metadata." +
+	datastore.HealthEventIdempotencyKeyMetadataField
+
+// mongoDuplicateKeyErrorCode is the MongoDB server error code for a unique
+// index violation (E11000).
+const mongoDuplicateKeyErrorCode = 11000
+
+// mongoDuplicateIndexNameRegex extracts the violated index name from a
+// duplicate-key error message, e.g.
+// "E11000 duplicate key error collection: db.coll index: <name> dup key: {...}".
+// The server does not expose the index name as a structured field on write
+// errors (only keyPattern/keyValue), so message parsing is the primary source.
+var mongoDuplicateIndexNameRegex = regexp.MustCompile(`index: (\S+)`)
+
+// InsertManyIdempotent inserts the documents in order, so a monitor's events
+// land in the order it sent them, and continues past a document that already
+// exists under the idempotency index. MongoDB keeps array order only for
+// ordered inserts, and an ordered insert stops at its first failure; when that
+// failure is such a duplicate (a resent batch) the insert resumes with the
+// documents after it. Any other per-document failure ends the insert and is
+// reported as a *datastore.BulkWriteFailure with the server's answer; the
+// resends skipped are counted in the result, so the caller can tell a resend
+// from a failure.
+func (c *MongoDBClient) InsertManyIdempotent(ctx context.Context, documents []any) (*InsertManyResult, error) {
+	return insertOrderedResumingDuplicates(ctx, documents,
+		func(ctx context.Context, documents []any) (*mongo.InsertManyResult, error) {
+			return c.mongoCol.InsertMany(ctx, documents)
+		})
+}
+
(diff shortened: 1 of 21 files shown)


store-client/pkg/datastore/idempotency_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"errors"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIdempotencyConstants(t *testing.T) {
```
XRFXLP: this looks like a redundant test as we're testing the value of a constant?

store-client/pkg/client/mongodb_idempotency.go
```diff
@@ -0,0 +1,418 @@
...
+// InsertManyIdempotent inserts the documents in order, so a monitor's events
+// land in the order it sent them, and continues past a document that already
+// exists under the idempotency index. MongoDB keeps array order only for
+// ordered inserts, and an ordered insert stops at its first failure; when that
+// failure is such a duplicate (a resent batch) the insert resumes with the
+// documents after it. Any other per-document failure ends the insert and is
+// reported as a *datastore.BulkWriteFailure with the server's answer; the
+// resends skipped are counted in the result, so the caller can tell a resend
+// from a failure.
+func (c *MongoDBClient) InsertManyIdempotent(ctx context.Context, documents []any) (*InsertManyResult, error) {
```
XRFXLP, line 57: The ADR specified unordered InsertMany, do we need to update that?

store-client/pkg/datastore/idempotency_test.go
KaivalyaMDabhadkar: Good catch, removed it

store-client/pkg/client/mongodb_idempotency.go
KaivalyaMDabhadkar: Good catch, the design doc still said unordered. We went ordered because mongo only keeps a batch's order for ordered inserts, and a batch can hold a fatal and a later healthy event for the same check, which consumers must see in that order. The only cost is one extra round trip per duplicate on a resend. I have updated the two lines in the design doc in this PR.

XRFXLP (approved):

PR #1817, opened 2026-09-11
feat(health-events-analyzer): concurrent node-partitioned event processing

### LINKED ISSUE #1800: [Feature]: Process health events concurrently, partitioned by node
### Prerequisites

### Code of Conduct

### Feature Summary

Let health-events-analyzer process events concurrently, partitioned by node. It currently consumes its change stream in a single serial loop, so its throughput is fixed at one event per handling time regardless of resources, and one replica cannot be scaled or supplemented.

### Problem/Use Case

**Measured throughput.** The component's own histogram reports **17.67 ms per event** over 104,090 events (`health_event_analyzer_event_handling_duration_seconds`, sum 1,838.81 s). Wall time held between 17.5 and 19.4 ms across offered rates from 1.8 to 55 events/s, so it is flat rather than degrading. That fixes the ceiling at roughly **57 events/s**.

**Why more resources do not help.** `store-client/pkg/client/event_processor.go` consumes the stream in one `for`/`select` loop calling `handleSingleEvent` synchronously; nothing in `health-events-analyzer/pkg/` starts a goroutine, an `errgroup`, or a `WaitGroup`. Each event runs one aggregation per rule (`reconciler.go:243` ranges over the rule set, `:449` calls `Aggregate`), and most of the 17.67 ms is spent waiting on MongoDB rather than computing: CPU is about 5.8 ms of it, so at saturation the container draws **0.2 cores against a 2-core limit with zero throttled CFS periods**. Raising CPU or memory changes nothing.

**Why replicas do not help either.** There is no leader election and no partitioning; `replicaCount` is 1 and all replicas would share a single resume token keyed by `clientName: health-events-analyzer`. A second replica would re-process every event and the two would overwrite each other's checkpoint.

**Scale implication.** At an ambient rate of 0.1 events per node per second, one replica saturates at roughly **570 nodes**. A 100,000-node fleet on that assumption offers 10,000 events/s against a 57/s ceiling.

**This is not a regression.** [#1676](https://github.com/NVIDIA/NVSentinel/pull/1676) improved this by roughly 10x -- the microbenchmark in [#1523](https://github.com/NVIDIA/NVSentinel/issues/1523) measured v1.16.0 at 186 ms/event and ~4.6 events/s, against 17.67 ms and ~57/s on v1.22.0. The remaining gap is structural rather than a matter of further query tuning.

### Proposed Solution

**Partition by node and process partitions concurrently.** #1676 already made this safe: it added an enforced same-node constraint as the mandatory first filter of every rule pipeline, so events for different nodes cannot influence each other's evaluation. Events for the same node must stay ordered relative to each other; events for different nodes need not be.

Concretely, a worker pool keyed by a hash of `healthevent.nodename`, with one in-flight event per node at a time. With the measured 17.67 ms per event, N workers give roughly N x 57 events/s, so 16 workers reach ~900/s and 176 reach the 10,000/s a 100,000-node fleet implies.

**Checkpointing is the part that needs design.** The resume token may only advance past an event once every earlier event has been handled, so out-of-order completion requires tracking a low-water mark rather than writing the token of whichever event finished last. The current code takes the simple path because completion is strictly ordered.

**Two smaller changes worth considering alongside it**, both suggested by the measurement rather than by the architecture:

1. Each event runs 8 aggregations, one per rule set, and most of the 17.67 ms is MongoDB round-trip. Rules sharing a match shape could be evaluated in one aggregation, cutting the per-event cost directly rather than parallelising it.
2. If partitioning by node is adopted, sharding across replicas becomes possible too -- separate resume tokens per shard, each covering a disjoint node range -- which would also remove the single point of failure.

**Two things to confirm before accepting this design.** Not every shipped rule has been verified as genuinely node-scoped -- `MultipleRemediations` in particular -- and a rule that correlates across nodes would break node-partitioning. And the 0.1 events per node per second above is a scale-test assumption, not a measurement from a production fleet; the real rate should be established before sizing a worker pool against it.

HesandaLiyanage:
## Summary

Closes #1800.

This PR implements concurrent node-partitioned event processing in `health-events-analyzer` and `store-client` to address the ~57 events/s serial processing bottleneck benchmarked in issue #1800.

Key highlights:
- **Node-Keyed Partitioning:** Events are routed by node name hash into dedicated FIFO worker channels. Events for the same node are guaranteed to process strictly sequentially, while events across different nodes process concurrently.
- **Low-Water Mark Checkpointing (`LowWaterMarkTracker`):** Guarantees that out-of-order event completion never skips earlier unresolved events during restarts or crashes.
- **Backpressure & Bounded Memory:** Enforces `MaxInFlight` limits to pause stream reading when worker queues or uncheckpointed events reach capacity.
- **ADR-056:** Documents the architecture, decisions, failure mitigations, and alternatives considered in `docs/designs/056-concurrent-node-partitioned-event-processing.md`.
- **100% Backwards Compatible:** Preserves default `workers: 1` in `EventProcessorConfig` and Helm values.

## Type of Change
- [x] ✨ New feature
- [x] 📚 Documentation

## Component(s) Affected
- [x] Health Monitors (`health-events-analyzer`)
- [x] Deployment/Config (`distros/kubernetes/nvsentinel/charts/health-events-analyzer`)
- [x] Other: `store-client`

## Testing
- [x] Tests pass locally (`store-client` and `health-events-analyzer` with `go test -race`)

## Checklist
- [x] Self-review completed
- [x] Documentation updated (ADR-056 added and indexed)
- [x] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Health event processing now runs concurrently across nodes while preserving per-node order.
  - Added configurable worker counts, in-flight event limits, per-event timeouts, and partitioning behavior.
  - Checkpointing safely tracks completed events, including out-of-order processing.
  - Cancelled or timed-out events remain eligible for retry.
  - Shutdown retries final checkpoints when needed.
  - Added Helm configuration for worker and in-flight event settings.

- **Documentation**
  - Added guidance on concurrent processing, checkpointing, scaling, and configuration options.

Here's the throughput scaling matrix across different worker counts, tested with events distributed across 100 simulated GPU nodes:

### 1. Concurrent Workload Benchmark (Simulated Per-Event Processing)

| Workers | Latency (ns/op) | Throughput (events/s) | Speedup vs 1 Worker | Scaling Efficiency |
| :--- | :--- | :--- | :--- | :--- |
| **1** (serial baseline) | 178,368 ns/op | ~5,606 events/s | **1.00x** | 100% |
| **2** | 75,812 ns/op | ~13,190 events/s | **2.35x** | ~100% |
| **4** | 46,106 ns/op | ~21,689 events/s | **3.87x** | 96.7% |
| **8** | 22,070 ns/op | ~45,310 events/s | **8.08x** | ~100% |
| **16** | 15,434 ns/op | ~64,792 events/s | **11.56x** | 72.2% |

---

### 2. Production Health-Events-Analyzer Projection (MongoDB Aggregation Baseline)

In production `health-events-analyzer`, each rule evaluation executes an aggregation pipeline (~17.5ms per evaluation as profiled in #1800). Under node-partitioned parallel execution, throughput scales as follows:

| Workers | Throughput (events/s) | Speedup vs Serial (1 Worker) |
| :--- | :--- | :--- |
| **1** (serial baseline) | ~57 events/s | 1.0x |
| **2** | ~113 events/s | ~1.98x |
| **4** | ~224 events/s | ~3.93x |
| **8** | ~440 events/s | ~7.72x |
| **16** | ~862 events/s | ~15.1x |
| **32** | ~1,680 events/s | ~29.5x |

### 3. Pipeline Engine Overhead

Without I/O blocking, the in-memory hashing, channel multiplexing, and `LowWaterMarkTracker` sequencing process **~450,000 – 550,000 events/second**, confirming the partitioning machinery itself introduces negligible overhead.

### DIFF
--- a/distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
+++ b/distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
@@ -38,6 +38,12 @@ logLevel: info
 # STORE_ONLY: observability-only behavior; event should be persisted/exported but should not modify cluster resources (i.e., no node conditions, no quarantine, no drain, no remediation).
 processingStrategy: EXECUTE_REMEDIATION
 
+# Number of concurrent workers for event processing partitioned by node (default: 1)
+workers: 1
+
+# Maximum number of uncheckpointed in-flight events before applying backpressure (default: 1000)
+maxInFlight: 1000
+
 # Enable flags for the rules to be evaluated.
 enableMultipleRemediationsRule: true
 enableRepeatedXIDErrorOnSameGPURule: true
--- a/docs/designs/056-concurrent-node-partitioned-event-processing.md
+++ b/docs/designs/056-concurrent-node-partitioned-event-processing.md
@@ -86,8 +86,7 @@ ChangeStreamWatcher (Events channel)
    - Implements `client.EventProcessor`.
    - Manages $N$ worker goroutines and per-worker buffered channels.
    - Maps events to workers via `hash(nodeName) % N` using FNV-32a. Events without a node name route to worker 0.
-   - Enforces backpressure: when uncheckpointed in-flight events reach `MaxInFlight`, suspends reading from the watcher.
-   - Handles poison events: when `MarkProcessedOnError=true` (as in `health-events-analyzer`), failed events are marked resolved in the tracker to prevent stream stalling.
+   - Handles poison events vs transient errors: when `MarkProcessedOnError=true` (as in `health-events-analyzer`), only terminal errors (e.g., permanent unmarshaling failures or handler rejections) are marked resolved in the tracker to prevent poison pills from blocking stream progress. Transient conditions—specifically context cancellations (`context.Canceled`) and timeouts (`context.DeadlineExceeded`)—are never marked as completed, ensuring the low-water mark does not advance past them and that they are retried on pod restart rather than permanently lost.
 
 3. **`health-events-analyzer` Integration**:
    - `health-events-analyzer/main.go`: adds CLI flags `--workers` (default `1`) and `--max-in-flight` (default `1000`).
--- a/docs/designs/056-concurrent-node-partitioned-event-processing.md
+++ b/docs/designs/056-concurrent-node-partitioned-event-processing.md
@@ -1,0 +1,153 @@
+# ADR-056: `Performance` — Concurrent Node-Partitioned Event Processing
+
+## Table of contents
+
+- [Context](#context)
+- [Decision](#decision)
+- [Implementation](#implementation)
+- [Rationale](#rationale)
+- [Consequences](#consequences)
+- [Alternatives Considered](#alternatives-considered)
+- [Testing](#testing)
+- [Rollout](#rollout)
+- [References](#references)
+
+## Context
+
+`health-events-analyzer` monitors the database change stream for non-fatal, transient health events, correlates them over time windows via rule-driven database aggregation queries, and elevates recurrent failure patterns to fatal events for automated remediation.
+
+In production and scale testing, `health-events-analyzer` operates on a strict serial processing loop. Metrics benchmarked in issue #1800 indicate:
+- Handling duration per event averages **17.67 ms** (sum 1,838.81 s over 104,090 events).
+- Wall-clock time remains constant (17.5–19.4 ms) across offered rates from 1.8 to 55 events/s.
+- This creates an effective processing ceiling of approximately **57 events/s** per replica.
+- Most of the 17.67 ms is spent in database network round-trips (I/O wait) executing aggregation pipelines; CPU usage accounts for only ~5.8 ms per event.
+- Vertical scaling (allocating more CPU/memory cores) does not increase throughput because execution is I/O-bound in a single synchronous `for/select` loop.
+- Horizontal scaling (`replicaCount > 1`) is unsupported: there is no partitioning or leader election; all replicas share a single resume token keyed by `clientName: health-events-analyzer`, leading to duplicate execution and conflicting checkpoints.
+- At an ambient rate of 0.1 events per node per second, a single replica saturates at ~570 nodes. A 100,000-node fleet implies an offered load of 10,000 events/s against a 57 events/s ceiling.
+
+PR #1676 established the prerequisite for parallelization: it enforced a mandatory same-node filter (`healthevent.nodename: <incoming_event.nodename>`) as the first stage of every rule pipeline. Events for different nodes do not influence each other's evaluations.
+
+## Decision
+
+1. **Partition events by node across a concurrent worker pool**:
+   Events are hashed by `healthevent.nodename` into dedicated FIFO worker channels. Events for the same node are processed serially in strict chronological order; events for distinct nodes are processed concurrently across $N$ workers.
+2. **Implement low-water mark checkpointing**:
+   Because concurrent execution allows out-of-order completion across nodes, the change stream resume token must not be updated to whichever event finished last. The resume token is advanced only to the highest contiguous sequence number where every earlier event has been fully resolved (completed or marked processed per error policy).
+3. **Encapsulate concurrent processing in `store-client`**:
+   The concurrent processor is implemented in `store-client/pkg/client` as `PartitionedEventProcessor`, satisfying the existing `client.EventProcessor` interface.
+4. **Preserve strict backward compatibility**:
+   Worker count defaults to `1` in `EventProcessorConfig` and Helm charts (`workers: 1`). When `workers <= 1`, behavior remains single-threaded.
+
+## Implementation
+
+### Architecture Overview
+
+```text
+ChangeStreamWatcher (Events channel)
+            │
+            ▼
+┌────────────────────────────────────────────────────────────────────────┐
+│                      PartitionedEventProcessor                         │
+│                                                                        │
+│   1. Assign monotonically increasing sequence S_k to incoming Event    │
+│   2. Register (S_k, Token_k, Pending) in LowWaterMarkTracker           │
+│   3. Apply backpressure if in-flight >= MaxInFlight                    │
+│   4. Route to worker: workerIdx = hash(Event.NodeName) % NumWorkers   │
+│   5. Enqueue to Worker Queue W_i                                       │
+│                                                                        │
+│   Worker Queues (FIFO per worker):                                     │
+│   ┌───────────────┐      ┌───────────────┐      ┌───────────────┐      │
+│   │   Worker 0    │      │   Worker 1    │ ...  │   Worker N-1  │      │
+│   │ (Nodes A, C)  │      │ (Nodes B, D)  │      │ (Nodes E, F)  │      │
+│   └───────┬───────┘      └───────┬───────┘      └───────┬───────┘      │
+│           │                      │                      │              │
+│           ▼                      ▼                      ▼              │
+│       EventHandler.ProcessEvent(ctx, &HealthEventWithStatus)           │
+│           │                      │                      │              │
+│           └──────────────────────┼──────────────────────┘              │
+│                                  ▼                                     │
+│                        LowWaterMarkTracker                             │
+│                  - MarkComplete(S_k)                                   │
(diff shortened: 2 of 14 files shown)


HesandaLiyanage:
@XRFXLP handled the CodeRabbit comments, PR's ready for review

docs/designs/056-concurrent-node-partitioned-event-processing.md
```diff
@@ -0,0 +1,159 @@
...
+   The concurrent processor is implemented in `store-client/pkg/client` as `PartitionedEventProcessor`, satisfying the existing `client.EventProcessor` interface.
+4. **Preserve strict backward compatibility**:
+   Worker count defaults to `1` in `EventProcessorConfig` and Helm charts (`workers: 1`). When `workers <= 1`, behavior remains single-threaded.
+
+## Implementation
+
+### Architecture Overview
+
+```text
+ChangeStreamWatcher (Events channel)
```
XRFXLP: Can we use mermaid diagram?

docs/designs/056-concurrent-node-partitioned-event-processing.md
```diff
@@ -0,0 +1,159 @@
...
+- [Rationale](#rationale)
+- [Consequences](#consequences)
+- [Alternatives Considered](#alternatives-considered)
+- [Testing](#testing)
+- [Rollout](#rollout)
+- [References](#references)
+
+## Context
+
+`health-events-analyzer` monitors the database change stream for non-fatal, transient health events, correlates them over time windows via rule-driven database aggregation queries, and elevates recurrent failure patterns to fatal events for automated remediation.
```
XRFXLP: nit: these reads off as <pattern of non-fatal AND transient health events> HEA -> <fatal events>. But HEA is pretty general as it can monitor fatal events as well.

health-events-analyzer/pkg/reconciler/reconciler.go
```diff
@@ -72,13 +74,28 @@ func NewReconciler(cfg HealthEventsAnalyzerReconcilerConfig) *Reconciler {
 	}
 }
 
-func newEventProcessorConfig() client.EventProcessorConfig {
+func newEventProcessorConfig(reconcilerCfg ...HealthEventsAnalyzerReconcilerConfig) client.EventProcessorConfig {
+	workers := 1
+	maxInFlight := 1000
+
+	if len(reconcilerCfg) > 0 {
```
XRFXLP: is there a possibility of case for which this if-clause is defending for? We've defaults of workers as 1 and MaxInFlight for 1000, right?

XRFXLP (commented):
Can you please add some details for matrix for the event throughput achieved through different count of workers?

docs/designs/056-concurrent-node-partitioned-event-processing.md
HesandaLiyanage: Sure I'll update this!!!

HesandaLiyanage:
> Can you please add some details for matrix for the event throughput achieved through different count of workers?

@XRFXLP 

Here's the throughput scaling matrix across different worker counts, tested with events distributed across 100 simulated GPU nodes:

### 1. Concurrent Workload Benchmark (Simulated Per-Event Processing)

| Workers | Latency (ns/op) | Throughput (events/s) | Speedup vs 1 Worker | Scaling Efficiency |
| :--- | :--- | :--- | :--- | :--- |
| **1** (serial baseline) | 178,368 ns/op | ~5,606 events/s | **1.00x** | 100% |
| **2** | 75,812 ns/op | ~13,190 events/s | **2.35x** | ~100% |
| **4** | 46,106 ns/op | ~21,689 events/s | **3.87x** | 96.7% |
| **8** | 22,070 ns/op | ~45,310 events/s | **8.08x** | ~100% |
| **16** | 15,434 ns/op | ~64,792 events/s | **11.56x** | 72.2% |

---

### 2. Production Health-Events-Analyzer Projection (MongoDB Aggregation Baseline)

In production `health-events-analyzer`, each rule evaluation executes an aggregation pipeline (~17.5ms per evaluation as profiled in #1800). Under node-partitioned parallel execution, throughput scales as follows:

| Workers | Throughput (events/s) | Speedup vs Serial (1 Worker) |
| :--- | :--- | :--- |
| **1** (serial baseline) | ~57 events/s | 1.0x |
| **2** | ~113 events/s | ~1.98x |
| **4** | ~224 events/s | ~3.93x |
| **8** | ~440 events/s | ~7.72x |
| **16** | ~862 events/s | ~15.1x |
| **32** | ~1,680 events/s | ~29.5x |

### 3. Pipeline Engine Overhead

Without I/O blocking, the in-memory hashing, channel multiplexing, and `LowWaterMarkTracker` sequencing process **~450,000 – 550,000 events/second**, confirming the partitioning machinery itself introduces negligible overhead.

docs/designs/056-concurrent-node-partitioned-event-processing.md
HesandaLiyanage: Makes sense. I'll do the changes

health-events-analyzer/pkg/reconciler/reconciler.go
HesandaLiyanage: The variadic `...` and `len > 0` check were only added so existing test calls (`newEventProcessorConfig()`) didn't need to be updated. In production, `r.config` is always passed.

The inner `Workers > 0` check just guards against zero-valued config structs (e.g. unit tests where flags aren't parsed).

I can simplify this to take `cfg HealthEventsAnalyzerReconcilerConfig` directly:

```
func newEventProcessorConfig(cfg HealthEventsAnalyzerReconcilerConfig) client.EventProcessorConfig {
    workers := cfg.Workers
    if workers <= 0 {
        workers = 1
    }
    maxInFlight := cfg.MaxInFlight
    if maxInFlight <= 0 {
        maxInFlight = 1000
    }
    return client.EventProcessorConfig{
        Workers:     workers,
        MaxInFlight: maxInFlight,
    }
}
```

Would you prefer that cleaner signature?

docs/designs/056-concurrent-node-partitioned-event-processing.md
HesandaLiyanage: Resolved

docs/designs/056-concurrent-node-partitioned-event-processing.md
HesandaLiyanage: Resolved

health-events-analyzer/pkg/reconciler/reconciler.go
XRFXLP: yeah, that would be good.

HesandaLiyanage: Resolved

HesandaLiyanage:
@XRFXLP All review comments are now resolved.

HesandaLiyanage:
@lalitadithya @XRFXLP All review comments are addressed and CI is green. CodeRabbit had a minor nitpick on the test we just discussed and updated, but leaving it as agreed to avoid scope creep.

distros/kubernetes/nvsentinel/charts/health-events-analyzer/values.yaml
```diff
@@ -38,6 +38,12 @@ logLevel: info
 # STORE_ONLY: observability-only behavior; event should be persisted/exported but should not modify cluster resources (i.e., no node conditions, no quarantine, no drain, no remediation).
 processingStrategy: EXECUTE_REMEDIATION
 
+# Number of concurrent workers for event processing partitioned by node (default: 1)
```
XRFXLP, line 41: Can you add this in values-full.yaml and docs/configuration along with the table on how to scale this paramters depending on the event rate in DB?

HesandaLiyanage: Resolved !!

XRFXLP (approved):

HesandaLiyanage:
@lalitadithya @XRFXLP Thank you so much for the thorough reviews and thoughtful feedback throughout this PR! I really enjoyed working on this. 
I'd love to continue contributing to NVSentinel, so please feel free to tag or assign me to any future issues , I'd be more than happy to pick them up!

PR #1821, opened 2026-09-11
feat: implement new node validation

natherz97:
## Summary

This PR adds support for new node validation by implementing the node-validation-controller. The purpose of this controller is to batch together a group of nodes matching a particular criteria and create a single ValidationRequest targeting all nodes in the batch. Note that enabling this functionality requires enabling the validation-request-controller and specifying spec.newNodeValidation as part of the ValidationConfiguration:
```
  newNodeValidation:
    condition: NewNodeValidated
    batchPeriodSeconds: 900
    criteria:
      - name: recently-joined
        expression: >-
          has(node.metadata.creationTimestamp) &&
          now() - timestamp(node.metadata.creationTimestamp) < duration("30m")
      - name: gpu-present
        expression: >-
          has(node.metadata.labels) && "nvidia.com/gpu.present" in node.metadata.labels
    newNodeTests:
      - dcgm-diag-test
```
This configuration will batch together nodes over a 15 minute period where all nodes were created within the past 30 minutes and have the nvidia.com/gpu.present=true label present. Nodes which already have the NewNodeValidated condition will not be considered eligible to prevent the same node from being included in multiple batches. This will result in a ValidationRequest with newNodeTests which targets all nodes in the batch.
```
apiVersion: lifecycle.nvsentinel.nvidia.com/v1alpha1
kind: ValidationRequest
metadata:
  name: validation-1757612345123456789 
  labels:
    app.kubernetes.io/managed-by: nvsentinel
    nvsentinel.nvidia.com/created-by: lifecycle-manager
spec:
  nodes:
    - name: gpu-node-01
    - name: gpu-node-02
  tests:
    - dcgm-diag-test
```
See the comment for Reconcile in nodevalidation_controller.go for a comprehensive overview for the controller's implementation.

## Testing

Unit test cases for the node-validation-controller:
1. Batches one node into a ValidationRequest
2. Batches multiple nodes into a ValidationRequest
3. Creates two ValidationRequests for two batches
4. Creates no ValidationRequest when batch becomes empty
5. Does not add ineligible nodes to a pending batch
6. Removes ineligible nodes from a pending batch before it is flushed
7. Removes ineligible nodes from the batch at flush time
8. Adds eligible node when criteria is empty
9. Records a successful batch in metrics
10. Removes a deleted node from a pending batch before it is flushed
11. Removes a deleted node from the batch at flush time
12. Recovers after encountering an AlreadyExists error on ValidationRequest creation
13. Retries updating the newNodeValidation.condition for all nodes in the batch after a failure
14. Uses the node batch defined in the ValidationRequest after encountering an AlreadyExists error

End-to-end cases:
1. Existing TestValidationController was extended to have the node-validation-controller automatically create a ValidationRequest for a new node rather than manually create the ValidationRequest. 

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [X] Core Services
- [ ] Health Monitors
- [ ] Fault Management
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Janitor
- [ ] Documentation/CI
- [ ] Other: ____________

## Testing
- [X] Tests pass locally

## Checklist
- [X] Self-review completed
- [X] Ready for review

## New Features
- Added automatic validation for newly joined GPU nodes, with configurable eligibility criteria, batching, diagnostic tests, and validation status updates.
- Added Kubernetes event reporting permissions and monitoring metrics for validation batches.

## Configuration
- Added `newNodeValidation` settings using numeric batch intervals and required validation conditions.
- Updated examples for timestamp and GPU-label checks.

## Bug Fixes
- Improved handling of deleted, stale, already-processed, or concurrently updated nodes during validation.
- Added safer rechecks before validation requests are created.


lifecycle-manager/internal/controller/nodevalidation_controller.go
```diff
@@ -0,0 +1,376 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	requeueDurationForBatch := r.addNodeToPendingBatch(node.Name)
+
+	return ctrl.Result{RequeueAfter: requeueDurationForBatch}, nil
+}
+
+func (r *NodeValidationReconciler) createValidationRequestForNewNodes(ctx context.Context) error {
+	candidateNames := make([]string, 0, len(r.nodesInBatch))
```
XRFXLP, line 165: nit: can we do: 
```
candidateNames := slices.Collect(maps.Keys(r.nodesInBatch))
``` 
?

lifecycle-manager/internal/controller/nodevalidation_controller.go
```diff
@@ -0,0 +1,376 @@
...
+				return fmt.Errorf("mark node %q validated for %q: %w", name, validationRequestName, err)
+			}
+		}
+
+		metrics.NewNodeValidationBatchSize.Observe(float64(len(targetedNodes)))
+	}
+
+	r.nodesInBatch = nil
+
+	metrics.NewNodeValidationBatchesTotal.WithLabelValues(metrics.StatusSuccess).Inc()
```
XRFXLP, line 198: is this intended to fire even if no nodes were eligible and no CR was created?

lifecycle-manager/internal/controller/nodevalidation_controller.go
```diff
@@ -0,0 +1,376 @@
...
+		APIReader:        apiReader,
+		Scheme:           scheme,
+		Config:           cfg,
+		CriteriaPrograms: programs,
+	}, nil
+}
+
+func (r *NodeValidationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Node{}).
```
XRFXLP, line 78: do we need any transform stripping to avoid caching all the fields of the node?

lifecycle-manager/internal/controller/nodevalidation_controller.go
natherz97: Nice! Updated.

lifecycle-manager/internal/controller/nodevalidation_controller.go
natherz97: Good callout, the original intent of this metric was to capture to any successful flush of the batch which does currently cover all nodes within the batch being ineligible at flush time. Agree this can be misleading, updated to only emit this metric when we successfully create a ValidationRequest. This means we'll emit NewNodeValidationBatchSize + NewNodeValidationBatchesTotal together in all cases.

lifecycle-manager/internal/controller/nodevalidation_controller.go
natherz97: Right now we have both the node-validation-controller and the validation-controller both sharing the same underlying controller-manager and use a shared informer cache for nodes.


1. We need the following fields for nodes from the node-validation-controller:
a. metadata: node name, creation timestamp for newNodeValidation.criteria, labels for newNodeValidation.criteria
b. status: required for newNodeValidation.condition we use to only trigger validation once

3. We need the following fields for nodes from the validation-controller:
a. metadata: node name, active-validation-request and validation-session annotations, the quarantineHealthEvent annotation, and managing the cordon labels.
b. status: node allocatable status
c. spec: cordon and taint state

Since we are leveraging metadata, spec, and status across the 2 controllers and since it's possible for a user to refer to arbitrary node fields in either newNodeValidation.criteria or config.readinessCriteria CEL rules, I don't think it would be trivial to apply a transformation. 

Let me know your thoughts, we could choose to exclude certain properties from being requested in the CEL rules and leverage a transform to remove those pre-defined properties (it looks like node.Status.Images could be a good candidate for this).

XRFXLP: I see.

We can do followup if the component is OOMing in large cluster, I'm working on a shared mechanism for caching only those fields which are referred by CEL.

tests/validation_request_test.go
```diff
@@ -50,27 +52,35 @@ func TestValidationController(t *testing.T) {
...
+			client, err := c.NewClient()
+			require.NoError(t, err, "failed to create kubernetes client")
 
-		_, err = helpers.CreateValidationRequest(ctx, client, vrName, []string{nodeName}, []string{"e2e-smoke-test"})
-		require.NoError(t, err, "ValidationRequest should be created successfully")
+			vr := helpers.WaitForValidationRequestForNode(ctx, t, client, nodeName)
+			require.NotNil(t, vr, "a ValidationRequest should be automatically for the new node")
 
-		helpers.WaitForValidationRequestPhase(ctx, t, client, vrName, "Running")
+			helpers.WaitForNodeConditionWithCheckName(ctx, t, client, nodeName, "NewNodeValidated", "", "",
```
XRFXLP: nit: do we need to clear this condition in teardown?

lifecycle-manager/internal/controller/nodevalidation_controller.go
```diff
@@ -0,0 +1,374 @@
...
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+
+			return nil, fmt.Errorf("get node %q: %w", name, err)
+		}
+
+		isEligible, err := r.isNodeEligibleForBatch(&node)
+		if err != nil {
+			return nil, err
```
XRFXLP, line 252: if there is some error with one node, won't it stop the validation for entire batch? Is that intended?

lifecycle-manager/internal/controller/nodevalidation_controller.go
```diff
@@ -0,0 +1,374 @@
...
+
+	return kubeclient.RetryNodePatch(func() error {
+		var node corev1.Node
+		if err := r.APIReader.Get(ctx, client.ObjectKey{Name: nodeName}, &node); err != nil {
+			return client.IgnoreNotFound(err)
+		}
+
+		original := node.DeepCopy()
+
+		if !setNodeCondition(&node, conditionType, message) {
```
XRFXLP, line 325: This is kinda confusing; default chart ships `NewNodeValidated` and we set that whenever we've created a new CR that is kinda misleading incase the test fails as it will be:
```
- type: NewNodeValidated
  status: true
```
And that contradicts the ADR as it says:
> Once validation completes successfully, the controller sets the condition to True on the node so that subsequent evaluations no longer match

Although it also has:
> * **Validating new nodes:** we will need the validation-controller to create ValidationRequests for new nodes. Creation of the CRD can be signaled by the absence of a node status condition called NewNodeValidated, which the controller then sets to True after the request is created.

So I'm not sure which one we're following.

tests/validation_request_test.go
natherz97: Added removal of NewNodeValidated status condition to Teardown.

lifecycle-manager/internal/controller/nodevalidation_controller.go
natherz97: This is a good callout. The only 2 cases where an individual node can block the batch from being flushed is either:
- The get node requests fails with an error that isn't a 404
- The CEL expression evaluation of the node object against CriteriaPrograms fails

If we proceeded with the batch without the given node, there's not a straightforward way to ensure that we re-reconcile the problematic node. For example if we hit this sequence:
- node1 fires reconciling 
- we detect that we're at flush time and proceed with trying to create a ValidationRequest for node1 and node2 in the batch
- we check that each node is eligible and get an error on node2

At this point, if we proceeded with the ValidationRequest ignoring node2, we wouldn't have a way to trigger node2 reconciling and have it added to the next batch by returning an error  (since an error returned on this reconcile would re-queue node1). I think we'd have to either make a write against node2 to force a reconcile or it looks like controller-runtime has a way to do this by writing to a channel you can set-up when you construct the controller. A different approach might be to keep any failed nodes within the batch after the flush and allow it to eligible for the next batch.

For simplicity, I think we should prevent the current batch from flushing so that we get the retry for all nodes (specifically node2 would have its get request + CEL expression retried in the example above). If we run into issue later I think we can add it as a follow-up. Let me know if you agree with that.

lifecycle-manager/internal/controller/nodevalidation_controller.go
natherz97: Thanks for calling that out, the second description is correct which is also the behavior currently implemented. I will rename the node status condition to NewNodeValidationRequested to make it clear that the status condition only corresponds to the ValidationRequest being created but not succeeding. I also fixed the conflicting ADR section.

lifecycle-manager/internal/controller/nodevalidation_controller.go
lalitadithya: answering on behalf of OP since if on vacation and off grid. Hope he doesn't mind. I think a follow up is fine for now.

lalitadithya:
can you please check test failures? they don't seem to be flakes as they are failing consistently

natherz97:
@lalitadithya pushed a revision to fix e2e tests, the node-validation-controller was targeting all Kind/KWOK nodes since they already had the gpu.present label added. I updated the new node criteria for the e2e tests to be a custom label which is added and removed by the TestValidationController test.

lalitadithya (approved):

PR #1837, opened 2026-09-15
feat(metadata-collector): support host-native authentication

### LINKED ISSUE #1836: feat(metadata-collector): support host authentication for pod-to-GPU mapping
### Problem

`metadata-collector` cannot keep pod-to-GPU mapping active when it runs as a host-native process. Hardware inventory completes, but mapper initialization calls `rest.InClusterConfig()` and exits outside a pod:

```text
unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
```

The kubelet HTTPS client also reads a fixed projected ServiceAccount token path. Setting API server environment variables alone does not provide host authentication.

### Reproduction

Run the v1.23.0 metadata-collector executable on a Linux GPU node with the NVIDIA driver and kubelet PodResources socket available, but without pod ServiceAccount files or in-cluster environment variables. Inventory is written, then the process exits during mapper initialization.

### Proposed change

- Add optional `--kubeconfig` for Kubernetes API access and `--kubelet-kubeconfig` for the kubelet endpoint, credentials, and TLS trust.
- Keep separate credentials and certificate authorities for the two endpoints.
- Preserve the current in-cluster defaults when flags are absent.
- Use client-go transports for explicit credentials and require verified HTTPS for host configuration.
- Keep hardware inventory, pod annotations, and pod-to-GPU mapping unchanged.
- Let the PodResources client handle delayed socket availability within a bounded, cancellable request. Preserve the existing consecutive-poll failure limit.
- Add tests for authentication isolation, TLS rejection, credential rotation, startup delay, recovery, and annotation updates.

This follows the explicit configuration approach in #1359. It does not disable pod mapping or add an inventory-only mode.

### Scope

No credential provisioning, automatic RBAC changes, remediation, packaging, release workflow changes, or cloud-specific integration. Operators remain responsible for credentials and authorization. A design record and implementation will accompany the PR for review.

sulixu:
## Summary

Closes #1836.

Allow metadata-collector to authenticate as a host-native process while retaining hardware inventory and pod-to-GPU mapping.

- Add `--kubeconfig` for Kubernetes API access and `--kubelet-kubeconfig` for kubelet HTTPS access.
- Keep endpoints, credentials, and CA trust separate. Explicit configurations require authenticated, verified HTTPS.
- Preserve in-cluster defaults and existing constructor calls. Do not implicitly load a home-directory kubeconfig.
- Keep inventory and annotation schemas unchanged. Do not add an inventory-only mode.
- Wait for delayed PodResources socket availability within bounded, cancellable requests. Retain the existing consecutive-poll failure limit.
- Test token and certificate file rotation, certificate validation, authorization errors, socket recovery, and annotation changes.

ADR-058 records the proposed design and extends the deployment options described in ADR-010. It does not change publisher authentication from ADR-052. Maintainer acceptance is pending.

Credential provisioning, RBAC changes, packaging, release workflows, and remediation are outside this PR.

## Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected

- [ ] Health Monitor
- [x] Core Service
- [ ] Fault Management
- [ ] Janitor
- [x] Deployment/Config
- [x] API/Interface
- [ ] Preflight
- [ ] Plugins
- [x] Documentation/CI
- [ ] New Component
- [ ] Multiple Components
- [ ] Other

## Testing

- [x] Tests pass locally

`make -C metadata-collector lint-test` passes: 104 tests with the race detector, no lint findings. The integration test uses a real local API server and RBAC. It verifies annotation addition, changes, removal, no-op updates, and permission denial. The Linux AMD64 executable builds and exposes both new flags.

The complete metadata-collector suite also passes on Linux AMD64 with `go test -race -count=1 -timeout 5m ./...`.

The full `make lint-test-all` check passes on Linux AMD64 with the pinned build tools. This includes protobuf generation, license headers, all 33 Go module checks, module lint and tests, Helm lint, and script lint. Existing skipped tests remain unchanged. No test assertions or version pins were changed to obtain this result.

Live GPU-node/systemd integration remains untested for this change.

## Checklist

- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added separate configuration options for Kubernetes API server and kubelet authentication.
  - Supports bearer tokens, client certificates, exec providers, and authentication plugins.
  - Preserves in-cluster authentication when explicit configuration is omitted.
  - Refreshes kubelet credentials between requests and retries.

- **Bug Fixes**
  - Improved validation for incomplete credentials, unsafe endpoints, invalid certificates, and missing token files.
  - Clarified permissions required to update pod GPU annotations.

- **Documentation**
  - Added guidance and an architecture decision record covering authentication, TLS, permissions, and failure handling.


metadata-collector/pkg/mapper/grpcclient.go
```diff
@@ -44,34 +45,42 @@ type kubeletGRPClient struct {
...
+// NewKubeletGRPClient connects lazily to the local PodResources socket.
+// Each request waits a bounded time for readiness. Cancellation closes the connection.
 func NewKubeletGRPClient(ctx context.Context) (KubeletGRPClient, error) {
-	_, err := os.Stat(podResourcesKubeletSocket)
-	if err != nil {
-		return nil, err
-	}
-
-	resolver.SetDefaultScheme("passthrough")
+	return newKubeletGRPClient(ctx, podResourcesKubeletSocket)
```
lalitadithya, line 54: nit: do we need this one liner function; can we instead just keep the older `NewKubeletGRPClient` since the function parameters and return hasn't changed?

metadata-collector/pkg/mapper/httpsclient.go
```diff
@@ -63,16 +64,44 @@ type kubeletHTTPSClient struct {
...
+		httpRoundTripper: transport,
+		listPodsURI:      strings.TrimRight(config.Host, "/") + "/pods",
+		listPodsBackoff:  defaultListPodsBackoff,
+		listPodsTimeout:  defaultListPodsTimeout,
+	}, nil
+}
+
+// NewKubeletHTTPSClient retains the in-cluster endpoint, token, and TLS behavior.
+// Requests use ctx for cancellation throughout the client's lifetime.
 func NewKubeletHTTPSClient(ctx context.Context) (KubeletHTTPSClient, error) {
```
lalitadithya, line 105: I think we should add the kubeconfig param to this function. Having two functions with the same name but with different parameters tends to cause confusion

lalitadithya (commented):
thank you for putting this together! added some style comments so that the interfaces come out a bit cleaner, other than that no other concerns from me

metadata-collector/pkg/mapper/httpsclient.go
```diff
@@ -63,16 +64,44 @@ type kubeletHTTPSClient struct {
...
 	listPodsBackoff   wait.Backoff
 	listPodsTimeout   time.Duration
 }
 
-// NewKubeletHTTPSClient creates an HTTPS client configured to communicate with the local
-// kubelet. The provided ctx is used for the lifetime of list-pods requests made through the client.
+// newKubeletHTTPSClient reads the kubelet endpoint, trust, and credentials from an explicit
+// kubeconfig. An empty path retains KUBELET_HOST and projected ServiceAccount token behavior.
+// Requests use ctx for cancellation throughout the client's lifetime.
+func newKubeletHTTPSClient(ctx context.Context, kubeconfigPath string) (KubeletHTTPSClient, error) {
```
natherz97: It would be great if we could use 1 constructor for the KubeletHTTPSClient and only change construction of the transport field. 

Additionally, can we remove the custom staticBearerToken + bearerTokenPath fields and construct a transport which directly sets these fields? Something like this so that we don't have to manually set the Authorization header below:
```
      &rest.Config{
              Host:            ...
              BearerTokenFile: bearerTokenPath,
              TLSClientConfig: ...
      }
```

metadata-collector/pkg/mapper/mapper.go
```diff
@@ -77,6 +103,56 @@ func NewPodDeviceMapper(ctx context.Context) (PodDeviceMapper, error) {
...
+
+	if !hasClientCredentials(config) {
+		return fmt.Errorf("must provide client credentials")
+	}
+
+	return nil
+}
+
+// hasClientCredentials permits client-go's static, file-based, and plugin credentials.
+func hasClientCredentials(config *rest.Config) bool {
```
natherz97, line 151: Do we need to check config.Username? I don't believe this would be relevant for apiserver nor kubelet authentication since we typically only see bearer token, certificate, or provider-based authentication.

Also it looks like we're checking for a client certificate with CertFile/CertData but we're not checking for ClientKey/ClientKeyData?

metadata-collector/pkg/mapper/grpcclient.go
sulixu: newKubeletGRPClient(ctx, socketPath) lets tests use a temporary socket

so test can use this private constructor

metadata-collector/pkg/mapper/mapper.go
sulixu: you are right, will remove username, and this helper func only checks that credentials are configured, will update

natherz97 (approved):

PR #1614, opened 2026-08-17
feat(fault-quarantine): remove node.status from fault-quarantine informer cache

XRFXLP:
## Summary

Strips `node.Status` from the FQ node informer cache using `SetTransform`. The informer receives full node objects from the API server but discards the status field before caching, so only `metadata` and `spec` are held in memory.

FQ rule evaluation uses only `metadata` and `spec` fields — `node.Status` is never read by any rule predicate. Stripping it at the cache layer has no functional impact.

### Memory impact at 99k nodes (scale benchmark)

| Build | Working set | Per-1k-nodes |
|---|---|---|
| v1.18.0 — no stripping | 3,909 MiB (3.8 GiB) | 38 MB |
| This PR — `SetTransform` strips `node.Status` | **864 MiB (0.84 GiB)** | 7.2 MB |
| **Reduction** | **4.5×** | |

> Measured on a KWOK-dominated cluster (99,373 KWOK nodes + 5 EC2 nodes). Production nodes carry 35+ Status conditions from NVSentinel health monitors, Osmo, and Skyhook — the absolute saving on production hardware is expected to be larger.

### Reference: benchmark KWOK node object

The cluster nodes are KWOK fakes enriched with realistic payloads. A representative node JSON (after dropping `managedFields`) is **13.7 KB**:

- `metadata`: 10 labels, 9 annotations — **1.5 KB**
- `status`: 4 conditions, 47 images, capacity/allocatable/nodeInfo — **12.2 KB (89% of total)**
- `spec`: taints only — included in the 1.5 KB above

After `SetTransform` strips `node.Status`, the cached object shrinks from **13.7 KB → 1.5 KB per node** (9×). This directly drives the 4.5× working-set reduction observed.

<details>
<summary>Example KWOK node JSON (images trimmed to 3 of 47)</summary>

```json
{
  "kind": "Node",
  "apiVersion": "v1",
  "metadata": {
    "name": "kwok-node-090515",
    "uid": "66f1177f-2e1c-4ccc-a0d2-16bd451bc4b7",
    "resourceVersion": "36654412",
    "creationTimestamp": "2026-07-27T04:40:55Z",
    "labels": {
      "k8saas.nvidia.com/cordon-by": "NVSentinel",
      "k8saas.nvidia.com/cordon-reason": "GPU-fatal-error-ruleset",
      "k8saas.nvidia.com/cordon-timestamp": "2026-07-28T07-05-14Z",
      "kwok.x-k8s.io/node": "fake",
[...]

PR #1842, opened 2026-09-15
perf(fault-quarantine): prune the node cache to the keys the rules read

### LINKED ISSUE #1810: [Feature]: Derive fault-quarantine's node cache transform from the CEL rules
### Prerequisites

### Code of Conduct

### Feature Summary

fault-quarantine's node informer transform is `stripNodeStatus` (`fault-quarantine/pkg/informer/node_informer.go:121`): it clears `node.status` and keeps labels, annotations and spec whole. The CEL rules are config loaded at startup, so the keys they read are knowable before the first object is cached -- the #1720 argument applied to fault-quarantine. Derive the retained key set from the rule ASTs, union it with the keys the reconciler reads directly, drop the rest.

### Problem/Use Case

fault-quarantine holds 17.88 GB at 100,005 nodes, 161,516 B per node -- second only to kubernetes-object-monitor. Measured on a real EC2 node from the fleet:

| Part of the object | Bytes | Map entries |
| --- | --- | --- |
| `status` | 14,388 | -- |
| `metadata.labels` | 5,628 | 102 |
| `metadata.annotations` | 3,238 | 14 |
| identity and spec | 480 | -- |
| total | 23,734 | |
| **retained after `stripNodeStatus`** | **9,346** | **116** |

9,346 retained bytes become 161,516 held, a 17x multiplier that comes from per-entry map and interface overhead. So map entries, not bytes, predict the saving -- and the current transform removes 60% of the bytes while leaving 116 of the entries.

The shipped ruleset reads two label keys and no annotations (`distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml:108`):

```
!('k8saas.nvidia.com/ManagedByNVSentinel' in node.metadata.labels && node.metadata.labels['k8saas.nvidia.com/ManagedByNVSentinel'] == "false") &&
!('nvsentinel.dgxc.nvidia.com/managed' in node.metadata.labels && node.metadata.labels['nvsentinel.dgxc.nvidia.com/managed'] == "false")
```

Both keys are absent on this node, which is what an opt-out label looks like -- it exists only on nodes someone deliberately excluded. So 102 of 102 label entries and 13 of 14 annotation entries are cached for nothing.

### Proposed Solution

Walk each enabled Node rule's AST, collect the literal keys read under `node.metadata.labels` and `node.metadata.annotations`, union with the fixed set below, and prune the rest in the transform. `NewNodeRuleEvaluator` already compiles every expression (`fault-quarantine/pkg/evaluator/rule_evaluator.go:139`), so the ASTs exist before the informer starts. The node variable is `cel.AnyType`, so `env.Check` validates no field paths; extraction walks the parsed tree.

This also shrinks per-evaluation cost: `getNode` runs `ToUnstructured` on the whole node on every rule evaluation (`rule_evaluator.go:229`).

#### Reusing the kubernetes-object-monitor machinery

`pkg/cel/fields.go` (`ResourceFieldPaths`, `LookupTargets`) and the `fieldTree` prefix tree in `pkg/initializer/transform.go` are generic over path shape and reusable. Three blockers:

1. **`ResourceVar = "resource"` is a package constant**, used by `isResourceVar` and the comprehension shadowing in `pushBinding`/`popBinding`. fault-quarantine binds `node`. Parameterise it.
2. **The transform is unstructured-only.** `newFieldPruningTransform` asserts `*unstructured.Unstructured` and passes structured objects through untouched, so on fault-quarantine's typed `SharedInformerFactory` it is a silent no-op. It needs a typed transform rewriting `node.Labels` and `node.Annotations` alongside the existing `node.Status = v1.NodeStatus{}` -- simpler than the generic `prune`, since only two maps are in play.
3. **`in` defeats the extractor.** Running the shipped predicate through `ResourceFieldPaths` returns `paths=[[metadata labels] [metadata labels k8saas.nvidia.com/ManagedByNVSentinel]]`. A recorded path covers its whole subtree, so every label is retained. `operators.Index` is the only operator `resolveChain` special-cases (`pkg/cel/fields.go:330`); `operators.In` falls through to `walkChildren`, which records the map operand whole. Fix: treat `<string literal> in <chain>` as a key access on that chain. Sound, because a map pruned to key `k` answers `'k' in m` identically.

Without blocker 3 the change still prunes annotations, but leaves the larger half untouched:

| After `stripNodeStatus` | Map entries | Bytes |
| --- | --- | --- |
| retained today | 116 | 9,346 |
| retained, extractor as-is | 103 | 6,453 |
| retained, with the `in` fix | ~1-3 | ~825 |

#### Keys the transform must retain regardless of the CEL

`getNode` reads the informer cache on the live path and only bypasses it during recovery (`rule_evaluator.go:203-217`). Two other consumers read the same cached object: `getNodeQuarantineAnnotations` (`fault-quarantine/pkg/reconciler/reconciler.go:2149`) pulls eight named annotation keys, and the informer's `quarantineAnnotationIndexFunc` indexer (`node_informer.go:100`) reads the same. The cordon path reads `spec.unschedulable` and `spec.taints`.

So the retained set is the CEL-derived keys plus: `quarantineHealthEvent`, `quarantineHealthEventAppliedTaints`, `quarantineHealthEventAppliedLabels`, `quarantineHealthEventIsCordoned`, `quarantineHealthEventCordonPreExisting`, `quarantinedNodeUncordonedManually`, `quarantinedNodeIsUntaintedManually`, `quarantineValidationHealthEvent`, `spec`, and the identity fields the informer keys on.

Dropping one of these fails quietly: fault-quarantine reads an empty quarantine annotation, concludes there is no quarantine in progress, and re-cordons a node it already handled. Only on already-quarantined nodes, which a fresh test cluster has none of.

#### How a rule must be written for pruning to hold

1. **Literal keys only.** `labels['nvidia.com/gpu.count']` records one key. A computed key, or one built with `ext.Strings()`, retains the whole map.
2. **Map-level operations retain the whole map.** `size(labels)`, `labels.exists(k, ...)`, or passing the map as a map. One such term in one rule disables pruning for that map across the ruleset.
3. **`has()` on the map itself changes meaning under pruning.** After pruning, `has(node.metadata.annotations)` is true iff some retained key survived, so its value depends on the transform rather than the node. lifecycle-manager's rule does this today. Test the key, not the map.
4. **An absent key is an error, not `false`.** `labels['k']` raises `no such key`, which `Evaluate` wraps as a `coldstart.PermanentError` -- failed permanently, not retried. `cel.OptionalTypes()` is not enabled, so `labels[?'k']` is unavailable, and `has(labels.k)` cannot express a key containing `/` or `.`. Every read needs an `in` guard, which is why blocker 3 must be fixed rather than worked around. Pre-existing hazard: `int(node.metadata.labels['nvidia.com/gpu.count'])` in `distros/kubernetes/nvsentinel/charts/labeler/values.yaml:68` is unguarded.
5. **Positive and negative rules fail in opposite directions.** A positive rule (`'k' in labels && labels['k'] == 'true'`) that loses its key stops matching: a missed cordon, silent, no wrong action. A negative rule -- the shipped opt-out form, `!('k' in labels && labels['k'] == "false")` -- that loses its key flips to true, and fault-quarantine cordons a node whose owner opted it out. Pruning failure is fail-open on exactly the nodes the label protects, and every shipped rule is of this form.

   Two consequences: extraction must retain the whole map for any shape it does not recognise, never an empty set; and startup must log the retained key set per map, as #1720 does.
6. **Rules are read once, at startup.** A ruleset edited in the ConfigMap without a restart runs against a cache pruned for the old rules, which is case 5. Rebuild the transform and resync on config change, or require a restart.

#### Scope

Recovery is unaffected -- the cold-start path reads live through `GetNodeDirect`, so a pruning bug can only surface on the live path. A regression test should drive the component from outside: ruleset and cluster state in, cordoned nodes and published events out, asserting nothing about which fields survived the transform.

Refs #1720

### Component

Fault Management

XRFXLP:
## Summary

Fixes #1810.

[#1614](https://github.com/NVIDIA/NVSentinel/pull/1614) took the first step on Fault Quarantine's informer cache: it cleared `node.status` and deliberately kept the rest, leaving the generality the CEL rule sets expose for later. This PR does that part, and keeps only the fields the rules read plus a fixed set of keys Fault Quarantine hardcodes its behaviour on.

The numbers say why the first step was not enough. Fault Quarantine holds 17.88 GB of node cache at 100,005 nodes, or 161,516 bytes per node. On a measured production node `status` is 14,388 bytes of 23,734, so clearing it drops 60% of the bytes, yet the 9,346 that remain still cost 161,516 bytes held. **Map entries, not bytes, predict the saving.** That node carried 116 of them — 102 labels and 14 annotations — while the shipped rule set reads two label keys and no annotations, and both label keys are normally absent, since an opt-out label exists only on nodes an operator excluded. So 102 of 102 label entries and 13 of 14 annotation entries were cached for nothing.

Rules are configuration loaded at startup, so the keys they read are knowable before the first object is cached. This change derives those keys from the Node rules' CEL, unions them with the keys Fault Quarantine reads or removes on its own account, and prunes the cached maps to that set. Where an expression does not name a field precisely, the cache falls back to the immediate parent and keeps that subtree whole:

```text
node.metadata.labels['nvidia.com/gpu.present']               ->  that one key
node.metadata.labels[node.spec.nodeName]                     ->  metadata.labels (whole) + spec.nodeName
size(node.metadata.labels) > 0                               ->  metadata.labels (whole)
node.metadata.labels.exists(k, k.startsWith('nvidia.com/'))  ->  metadata.labels (whole)
node.status.conditions.exists(c, c.type == 'Ready')          ->  status.conditions (whole)
size(node), or an expression that does not parse             ->  cache the node whole
```

A retained path stands for the whole subtree beneath it, which is what makes comprehensions safe without per-element handling. Rule sets are walked whether or not they are enabled, which over-retains for a disabled set but cannot make a rule evaluate against a dropped key. Because two rules that select the same nodes can prune differently, this is now documented for operators in `docs/fault-quarantine.md`.

### Reusing what already existed

Kubernetes Object Monitor already derives cache fields from policy CEL, so the AST walk existed and was tested. Two things had to change for Fault Quarantine to share it:

- The walk moves to `commons/pkg/celfields`, parameterised on the bound variable name, since Kubernetes Object Monitor binds `resource` and Fault Quarantine binds `node`. `environment.go` stays put, because `resource`, `now` and `lookup()` are that monitor's vocabulary rather than shared. Its derived fields are byte-identical before and after the move.
- `'k' in <chain>` now records a read of key `k` rather than the whole map. A map pruned to `k` answers `'k' in m` exactly as the unpruned map does, and this is not cosmetic: retained entries drop from 103 to roughly 1-3. It is also what makes guarded rules expressible, since an absent key raises `no such key` and every safe read of an optional label needs an `in` guard.

Fault Quarantine's own transform is typed rather than unstructured, so it could not reuse `newFieldPruningTransform`, which asserts `*unstructured.Unstructured` and would pass typed objects through untouched.

## FQ specific handling

Fault Quarantine reads and writes keys no expression mentions, so these are retained for every node whatever the rules say.

**Annotations** — all eight quarantine keys: `quarantineHealthEvent`, `quarantineHealthEventAppliedTaints`, `quarantineHealthEventAppliedLabels`, `quarantineHealthEventIsCordoned`, `quarantineHealthEventCordonPreExisting`, `quarantinedNodeUncordonedManually`, `quarantinedNodeUntaintedManually` and `quarantineValidationHealthEvent`. That list moves out of `getNodeQuarantineAnnotations` into `common.QuarantineAnnotationKeys`, because the cache must retain exactly what the reconciler reads: with two copies, a key added to one would vanish from the cache, and Fault Quarantine would conclude no quarantine is in progress and re-cordon a node it had already handled.

**Labels**

- `dgxc.nvidia.com/nvsentinel-state`, the node state label.
- The GPU node label from `--gpu-node-label-key`. `GetNodeCounts` selects on it to size the circuit breaker's denominator, and a label selector runs against the cached object, so a pruned key would make the breaker see an empty fleet.
- The six cordon and uncordon bookkeeping labels built from the configured `label-prefix`.
- Every `label.key` configured on a rule set.

**Labels read per node** — a rule set's label is recorded in the applied-labels annotation, and unquarantine reads it back to remove it. Configured labels are already retained, so the per-object read covers labels configuration no longer mentions: a rule set renamed or removed since a node was quarantined leaves a label whose key survives only on the node, and cleanup still has to remove it. Only an already-quarantined node carries the annotation, so the unquarantined majority that drives the 17.88 GB still prunes fully.

`spec` and the identity metadata stay whole: `spec` is 480 bytes of 23,734, and the cordon path, manual-untaint detection and pre-existing-taint marking all read it.

This is also why the evaluator is touched. Pruning can empty a map, and `Labels` and `Annotations` carry `omitempty`, so the conversion for CEL would drop the field entirely and an absent map raises `no such key` instead of evaluating the guard as false. `restorePrunedMetadataMaps` restores an emptied map before evaluation, and only one the node actually had, so a rule asking whether a node has any labels at all still reads the node rather than the cache.

## Two ways dropping a key could cordon a healthy node

Kubernetes Object Monitor only reads its cache. Fault Quarantine writes, and the cached node is the base for the write, so a dropped key changes what it *does*, not only what it sees. Both cases below can take a healthy node out of service, and the design is built around them.

**A dropped key makes an opt-out rule match.** A positive rule that loses its key stops matching: a missed cordon, silent but harmless. A negative rule that loses its key evaluates to true — and every shipped rule is negative, of the form `!('k' in labels && labels['k'] == "false")` — so Fault Quarantine would cordon a node whose owner had explicitly opted it out. Pruning therefore errs in one direction only: the derivation keeps the whole map for any shape it does not recognise and never produces an empty set, the zero value of `Keys` keeps everything, and every path that fails to derive a set lands there.

**A dropped key makes a deletion silently do nothing.** `UpdateNode` patches against a deep copy of the cached node, and `removeLabels` deletes from that copy. Deleting a key pruning already removed changes nothing, so the diff comes out empty and no `null` is emitted: the label survives on the live node while the log reports it removed. The retained set therefore covers every key Fault Quarantine can **remove**, not only every key it reads, which is why the state label and the cordon bookkeeping labels are retained though nothing reads them from the cache.

The retained set is logged per map at startup, so an operator who edits a rule set can see the effect without reading the code. Recovery reads live through `GetNodeDirect` rather than the cache, so it sees an opt-out the cache does not hold.

### Compatibility 

No compatibility or functionality surface is touched

## Measured on the scale cluster

10,005 KWOK nodes, one replica, no resource limits, no event load, fresh pod per arm, working set read per pod.

| Arm | Peak | Settled | Bytes/node |
| --- | --- | --- | --- |
| main | 1.758 GB | 1.758 GB | 175,671 |
| this branch | 0.377 GB | 0.373 GB | 37,257 |

**4.7x less memory, 138 KB per node.** The peak falls as far as the steady state, since the transform runs as each object enters the cache and the initial LIST never accumulates whole nodes, so request and limit both come down.

Derivation on the deployed ruleset retains 9 labels and 8 annotations — the one Node rule's label, `nvidia.com/gpu.present` for the circuit breaker's denominator, and the cordon keys FQ must be able to remove.

Rulesets used for benchmarking:
```
Data
====
config.toml:
----
label-prefix = "k8saas.nvidia.com/"

[circuitBreaker]
percentage = 50
duration = "5m"

[[rule-sets]]
  enabled = true
  version = "1"
  name = "GPU fatal error ruleset"
  priority = 0

  [[rule-sets.match.all]]
    kind = "HealthEvent"
    expression = "event.agent == 'gpu-health-monitor' && event.componentClass == 'GPU' && event.isFatal == true"

  [[rule-sets.match.all]]
    kind = "Node"
    expression = "!('k8saas.nvidia.com/ManagedByNVSentinel' in node.metadata.labels && node.metadata.labels['k8saas.nvidia.com/ManagedByNVSentinel'] == 'false')"

  [rule-sets.cordon]
    shouldCordon = true
```

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [x] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Health Monitor
- [ ] Core Service
- [ ] Fault Management
- [ ] Janitor
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Documentation/CI
- [ ] New Component
- [x] Multiple Components
- [ ] Other

## Testing
- [x] Tests pass locally

The whole reconciler E2E suite now builds its informer from derived keys, so it runs against a real pruned cache and any read or removal it already covers that the retained set misses fails there. Two black-box tests were added on top, driving the component from outside: an opted-out node carrying 100 labels is never cordoned, and a node that did not opt out still is, the second keeping the first from passing vacuously. That pair was mutation-checked by dropping the CEL-derived keys, and failed as intended.

Unit tests pin the fail-safe direction: map-level reads, comprehensions, runtime-computed keys, whole-object reads and unparseable expressions each retain the whole map, and the removable keys are asserted to survive pruning. Three further tests cover the edges — a node pruned to empty maps still evaluates the shipped opt-out guards, a node with genuinely no annotations still reports them absent, and a recovery read honours an opt-out the pruned cache does not hold. The last was mutation-checked by disabling the recovery branch.

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Node caches now retain only labels and annotations required by configured rules and operations.
  * Quarantine metadata and dynamically applied labels remain available after cache filtering.
  * CEL field and lookup analysis supports configurable identifiers.

* **Bug Fixes**
  * Improved opt-out and quarantine decisions for nodes with many labels.
  * Preserved individual metadata entries when generating node updates.
  * Added safer fallback behavior when rules or metadata cannot be analyzed.

* **Documentation**
  * Documented cached-node field retention, memory impact, and map-key behavior.


fault-quarantine/pkg/nodecache/nodecache.go
```diff
@@ -0,0 +1,386 @@
...
+func (k *Keys) retainAllLabels() {
+	k.pruneLabels = false
+}
+
+func (k *Keys) retainAllAnnotations() {
+	k.pruneAnnotations = false
+}
+
+// log reports what the cache will keep. An operator who edits a ruleset needs
+// to see the effect without reading the code.
```
lalitadithya: nit: can we frame this a bit better: maybe just say "log the paths the cache will keep"

docs/fault-quarantine.md
```diff
@@ -116,6 +116,48 @@ Rules are defined using rulesets that evaluate CEL expressions. Each ruleset has
 - **Label Prefix**: Customize the prefix for tracking labels and annotations on nodes
 - **Multiple Rulesets**: Define different rules for different failure types with CEL expressions that access either the health event or Node metadata/spec
 
+#### Cached Node Fields
```
lalitadithya, line 119: can we link to this in https://github.com/NVIDIA/NVSentinel/blob/main/distros/kubernetes/nvsentinel/charts/fault-quarantine/values.yaml so that people who want to update the rules know that they have some stuff they need to read though for guidance?

fault-quarantine/pkg/nodecache/nodecache.go
XRFXLP: done

docs/fault-quarantine.md
XRFXLP: linked

KaivalyaMDabhadkar (dismissed):

fault-quarantine/pkg/evaluator/rule_evaluator.go
```diff
@@ -231,11 +231,46 @@ func (nm *NodeRuleEvaluator) getNode(ctx context.Context, nodeName string) (map[
...
+func restorePrunedMetadataMaps(unstructuredNode map[string]any, node *corev1.Node) {
+	metadata, ok := unstructuredNode["metadata"].(map[string]any)
+	if !ok {
+		return
+	}
+
+	restore := func(field string, source map[string]string) {
+		if source == nil {
+			return
+		}
```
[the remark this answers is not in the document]
XRFXLP: should've been fixed.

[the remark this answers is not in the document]
XRFXLP: isn't that's what supposed to happen? if Labels == nil then metadata.Label need to be omitted because labels are not htere?

lalitadithya (approved):

PR #1843, opened 2026-09-15
fix(fault-management): report datastore connectivity in readiness probes

### LINKED ISSUE #1839: [Bug]: readiness probes in fault-quarantine, node-drainer and fault-remediation cannot report datastore loss
### Prerequisites

### Code of Conduct

### Bug Description

The readiness probe for `fault-quarantine`, `node-drainer` and `fault-remediation` always succeeds. It returns HTTP 200 without checking anything, so `ready=True` does not mean the component reached the event store.

The probe endpoints are constants. `fault-quarantine` and `node-drainer` serve readiness from `/healthz`, which `commons/pkg/server.WithSimpleHealth()` registers as an unconditional 200. `fault-remediation` serves readiness from `/readyz`, registered as `healthz.Ping`, which controller-runtime defines as `func(_ *http.Request) error { return nil }`. For `fault-quarantine` and `node-drainer` the liveness and readiness probes name the same endpoint, so the two signals are identical.

The HTTP server also binds before the datastore connection starts. `node-drainer` and `fault-remediation` do this so a slow MongoDB initialization cannot trip the liveness probe. `fault-quarantine` gets the same ordering because `InitializeAll` does not touch the datastore — `datastore.NewDataStore` runs inside `Reconciler.Start`, concurrent with the metrics server.

A pod therefore reports Ready while it processes no events. During the 100k-node scale test this produced two false recovery calls.

`commons/pkg/server` already defines `HealthChecker` and `ReadinessChecker`, with `WithHealthCheck` and `WithReadinessCheck` to register them. No component wires a datastore check into either.

### Component

Fault Management

### Steps to Reproduce

1. Deploy NVSentinel with the MongoDB store.
2. Block the pods' egress to MongoDB: apply a `NetworkPolicy` denying the MongoDB port, or scale the replica set to zero.
3. Restart `fault-quarantine`, `node-drainer` and `fault-remediation`.
4. Run `kubectl get pods` and read the READY column. The containers report `1/1`.
5. Run `kubectl logs` and confirm `Successfully pinged database to confirm connectivity` never appears.
6. Publish a health event. No component acts on it.

### Environment

- NVSentinel version: `main` (reproduced against the scale-test build)
- Kubernetes version: 1.34
- Deployment method: Helm, `distros/kubernetes/nvsentinel`

### Logs/Output

```text
# Pod is Ready with no datastore connection:
$ kubectl exec <node-drainer-pod> -- curl -s -o /dev/null -w '%{http_code}' localhost:2112/healthz
200

# The connectivity confirmation never logs:
$ kubectl logs <node-drainer-pod> | grep 'Successfully pinged database'
(no output)
```

### Suggested fix

Keep liveness trivial so slow initialization cannot kill the pod. Make readiness report the real state: give each component a `ReadinessChecker` that fails until the datastore ping succeeds and the change stream is open, register it with `server.WithReadinessCheck`, and add the equivalent `AddReadyzCheck` to `fault-remediation`. A `datastore_connected` gauge alongside it lets alerts catch the condition without polling a probe.

Two decisions for the reviewer. Meaningful readiness makes these pods `NotReady` during initialization, which changes Deployment rollout behaviour, so confirm nothing keys off them being Ready. The check also runs every 10 seconds against a store already carrying the whole fleet's event load, so it has to stay cheap.

### Not covered

Unexpected change-stream death is already handled — all three components exit when the event channel closes, added in `7771e9b8`. This issue is only about the readiness signal.

HesandaLiyanage:
Fixes #1839

### Motivation

Currently, the readiness probes for `fault-quarantine`, `node-drainer`, and `fault-remediation` always report HTTP 200 without verifying connectivity to the event store:
- `fault-quarantine` and `node-drainer` serve readiness from `/healthz`, which `server.WithSimpleHealth()` registers as an unconditional 200.
- `fault-remediation` serves readiness from `/readyz`, registered as `healthz.Ping`, which controller-runtime defines as `func(_ *http.Request) error { return nil }`.
- In all three components, the HTTP / metrics server binds before datastore initialization completes to protect liveness probes against slow database initialization. Consequently, pods report `Ready=True` even when completely disconnected from MongoDB/PostgreSQL or still attempting to initialize.

### Changes

1. **`DatastoreReadinessChecker` in `commons/pkg/server`:**
   - Implements `server.ReadinessChecker` (`Ready(ctx)`), controller-runtime `healthz.Checker` (`Check(req)`), and `prometheus.Collector`.
   - Uses `lagstate.Provider.LagState()` (introduced in ADR-054) to determine readiness without issuing extra active database pings on 10s kubelet probe intervals.
   - Reports not ready (HTTP 503) while the watcher is unassigned or before the initial poll completes (both `lastEmptyBatch` and `lastEventRead` are zero).
   - Reports ready (HTTP 200) once affirmative evidence of completing a batch/poll is recorded.
   - Exports the `datastore_connected` gauge metric (1 when ready, 0 when not ready) evaluated at scrape time. Handles duplicate registrations gracefully by returning the existing collector.
2. **Component Integration:**
   - **`node-drainer`:** Registers `/readyz` via `server.WithReadinessCheck` and connects `components.EventWatcher`.
   - **`fault-quarantine`:** Registers `/readyz` via `server.WithReadinessCheck` and connects the watcher inside `Reconciler.Start()`.
   - **`fault-remediation`:** Replaces `healthz.Ping` on `/readyz` with `checker.Check`, and registers `datastore_connected` on `crmetrics.Registry`.
3. **Helm Charts:**
   - Updates `charts/fault-quarantine/templates/deployment.yaml` and `charts/node-drainer/templates/deployment.yaml` to point `readinessProbe.httpGet.path` to `/readyz` (keeping `livenessProbe` on `/healthz`). `fault-remediation` already targets `/readyz`.
4. **Unit Tests:**
   - Unit tests covering lifecycle transitions, unwrapped watchers, server integration, duplicate metric registration, and controller-runtime registry export across `commons`, `node-drainer`, `fault-quarantine`, and `fault-remediation`.

---

#### 1. Baseline on `main` (Silent Failure Behavior)
On `main`, readiness probes for `fault-quarantine`, `node-drainer`, and `fault-remediation` query unconditional 200 endpoints (`/healthz` or `healthz.Ping`):
- **Startup with Datastore Down:** When datastore egress is blocked (e.g. NetworkPolicy or MongoDB offline), the HTTP server binds immediately and serves `200 OK` on `/healthz`. Kubelet marks the pod `READY: 1/1` and `STATUS: Running`.
- **Silent Failure:** In the background, `datastore.NewDataStore` fails or hangs attempting to connect. Because Kubernetes considers the pod healthy (`1/1 Ready`), deployment rollouts proceed, traffic/events are silently dropped, and `!Ready` alerts never fire.
- **Restart / Crash Loop Masking:** When FQ eventually crashes due to connection failure (`os.Exit(1)`) and is restarted by Kubelet, it immediately reports `1/1 Ready` again on `/healthz` during startup until the next crash. This masks the underlying datastore failure and presents a false recovery signal to the cluster.

#### 2. Behavior with this PR (Active Crash & Explicit Failure Detection)
With this PR, readiness is decoupled from liveness and bound to `/readyz` via `DatastoreReadinessChecker` and `LagState()`:
- **Startup with Datastore Down:** `/readyz` immediately returns `HTTP 503 Service Unavailable` (`datastore watcher initializing` / `datastore watcher awaiting first poll`).
- **Active Failure:** The pod remains `READY: 0/1`. Kubernetes knows the pod cannot process work:
  - Deployment rollouts halt immediately rather than replacing healthy pods with non-functional replicas.
  - Metrics scrape reports `datastore_connected == 0`, alerting operators immediately.
- **Restart / Crash:** When FQ restarts while the datastore is unavailable, the pod is never marked `Ready` (`0/1`), transforming the previous silent failure into an active, observable failure state that accurately reflects the component's crash loop / datastore loss.
- **Recovery:** As soon as datastore connectivity is established and the watcher records its first batch (`lastEmptyBatch` or `lastEventRead`), `/readyz` transitions to `HTTP 200 OK` and `datastore_connected` flips to `1`.

#### 3. Test Coverage Summary
- **`commons/pkg/server/datastore_readiness_test.go`**:
  - `TestDatastoreReadinessChecker_Lifecycle`: Verifies transition from unassigned provider (503, metric=0) -> zero timestamps (503, metric=0) -> first empty batch observed (200, metric=1) -> event read observed (200, metric=1).
  - `TestDatastoreReadinessChecker_SetWatcher`: Verifies unwrapping wrappers to extract `lagstate.Provider`.
  - `TestDatastoreReadinessChecker_ServerIntegration`: Verifies `/healthz` returns 200 (liveness OK) while `/readyz` returns 503 until datastore connection completes, then flips to 200.
  - `TestDatastoreReadinessChecker_DuplicateRegistration`: Verifies existing collector reuse on shared registry.
- **`node-drainer/main_test.go`**: Verifies `/readyz` endpoint integration and metric registration on drainer server.
- **`fault-quarantine/pkg/reconciler/readiness_test.go`**: Verifies `Reconciler.Start` connects the watcher to the readiness checker upon change stream start.
- **`fault-remediation/pkg/initializer/lag_metrics_test.go`**: Verifies controller-runtime `/readyz` integration and `datastore_connected` collector export.

## Summary by CodeRabbit

- **New Features**
  - Added datastore-aware readiness checks for fault quarantine, fault remediation, and node drainer services.
  - Added the `datastore_connected` Prometheus metric.
  - Readiness now reflects datastore watcher connectivity.

- **Bug Fixes**
  - Kubernetes readiness probes now use `/readyz` for fault quarantine and node drainer.
  - Improved KWOK node creation validation across ARM64 and AMD64 environments, including clearer failure handling.

- **Tests**
  - Added coverage for datastore readiness, health endpoints, metrics, and watcher connectivity.


XRFXLP:
Can you please add the details of the testing around the issue -- silent failures in the PR description?

This should include a scenerio where FQ that were built on top of main commit would exhibit the silent failure behaviour -- pod shows Running but crashes when restarted and how this PR changes to more of an active crash?

HesandaLiyanage:
> Can you please add the details of the testing around the issue -- silent failures in the PR description?
> 
> This should include a scenerio where FQ that were built on top of main commit would exhibit the silent failure behaviour -- pod shows Running but crashes when restarted and how this PR changes to more of an active crash?

I have updated the PR description

lalitadithya:
can you please share why we need changed to the CI/Tilt setup? I'm a little concerned that this might be some regression that we are missing since the same works on main without any issues

HesandaLiyanage:
> can you please share why we need changed to the CI/Tilt setup? I'm a little concerned that this might be some regression that we are missing since the same works on main without any issues
@lalitadithya 
Sorry for the late reply! You're completely right to call this out.

The earlier Makefile change was a misguided attempt to work around a timeout caused by our initial readiness probe logic (awaiting first poll), which deadlocked whenever event consumption was paused (e.g. during circuit breaker startup/tripped states).

This is what i did 

  1. Fixed the readiness probe: DatastoreReadinessChecker.Ready now simply checks that the
  datastore watcher is connected (provider != nil), as requested in #1839.
  2. Reverted Makefile: It is now 100% identical to main.
  3. Only remaining script change: tilt/create-kwok-nodes.sh was updated to substitute the host
  architecture (arm64/amd64), matching what tilt/Tiltfile already does.

  Happy to revert the create-kwok-nodes.sh change as well if you prefer to keep this PR strictly
  scoped to the Go probe code.

lalitadithya:
Thanks for the details! 

> Happy to revert the create-kwok-nodes.sh change as well if you prefer to keep this PR strictly
scoped to the Go probe code.

Yeah, that would be preferrable, let's keep this PR strictly for the go probe code.

commons/pkg/server/datastore_readiness.go
```diff
@@ -0,0 +1,147 @@
...
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	// DatastoreConnectedMetricName is the Prometheus metric reporting datastore readiness.
+	DatastoreConnectedMetricName = "datastore_connected"
+
+	datastoreConnectedHelp = "Reports 1 if the datastore watcher has connected, 0 otherwise."
```
XRFXLP: nit: we don't need to declare this as variable if this is used only once.

commons/pkg/server/datastore_readiness.go
```diff
@@ -0,0 +1,147 @@
...
+		Unwrap() any
+	}
+
+	if u, ok := watcher.(unwrapper); ok {
+		c.SetWatcher(u.Unwrap())
+	}
+}
+
+// Ready implements ReadinessChecker. It reports ready once the datastore watcher
+// has established connection.
```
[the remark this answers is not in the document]
HesandaLiyanage: This suggestion is not valid and should not be applied:

  1. Connectivity is already proven: The watcher constructor pings the database and confirms
  connectivity before returning. If the database is unreachable, SetWatcher is never called and
  /readyz stays 503.
  2. First-poll requirement deadlocks CI: In fault-quarantine, event dequeuing is intentionally
  halted when the circuit breaker is in TRIPPED state. Requiring polling timestamps before
  becoming ready keeps /readyz at 503 indefinitely, which deadlocks Kubernetes deployment
  rollouts and fails E2E tests with 10-minute timeouts.

  Checking provider != nil satisfies #1839 without creating startup deadlocks.

commons/pkg/server/datastore_readiness.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if provider, ok := watcher.(LagStateProvider); ok {
+		c.SetLagProvider(provider)
+
+		return
+	}
+
+	type unwrapper interface {
+		Unwrap() any
```
XRFXLP: Don't we need exact signature to satisy the interface? Moreover when this would be executed? As all the current modules would fall through:
```
	if provider, ok := watcher.(LagStateProvider); ok {
		c.SetLagProvider(provider)

		return
	}
```

commons/pkg/server/datastore_readiness.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if u, ok := watcher.(unwrapper); ok {
+		c.SetWatcher(u.Unwrap())
+	}
+}
+
+// Ready implements ReadinessChecker. It reports ready once the datastore watcher
+// has established connection.
+func (c *DatastoreReadinessChecker) Ready(_ context.Context) error {
```
XRFXLP, line 99: this only checks the initalization, right? which was already being caught by the existing mechanism, what if connectivity is healthy at the start but dies down during running?

XRFXLP (changes requested):

commons/pkg/server/datastore_readiness.go
HesandaLiyanage: yes that makes sense. ill remove the unused unwrapper block

commons/pkg/server/datastore_readiness.go
HesandaLiyanage: Fair, but this was actually covered by #1839. Before this change /healthz just returned 200 unconditionally, so if Mongo was unreachable at startup the pod would still show 1/1 Ready without ever processing anything. That's the gap this closes, 503 until SetWatcher actually connects.

For the mid run case, if the connection drops after startup the change stream dies and the component exits fatally (see 7771e9b8), so k8s restarts the pod anyway. If Mongo's still down at that point, readiness just keeps it NotReady on the restart. Didn't want to add active pings every 10s against a store that's already handling the whole fleet's events, so this leans on the existing fatal exit behavior instead of duplicating that check.

XRFXLP (approved):

PR #1845, opened 2026-09-15
feat(platform-connectors): one handler and connector set for both roles, on-change node updates and keyed inserts

KaivalyaMDabhadkar:
## Summary

Part of ADR 052, the deployment platform connector. #1796 added the client library and #1814 the datastore side. This PR prepares the platform-connectors binary for its second role without adding it yet: the request handler and the connectors become one code path that both roles can use, and the connectors and pipeline stages are sized and hardened for a fleet. The deployment role itself (`PC_MODE=deployment`, the `pkg/central` package) comes in the next PR, which builds only on what is here. The wire contract is unchanged: the same `PlatformConnector` gRPC service, the same protobuf.

Most of this is internal, but a few changes are visible on the DaemonSet that runs today. They are listed under "What the DaemonSet sees differently" so they can be judged on their own.

### What changes

- **One connector interface** (`pkg/connectors`): `Connector.ProcessBatch` receives one batch. `Set` hands a batch to several connectors at the same time and fails with the first error, which cancels the others. `BestEffort` wraps a connector whose failure or timeout must not fail the batch; it is logged and counted in `platform_connector_best_effort_failures_total`. A ring buffer implements the interface by queueing the batch and answering at once, so the DaemonSet's handler hands each batch to a `Set` of its ring buffers, as before in effect.
- **One request handler** (`pkg/server`): validate the batch, run the pipeline, hand it to the connector, answer with its result. Outcomes are counted in `platform_connector_request_duration_seconds{outcome}`. The handler joins the trace context a caller sends in the request metadata, and stamps an event without `generatedTimestamp` with the arrival time (the analyzer cannot process an event without one). The payload log level is a field, so the deployment role can log at Debug.
- **Pipeline** (`pkg/pipeline`): `ProcessBatch` runs a whole batch, with an optional batch budget under which transformers that implement `Prewarm` prepare for the batch first (the metadata augmentor reads every node the batch names concurrently) and report what the budget cut short, so the caller can defer the batch before anything is marked. What the preparation read is kept in a per-batch scope, so the transformers use it even if the shared cache dropped it. Without a budget, the DaemonSet's case, the events are processed as they are. The pipeline construction from `config.json` moves into the package.
- **Store connector**: `InsertBatch` writes one batch inside the caller's request and reports `stored` or `duplicate`; `VerifyIdempotencyIndex` checks the idempotency index of #1814 with its full definition. Both roles now insert with the idempotent insert: the deployment stamps keys per caller before the pipeline runs (next PR), the DaemonSet's store loop keys each queued batch itself on its first attempt and keeps the key across requeues. A duplicate on another unique index is answered with a terminal status; an empty batch is a success without a datastore call.
- **Kubernetes connector**: node conditions and Events are written only when a batch changes what the node shows, Event names are derived from the fault, the Event memory is bounded per node and check, a batch is processed in timestamp order on both paths, and `ProcessBatch` lets the deployment role call it inside a request. `NodeEventMemorySize` sizes the memory for a fleet.
- **Dedup**: the tracker is bounded (`maxEntries`, default 100000) and indexed per node and check, so a recovery visits only its own bucket, and it keys on the idempotency key so a resend of the same batch is never mistaken for a repeat of another.
- **Metadata augmentor**: one read per node even when many batches name it at the same time (singleflight instead of one mutex), a read bounded to `lookupTimeout` (3 s) and detached from the caller, `Prewarm` for batches, fleet-sized client rate limit and cache through `pipeline.Options`, and a node label named like the reserved idempotency key field is ignored.
- **Prometheus connector and gRPC sink** gain `ProcessBatch`, so the deployment role can call them per batch.
- controller-runtime's logger is routed through slog, so the certificate watcher's lines appear and its "SetLogger was never called" warning is gone.
- `go.mod` gains `go-grpc-middleware/v2` (indirect, through the client library) and makes `logr` a direct requirement.

### What the DaemonSet sees differently

Deliberate, each small, none changes the wire contract.

- **Node updates only on change.** A batch that leaves every condition's status, reason and message as they are costs no node status write (`k8s_platform_connector_node_condition_update_total{status="skipped"}`), and a repeat of a non-fatal fault whose Event was written less than 10 minutes ago costs no Event write. A same-fault repeat with other diagnostic text does not rewrite the message; a recovery makes the fault's next occurrence a change again. The condition's heartbeat time therefore moves only with a change (nothing in the repository reads it), and Event counts grow at most every 10 minutes instead of on every poll. On a fleet this is the largest steady-state API server load the connector produces.
- **Keyed inserts.** Documents written by the DaemonSet carry `metadata.idempotencyKey` (`instanceID#sequence#index`), a batch retried after a partial write stores only the missing documents instead of the first ones twice, `platform_connector_store_batches_total` exists on its pods, and on PostgreSQL the insert is one row at a time. A key that arrives in an event (a derived event copied from a stored document) is overwritten instead of stored.
- **Node Event names are derived from the fault** (`<node>.<16 hex>` of node, check, reason and message) instead of the write time, so a connector that lost its memory finds the Event it or an earlier life of the same pod wrote, and refreshes it instead of writing a second one. A batch is processed in timestamp order on both the condition and the Event path.
- **Dropped labels**: `node_name` on `k8s_platform_connector_node_event_operations_total` and `node` on the dedup counter, because one deployment replica would carry a series per node of the fleet.
- **Smaller ones**: a node metadata read is bounded to 3 s and detached from the caller (then fails open, as before); an event without `generatedTimestamp` is stamped with the arrival time; an empty batch is a success without a datastore call; trace context sent in the request metadata joins the trace instead of starting a new one; the per-event "Metadata augmented" line is at Debug.

### What is not in this PR

The deployment role itself (`pkg/central`: configuration, TLS listener, token validation, the idempotency-key and readiness interceptors, the no-local-node branch of `pkg/auth`) is the next PR. The Helm chart, the monitors' switch to publishing directly, the Tilt environment and e2e tests, and the operator documentation come after it. Reviewing this PR does not require them.

### Tests

Unit tests, all passing, lint clean, `go mod tidy` clean:

- `pkg/connectors/connector_test.go`: every member gets the batch, members run at the same time, the first error cancels the rest, a best-effort member's failure or timeout is counted and does not fail the batch, the caller's cancellation reaches the members.
- `pkg/server/platform_connector_server_test.go`: acknowledged only when the connector accepts; a connector failure is not acknowledged; rejection happens before any side effect; a cancelled caller is reported as such; the caller's trace is joined; a retry after a failed write keeps the processing strategy; one budget covers the whole batch and a spent metadata budget defers it; queues acknowledge at once and hold the batch; a status the connector chose is passed through; defaults and timestamp stamping.
- `pkg/connectors/kubernetes` (envtest): a repeat costs no status write, also when the message is saturated and compacted; a new fault joining or a partial recovery is a change; Events are written on change and refreshed after the interval; Event names are derived from the fault; a connector without memory refreshes the existing Event; the Event memory is bounded; an update racing a deletion recreates the Event.
- `pkg/connectors/store`: a duplicate-only replay is success, a partial replay stores the missing events, an empty batch is terminal success, a duplicate on another index stays an error and is answered with a terminal status; the ring buffer loop keys the stored copies and keeps the key across requeues; a batch keyed by the caller keeps its keys.
- `pkg/transformers/dedup` and `pkg/transformers/metadata`: key-aware and bounded dedup with an index that stays in step; one read per node under concurrent batches, a slow node does not block others, a spent budget starts no reads and is reported, what the preparation read outlives the shared cache for the batch.
- `pkg/pipeline`, `pkg/ringbuffer`, `pkg/connectors/prom`: batch processing with and without a budget, the batch scope, queueing with the caller's trace, counting without a ring buffer.

Manual testing on kind with the DaemonSet role (every publisher on the socket), on Mongo and on PostgreSQL with certificate auth:

- a repeated XID costs no status write and keeps the condition heartbeat and the node's resourceVersion, while both batches are stored under their own keys; a repeated non-fatal fault is one Event whose count grows only after a recovery;
- an XID published during an 80 s Mongo outage is retried by the store loop and stored exactly once after Mongo returns;
- every document written by the DaemonSet is keyed, on both datastores;
- the upstream e2e tests for dedup, the syslog monitor (XID without metadata, cancellation uncordons the node), the gpu monitor and repeated XIDs pass through the socket; no connector pod restarted or logged an error.

With the deployment role PR applied on top, the same code ran the deployment scenarios (contract, fan-out, dedup, Event names across replicas, rolling restarts, Mongo outages under keyed traffic, the deployment e2e tests) on kind, and an earlier version on four dev clusters.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Health Monitor
- [x] Core Service
- [ ] Fault Management
- [ ] Janitor
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Documentation/CI
- [ ] New Component
- [ ] Multiple Components
- [ ] Other

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added configurable limits for deduplication memory, Kubernetes metadata caching, lookup timeouts, and client request capacity.
  - Added support for loading connector pipelines directly from raw configuration.
  - Added batch processing across connectors, Prometheus, storage, and gRPC sinks.

- **Bug Fixes**
  - Prevented duplicate Kubernetes Events and unchanged condition updates from creating unnecessary writes.
  - Preserved batch identity across retries and handled duplicate datastore submissions safely.
  - Connector processing now waits for all connectors and reports every failure.

- **Documentation**
  - Expanded metrics documentation for connector outcomes, skipped operations, and datastore batches.


platform-connectors/pkg/connectors/kubernetes/k8s_connector.go
```diff
@@ -108,6 +121,14 @@ func InitializeK8sConnector(ctx context.Context, ringbuffer *ringbuffer.RingBuff
 	return kubernetesConnector, clientSet, nil
 }
 
+// ProcessBatch applies one batch to the cluster: node conditions and
+// Kubernetes Events for every processable event. It is the entry point for
+// callers that hold no queue (the deployment platform connector) and does
+// exactly what one iteration of FetchAndProcessHealthMetric does.
+func (r *K8sConnector) ProcessBatch(ctx context.Context, healthEvents *protos.HealthEvents) error {
+	return r.processHealthEvents(ctx, healthEvents)
```
lalitadithya, line 125: Don't we also need to start the tracing span?

```
			batchCtx, span := tracing.StartSpanWithLinkFromSpanContext(
				ctx, queuedHealthEvents.ParentSpanContext, "platform_connector.k8s.fetch_and_process_health_metric")
```

platform-connectors/pkg/server/platform_connector_server.go
```diff
@@ -17,84 +17,186 @@ package server
...
+// datastore write.
 type PlatformConnectorServer struct {
 	pb.UnimplementedPlatformConnectorServer
-	Pipeline *pipeline.Pipeline
+	Pipeline  *pipeline.Pipeline
+	Connector connectors.Connector
+	// PayloadLogLevel is the level of the per-batch log line that carries the
+	// whole payload. The zero value, Info, is the node-local role's level; the
+	// deployment role serves the whole fleet and logs it at Debug.
+	PayloadLogLevel slog.Level
```
lalitadithya: this feels a bit like premature optimization, can we drop this for now?

platform-connectors/pkg/server/platform_connector_server.go
```diff
@@ -17,84 +17,186 @@ package server
...
 	defer span.End()
 
-	eventCount := len(he.Events)
+	eventCount := len(he.GetEvents())
 	span.SetAttributes(
 		attribute.Int("platform_connector.grpc.event_count", eventCount),
 	)
 
-	slog.InfoContext(ctx, "Health events received", "events", he)
+	slog.Log(ctx, p.PayloadLogLevel, "Health events received", "events", he)
```
lalitadithya: if this going to be spammy, we can just reduce this to debug

platform-connectors/pkg/transformers/metadata/transformer.go
```diff
@@ -141,29 +181,165 @@ func (a *Augmentor) Name() string {
...
+
+// Prewarm reads the metadata of every distinct uncached node in the batch
+// concurrently, up to maxPrewarmReads at a time, so the per-event pass finds
+// it cached. The events of a batch are processed one after another inside one
+// budget; without this, a cross-node batch naming many uncached nodes would
+// spend the budget on the first few. A read that fails inside the budget is
+// left to the per-event pass, which reports it and fails open; the nodes the
+// budget ended before reading are reported as ErrBudgetExhausted instead, so
+// the batch is deferred rather than passing the skip-label gate unread.
+func (a *Augmentor) Prewarm(ctx context.Context, events []*pb.HealthEvent) error {
```
lalitadithya: I'm not quite able to follow what is the use case for this, do we have some data showing why we need this?

platform-connectors/pkg/connectors/connector.go
```diff
@@ -0,0 +1,155 @@
...
+
+// BestEffort returns a Connector that gives c at most timeout per batch (no
+// bound when timeout is zero) and never fails the batch: a failure or a
+// timeout is logged and counted under name, and the batch is acknowledged
+// anyway. It is for work that repairs
+// itself, like node conditions the next report rewrites, and for forwards
+// nobody waits for. When the caller's own context ends first, the connector
+// was cut short with the request and nothing is counted: the batch is not
+// acknowledged and will be resent.
+func BestEffort(name string, c Connector, timeout time.Duration) Connector {
```
lalitadithya: where is this used, I'm not able to find any references to this

platform-connectors/pkg/connectors/connector.go
```diff
@@ -0,0 +1,155 @@
...
+			err := c.ProcessBatch(memberCtx, he)
+			if err == nil {
+				return
+			}
+
+			mu.Lock()
+			if first == nil {
+				first = err
+			}
+			mu.Unlock()
```
lalitadithya: if everything is idempotent, what is the use case to cancel on the first error? I think we should collect all the errors into a error group and return that

platform-connectors/pkg/server/platform_connector_server.go
```diff
@@ -17,84 +17,186 @@ package server
...
 	}
 
-	// Enqueue with trace context so store and K8s connectors continue this trace
-	parentSC := span.SpanContext()
-	item := &ringbuffer.QueuedHealthEvents{Events: he, ParentSpanContext: parentSC}
+	return outcomeOK, nil
+}
 
-	for _, buffer := range ringBufferQueue {
-		buffer.Enqueue(item)
```
lalitadithya, line 92: won't removing `buffer.Enqueue` remove the buffer entirely?

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -490,46 +546,175 @@ func messageMatchesAnyErrorCode(msg string, errorCodes []string) bool {
...
+// bounds how long a replica can miss a recurrence when the recovery in
+// between was reported to another replica.
+const nodeEventRefreshInterval = 10 * time.Minute
+
+// maxRememberedMessagesPerCheck bounds the Events remembered for one check
+// on one node. A check reports a handful of distinct faults at a time, and
+// the message is producer-controlled, so the map must not grow with it; past
+// the bound the check's memory is dropped and its faults are announced again,
+// which costs one extra Event write each.
+const maxRememberedMessagesPerCheck = 32
```
lalitadithya: Do we have some data to justify why we need these caps? We have supressed the high frequency watches by default now, so we should no longer be getting a continuous stream of events. Do we know at what limits does the current mechanism fall over?

platform-connectors/pkg/connectors/kubernetes/k8s_connector.go
KaivalyaMDabhadkar: I believe this is already covered: the linked span in the loop exists because the queue breaks the context chain, whereas ProcessBatch runs inside the request, and processHealthEvents opens its own span as a child of the request span

platform-connectors/pkg/server/platform_connector_server.go
KaivalyaMDabhadkar: Dropped

platform-connectors/pkg/server/platform_connector_server.go
KaivalyaMDabhadkar: Done

platform-connectors/pkg/connectors/connector.go
KaivalyaMDabhadkar: This was actually going to be used in the next PR, had missed removing this from this PR. Removed now

platform-connectors/pkg/connectors/connector.go
KaivalyaMDabhadkar: Agreed, done

platform-connectors/pkg/transformers/metadata/transformer.go
KaivalyaMDabhadkar: I had originally added it for cross-node batches naming many uncached nodes, where sequential lookups would spend the whole budget on the first few, but no publisher actually sends such batches (each sends one event per batch) so I removed Prewarm

platform-connectors/pkg/server/platform_connector_server.go
KaivalyaMDabhadkar: No the buffers are still there, the enqueue just moved. The handler now calls Connector.ProcessBatch. In the dameonset, that Connector is the set of ring buffers and main.go collects the ring buffer of each enabled connector into set.queues and passes it to the handler. RingBuffer.ProcessBatch (pkg/ringbuffer/ring_buffer.go) is what the old loop did, it calls Enqueue with the batch and returns right away. So every batch still goes into every ring buffer, the same as before

lalitadithya: i'm confused now, where does the sync path come in because I'm looking a if check somewhere and I can't seem to find it. It is entirely possible I'm looking in the wrong place

KaivalyaMDabhadkar: No you are in the right place, there is no if check currently. The handler always calls Connector.ProcessBatch and what that does depends on the Connector that main.go passes in. In this PR that is always the ring buffers, whose ProcessBatch just enqueues and returns, so nothing changes in this PR. The sync path comes in the next PR, where the deployment role passes the real connectors instead. Sorry for the confusion, should have clarified that in the previous comment

lalitadithya: ah, okay, it makes sense now. I tried to trace the new path in this PR and got totally lost

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
KaivalyaMDabhadkar: `We have supressed the high frequency watches by default now, so we should no longer be getting a continuous stream of events`
Cool cool, makes sense. I removed the caps

docs/METRICS.md
```diff
@@ -220,6 +220,15 @@ sum by (agent, check_name) (rate(health_events_total{recommended_action!="NONE",
 sum by (node) (rate(health_events_total{recommended_action!="NONE"}[1h]))
 ```
 
+### Platform Connector Request Metrics
+
+Both roles of the platform connector, the node-local DaemonSet and the deployment platform connector, expose these for the batches that reach the request handler. Alert on `failed` outcomes.
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Right, fixed. the section now describes the node-local connector only

platform-connectors/pkg/connectors/store/store_connector.go
```diff
@@ -183,10 +230,60 @@ func (r *DatabaseStoreConnector) Disconnect(ctx context.Context) error {
 	return nil
 }
 
+// ProcessBatch writes one batch inside the caller's request, the deployment
+// platform connector's path: the events already carry their idempotency keys
+// (stamped before the pipeline ran), the reply follows the returned error,
+// and a resend whose events already exist is a success, counted as a
+// duplicate.
+func (r *DatabaseStoreConnector) ProcessBatch(ctx context.Context, healthEvents *protos.HealthEvents) error {
+	outcome, err := r.insertHealthEvents(ctx, healthEvents, "")
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: The socket-mode client sends no idempotency key: the shared client attaches the header only when it publishes directly to the deployment platform connector, so on the socket there is no publisher key to preserve, and a publisher resend is a new batch there, exactly as on main today. The key the store loop stamps protects its own retries after a partial write, which is what the PR claims. Keyed publisher retries belong to the deployment role, whose interceptor requires the header and derives the per-event keys before the pipeline; that is the next PR.

platform-connectors/pkg/connectors/kubernetes/process_node_events.go
```diff
@@ -314,6 +364,14 @@ func (r *K8sConnector) addMessageIfNotExist(messages []string, healthEvent *prot
 		if fmt.Sprintf("%s;", msg) == newMessage {
 			return messages
 		}
+
+		// An entry naming the same fault (error codes, entity, action) is
+		// that fault, whatever its text. Compaction at the length cap rewrites
+		// the text, so without this a saturated message would gain the fault
+		// again, move it, and count as a change on every repeat.
+		if messagesMatchByIdentity(msg, newMessage[:len(newMessage)-1]) {
```
deesharma24, line 372: Can we update both extractMessageIdentity and messagesMatchByIdentity? The parser currently extracts only GPU and PCI entities, missing NVLINK and NVSWITCH. The matcher then considers one shared entity enough to identify a duplicate.

For example, two SXID faults can share the same error code and switch PCI address but affect different GPUs and links. This check treats them as the same fault, so the second fault is missing from the node condition.
We should ideally extract all relevant entities and compare their complete sets.

KaivalyaMDabhadkar: That's a really great catch. Fixed now, the identity now includes all entity tokens and two entries match only when their complete entity sets are equal, with tests for your SXID example

deesharma24 (approved):

XRFXLP (approved):

PR #1858, opened 2026-09-17
feat(platform-connectors): deployment role with token auth, idempotent writes and index-gated readiness

KaivalyaMDabhadkar:
## Summary

Part of ADR 052, the deployment platform connector. #1796 added the client library, #1814 the datastore side and #1845 the connector interface, handler and pipeline both roles share. This PR adds the second role: with `PC_MODE=deployment` the platform-connectors binary runs as a central Deployment that the monitors reach over the network with TLS and a token. With `PC_MODE` unset the binary is the node-local DaemonSet connector, unchanged by this PR, so nothing changes at runtime until the Helm chart PR sets the mode. The wire contract is unchanged: the same `PlatformConnector` gRPC service, the same protobuf.

### What changes

- **The deployment role, `pkg/central`**: configuration from environment variables (validated at start, a wrong value refuses to start and names the variable), the TLS listener with certificates reloaded from `TLS_CERT_DIR` when cert-manager rotates them, the TokenReview validator sized for a fleet (`TOKEN_CACHE_SIZE`, `TOKENREVIEW_QPS`, `TOKENREVIEW_BURST`), and three interceptors in front of the shared handler: authentication, the idempotency key, and readiness. Plaintext is refused unless `TLS_INSECURE_DEVELOPMENT_MODE=true`, which logs a warning on every start. Connections are recycled after `MAX_CONNECTION_AGE` or `MAX_CONNECTION_IDLE` so a rollout spreads the monitors over the new replicas. `main.go` selects the role from `PC_MODE`.
- **The connectors run inside the request.** The handler hands each batch to a `Set` of the store connector, the Kubernetes connector and the gRPC sink wrapped in `BestEffort`, and the Prometheus connector. `BestEffort` is new in `pkg/connectors`: it gives the wrapped connector `CONDITION_UPDATE_TIMEOUT` per batch and never fails the batch; a failure or timeout is logged and counted in `platform_connector_best_effort_failures_total`, not returned, because the stored event is the source of truth. The reply follows the datastore write.
- **Authentication is the DaemonSet's node-binding interceptor with no local node.** `pkg/auth` gains one branch: when `Config.NodeName` is empty, every caller must present a token (`Unauthenticated` otherwise), the token must be bound to a scheduled pod and carry a node claim, that claim is the node the caller may name, and `AllowedServiceAccounts` gates who may call at all. Identities on `CrossNodeServiceAccounts` may name any node but must name one. Fail-open and audit mode are refused without a local node. `pkg/central/auth.go` only wires this configuration; there is no second implementation. New violation reasons: `token_missing`, `unbound_token`, `node_claim_absent`, `identity_not_allowed`.
- **Writes are idempotent.** The `idempotency-key` header is mandatory (`InvalidArgument` when missing, malformed or repeated, counted in `platform_connector_deployment_refusals_total{reason="idempotency_key"}`). The interceptor writes a key per event, `callerPodUID#clientKey#index`, into the event's metadata before the pipeline runs, and the store connector inserts with the idempotent insert from #1814. A resend is acknowledged as a duplicate and stores nothing; a resend after a partial write stores the missing events. A batch is acknowledged only once it is in the datastore; a store failure is answered `Unavailable` and the caller retries with the same key. A duplicate on another unique index is answered `InvalidArgument`, because a resend cannot fix it.
- **Readiness follows the index.** A replica refuses writes (`Unavailable`, counted as `index_unverified`) until it has verified the idempotency index the datastore setup creates, checking every 5 s. Once ready it re-checks every 5 minutes: a datastore restored without the index turns the replica unready until the setup recreates it. A datastore outage changes nothing, the replica stays ready and the callers' retries cover it. `/readyz` reflects this. The index itself is created by the Mongo setup Job's script and by the PostgreSQL table setup, in the chart and store-client PRs; the connector never creates it.
- **Fleet sizing**: the metadata augmentor's client rate limit and cache (`NODE_METADATA_CACHE_SIZE`, `NODE_METADATA_CACHE_TTL`), since one replica serves the whole fleet.
- **Metrics**, documented in `docs/METRICS.md`: `platform_connector_deployment_refusals_total` by reason, and the auth rows gain the new reasons; the request and store metrics are those of #1845; `platform_connector_best_effort_failures_total` by connector and reason (`failed`, `timeout`) is new here.

### What is not in this PR

The Helm chart (Deployment, Service, certificate, NetworkPolicy, publisher wiring, the index block of the Mongo setup script), the PostgreSQL index in the table setup, the monitors' switch to publishing directly, the Tilt environment and e2e tests, and the operator documentation come in the remaining PRs of the series. Reviewing this PR does not require them.

### Tests

Unit tests, all passing, lint clean, `go mod tidy` clean:

- `pkg/central/config_test.go`: defaults, overrides and rejected values for every variable.
- `pkg/central/auth_test.go`: the wiring through the shared interceptor: missing and malformed token, unbound token, identity off the allowlist, another node, own node with the caller identity in the context, cross-node publisher, cross-node token without a node claim, non-batch requests still gated.
- `pkg/central/interceptors_test.go`: the key is required, one header only and of the documented shape; every event is stamped; a missing caller identity is Internal; writes are refused until the index is verified.
- `pkg/central/server_test.go`: the readiness loop follows definitive answers and survives a stalled check; the interceptor chain as wired, over an in-memory connection.
- `pkg/central/tls_test.go`: serving with a reloaded certificate.
- `pkg/connectors/connector_test.go`: a best-effort member's failure or timeout is counted and does not fail the batch; a cancelled request is not counted as its failure; a store failure leaves a best-effort member to finish its work.
- `pkg/auth/nodebinding_test.go`: the fleet configuration (token required, unbound and claimless tokens refused, allowlist, cross-node scope) next to the existing node-local tests; fail-open and audit mode refused without a local node.

Manual testing on kind with the whole series applied, on Mongo and on PostgreSQL with certificate auth, two replicas behind the Service:

- both replicas ready only after the idempotency index existed, with the index-lost and index-recreated transitions observed live on both datastores (a wrong index definition turned both replicas unready and blocked the publishers until the setup Job recreated it);
- the request contract with real pod-bound tokens: healthy probe stored, replay stored once and counted as a duplicate, no token, unbound cross-node token, wrong audience, identity off the allowlist, node-scoped token naming another node, missing or malformed or repeated key, CUSTOM action without text, plaintext, empty batch, missing timestamp, a batch naming every node, a batch of 200 events, a node that does not exist;
- central dedup, derived Event names across replicas, rolling restarts and Mongo outages under keyed traffic with exactly one document per key;
- a 150 s connector outage, a 1 minute retry window, SIGTERM while blocked in the Go and the Python client, token rotation across expiry;
- the deployment e2e tests on Mongo and PostgreSQL.

After the review round on #1845 (the connector set waits for every member, no batch budget, the time-bounded Event memory, the complete-entity identity match) the deployment-mode regression ran again on kind with this exact code: the request contract, the 12-node and 200-event batches, central dedup, derived Event names across replicas, a rolling restart and a Mongo outage under keyed traffic, and the three deployment e2e tests, all passing.

An earlier version of this server, before the connector set and the auth fold, also ran the same scenarios on four dev clusters (OCI, GKE, an on-prem datacenter with Calico, EKS), including CA rotation by cert-manager, token rotation by the kubelet, a Mongo primary failover and the gRPC NetworkPolicy on the three enforcing CNIs.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Health Monitor
- [x] Core Service
- [ ] Fault Management
- [ ] Janitor
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Documentation/CI
- [ ] New Component
- [ ] Multiple Components
- [ ] Other

## Testing
- [x] Tests pass locally

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **New Features**
  - Added deployment-mode platform connector support with authenticated, pod-bound requests and node-scoped authorization.
  - Added readiness protection and idempotency-key validation for deployment batches.
  - Added TLS certificate rotation, health, metrics, and probe endpoints.
  - Added best-effort connector handling so optional connector failures do not fail batches.
  - Added configuration support for gRPC sink and Kubernetes connectors.

- **Documentation**
  - Expanded authentication, request, deployment, refusal, audit, and connector metrics documentation.

- **Tests**
  - Added coverage for authentication, configuration, readiness, idempotency, TLS, and connector failure handling.


platform-connectors/pkg/auth/nodebinding.go
```diff
@@ -209,6 +229,16 @@ type Config struct {
 	// which is the same treatment an anonymous caller gets.
 	CrossNodeServiceAccounts []string
 
+	// AllowedServiceAccounts, when set, lists the canonical usernames that may
```
lalitadithya: I'm not quite sure we need this, the only two checks we need in the new "strict" mode is that a token must be present and the token has a node name if is not from `CrossNodeServiceAccounts`. Additional filtering is good, but in this case will cause churn because in this case every time a monitor is added/updated, the deployment will also have to be rolled, and this would result in dropped caches etc.

platform-connectors/pkg/central/auth.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// newAuthInterceptor builds caller authentication for the deployment role
+// from the node-binding interceptor the DaemonSet uses, in its configuration
+// without a local node: every caller must present a pod-bound token, its
+// events are pinned to the node the token claims, the listed cross-node
+// publishers may name any node, and only the allowed publishers may call at
+// all. The interceptor validates both lists: a malformed username, or a
+// cross-node entry missing from the allowlist, refuses to start.
+func newAuthInterceptor(cfg *config, validator auth.TokenValidator) (grpc.UnaryServerInterceptor, error) {
```
lalitadithya: isn't this the same as https://github.com/KaivalyaMDabhadkar/NVSentinel/blob/822d8e174c675bc0a56f7fe2ffa0b9d77eac232b/platform-connectors/main.go#L377? should we try and combine?

platform-connectors/pkg/central/config.go
```diff
@@ -0,0 +1,329 @@
...
+	for _, w := range strings.Split(raw, ",") {
+		if w = strings.TrimSpace(w); w != "" {
+			out = append(out, w)
+		}
+	}
+
+	return out
+}
+
+func envInt(key string, def int) (int, error) {
```
lalitadithya: can we use functions in commons/pkg/envutil/envutil.go or add to it for this?

platform-connectors/pkg/central/server.go
```diff
@@ -0,0 +1,550 @@
...
+// initComponents builds the store connector, the other connectors, the
+// pipeline and the request handler from the environment and the shared
+// config.json, read with the same code as the DaemonSet role. The connectors
+// go into one set without queues: the store's result is the reply, and every
+// other connector is best effort inside the condition update timeout, so a
+// slow node update or sink never fails a batch that is safely stored.
+func initComponents(ctx context.Context, cfg *config) (*components, error) {
+	c := &components{cfg: cfg, ready: &readiness{}}
+
+	storeConnector, err := store.InitializeDatabaseStoreConnector(ctx, nil, cfg.certMountPath, 0)
```
lalitadithya: this make the store connector mandatory, why? Maybe some customers want to use the deployment version with just the kubernetes connector, can we try and retain the same init and config behavior we have for the daemonsets please?

platform-connectors/pkg/central/config.go
KaivalyaMDabhadkar: Done

platform-connectors/pkg/central/server.go
KaivalyaMDabhadkar: Fixed

platform-connectors/pkg/central/auth.go
KaivalyaMDabhadkar: Good catch, yes they were doing the same thing so I combined them. The daemonset's parsing of the auth settings moved into a shared reader in pkg/auth, and both roles now read the same config.json keys with one audience. The deployment side keeps only the small part that is different without a local node which is the allowlist and enforce mode.

platform-connectors/pkg/auth/nodebinding.go
KaivalyaMDabhadkar: You are right about the churn. I fixed that on the chart side, the allowlist now names every bundled publisher whether it is enabled or not, so enabling or disabling a monitor no longer changes the configmap and no longer restarts the deployment platform connector pods. I kept the check itself though. Any pod in the cluster can request a projected token for our audience, so without the allowlist any pod could publish events about its own node, which the socket never allowed. With the churn gone I think the check is worth keeping

lalitadithya: > I kept the check itself though. Any pod in the cluster can request a projected token for our audience, so without the allowlist any pod could publish events about its own node, which the socket never allowed.

tbh they can do that today with the daemonset mode. so I'm not sure how is this increasing security,

platform-connectors/pkg/central/server.go
```diff
@@ -0,0 +1,550 @@
...
+	}
+
+	if configfile.Bool(raw, "enableGRPCSinkConnector") {
+		sink, err := grpcsink.SettingsFromConfig(raw)
+		if err != nil {
+			return nil, err
+		}
+
+		// No queue, so no retry loop: BestEffort logs and counts a failure.
+		c.sink, err = grpcsink.InitializeGRPCSinkConnector(nil, sink.Target, 0, sink.TokenPath)
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: This is the existing ADR-033 gRPC sink connector and this PR does not change it, the daemonset already dials the sink the same way with the same token today

platform-connectors/pkg/auth/nodebinding.go
KaivalyaMDabhadkar: Fair enough, I have dropped it

docs/METRICS.md
```diff
@@ -171,8 +171,8 @@ submitting health events naming another node. See
...
... as retryable `Unavailable` rather than counted as `node_mismatch`, since the caller might really be an allowlisted cross-node publisher the outage prevented from being verified; under `audit` that same `Unavailable` is recorded and then let through like any other audited violation, so the event is forwarded with whatever node name it carried. `missing_node_name` is the one violation `mode: audit` does not let through: a blank node name from a caller resolved to **verified** cross-node scope is always rejected, in both modes, because nothing downstream can handle it. This does not cover a caller that fell back to a degraded node-local scope under `failOpenOnUnavailable`: its identity was never verified as cross-node, so a blank name from it is stamped like any other node-local caller's. Exclude those three when alerting on suspected credential abuse, because a control-plane outage increments them for every in-flight request and would otherwise look identical to an attack: `sum(rate(platform_connector_auth_violations_total{reason=~"node_mismatch\|token_invalid\|node_claim_mismatch\|unbound_cross_node_token\|malformed_credentials\|cross_node_claim_absent\|missing_node_name"}[5m]))` |
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Fixed, the selector now includes token_missing, unbound_token and node_claim_absent

platform-connectors/pkg/central/interceptors.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+// readinessInterceptor refuses batches with a retryable status while the
+// replica has not verified the idempotency index, so a resend is never stored
+// twice before the index exists. Taking the replica out of the Service is not
+// enough, because established connections keep sending to it.
+func readinessInterceptor(ready *readiness) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req any, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
+		if !ready.writesAllowed() {
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Without a store the replica already starts with the index marked verified in initComponents, so batches are accepted. I moved that into a small newReadiness function and added a test for both cases

platform-connectors/pkg/central/server.go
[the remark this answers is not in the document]
KaivalyaMDabhadkar: The deployment role uses the sink connector unchanged, so it dials the target exactly the way the daemonset does today. The plaintext default is the documented ADR-033 behaviour of that connector, not something this PR introduces. Adding TLS to the sink is a change to the connector itself for both roles, so it stays out of this PR. Resolving this thread.

platform-connectors/pkg/central/server.go
```diff
@@ -0,0 +1,601 @@
...
+
+	return settings, nil
+}
+
+// appendOptionalConnectors adds the connectors config.json enables to set,
+// built without queues: the k8s connector updates node conditions and writes
+// Events inside the request, on change, and the gRPC sink is called once per
+// batch, both run as member describes; the Prometheus connector counts every
+// batch.
+func (c *components) appendOptionalConnectors(
```
lalitadithya, line 348: a lot of this logic seems to be similar to what we have with the daemonset mode as well. I was wondering why can't we have a single code path that can do the initialization and then we pick either best effort or queue connector based on the flag?

platform-connectors/pkg/central/config.go
```diff
@@ -0,0 +1,254 @@
...
+	if cfg.tokenReviewBurst, err = envPositiveInt("TOKENREVIEW_BURST", defaultTokenReviewBurst); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// loadFleetSizingEnv reads the overrides for values the shared config.json
+// sizes for one node. Unset (zero) keeps the config.json value.
+func loadFleetSizingEnv(cfg *config) error {
```
lalitadithya, line 183: I'm not really a fan of the overrides like this since this add one more layer to the helm overrides that we already have. Why can't these be handled in the helm layer itself?

platform-connectors/pkg/central/server.go
```diff
@@ -0,0 +1,601 @@
...
+
+	return nil
+}
+
+// indexVerifier is the one store connector call the readiness loop needs.
+type indexVerifier interface {
+	VerifyIdempotencyIndex(ctx context.Context) error
+}
+
+// verifyIndexLoop keeps readiness tied to the idempotency index for as long
```
lalitadithya, line 145: can you please share why we need to run this continuously? Under what cases would the index be dropped? I feel the only case would be an operator action and in such case it is a deliberate operation taken by the user of the system.

platform-connectors/pkg/central/server.go
```diff
@@ -0,0 +1,601 @@
...
+	}
+
+	if cfg.grpcWriteBufferBytes > 0 {
+		opts = append(opts, grpc.WriteBufferSize(cfg.grpcWriteBufferBytes))
+	}
+
+	if cfg.tlsCertDir == "" {
+		// loadConfigFromEnv only allows this with the explicitly named
+		// insecure development mode.
+		slog.Warn("Serving PLAINTEXT: TLS_INSECURE_DEVELOPMENT_MODE is set; never use this outside development")
```
lalitadithya, line 238: this is a confusing override, can we keep all the required configuration options in one place -- the config file?

platform-connectors/pkg/central/server.go
```diff
@@ -0,0 +1,601 @@
...
+
+	c.grpcServer = grpc.NewServer(serverOpts...)
+	pb.RegisterPlatformConnectorServer(c.grpcServer, c.handler)
+
+	return serve(ctx, cancel, c, certWatcher)
+}
+
+// serve runs the gRPC server, the metrics server, the certificate watcher and
+// the signal handler until one of them stops the group.
+func serve(ctx context.Context, cancel context.CancelFunc, c *components, certWatcher *certwatcher.CertWatcher) error {
```
lalitadithya, line 473: a lot of this function seems to be similar to what we already have scaffolded for the daemonset platform connector. Some of the function is different, but those parts I believe can be wrapped behind an interface or a simple var switch. Can you please check if we can merge the two code paths?

platform-connectors/pkg/central/server.go
KaivalyaMDabhadkar: Done. Now both roles build their connectors through one builder in pkg/bootstrap. The only switch is whether the connectors run from queues or inside the request

platform-connectors/pkg/central/config.go
KaivalyaMDabhadkar: Yeah, you are right. I have removed the fleet sizing env overrides

platform-connectors/pkg/central/server.go
KaivalyaMDabhadkar: Agreed with you, it was not ncessary. removed the recheck

platform-connectors/pkg/central/server.go
KaivalyaMDabhadkar: Done. Every deployment tunable now lives in config.json

platform-connectors/pkg/central/server.go
KaivalyaMDabhadkar: Done. There is now one run loop in pkg/bootstrap for both roles, and a role only supplies its listener, interceptors and transport security

docs/configuration/platform-connectors.md
```diff
@@ -236,7 +236,7 @@ Healthy events are not downgraded by deduplication. Before they continue downstr
 ### Operational Notes
 
 - Dedup state is in-memory only and is cleared on platform-connectors pod restart.
-- The dedup counter is exposed as `nvsentinel_platform_connector_dedup_store_and_analyse_total{check,node,err_code}`.
+- The dedup counter is exposed as `nvsentinel_platform_connector_dedup_store_and_analyse_total{check,err_code}`. It carries no node label, because the deployment platform connector runs the transformer for the whole fleet; the node is in the log line.
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: This label is not new here. It comes from the original dedup transformer, and this series only dropped the node label, which was the one that grew with the fleet. The values are XID and SXID numbers, a small fixed set, so I am leaving it as it is

platform-connectors/pkg/connectors/connector.go
```diff
@@ -64,3 +73,61 @@ func (s Set) ProcessBatch(ctx context.Context, he *pb.HealthEvents) error {
...
+		"by connector and reason (failed, timeout)",
+}, []string{"connector", "reason"})
+
+// BestEffort returns a Connector that gives c at most timeout per batch
+// (which must be positive) and never fails the batch: a failure or a
+// timeout is logged and counted under name, and the batch is acknowledged
+// anyway. It is for work that repairs itself, like node conditions the next
+// report rewrites, and for forwards nobody waits for. When the caller's own
+// context ends first, the connector was cut short with the request and
+// nothing is counted: the batch is not acknowledged and will be resent.
```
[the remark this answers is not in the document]
KaivalyaMDabhadkar: Returning nil there is intended. When the caller's context ends, the caller has already given up on the attempt and resends the batch under the same idempotency key, so a cut-short node update is not a failure worth counting and the request outcome stays with the store write. I reworded the comment to say exactly that.

platform-connectors/pkg/central/role.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+// waitForIndex opens the gate once the idempotency index matches its expected
+// definition, checking every `every` until it does: this is what orders the
+// datastore setup before any client traffic. Each check has its own deadline.
+// The wait ends with ctx. The index is not re-checked afterwards: it can only
+// go missing through an operator action on the datastore, and the restore
+// procedure re-runs the setup and restarts the replicas, which verify it
+// again at start.
+func waitForIndex(ctx context.Context, verifier indexVerifier, gate *indexGate, every, attemptTimeout time.Duration) {
```
lalitadithya: can this loop be made to fail loudly after a certain number of retries (CrashLoopBackOff)?

platform-connectors/main.go
```diff
@@ -19,887 +19,150 @@ import (
...
-	}
-
-	return validator, nil
-}
+// One image, two roles. The process setup here is shared; the roles differ
+// in what they hand bootstrap.Run. The DaemonSet keeps the tracing service
+// name it has always had.
+func main() {
+	f := parseFlags()
+	mode := os.Getenv("PC_MODE")
```
lalitadithya: should this also be a flag or in the config file?

platform-connectors/pkg/central/role.go
KaivalyaMDabhadkar: Good idea, done

platform-connectors/main.go
KaivalyaMDabhadkar: Agreed, made the change

lalitadithya (approved):

PR #1873, opened 2026-09-17
fix(fault-quarantine): remove cordon labels on manual uncordon

### LINKED ISSUE #1872: [Bug]: manual uncordon leaves cordon-by/cordon-reason/cordon-timestamp labels on the node
## Bug Description

When a quarantined node is restored with `kubectl uncordon` — the cancellation
path documented in `docs/cancelling-breakfix.md` — fault-quarantine cleans up
almost everything it applied, but three labels are left behind:

- `<label-prefix>cordon-by`
- `<label-prefix>cordon-reason`
- `<label-prefix>cordon-timestamp`

The node returns to service permanently labelled `cordon-by=NVSentinel`. Anything
that reads those labels to attribute a cordon will then mis-attribute the node's
*next* cordon, whatever its real origin — including
`tests/scale-tests/cmd/fqm-scale-test/main.go`, which selects on
`k8saas.nvidia.com/cordon-by=NVSentinel`.

The automatic uncordon path does remove these three labels. Only the manual one
leaves them.

Expected: both paths leave the node in the same state.

## Component

Fault Management

## Steps to Reproduce

Any fatal event that cordons a node works. I used an injected XID so the
reproduction is self-contained — writing a fake `NVRM: Xid` line into the node's
kernel ring buffer via `/dev/kmsg`, which `syslog-health-monitor` picks up like a
real one.

1. Pick a healthy node with no `cordon-*` labels:

   ```console
   $ kubectl get node <node> -o json | jq '.metadata.labels | with_entries(select(.key|test("cordon")))'
   {}
   ```

2. Trigger a fatal event that the ruleset cordons on. With an injected XID, from
   a privileged pod on the node:

   ```sh
   echo "gpu 0: NVRM: Xid (PCI:0008:06:00): 79, pid=0, name=xid-test, GPU has fallen off the bus" > /host/dev/kmsg
   ```

   (The `gpu 0: ` prefix is only needed for the injection itself: journald turns
   the first `word: ` of a userspace `/dev/kmsg` write into `SYSLOG_IDENTIFIER`
   and strips it from `MESSAGE`, so a bare `NVRM: Xid ...` never matches
   `XIDPattern`. Irrelevant to this bug.)

3. The node is cordoned and the labels are applied:

   ```console
   $ kubectl get node <node> -o json | jq -r '.metadata.labels | to_entries[] | select(.key|test("cordon")) | "\(.key) = \(.value)"'
   k8saas.nvidia.com/cordon-by = NVSentinel
   k8saas.nvidia.com/cordon-reason = Syslog-fatal-error-ruleset
   k8saas.nvidia.com/cordon-timestamp = 2026-09-17T14-39-51Z
   ```

4. Cancel the breakfix as documented:

   ```sh
   kubectl uncordon <node>
   ```

5. Everything else is cleaned within ~400 ms — both `quarantineHealthEvent*`
   annotations, the `NoSchedule` taint, the DrainRequest CR, the
   `dgxc.nvidia.com/nvsentinel-state` label, and the health event is marked
   cancelled. The three `cordon-*` labels remain:

   ```console
   $ kubectl get node <node> -o json | jq -r '.metadata.labels | to_entries[] | select(.key|test("cordon")) | "\(.key) = \(.value)"'
   k8saas.nvidia.com/cordon-by = NVSentinel
   k8saas.nvidia.com/cordon-reason = Syslog-fatal-error-ruleset
   k8saas.nvidia.com/cordon-timestamp = 2026-09-17T14-39-51Z
   ```

Ran 3/3 on the same node, deleting the labels by hand between runs each time.
They came back with the current cordon's timestamp every run, so this is not a
stale residue from an earlier state. They were also still present 15 minutes
after an uncordon, so it is not a slower reconcile catching up later.

## Environment

- NVSentinel version: v1.23.0 (`ghcr.io/nvidia/nvsentinel/fault-quarantine:v1.23.0`)
- Kubernetes version: v1.35.3
- Deployment method: Helm chart via Argo CD, `label-prefix = "k8saas.nvidia.com/"`

## Logs/Output

fault-quarantine, one full manual-uncordon handling. Note the single
`Removing label key from node` line, for `nvsentinel-state` only:

```
{"time":"2026-09-17T14:10:15.963Z","level":"INFO","msg":"Detected manual uncordon of FQ-quarantined node","module":"fault-quarantine","version":"v1.23.0","node":"<node>"}
{"time":"2026-09-17T14:10:16.136Z","level":"INFO","msg":"Removing label key from node","module":"fault-quarantine","version":"v1.23.0","key":"dgxc.nvidia.com/nvsentinel-state","node":"<node>"}
{"time":"2026-09-17T14:10:16.313Z","level":"INFO","msg":"Set currentQuarantinedNodes to 0 for manually uncordoned node","module":"fault-quarantine","version":"v1.23.0","node":"<node>"}
{"time":"2026-09-17T14:10:16.337Z","level":"INFO","msg":"Updated quarantining events to cancelled status","module":"fault-quarantine","version":"v1.23.0","node":"<node>","firstEventId":"...","documentsUpdated":1}
{"time":"2026-09-17T14:10:16.337Z","level":"INFO","msg":"Successfully completed manual uncordon handling","module":"fault-quarantine","version":"v1.23.0","node":"<node>"}
```

iacker:
## Summary

`fault-quarantine` cleans up almost everything on a manual `kubectl uncordon`, but the manual path only removed the `nvsentinel-state` label and left `cordon-by`, `cordon-reason`, and `cordon-timestamp` behind. The node returned to service permanently labelled `cordon-by=NVSentinel`, so anything attributing a cordon from that label (including `tests/scale-tests/cmd/fqm-scale-test`) would mis-attribute the node's next cordon. The automatic uncordon path already removes these three labels.

This makes `handleManualUncordon` remove the same three labels, but only when NVSentinel set `cordon-by`, per the maintainer note on the issue: some users reuse the key for other purposes, so a blind removal would drop their label.

Fixes #1872

## Type of Change
- [x] 🐛 Bug fix

## Component(s) Affected
- [x] Fault Management

## Testing
- [x] Tests pass locally

`go test ./pkg/reconciler/` passes (envtest). Added two e2e tests: one asserting the NVSentinel-set labels are removed on manual uncordon, one asserting a foreign `cordon-by` value is preserved. Updated `TestE2ECordonAndTaint_ManualUncordon`, which asserted the old orphaned-label behavior, to expect the labels to be gone. `golangci-lint` v2.13.1 reports 0 issues.

## Checklist
- [x] Self-review completed
- [x] Ready for review

## Summary by CodeRabbit

- **Bug Fixes**
  - Manual uncordoning now removes NVSentinel-applied cordon labels, including the source, reason, and timestamp.
  - Cordon labels managed by other users remain unchanged.
  - Cleanup is skipped when the affected node cannot be retrieved.


lalitadithya:
hey @iacker , can you please rebase and see if the rabbit comments are worth addressing?

iacker:
Rebased on main, tests pass. The rabbit comment is valid. nvsentinelCordonLabelsToRemove checks cordon-by once, but the retry callback in HandleManualUncordonCleanup then removes the labels unconditionally on the fresh node. A concurrent owner change could drop a third party's cordon-by. The clean fix rechecks ownership inside the callback, which touches the shared cleanup path. Do it here or in a follow-up?

lalitadithya:
> A concurrent owner change could drop a third party's cordon-by. The clean fix rechecks ownership inside the callback, which touches the shared cleanup path. Do it here or in a follow-up?

either is fine, your call

iacker:
Thanks @lalitadithya — rebased onto latest `main`, and both CodeRabbit comments were worth addressing:

**Major (ownership re-check on retry).** Good catch. The cordon labels were removed from a set computed once before the update, but `NodePatcher.Patch` re-invokes the mutate callback with a freshly fetched Node on every conflict retry — so a concurrent owner change could have let a retry strip another controller's `cordon-by`/`reason`/`timestamp`. Fixed by moving the ownership guard into the callback via a small `ConditionalLabelRemoval`, so `cordon-by=NVSentinel` is re-checked against the live Node on each attempt. Removed the now-unused `nvsentinelCordonLabelsToRemove` helper.

**Minor (test coverage).** The foreign-label test only seeded and asserted `cordon-by`, so it could have passed even if `reason`/`timestamp` were wrongly removed. It now seeds and asserts all three foreign cordon labels.

Verified locally against envtest — `TestE2E_ManualUncordonRemovesCordonLabels` and `TestE2E_ManualUncordonPreservesForeignCordonByLabel` both pass, and the wider `ManualUncordon`/`ManualUntaint`/`CordonAndTaint` set is green. PTAL when you get a chance.

KaivalyaMDabhadkar (approved):

PR #1874, opened 2026-09-17
feat: label dcgm_health_active_events with the error code

### LINKED ISSUE #1862: [Feature]: the DCGM error code is on suppressed incidents but not on the ones that fire
### Prerequisites

### Code of Conduct

### Feature Summary

The DCGM error code is attached to the metrics for incidents that were **discarded**, and to nothing that **fires**. `dcgm_health_check_suppressed_incidents{error_code}` and `dcgm_health_check_debounced_incidents{error_code, gpu_id}` both carry it, while the two metrics an operator actually alerts on do not:

```python
# platform_connector/metrics.py:38
dcgm_health_active_events = Gauge(..., labelnames=["event_type", "gpu_id"])
```
```go
// platform-connectors/pkg/connectors/prom/metrics.go:43
[]string{"node", "agent", "check_name", "recommended_action", "is_fatal", "is_healthy"}
```

So you can see the code for a finding that was thrown away, and not for the one that paged you. The value is already in hand at the point of emission; it is simply not on the series that fires.

### Problem/Use Case

`check_name` plus `recommended_action` is not enough to decide what to do, because one check maps to several codes with opposite operational responses. Concrete, from a 270-node GB200 cluster this week.

**Today**, 11 active `GpuNvlinkWatch`, every one `recommended_action=CONTACT_SUPPORT`:

```
DCGM_EXP_GPU_HEALTH_STATUS{health_watch="NVLINK"} > 0
  -> DCGM_FR_NVLINK_EFFECTIVE_BER_THRESHOLD   x11
```

One marginal lane per flagged GPU. On one example node, GPU 0 link 0 shows `Effective BER 3e-11` with 330 successful link-recovery events, while its other 17 links and every link on the other three GPUs sit at the measurement floor with zero recoveries. All links up at full 53.125 GB/s. **Correct response: inspect or RMA that one connector.**

**Two weeks earlier**, the same `check_name` with the same `recommended_action` on the same cluster was:

```
  -> DCGM_FR_FABRIC_PROBE_STATE               rack-wide, all 4 GPUs, post-reboot
```

That one is the IMEX fabric probe re-running after a reboot on a cluster that deliberately does not run IMEX. **Correct response: ignore it.**

Identical `check_name`, identical `recommended_action`, opposite actions, and the metric cannot tell them apart. To find out which one I was looking at I had to leave NVSentinel entirely and scrape `DCGM_EXP_GPU_HEALTH_STATUS` from dcgm-exporter in another namespace. That is a poor position from which to write an alert rule, and it is the reason we have not enabled paging on this signal.

### Proposed Solution

Add the code to the firing metrics, in the narrow form that avoids the cardinality problem discussed on #1687.

**On `dcgm_health_active_events`, add `error_code`.** These are DCGM `DCGM_FR_*` constants: a closed, small set enumerated by DCGM itself, and the gauge is already per `(event_type, gpu_id)` on a DaemonSet so each pod only ever reports its own node. This is where the value is highest and the cost is lowest.

**I would leave `health_events_total` alone unless the suffix problem is solved first.** For the record, I argued against putting `error_code` there on #1687 and I still think that was right: syslog codes arrive suffixed, for example `145.RLW_SRC_TRACK`, so the set is not bounded. If it is wanted there eventually, the way in is a normalised label, `error_code` truncated at the first `.` with the detail staying in the event body, rather than the raw string.

An alternative that solves the same problem without touching label sets: put the code in the health event payload the connector publishes, so it lands in the datastore and the dashboard can read it per event. That would suit the "why is this node flagged" question even better, though it does not help alert rules.

### Why this is worth doing

- It is the difference between an alert that names a physical action and one that says "some NVLink thing happened on some GPU".
- It costs nothing to collect, since the code is already available where both existing counters attach it.
- It removes the need to correlate NVSentinel against dcgm-exporter by hand to interpret NVSentinel's own output.

Follow-on to #1687, which added `recommended_action` and made the one-to-many relationship visible in the first place. Happy to open the PR if the `dcgm_health_active_events` shape is acceptable.

### Component

gpu-health-monitor

### Version

v1.23.0

lfriedman-netllama:
Closes #1862.

`dcgm_health_active_events` reports which watch and which GPU has an active event, but not which error code raised it. On an NVL72 fleet every active `GpuNvlinkWatch` series looks identical, so an effective-BER threshold breach and a link-down are indistinguishable from the metric; telling them apart means going to the datastore. Adding `error_code` makes the gauge answer "which fault" directly.

The label value is `failure_details.code`, the same string already used as the `errorCode` on the published `HealthEvent` and as the key into the DCGM error map, so it needs no new plumbing or translation.

### The part worth reviewing

This is a gauge, so every clear has to zero the same label set it set. A single check can be latched by more than one code, and `GpuDcgmConnectivityFailure` is reachable with both `DCGM_CONNECTIVITY_ERROR` and `DCGM_PROBE_HANG`, so a clear that assumed one code would leave the others reading 1 for the lifetime of the process.

Rather than guess, every clear path now enumerates the codes the entity cache recorded. That state already exists as `EntityCacheEntry.active_errors` and is read before the cache entry is reset:

- the DCGM watch path carries the code through `pending_metric_updates`, one entry per code on recovery
- the connectivity-restored and probe-responsive paths call a small `_clear_active_event_metric` helper

### Cardinality

Bounded, and narrower than it looks. The label multiplies existing series only by the codes actually observed on a node, and a node has a handful of distinct DCGM error codes active at a time, drawn from a fixed enum. For reference, on a 271-node GB200 fleet the gauge currently carries 11 active series and every one of them is a single code, so this is a rename in practice there.

I did consider the opt-in shape used for `rule_matched_entity_total` in #1812, but that label was `entity_value`, which is genuinely unbounded. This one is not, and gating it would mean the default deployment keeps the ambiguity the issue is about. Happy to put it behind a flag if you would rather.

### Tests

`test_active_events_metric_carries_error_code_and_clears_every_code` drives a GPU with two simultaneous codes through failure and recovery and asserts each code gets its own series and that both are zeroed. It is decisive: narrowing the clear to a single assumed code leaves `GPU_ERROR_2` at 1 and the test fails.

The existing `test_dcgm_unresponsive_state_survives_process_restart` assertion is updated for the new label set. Full `test_platform_connector.py` suite passes, 48 passed / 5 subtests. `docs/METRICS.md` updated.

## Summary by CodeRabbit

- **New Features**
  - Health-event monitoring metrics now identify individual error codes alongside event type and GPU.
  - Separate metric series are maintained for multiple simultaneous error codes, enabling more precise monitoring and alerting.

- **Bug Fixes**
  - Resolved an issue where recovered health events could leave code-specific metrics active.
  - Metrics now reset correctly for all recovered error codes, including connectivity-related events.

- **Documentation**
  - Updated metric documentation to describe the new `error_code` label.


lalitadithya:
hey @lfriedman-netllama , can you please complete DCO -- https://github.com/NVIDIA/NVSentinel/pull/1874/checks?check_run_id=105279226903

lalitadithya:
hey, sorry about this, we merged a DCGM upgrade which has resulted in conflicts, can you please help check?

lfriedman-netllama:
Both done, thanks for the heads up.

DCO is signed off and passing. The conflict from the DCGM 4.7.0 merge is resolved in `037e2f6e`, and the PR is mergeable again.

Worth flagging what the conflict actually was, because it changed the patch rather than just moving it. #1747 added a per-code partial recovery path: when a GPU still has some codes active but others have cleared, it now emits a healthy event per recovered code. That is a **new clear path for this gauge**, so each recovered code now zeroes its own `error_code` series:

```python
for recovered_code in sorted(recovered_errors):
    health_events.append(...)
    pending_metric_updates.append((check_name, gpu_id, recovered_code, 0))
```

Without that, a partially recovered GPU would keep reporting 1 for codes that had already cleared, which is exactly the orphaning this PR exists to avoid. So the merge made the change more complete, not just compatible.

Full `gpu-health-monitor` suite passes on the merge: 172 passed, 5 subtests. The new test also moved onto the `PlatformConnectorConfig` constructor.

lalitadithya (approved):

Issue #1860, opened 2026-09-17
[Feature]: GPUCluster (DRA) mode support in labeler


tanishagoyal2:
### Prerequisites

- [x] I searched existing issues

### Code of Conduct

- [x] I agree to follow NVSentinel's Code of Conduct

### Feature Summary

Add GPU Operator GPUCluster (DRA) mode support to the labeler's GPU device-count labels, so gpu.count.current and gpu.count.expected keep working for both modes (ClusterPolicy and GPUCluster).

### Problem/Use Case

The labeler derives gpu.count.current / gpu.count.expected from the GPU Feature Discovery label nvidia.com/gpu.count. GPU Operator 26.7's GPUCluster mode does not deploy GFD, so on those nodes the class fails every reconcile and the count labels are never written or freeze at their last value, leaving GPU loss undetected. The GPU inventory in that mode is published instead as a per-node ResourceSlice from the NVIDIA DRA driver (gpu.nvidia.com), with one device entry per GPU. The labeler already supports ResourceSlice-based expressions, but a single class cannot serve both modes: a class whose expression references resourceSlices is skipped when the node has no slices, which is every ClusterPolicy node.

### Proposed Solution

Configure two gpu classes writing the same label pair: the existing GFD-label expression, and a DRA expression summing gpu-type devices from the node's ResourceSlices for driver gpu.nvidia.com. Exactly one succeeds per node, the other is skipped, so the labels never conflict.

### Component

Core Service

PR #1877, opened 2026-09-18
feat: labeler support gpuCluster DRA mode

tanishagoyal2:
## Problem Summary

GPU Operator 26.7's GPUCluster (DRA) mode does not deploy GPU Feature Discovery, so the labeler's gpu class, which reads nvidia.com/gpu.count, fails on every node in that mode. The gpu.count.current and gpu.count.expected labels are never written or freeze at their last value, and GPU-loss detection goes blind. In that mode the GPU inventory is published instead as resource.k8s.io/v1 ResourceSlices from driver gpu.nvidia.com. Fixes [#1860](https://github.com/kubernetes/node-problem-detector/issues/1860).

## Decision

Ship two enabled GPU classes writing the same label pair, gpu (GFD label) and gpu-dra (ResourceSlices), so one default configuration serves both operator modes with no per-cluster values. Exactly one succeeds per node:

1. New resourceSliceDriver class setting filters slices to one DRA driver and treats a node with none from that driver as a missing source, so gpu-dra is skipped on ClusterPolicy nodes even when another DRA driver publishes slices.
2. ResourceSlice pools are normalised before evaluation: only the highest spec.pool.generation counts, and the class is skipped until all resourceSliceCount slices of that generation are visible (no double or short counting during driver republish).
3. The resource.k8s.io/v1 informer is created only if the API server serves resourceslices, so Kubernetes 1.33 clusters keep working.

## Testing
Step | GPU Operator state | Result
-- | -- | --
1 | GPUCluster (DRA) mode, operator 26.7.0 | gpu-dra wrote 8/8 on all GPU nodes from gpu.nvidia.com slices (8 type=gpu devices each). gpu skipped with evaluation_error. gpu-dra never skipped on a GPU node. A node that had no count labels before gained 8/8.
2 | Switched to ClusterPolicy, operands at 26.3.3 | ClusterPolicy operands, nvidia RuntimeClass and GFD labels returned. gpu wrote 8/8 from nvidia.com/gpu.count; gpu-dra skipped with missing_source.
3 | ClusterPolicy operands upgraded to 26.7.0 (driver 595.91.07, GFD v0.20.0, DCGM 4.6.0) | Driver upgrade rolled through both reachable nodes. Count labels held their last value while operands were down and no false GPU-loss signal was produced. GFD v0.20.0 still writes nvidia.com/gpu.count; gpu wrote 8/8, gpu-dra skipped.
4 | Switched back to GPUCluster (DRA) mode | NFD removed GFD labels on ClusterPolicy deletion, so no stale source remained. gpu-dra resumed writing the moment slices appeared; gpu skipped.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] Health Monitor
- [x] Core Service
- [ ] Fault Management
- [ ] Janitor
- [ ] Deployment/Config
- [ ] API/Interface
- [ ] Preflight
- [ ] Plugins
- [ ] Documentation/CI
- [ ] New Component
- [ ] Multiple Components
- [ ] Other

## Testing
- [x] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added GPU device-count discovery from Kubernetes ResourceSlices produced by the NVIDIA GPU driver.
  * Added filtering by driver and GPU device type, with grouping by instance type.
  * GPU count labels can use ResourceSlice data alongside existing sources.

* **Bug Fixes**
  * ResourceSlice counting now uses only the latest complete pool generation.
  * Incomplete or unavailable data is skipped without publishing partial counts.
  * Processing is skipped when the required Kubernetes API is unavailable.

* **Documentation**
  * Expanded guidance for ResourceSlice configuration, completeness requirements, and generation handling.


tanishagoyal2:
/ok to test https://github.com/NVIDIA/NVSentinel/pull/1877/commits/4124ae39947df91b8c4904a87d8a224d988a4ec3

tanishagoyal2:
/ok to test https://github.com/NVIDIA/NVSentinel/pull/1877/commits/4124ae39947df91b8c4904a87d8a224d988a4ec3

distros/kubernetes/nvsentinel/charts/labeler/values.yaml
```diff
@@ -66,6 +66,32 @@ expectedDeviceCounts:
       expectedCountOverrides: []
       currentExpression: |
         int(node.metadata.labels['nvidia.com/gpu.count'])
+    
+    # GPU Operator GPUCluster (DRA) mode publishes the GPU inventory as
+    # ResourceSlices from the gpu.nvidia.com driver instead of GFD labels. This
+    # class writes the same label pair from those slices.
+    - name: gpu-dra
+      enabled: true
```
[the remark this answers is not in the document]
tanishagoyal2: This was the design decision we took to keep two classes applying same labels. However in any mode only type of class will be evaluated to true so, there will be no overriding of labels. Also, its not possible to have mix nodes thing, as we can have either GPUCluster or clusterPolicy enabled at a time.

labeler/pkg/devicecounts/device_counts.go
```diff
@@ -773,8 +778,31 @@ func (class compiledClass) expectedOverride(node *corev1.Node) (int, bool) {
...
+) []*resourcev1.ResourceSlice {
+	if class.ResourceSliceDriver == "" {
+		return resourceSlices
+	}
+
+	filtered := make([]*resourcev1.ResourceSlice, 0, len(resourceSlices))
+
+	for _, resourceSlice := range resourceSlices {
+		if resourceSlice != nil && resourceSlice.Spec.Driver == class.ResourceSliceDriver {
+			filtered = append(filtered, resourceSlice)
```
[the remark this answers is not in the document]
tanishagoyal2: Updated the code to handle slice generation

labeler/pkg/labeler/labeler.go
```diff
@@ -389,6 +407,28 @@ func createNodeInformer(
...
+
+	for _, resource := range resources.APIResources {
+		if resource.Name == resourceSliceResourceName {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
```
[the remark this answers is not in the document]
tanishagoyal2: added test. please check

tanishagoyal2:
/ok to test https://github.com/NVIDIA/NVSentinel/pull/1877/commits/37262daf426b3b9949dc9f26d1be930947999a18

labeler/pkg/devicecounts/resource_slices.go
```diff
@@ -75,3 +75,64 @@ func resourceSliceNodeName(resourceSlice *resourcev1.ResourceSlice) (string, boo
...
+		}
+
+		observed[key]++
+		expected[key] = resourceSlice.Spec.Pool.ResourceSliceCount
+		current = append(current, resourceSlice)
+	}
+
+	for key, want := range expected {
+		if observed[key] != want {
+			return current, false
```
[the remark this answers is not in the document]
tanishagoyal2: Node selection is a pool-level property: spec.nodeName "identifies the node which provides the resources in this pool", so a node-local pool cannot span nodes and the node-indexed slice set is the full pool. Completeness per node is therefore exact, which is also how the scheduler's pool gathering works. Pools using nodeSelector, allNodes or per-device node selection have no spec.nodeName, are not indexed, and are out of scope for a per-node count class rather than miscounted. Keeping the indexed lookup; a full-inventory scan per reconcile would remove the O(N·K) bound the index provides.

labeler/pkg/labeler/labeler.go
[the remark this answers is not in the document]
tanishagoyal2: added. Please check now

tanishagoyal2:
/ok to test https://github.com/NVIDIA/NVSentinel/pull/1877/commits/d42c44bab4ecf337e8536f70e2cc755ab0dfc6d5

docs/designs/057-labeler-gpucluster-dra-gpu-counts.md
```diff
@@ -0,0 +1,168 @@
...
+| ClusterPolicy GPU node + foreign DRA driver | writes labels | skipped: slices filtered to none |
+| GPUCluster GPU node | skipped: no `nvidia.com/gpu.count` label | writes labels |
+| GPUCluster GPU node, DRA pool mid-publication | skipped | skipped: pool incomplete, labels keep last value |
+| Kubernetes < 1.34 | writes labels | skipped: no informer |
+| CPU node | skipped | skipped |
+
+Classes sharing a label pair are evaluated in configuration order and the
+last to succeed wins; there is no precedence rule. Sufficient while at most
+one can succeed per node, which the driver filter and the GFD-label dependency
+guarantee for the shipped pair; marked in code as a deliberate ceiling.
```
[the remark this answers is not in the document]
tanishagoyal2: a lingering nvidia.com/gpu.count writes the same value GFD last reported, so nothing false is emitted. NFD removes the GFD labels when ClusterPolicy is deleted (observed on the canary within seconds, both directions), so the window closes before the DRA source appears. Gating gpu on a mode label is the central-detection alternative rejected in the ADR:

docs/designs/057-labeler-gpucluster-dra-gpu-counts.md
lalitadithya, line 1: please do not word wrap, and also can you update the ADR index to include this new ADR?

docs/designs/057-labeler-gpucluster-dra-gpu-counts.md
```diff
@@ -0,0 +1,168 @@
...
+- API discovery before the `resource.k8s.io/v1` informer is created, so on API
+  servers that do not serve `resourceslices` the labeler logs a warning and
+  runs with the DRA class permanently skipped instead of never syncing.
+
+## Implementation
+
+All changes live in the `labeler` module and its Helm chart; the code is the
+reference, this section only records the shape.
+
+- `labeler/pkg/devicecounts`: `ClassConfig.ResourceSliceDriver` (TOML
```
lalitadithya, line 54: the NIC DRA driver expression was written without this, do we know why we need this specifically for this new field for the GPU DRA driver? Can we add that info to the ADR?

lalitadithya (commented):
Can we add a quick memory comparison since we are now enabling a new DRA based rule by default? More specifically, I'm looking to see on a equally sized cluster would the DRA based rule require more memory or not

docs/designs/057-labeler-gpucluster-dra-gpu-counts.md
tanishagoyal2: Yes, we would need resourceSliceDriver for NIC as well. Filtering inside CEL selects the right slices but can't skip. On a node that has slices from some other driver but none from dra.networking.k8s.aws, the class is not treated as missing its source, the filter matches nothing, sum returns 0, and nic.count.current=0 is written. 
I'll update about it in doc and will update the commented example for nic in values.yaml

lalitadithya: > On a node that has slices from some other driver but none from dra.networking.k8s.aws, the class is not treated as missing its source, the filter matches nothing, sum returns 0, and nic.count.current=0 is written.

but this could also mean that all the hardware on the node is broken right? I guess there are two cases here, a CPU node that has no NIC/GPU and an accelerated node that all it's hardware nonfunctional, how would we differentiate between the two?

tanishagoyal2: We can't do much in that case. IMO these labels are to show current GPU or NICs count, not to find if driver is running on node or not. 
A GPU node in GPUCluster with a dead driver still gets gpu.present=true from NFD, gets driver.installed=false from the labeler's driver watch, and kubelet removes the driver's ResourceSlices when the plugin de-registers, so gpu-dra skips and the count labels keep their last value.

lalitadithya: > IMO these labels are to show current GPU or NICs count, not to find if driver is running on node or not.

correct, but this is orthogonal to the driver install. It is possible for the driver to be installed, but no devices being reported on the node, for example, all the GPUs fell off or the GPU ran into a XID and then didn't recover after reboot.

tanishagoyal2: `all the GPUs fell off or the GPU ran into a XID` But that case will be handled by SHM, right? by detecting XID error

lalitadithya: > But that case will be handled by SHM, right? by detecting XID error

but not for the case if the GPU doesn't register post reboot which was one of the main cases for which we implemented this label

tanishagoyal2:
Closing this PR, Raised a new PR with minimal changes limited to config changes only. https://github.com/NVIDIA/NVSentinel/pull/1890
The automatic mode detection will be handled in separate PR and is being tracked with this issue https://github.com/NVIDIA/NVSentinel/issues/1892

PR #1879, opened 2026-09-18
fix(mongodb-store): allow scraping exporter from outside release namespace

### LINKED ISSUE #1878: [Bug]: mongodb-networkpolicy opens the exporter port 9216 but only to the release namespace, so a Prometheus deployed elsewhere can never scrape it
### What happened

`charts/mongodb-store/templates/networkpolicy.yaml` opens the mongodb_exporter port on the psmdb pods, and then restricts the source to the release namespace, hardcoded:

```yaml
  ingress:
    - ports:
        - port: 27017
          protocol: TCP
{{- if .Values.usePerconaOperator }}
        - port: 9216
          protocol: TCP
{{- end }}
      from:
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: {{ .Release.Namespace }}
```

Port 9216 is only useful to a metrics scraper, and a Prometheus is very rarely deployed into the NVSentinel release namespace. In the common layout, where Prometheus lives in its own namespace, the chart therefore exposes the exporter port and simultaneously forbids the only thing that would consume it. There is no values override for the source, so a deployer cannot fix it without patching the template or adding a second policy out of band.

### Why it is easy to misdiagnose

Because a NetworkPolicy drops rather than refuses, the scrape fails as a timeout, not a connection error. Prometheus reports:

```
podMonitor/<ns>/mongodb-exporter/0  health=down
  Get "http://10.240.224.50:9216/metrics": context deadline exceeded
```

which reads as a slow or wedged exporter. It is not. On our cluster the same endpoint, measured over a port-forward, answers in **0.9 to 1.7 s** with **6,500 series** and `mongodb_up 1`. This cost real time to track down, because every signal pointed at the exporter: the container is healthy, the PodMonitor is selected, the targets are registered, and the endpoint is fast when reached directly.

It is also the only NetworkPolicy the chart ships, so every other NVSentinel pod scrapes normally and only the datastore is unreachable, which makes it look like an exporter-specific fault rather than a policy one.

### Suggested fix

Make the ingress source configurable, defaulting to current behaviour so nothing changes for existing installs. Something like:

```yaml
      from:
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: {{ .Release.Namespace }}
{{- range .Values.mongodb.networkPolicy.additionalScrapeNamespaces }}
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: {{ . | quote }}
{{- end }}
```

A `podSelector`-based alternative would also work if you would rather match Prometheus by label than by namespace.

Two smaller options, if a values knob is unwanted:

- split 9216 into its own ingress rule whose source is configurable independently of 27017, since the two ports have genuinely different consumers
- document the limitation next to the exporter settings, so the deployer knows a supplementary policy is required

### Workaround, for anyone hitting this

NetworkPolicy allow rules are additive, so a second policy grants the access without touching the chart's:

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: mongodb-exporter-scrape
  namespace: nvsentinel
spec:
  podSelector:
    matchLabels:
      app.kubernetes.io/name: percona-server-mongodb
      app.kubernetes.io/component: mongod
  policyTypes:
    - Ingress
  ingress:
    - ports:
        - port: 9216
          protocol: TCP
      from:
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: <your prometheus namespace>
```

Verified on a 271-node GB200 cluster: all three mongod targets went from `down` to `up` with this applied, and `mongodb_up` reached Prometheus for all three members.

### Related

The Percona exporter sidecar also ships without a client certificate, so `mongodb_up` is 0 on any TLS-enabled install even before the network policy matters. That is a separate defect and I have not filed it yet; happy to do so if useful. Both had to be fixed before this cluster had any datastore metrics at all.

### Environment

- NVSentinel v1.23.0, `usePerconaOperator: true`
- psmdb operator 1.23.0, crVersion 1.23.0, mongod 8.0.26-11
- Prometheus in a separate namespace, selected via PodMonitor

HesandaLiyanage:
### Problem Summary

In `charts/mongodb-store/templates/networkpolicy.yaml`, when `usePerconaOperator` is enabled, port 9216 (`mongodb_exporter`) was bundled into the same ingress rule as port 27017 (MongoDB database port), restricting ingress strictly to `.Release.Namespace`.

Because Prometheus is typically deployed in an external monitoring namespace (such as `monitoring`, `prometheus`, or `gmp-system`), scrape requests to port 9216 timed out (`context deadline exceeded`). Additionally, there was no values knob to allow external scrape sources.

### Solution

1. **Separated Ingress Rules (Least Privilege):** Kept port 27017 in its own ingress rule strictly restricted to the release namespace so database traffic is never exposed to external scrape sources.
2. **Configurable Scrape Sources:** Moved port 9216 into its own ingress rule under `{{- if .Values.usePerconaOperator }}` and added support for `networkPolicy.additionalScrapeNamespaces` (and custom `networkPolicy.additionalScrapeRules`).
3. **Backwards Compatibility:** Defaulted `additionalScrapeNamespaces` and `additionalScrapeRules` to empty lists (`[]`) so `.Release.Namespace` is always permitted and existing installations remain unaffected.
4. **Docs & Tests:**
   - Added documentation in `docs/configuration/mongodb-store.md`.
   - Added unit tests in `charts/mongodb-store/tests/networkpolicy_test.yaml` verifying default renders, port separation, and additional scrape namespaces.

Fixes #1878

## Summary by CodeRabbit

- **New Features**
  - MongoDB traffic is permitted from the release namespace.
  - Metrics access can be extended to additional namespaces and custom ingress rules on port 9216.
  - Percona deployments apply separate access rules for MongoDB and metrics traffic.
  - Network policy settings support primary or MongoDB-specific configuration, with primary settings taking precedence.

- **Documentation**
  - Added guidance and examples for cross-namespace Prometheus access and Percona-specific metrics settings.

- **Tests**
  - Added coverage for default, Percona, namespace-based, and custom metrics ingress scenarios.


HesandaLiyanage:
@lalitadithya I think there is a transient flake, can you rerun tests? one e2e failing just after 5minutes

lalitadithya (approved):

Issue #1851, opened 2026-09-16
[Bug]: HEA does ~21 sequential Mongo round trips per event; rule queries also pay 3x multikey key-examination


lfriedman-netllama:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Code of Conduct

- [x] I agree to follow NVSentinel's Code of Conduct

### Bug Description

#1676 scoped HEA rule evaluation to the incoming event's node, which removed the full-collection COLLSCAN reported in #1593. The remaining half of that scan is still there: **the mandatory first stage bounds by node but not by time, and the per-rule time bound sits in the next stage written with `$expr`, where it cannot contribute index bounds.** Every rule evaluation therefore reads every event retained for that node and filters in memory, so cost grows with retention and with the node's fault rate.

On a 270-node GB200 cluster running v1.23.0 this has put the analyzer **14.4 days behind and falling further behind by a full day per day**. It is not failing: zero restarts, and `health_event_analyzer_events_received_total` minus `health_event_analyzer_events_successfully_processed_total` is 0 over 6h. It is purely throughput-bound.

### Measurements

From a production cluster, v1.23.0, 270 nodes, MongoDB datastore with the chart's default 30-day `collectionExpirySeconds`:

| Metric | Value |
| --- | --- |
| `change_stream_lag_seconds` (health-events-analyzer) | 1,245,263 s = **14.4 days** |
| Lag growth | **+0.9997 day per day** over 24h, i.e. the read position advances ~27 s of event time per day |
| `health_event_analyzer_event_handling_duration_seconds` mean | **119.6 s per event** |
| `mongo_query_execution_duration_seconds_sum` rate | **0.991 s of query time per second of wall clock** (99.1% saturated, serial) |
| Queries per event | **21.6**, matching the 22 configured rules |
| Throughput | ~30 events/hour, which 119.6 s/event fully explains |
| Queries exceeding the 10 s top histogram bucket | **~90 per hour**, so the tail is unbounded |

The growth over time is the part that identifies it as a scan rather than a seek. Mean query duration, same cluster, same ruleset:

```
2026-08-18   0.82 s
2026-08-25   1.15 s
2026-09-09   3.72 s
2026-09-16   4.65 s     (dips in between correspond to restarts)
```

A query bounded to a 24-hour or 7-day window should stay flat as the collection grows. Growing cost means it is reading proportionally more data.

### Mechanism

[...]

PR #1880, opened 2026-09-18
perf(health-events-analyzer): bound the sticky XID window with a scalar

lfriedman-netllama:
Refs #1851. Single PR across all four affected rules, as agreed on the issue.

## What was slow

The sticky XID check pushed `$$ROOT` over `{"documents": ["unbounded", -1]}`, materialising every previous event as a full document on every document, then scanned that array with `$anyElementTrue`/`$map`. Quadratic twice over, in data volume and in work.

The key observation: the three hour bound was **already inside the map**, so only the *most recent* previous sticky XID can ever satisfy it. Everything older is scanned and discarded. So a scalar `$max` of the sticky timestamps answers exactly the same question:

```js
// replaces the $push of $$ROOT
{"$setWindowFields": {
  "sortBy": {"healthevent.generatedtimestamp.seconds": 1},
  "output": {"lastStickyTimestamp": {
    "$max": {"$cond": {"if": "$isStickyXid",
                       "then": "$healthevent.generatedtimestamp.seconds",
                       "else": null}},
    "window": {"documents": ["unbounded", -1]}}}}},

// replaces the $anyElementTrue/$map scan
{"$addFields": {"stickyXidWithin3Hours": {"$cond": {
  "if": "$isStickyXid",
  "then": {"$and": [
    {"$ne": ["$lastStickyTimestamp", null]},
    {"$lte": [{"$subtract": ["$healthevent.generatedtimestamp.seconds", "$lastStickyTimestamp"]}, 10800]}
  ]},
  "else": false}}}}
```

`$max` ignores nulls, so `lastStickyTimestamp` is null exactly when no previous event was sticky. The **same window** is kept, so the previous-only boundary you asked to preserve and the tie handling are both unchanged by construction rather than by argument. `allPreviousEvents` disappears, so nothing downstream carries prefix arrays, and the `burstId` stage is untouched.

## Measurements

Production collection, busiest node (27,321 XID events), slice pinned by `_id` and `sortBy` ties broken by `_id` so neither form can benefit from arbitrary ordering:

| N | as shipped | scalar |
| --- | --- | --- |
| 200 | 1,138 ms | 1,176 ms |
| 800 | 5,406 ms | 1,120 ms |
| 1,600 | 18,445 ms | 1,164 ms |
| 3,200 | ~90,000 ms | |
| 6,400 | exceeded the client connection timeout | |

The shipped form grows superlinearly (4.75x for 4x the input, then 3.4x for a further 2x); the scalar form is flat, dominated by fixed cost at these sizes.

## Equivalence

Checked per document on the same slices, comparing **both** `stickyXidWithin3Hours` and `burstId`:

| N | sticky events in slice | `stickyXidWithin3Hours` differs | `burstId` differs |
| --- | --- | --- | --- |
| 200 | 199 | 0 | 0 |
| 800 | 799 | 0 | 0 |
| 1,600 | 1,599 | 0 | 0 |

Repeated with the shipped list `["74","79","95","109","119"]`, under which no event on that node is sticky so the branch is inert: also 0 and 0 at all three sizes. So the replacement is exercised in both directions and agrees everywhere.

One caveat on how easily this can be measured wrong, in case it saves someone time: `$$ROOT` in a Kubernetes Job manifest is eaten by `$(...)` expansion and arrives as `$ROOT`, a missing field, which makes `$push` return an empty array and the shipped form look fast and inert. Any probe needs a control proving the push actually collected something.

## Scope

All four rules carrying the pattern: `RepeatedXIDErrorOnSameGPU`, `RepeatedXID31OnSameGPU`, `RepeatedXID13OnSameGPCAndTPC`, `XIDErrorSoloNoBurst`. The stage pair was byte-identical across all four, so one transformation covers them, applied with an exact-count assertion rather than a regex.

**Each rule keeps its own bound**: 10800s in the chart values, and in `values-tilt.yaml` 30s for three rules and 20s for `XIDErrorSoloNoBurst`. That last one is why the replacement is done per bound.

## The coupling I nearly broke, now tested

`ParseXidConfigFromPipeline` reads the sticky window back **out of the rule text** by looking for an `$lte` under the `stickyXidWithin3Hours` key, which is what lets the Go detector mirror the rule on PostgreSQL. That made the rule text and the parser an interface, and it had no test coverage at all, so a change here could silently leave the detector on its defaults.

The new formulation keeps the `$addFields` key name and the `$lte` bound, and `findLteInValue` recurses generically, so the changed nesting is fine. `parseSetWindowFieldsStage` only reads `output["burstId"]`, which is untouched.

Added `xid_sticky_window_parse_test.go` to pin it. The 30 second case is the decisive one: 3 hours is the default, so a parse that found nothing would still report 3 hours and look correct. Replacing that `$lte` with something unparseable makes the test fail with `expected: 30s, actual: 3h0m0s`.

Also verified: both values files still parse as TOML, all 22 rules and 174 stages are valid JSON, and no `allPreviousEvents` reference remains in either. The `$push: "$$ROOT"` case in `postgresql_client_test.go` is left alone, since that tests the translator's capability rather than any shipped rule.

`go test ./pkg/analyzer/` passes, gofmt and vet clean.

## Summary by CodeRabbit

- **Performance**
  - Improved XID burst and sticky-event processing by reducing memory usage when tracking prior events.
  - Processing now retains only the latest relevant sticky-event timestamp instead of materializing the full event history.

- **Behavior**
  - Preserved existing sticky-event detection windows, including the three-hour rule and burst windows of 30 or 20 seconds.
  - Existing burst thresholds and event-matching behavior remain unchanged.


lfriedman-netllama:
CI is complete on your merge commit `ba5484f1`: **84 passed, 3 skipped, 0 failing, and all 8 E2E suites green**, including both PostgreSQL ones.

| suite | result |
| --- | --- |
| AMD64 + MongoDB | SUCCESS |
| AMD64 + MongoDB (Percona) | SUCCESS |
| AMD64 + MongoDB (no TLS) | SUCCESS |
| AMD64 + PostgreSQL | SUCCESS |
| ARM64 + MongoDB | SUCCESS |
| ARM64 + MongoDB (Percona) | SUCCESS |
| ARM64 + MongoDB (no TLS) | SUCCESS |
| ARM64 + PostgreSQL | SUCCESS |

That was the condition you set on #1851 for doing this as a single PR rather than four, so flagging it explicitly. CodeRabbit also came back with no actionable comments.

Two notes on why the E2E result is meaningful here rather than incidental:

- the MongoDB analyzer suites exercise the sticky path directly, with XID 119 and 79 inside a 30 second sticky window, which is the branch this PR rewrites. They are also the suites that depend on `values-tilt.yaml` keeping its per-rule bounds, including the 20 second one on `XIDErrorSoloNoBurst` that differs from the other three.
- the PostgreSQL suites pass because `ParseXidConfigFromPipeline` still recovers the sticky window from the new stage text. That is the coupling the new unit test pins, and it is the part that would have failed silently rather than loudly if I had got the nesting wrong.

Nothing outstanding from me on this one.

lalitadithya (approved):

